PR #11, opened 2019-01-31
Cleanup FE icache

black-parrot:


black-parrot:
@Tommy @zaazad

black-parrot:
@tommydcjung

bp_fe/src/v/bp_fe_cce_lce_data_cmd.v
```diff
@@ -0,0 +1,74 @@
+/**
+ *
+ * bp_fe_lce_cce_req.v
+*/
+
+`ifndef BP_CCE_MSG_VH
+`define BP_CCE_MSG_VH
```
tommydcjung: I think these header guards are unnecessary. In fact, I believe these have been causing issues with compilation...

bp_fe/src/v/bp_fe_cce_lce_cmd.v
```diff
@@ -0,0 +1,328 @@
...
+                                                                        ,lce_assoc_p
+                                                                       )
+    , parameter bp_lce_lce_tr_resp_width_lp=`bp_lce_lce_tr_resp_width(num_lce_p
+                                                                      ,lce_addr_width_p
+                                                                      ,lce_data_width_p
+                                                                      ,lce_assoc_p
+                                                                     )    
+    , localparam lce_id_width_lp=`bp_lce_id_width
+   )
+   (input logic                                                  clk_i
```
tommydcjung: 'logic' keywork is necessary if one wants to make assignment in "always_ff" or "always_comb", since input ports are never to be assigned something, so I think 'logic' keyword is unnecessary for input ports.

bp_fe/src/v/bp_fe_cce_lce_data_cmd.v
spikemar: resolved

bp_fe/src/v/bp_fe_cce_lce_cmd.v
spikemar: resolved

bp_fe/src/v/bp_fe_icache.v
```diff
@@ -0,0 +1,611 @@
...
+  // meta_data mem
+  assign meta_data_mem_v_li = v_tv_r | meta_data_mem_pkt_yumi_li;
+  assign meta_data_mem_w_li = (v_tv_r & ~miss_v) | meta_data_mem_pkt_yumi_li;
+  assign meta_data_mem_addr_li = v_tv_r ? addr_index_tv : meta_data_mem_pkt.index;
+
+  logic [lg_lce_assoc_lp-1:0] lru_decode_way_li;
+  logic [lce_assoc_p-2:0]     lru_decode_data_lo;
+  logic [lce_assoc_p-2:0]     lru_decode_mask_lo;
+
+   bp_be_dcache_lru_decode #(
```
tommydcjung: maybe this module should go to somewhere common...

spikemar: Possibly, let's think about this for the next round.

PR #13, opened 2019-02-14
Adding ability to check for synthesizability 

stdavids:


TavioGuarino:
@stdavids Regarding synthesizability, do you know if some of the pre-processor synthesis directives that guard non-synthesizabile code would cause problems if users use a non-synopsis design compiler? 

Example attached from: bsg_ip_cores / bsg_noc / bsg_mesh_router.v (which I used)

synopsys

stdavids:
From my experience many synthesis tools will accept a variety of the common "translate_on/translate_off" directives.

dpetrisko:
Verilator has a separate flag. I was thinking it might be a good idea to define macros ‘bsg_nonsynth_start and _stop that encompass common flags

TavioGuarino:
@stdavids  Awesome, I was just worried about compatibility issues.
@dpetrisko  I think that is a good idea for sure

PR #22, opened 2019-03-03
Pparrot mittal tang dcache clockgate

tanglingshu:
clock gating dcache's data_mem, stat_mem, tag_mem


taylor-bsg:
Good changes!

Do you have the PPA analysis? In terms of the change, I suggest refactoring it a little:

Add a bsg_clkgate to bsg_ip_cores/misc.

Modify the bsg rams to include a clkgate_p parameter whose default is 0. 

Based on this, you can use a generate statement to instantiate inside the ram module the bsg_clkgate.

Then just invoke .clkgate(1) as a parameter to the rams. The benefit of this approach is 1) code can easily be adapted to use the clock gating cell if it is available in the standard cell library, and 2) you make it easy for others to enable clock gating in other rams.

black-parrot:
Per discussion, closing so that somebody may push the desired change.

PR #23, opened 2019-03-04
Changed the negedge FF to a negative D-latch 

tanglingshu:
The clock gating method used here is a integrated clock gating (ICG) cell with a negative D-latch and an AND gate. My previous design was with a negedge FF which may cause some issue. However, I don't know if you should pull this change or if I should implement a bsg_clkgate module for bsg/misc which then can be instantiated for all bsg_mem (suggested by Professor Michael Taylor). My only problem with the latter is that I'm not 100% sure that my implementation of ICG is robust enough for the bsg_ip_cores.

### DIFF
--- a/bp_be/src/v/bp_be_mmu/bp_be_dcache/bp_be_dcache.v
+++ b/bp_be/src/v/bp_be_mmu/bp_be_dcache/bp_be_dcache.v
@@ -217,12 +217,23 @@ module bp_be_dcache
   bp_be_dcache_tag_info_s [ways_p-1:0] tag_mem_mask_li;
   bp_be_dcache_tag_info_s [ways_p-1:0] tag_mem_data_lo;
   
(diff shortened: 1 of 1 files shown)


black-parrot:
You can commit the addition to the class bsg_ip_core repo on a branch as well. For things like this, you definitely want the indirection, so that if there is a bug in how we do gating, we don’t have to track down every place in Black Parrot where somebody tried to add a clock gate. The default implementation for the clock-gate could be a wire; which is replaced via the hard directory for a particular process technology. Let’s ask Scott what his flow is for committing changes to bsg_ip_cores for the class.

stdavids:
Yeah, you can push a branch to the class bsg_ip_cores repo. You should have permissions.

black-parrot:
Per discussion, closing.

PR #24, opened 2019-03-04
 bp_be_int_alu -> added if else in that module. No need to go through…

niharika77:
bp_be_int_alu → both cases are mutually exclusive and end up selecting only one as the output of the module. 
                   Added, if else there → No need to go through both the case statements.

Example:
                If, opw_v_i = 0
                    op_i = e_int_op_sgeu
Earlier → parsing through both the case statements till the end.
Now → Skip the first case and move on to the second one directly [else part]

RESULT: There is an overall improvement, i.e, both in power and time.
                But the magnitude of it is very less.

### DIFF
--- a/bp_be/src/v/bp_be_calculator/bp_be_int_alu.v
+++ b/bp_be/src/v/bp_be_calculator/bp_be_int_alu.v
@@ -49,7 +49,7 @@ module bp_be_int_alu
    , input                          opw_v_i
 
    // Result
-   , output [reg_data_width_lp-1:0] result_o
+   , output logic [reg_data_width_lp-1:0] result_o
    );
 
 // Intermediate connections
@@ -76,6 +76,8 @@ always_comb
     // These two case statements are mutually exclusive, but we separate them because they 
     //   assign to different results
     // Calculate result for 32-bit operations
+  if(opw_v_i)
+   begin
     unique case (op_i)
       e_int_op_add : resultw_sgn = src1_w_sgn +   src2_w_sgn;
       e_int_op_sub : resultw_sgn = src1_w_sgn -   src2_w_sgn;
@@ -84,8 +86,11 @@ always_comb
       e_int_op_sra : resultw_sgn = src1_w_sgn >>> shamtw;
       default      : resultw_sgn = '0;
     endcase
-  
+    result_o = reg_data_width_lp'(resultw_sgn);
+  end
     // Calculate result for 64-bit operations
+  else
+   begin  
     unique case (op_i)
       e_int_op_add       : result_sgn = src1_sgn +   src2_sgn;
       e_int_op_sub       : result_sgn = src1_sgn -   src2_sgn;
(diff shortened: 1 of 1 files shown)


stdavids:
Interesting, what were the actual power, area and timing improvements?

From a hardware description standpoint the two are _very_ similar, for example if I were to draw the schematics I'm not sure I would draw them differently. However, it is possible that DC did parse it slightly differently and found a small optimization.

niharika77:
Please find the attached document.
Optimization idea 2.

stdavids:
Seems like the attachment got dropped? You could try using the github website to attach a document.

dpetrisko:
The difference could be in the X handing behavior. ifs and ternaries have different X pessimism requirements and could cause very slight performance differences 

Dan

taylor-bsg:
@niharika77  Let us know when you can share the PPA results. It might also be useful to show the difference in the area report. I have posted the baseline area report below. I think that there is optimization opportunity here, based on the synthesis report -- I see 4 different comparators, and two DP_OPs (one appears to be 64 bit and the other is 32 bit). If you can merge the common operators (i.e. the 32-bit and 64-bit adds into a single adder, by preprocessing the inputs of the 32-bit one so that you can feed into the 64-bit adder), I think you could save power and energy.

            bp_be_int_alu_0                                                                                                                                                            5127       5127          0        36651.11     36651.11         0.00
                 bp_be_int_alu_DW_cmp_J28_0_0                                                                                                                                            348        348          0         2282.80      2282.80         0.00
                 bp_be_int_alu_DW_cmp_J29_0_0                                                                                                                                            323        323          0         2152.86      2152.86         0.00
                 bp_be_int_alu_DW_cmp_0_0                                                                                                                                                335        335          0         2182.35      2182.35         0.00
                 bp_be_int_alu_DW_cmp_J27_0_0                                                                                                                                            347        347          0         2257.92      2257.92         0.00
                 bp_be_int_alu_DP_OP_37_128_1253_J36_1_0                                                                                                                                 592        592          0         5128.70      5128.70         0.00
                 bp_be_int_alu_DP_OP_40_129_318_J37_1_0                                                                                                                                  298        298          0         2549.15      2549.15         0.00

niharika77:
[EE525 Final Project Log.pdf](https://github.com/black-parrot/pre-alpha-release/files/2932701/EE525.Final.Project.Log.pdf)

Sorry for the delay, I did send the file to Scott. 
I wasn't able to send the file through gmail to everyone. Let me know, whether you can see it or not!

niharika77:
I haven't looked at the area report but will do that and will update you.

niharika77:
**Area Report Post_synthesis**

bp_be_int_alu_0                                                                                                                                                         5129       5129          0        36516.56     36516.56         0.00 
bp_be_int_alu_DW_cmp_J28_0_0                                                                                                                                            347        347          0         2292.02      2292.02         0.00 
bp_be_int_alu_DW_cmp_J29_0_0                                                                                                                                            323        323          0         2142.72      2142.72         0.00 
bp_be_int_alu_DW_cmp_0_0                                                                                                                                                334        334          0         2175.90      2175.90         0.00 
bp_be_int_alu_DW_cmp_J27_0_0                                                                                                                                            347        347          0         2224.74      2224.74         0.00 
bp_be_int_alu_DP_OP_37_128_1253_J36_1_0                                                                                                                                 608        608          0         5138.84      5138.84         0.00 
bp_be_int_alu_DP_OP_40_129_318_J37_1_0                                                                                                                                  290        290          0         2503.07      2503.07         0.00 



We can see area too gets better but with small magnitude.

PR #26, opened 2019-03-05
TeamOne (ME): reduced fifo depth in cce

derekcom17:
**FIFOs in bp_cce_top have been reduced from two deep to one deep.** 

PPA results to memory end: 
- Power: reduced by 6.5%
- Performance: reduced by 1.3% _(IPC did not change)_
- Area: reduced by 3.8%

### DIFF
--- a/bp_me/src/v/cce/bp_cce_top.v
+++ b/bp_me/src/v/cce/bp_cce_top.v
@@ -138,10 +138,14 @@ module bp_cce_top
   logic                                          mem_data_cmd_v_from_cce;
   logic                                          mem_data_cmd_ready_to_cce;
 
+  localparam data_fifo_depth_lp = 1; // These fifos are very wide
+  localparam  cmd_fifo_depth_lp = 1; // These fifos are less wide
+
   // Inbound LCE to CCE
-  bsg_two_fifo
+  bsg_fifo_1r1w_small
     #(.width_p(bp_lce_cce_req_width_lp)
       ,.ready_THEN_valid_p(1) // ready-then-valid
+      ,.els_p(cmd_fifo_depth_lp) // can be as low as 1
       )
     lce_cce_req_fifo
      (.clk_i(clk_i)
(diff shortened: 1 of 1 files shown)


taylor-bsg:
Reducing the area of the FIFOs is a reasonable direction to take for reducing PPA and I commend you for the proposal.  However, the changes need to include comments that convince us why the changes do not introduce correctness or performance bugs into black parrot, and it cannot rely solely on "the benchmark ran okay". In general, with on-chip networks, we want to avoid deadlock and congestion in the network. If the FIFOs are not sized correctly, then packets will back up into the network and cause other cores to stall on unrelated traffic. So for example, on the network where LCEs send requests to CCE's, if we have a 1 element FIFO, then if multiple CCE's send a request to the LCE, then traffic will backup into the network, blocking the links in the network, preventing other cores from doing cache misses to unrelated CCE's. On the other hand, on some networks, I think we know that a CCE has a request out that has at most packet of response, and in that case only 1 word is required. The full-on version of this would be an argument that establishes the relationship between CCE/LCE count and the number of words in each FIFO. But a simpler piece to chew off would be identifying those CCE fifos that only need one element because we know they can only ever have one outstanding message.

On our side, it seems like we could definitely facilitate the memory end team by providing traces that will run on say 4-cores and say, 2 CCEs?

derekcom17:
Thank you for the response.

In this branch, the only FIFOs we changed were between the CCE and its L2.
These FIFOs can be found in pb_cce_top.

Changing the FIFOs in the network was another idea we were working on.
Thank you for the advice.

dpetrisko:
Traces are hard / impossible to create for this case, as the ordering is non-deterministic with multiple cores working on the same data. 

The ME random stress testbench should be able to test more LCEs, although I am not sure if it has ever been tested with more than one CCE.

black-parrot:
@dpetrisko Agreed that there is an element of non-determinism, if we are comparing different implementations, and there is sharing. But we need to have some form of testing that people can use to sanity check their changes, especially something that uses more than 1 core. Putting aside the question of how we test all cases in a general manner, one test/benchmark I could see helping for the kind of optimization being proposed would be to have a bunch of cores that are using independent memory regions that are striped across the 2 CCE's. So there would be no invalidation traffic or sharing issues. Then we could play a bunch of traces against the memory system and compare the performance.

taylor-bsg:
We are closing this one per the discussion in class. Thanks again for submitting it!

PR #29, opened 2019-03-05
Change BTB indexing scheme

dpetrisko:
Indexing not by the bottom two bits of the PC, because these are constant.  As a result, we can shrink the BTB by 75% for no performance impact.  Additionally, queues are smaller because of less branch_metadata.

This results in ~10% FE area decrease and ~55% BTB area decrease in post_synth results


iihihiuh:
Hi, Dan,
I have some thoughts about bht & btb indexing.

I think in the future if compressed instruction is supported (aligned on 2 bytes), it might be good to include feature to select index depending on whether current core is supporting compressed instruction.

black-parrot:
@iihihiuh Thanks Yongqin -- question: in the compressed instruction set, are there any branches that are encoded in 16 bits? If not, then we probably can just use word indexing?

iihihiuh:
Yes, there are a number of them. There are compressed beqz, bnez, and some other jump with/without link. Each of them could also jump to 2 byte aligned address. 

Even if there are no compressed branch instruction, a 4-byte branch instruction could also happen to have an address aligned to 2 byte if compressed is supported.

dpetrisko:
1. You are right, but we don't have compressed instructions right now
2. It's fairly standard to use even higher index bits.  Since branches are rarely back to back, the destructive interference is usually not a problem.  Without doing a larger study, I didn't want to get more aggressive.  Possibly the offset should be a parameter

taylor-bsg:
@iihihiuh Then what you say makes sense. We could address both concerns by XORing some high bits and the lowest bit. But as @dpetrisko suggests let's wait until the compressed set is implemented.

taylor-bsg:
@stdavids Good to merge?

taylor-bsg:
@dpetrisko what about power?

dpetrisko:
I'll collect the before and after power tomorrow.  One concern I have about merging is that this requires a new hardened memory.  Should I push that memory config to the class designs + bsg_ip_cores? @stdavids

black-parrot:
What is the new hardened memory that is required?

dpetrisko:
64x128 instead of 64x512

taylor-bsg:
Ah, okay, got the two threads mixed up.  Should it really be 64 bits wide? What is eaddr width anyways? Presumably it should be as wide as the virtual address space?

M

dpetrisko:
True, yes. There’s a large set of eaddr->vaddr optimizations (in both the FE and BE). It requires some thought as to where each is allowed according to the spec. Technically, the calculated address is an eaddr, which is defined by the spec to be vaddr sign extended to 64 bits. 

If the BTB is filled by a vaddr where the calculated address overflows (and thus in incorrect), that’s probably a problem to be dealt with in the BE? But in either case the entry stored in the BTB will be semantically different than the calculated address. Since it’s speculative, I don’t believe this would cause correctness problems, but I left things like that because I haven’t thought too carefully on the implications.

black-parrot:
Okay I'll make an issue.

stdavids:
Regarding new memories, you can just push to designs and bsg_ip_cores. Just add your SRAM though, don't delete the older one.

black-parrot:
@dpetrisko are the commits to designs and bsg_ip_cores done? Do you want to submit this as a branch to your critical path optimization? Or do you want to abandon this change and submit your new FE changes?

dpetrisko:
Let's abandon and submit new changes.  I'm also changing the indexing scheme in those

PR #30, opened 2019-03-05
unifying lce-cce port names

tommydcjung:


dpetrisko:
Looks good to me.  Please keep the comment header for now though.  I'm still debating about what to do for inline documentation

tommydcjung:
okay, but if we are going to have comment header, it should be filled out.

PR #31, opened 2019-03-05
Balanced pipelined logic in dcache

iihihiuh:
In our floor plan, some logic on critical path can be moved into earlier pipeline stage. Right now our chip can run 0.7 percent faster than our baseline. Moreover, power gets down 3 percent and area shrinks by 2.13 precent.

### DIFF
--- a/bp_be/src/v/bp_be_mmu/bp_be_dcache/bp_be_dcache.v
+++ b/bp_be/src/v/bp_be_mmu/bp_be_dcache/bp_be_dcache.v
@@ -181,7 +181,12 @@ module bp_be_dcache
   logic byte_op_tl_r;
   logic [bp_page_offset_width_gp-1:0] page_offset_tl_r;
   logic [data_width_p-1:0] data_tl_r;
+  logic [paddr_width_p-1:0] paddr_tl;
+  logic [tag_width_lp-1:0] addr_tag_tl;
 
+   assign paddr_tl = {ptag_i, page_offset_tl_r};
+   
+  assign addr_tag_tl = paddr_tl[block_offset_width_lp+index_width_lp+:tag_width_lp]; 
   assign tl_we = v_i & ready_o & ~poison_i;
  
   always_ff @ (posedge clk_i) begin
@@ -277,15 +282,23 @@ module bp_be_dcache
   logic [index_width_lp-1:0] addr_index_tv;
   logic [word_offset_width_lp-1:0] addr_word_offset_tv;
 
+  logic [ways_p-1:0] load_hit_tv;
+  logic [ways_p-1:0] load_hit_tl;
+  logic [ways_p-1:0] store_hit_tv;
+  logic [ways_p-1:0] store_hit_tl;
+
   assign tv_we = v_tl_r & ~poison_i & ~tlb_miss_i;
 
   always_ff @ (posedge clk_i) begin
     if (reset_i) begin
       v_tv_r <= 1'b0;
+      store_hit_tv <= '0; 
+      load_hit_tv <= '0; 
     end
     else begin
       v_tv_r <= tv_we;
-
+      store_hit_tv <= store_hit_tl; 
+      load_hit_tv <= load_hit_tl; 
       if (tv_we) begin
         load_op_tv_r <= load_op_tl_r;
         store_op_tv_r <= store_op_tl_r;
@@ -311,11 +324,10 @@ module bp_be_dcache
   assign addr_tag_tv = paddr_tv_r[block_offset_width_lp+index_width_lp+:tag_width_lp];
   assign addr_index_tv = paddr_tv_r[block_offset_width_lp+:index_width_lp];
   assign addr_word_offset_tv = paddr_tv_r[byte_offset_width_lp+:word_offset_width_lp];
-
+   
   // miss_detect
   //
-  logic [ways_p-1:0] load_hit_tv;
-  logic [ways_p-1:0] store_hit_tv;
+  logic [ways_p-1:0] way_hit_tl; 
   logic [ways_p-1:0] invalid_tv;
   logic load_miss_tv;
   logic store_miss_tv;
@@ -325,10 +337,11 @@ module bp_be_dcache
   logic [way_id_width_lp-1:0] store_hit_way;
 
   for (genvar i = 0; i < ways_p; i++) begin
-    assign load_hit_tv[i] = (addr_tag_tv == tag_info_tv_r[i].tag)
-      & (tag_info_tv_r[i].coh_state != e_MESI_I);
-    assign store_hit_tv[i] = (addr_tag_tv == tag_info_tv_r[i].tag)
-      & (tag_info_tv_r[i].coh_state == e_MESI_E);
+    assign way_hit_tl[i] = addr_tag_tl == tag_mem_data_lo[i].tag; 
+    assign load_hit_tl[i] = way_hit_tl[i]
+      & (tag_mem_data_lo[i].coh_state != e_MESI_I);
+    assign store_hit_tl[i] = way_hit_tl[i]
+      & (tag_mem_data_lo[i].coh_state == e_MESI_E);
     assign invalid_tv[i] = (tag_info_tv_r[i].coh_state == e_MESI_I);
   end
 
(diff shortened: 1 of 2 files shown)


bp_be/src/v/bp_be_mmu/bp_be_dcache/bp_be_dcache.v
```diff
@@ -277,15 +282,23 @@ module bp_be_dcache
...
   always_ff @ (posedge clk_i) begin
     if (reset_i) begin
       v_tv_r <= 1'b0;
+      store_hit_tv <= '0; 
+      load_hit_tv <= '0; 
     end
     else begin
       v_tv_r <= tv_we;
-
+      store_hit_tv <= store_hit_tl;
```
tommydcjung: shouldn't this be in if (tv_we) block also?

bp_be/src/v/bp_be_mmu/bp_be_dcache/bp_be_dcache.v
```diff
@@ -311,11 +324,10 @@ module bp_be_dcache
   assign addr_tag_tv = paddr_tv_r[block_offset_width_lp+index_width_lp+:tag_width_lp];
   assign addr_index_tv = paddr_tv_r[block_offset_width_lp+:index_width_lp];
   assign addr_word_offset_tv = paddr_tv_r[byte_offset_width_lp+:word_offset_width_lp];
-
+   
   // miss_detect
   //
-  logic [ways_p-1:0] load_hit_tv;
-  logic [ways_p-1:0] store_hit_tv;
+  logic [ways_p-1:0] way_hit_tl;
```
tommydcjung: instead of "way_hit_tl", call it "tag_hit_tl"

bp_be/src/v/bp_be_mmu/bp_be_dcache/bp_be_dcache.v
```diff
@@ -325,10 +337,11 @@ module bp_be_dcache
   logic [way_id_width_lp-1:0] store_hit_way;
 
   for (genvar i = 0; i < ways_p; i++) begin
-    assign load_hit_tv[i] = (addr_tag_tv == tag_info_tv_r[i].tag)
-      & (tag_info_tv_r[i].coh_state != e_MESI_I);
-    assign store_hit_tv[i] = (addr_tag_tv == tag_info_tv_r[i].tag)
-      & (tag_info_tv_r[i].coh_state == e_MESI_E);
+    assign way_hit_tl[i] = addr_tag_tl == tag_mem_data_lo[i].tag;
```
tommydcjung: maybe it's better to separate the computation done in TL stage and TV stage in a separate for-loop. move up the TL stage for-loop to appropriate location.

tommydcjung:
Since we are accessing TAG SRAM in the TL stage, I thought it would be better to flop the SRAM output and do the tag comparison in the TV stage. Also, PTAG comes from TLB, which might be another source of delay.

bp_be/src/v/bp_be_mmu/bp_be_dcache/bp_be_dcache.v
```diff
@@ -277,15 +282,23 @@ module bp_be_dcache
...
+  logic [ways_p-1:0] load_hit_tl;
+  logic [ways_p-1:0] store_hit_tv;
+  logic [ways_p-1:0] store_hit_tl;
+
   assign tv_we = v_tl_r & ~poison_i & ~tlb_miss_i;
 
   always_ff @ (posedge clk_i) begin
     if (reset_i) begin
       v_tv_r <= 1'b0;
+      store_hit_tv <= '0;
```
tommydcjung: if this is a register, it needs to have suffix "_r"

iihihiuh:
Hi Tommy,

I think the ram we are using is synchronous sram, it will output data_o on the next cycle. Flopping the output of a synchronous sram is the same as adding another pipelined stage with no logic. Thus, I think this dcache is effective three stage pipelined. I moved some logics from the third stage to the second stage.

tommydcjung:
cache_pipeline

thanks Yongqin. From what I understand, the combination logic is moved to the output of the SRAM, as shown in the diagram above. Intuitively, I thought this would add more delay to SRAM read, and add more capacitive load to the output of SRAM, so I thought it could be worse for the delay, but I could be wrong. If the simulation results suggests that there is an improvement then we should definitely make a change.

iihihiuh:
Thanks Tommy!

taylor-bsg:
@iihihiuh Quick question -- what is the old and new cycle time with this change? Was the measurement after synthesis, APR, etc?  What is responsible for the area and power decrease?

In the Raw processor, we put the tag check in the same stage with the tags as well. The reason is that the tags were a smaller SRAM than the data lookup (i.e. 38bx512 versus 64bx4096), so there was more room in the critical path. In this case, it seems like the tags are dependent also on the output on the VM stage, so it will depend a little bit on what we do there.  In our design, what are the relative sizes and widths of the tag ram and the data "ramlets"?

bp_be/src/v/bp_be_mmu/bp_be_dcache/bp_be_dcache.v
```diff
@@ -325,10 +337,11 @@ module bp_be_dcache
...
-    assign load_hit_tv[i] = (addr_tag_tv == tag_info_tv_r[i].tag)
-      & (tag_info_tv_r[i].coh_state != e_MESI_I);
-    assign store_hit_tv[i] = (addr_tag_tv == tag_info_tv_r[i].tag)
-      & (tag_info_tv_r[i].coh_state == e_MESI_E);
+    assign way_hit_tl[i] = addr_tag_tl == tag_mem_data_lo[i].tag; 
+    assign load_hit_tl[i] = way_hit_tl[i]
+      & (tag_mem_data_lo[i].coh_state != e_MESI_I);
+    assign store_hit_tl[i] = way_hit_tl[i]
+      & (tag_mem_data_lo[i].coh_state == e_MESI_E);
     assign invalid_tv[i] = (tag_info_tv_r[i].coh_state == e_MESI_I);
```
tommydcjung: maybe invalid_tv should move as well?

bp_be/src/v/bp_be_mmu/bp_be_dcache/bp_be_dcache.v
```diff
@@ -325,10 +337,11 @@ module bp_be_dcache
   logic [way_id_width_lp-1:0] store_hit_way;
 
   for (genvar i = 0; i < ways_p; i++) begin
-    assign load_hit_tv[i] = (addr_tag_tv == tag_info_tv_r[i].tag)
-      & (tag_info_tv_r[i].coh_state != e_MESI_I);
-    assign store_hit_tv[i] = (addr_tag_tv == tag_info_tv_r[i].tag)
-      & (tag_info_tv_r[i].coh_state == e_MESI_E);
```
tommydcjung: is tag_info_tv_r still needed? might be good idea to check if there are other registers that can be removed.

iihihiuh:
@taylor-bsg Cycle time is unaffected. All my measurements are taken after place and route. I will post those measurements after I am done with my other homework.

iihihiuh:
@taylor-bsg After merging Dan's changes in pull request #51, the path we are optimizing became the critical path. We can run through, our chip with tighter constrain.

Performance wise:
Cycle counts are unaffected.
Before : 2.5601 ns
After : 2.491 ns
Improvement: 2.6 %
The path we optimized got 5.0 % up

Area:
Before : 1283418.5
After : 1283656.3050 um^2
Improvement: - 0.018 %.

black-parrot:
Great! Can you do a pull request on top of Dan’s branch? Then we can accept it. Make sure to run all of the instruction tests and all of the benchmarks.

dpetrisko:
Hm, I was able to get a better critical path than 2.4 with my changes alone. Are you sure you’re pushing the synthesis tool enough?

iihihiuh:
> Hm, I was able to get a better critical path than 2.4 with my changes alone. Are you sure you’re pushing the synthesis tool enough?
> […](#)
> On Mar 7, 2019, at 7:54 PM, black-parrot ***@***.***> wrote: Great! Can you do a pull request on top of Dan’s branch? Then we can accept it. Make sure to run all of the instruction tests and all of the benchmarks. — You are receiving this because you are subscribed to this thread. Reply to this email directly, view it on GitHub, or mute the thread.

iihihiuh:
@dpetrisko

black-parrot:
Closing; @iihihiuh can resubmit a pull request off of @dpetrisko 's branch if his change results in improved cycle time.

PR #35, opened 2019-03-06
Size reduction of bp_me network mesh_router_buffered fifos

chaseemory:
Throughout the network there are many fifos, many of which are nearly 600bits wide to accommodate the width of the cache lines in Black Parrot. Reducing the depth of these fifos has a massive improvement to power and area versus the baseline. Performance impacts were 0 at least in the design given in the lab. This may not be the case with other BP implementations in the future as congestion in the network caused by backpressure in these fifos could potentially impact performance negatively. The overall PPA could still improve if the benefits to power and area outweighed the performance degradation. 

Improvements vs Baseline

Area:     %13.1 and 315k um^2
Power:   %31.2 and .71e5uW
Performance had 0 change using the benchmarks provided.

### DIFF
--- a/bp_me/src/v/network/bsg_fifo_1r1w_small.v
+++ b/bp_me/src/v/network/bsg_fifo_1r1w_small.v
@@ -1,0 +1,108 @@
+// bsg_fifo_1r1w_small
+//
+// bsg_fifo with 1 read and 1 write, using
+// 1-write 1-async-read register file implementation.
+//
+// used for smaller fifos.
+//
+// input handshake protocol (based on ready_THEN_valid_p parameter):
+//     valid-and-ready or
+//     ready-then-valid
+//
+// output protocol is valid-yumi (like typical fifo)
+//                    aka valid-then-ready
+//
+//
+
+module bsg_fifo_1r1w_small #( parameter width_p      = -1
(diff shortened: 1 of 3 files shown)


taylor-bsg:
It would be better to offer this as a pull request to the class repo bsg_ip_cores since we can't see how you changed those modules from github. What did you change in buffered router and the 1r1w FIFO?

dpetrisko:
One way to see the changes would be to have one commit with the original router. Then make a second commit with the modified router.

chaseemory:
Yeah, originally I was trying to keep all changes in the BP folder so they could all be submitted as pull requests here, but Scott has mentioned that we can submit pull requests for any of the repos in the flow or bsg_ip_cores. 

I didn't change anything in the fifo, I just instantiated a bsg_fifo_1r1w_small and set its els_p to 1, instead of the bsg_two_fifo that was there originally. Including the fifo file was a mistake, it is not used at all.

The timing information for the optimization is also based on icc timing reports instead of PrimeTime, due to signal integrity issues causing problems with certain critical paths.

this optimization also decreases the DRC errors from ~7k to ~700 :D

taylor-bsg:
Per our discussion, we will close this request. Thanks again for this submission.

PR #60, opened 2019-03-07
Improved mIPC and Reworked the PC pipeline in "bp_fe_pc_gen"

farzamgl:
The original pc gen's pc pipeline used to stall on a pc_redirect until it went low. Now we switch the PC to redirected pc at the first cycle of pc redirect signal going high but in the next cycles, we keep the pipeline going and load the pc with the next_pc.

Also while the icache_miss signal is high, the pc pipeline used to constantly loop between pc, last_pc, and icache_miss_pc resulting in unnecessary power consumption and possibly wrong data on the icache_miss_pc signal when the icache_miss goes low. We changed the enable signals on these 3 registers to maintain the missed PC during that time and avoid switching.

Performance (Instr. per Second): 2.88x
Area: 1.07x
Power(Watt per Instr.): 0.38x

### DIFF
--- a/bp_fe/src/v/bp_fe_pc_gen.v
+++ b/bp_fe/src/v/bp_fe_pc_gen.v
@@ -112,12 +112,12 @@ assign pc_gen_fe_o     = pc_gen_queue;
 assign fe_pc_gen_cmd   = fe_pc_gen_i;
 assign icache_pc_gen   = icache_pc_gen_i;
 
-assign misalignment          = fe_pc_gen_v_i
-                               && fe_pc_gen_cmd.pc_redirect_valid 
-                               && ~fe_pc_gen_cmd.pc[3:0] == 4'h0 
-                               && ~fe_pc_gen_cmd.pc[3:0] == 4'h4
-                               && ~fe_pc_gen_cmd.pc[3:0] == 4'h8
-                               && ~fe_pc_gen_cmd.pc[3:0] == 4'hC;
+assign misalignment    = fe_pc_gen_v_i
+                         && fe_pc_gen_cmd.pc_redirect_valid 
+                         && ~fe_pc_gen_cmd.pc[3:0] == 4'h0 
+                         && ~fe_pc_gen_cmd.pc[3:0] == 4'h4
+                         && ~fe_pc_gen_cmd.pc[3:0] == 4'h8
+                         && ~fe_pc_gen_cmd.pc[3:0] == 4'hC;
 							   
 
 //register the pc redirect signal
@@ -199,25 +199,9 @@ always_ff @(posedge clk_i) begin
     pc <= bp_first_pc_p;
   end
   else if(pc_gen_icache_ready_i && pc_gen_fe_ready_i) begin
-    pc <= next_pc;
-  end
-end
-
-always_ff @(posedge clk_i) begin
-  if(reset_i) begin
-    last_pc <= bp_first_pc_p;
-  end
-  else if(pc_gen_icache_ready_i && pc_gen_fe_ready_i) begin
-    last_pc <= pc;
-  end
-end
-
-always_ff @(posedge clk_i) begin
-  if(reset_i) begin
-    icache_miss_pc <= bp_first_pc_p;
-  end
-  else if(pc_gen_icache_ready_i) begin
-    icache_miss_pc <= last_pc;
+    pc             <= next_pc;
+	last_pc        <= pc;
+	icache_miss_pc <= last_pc;
   end
 end
   
--- a/bp_fe/src/v/bp_fe_pc_gen.v
+++ b/bp_fe/src/v/bp_fe_pc_gen.v
@@ -54,21 +54,6 @@ module bp_fe_pc_gen
    , input                                           fe_pc_gen_v_i
    , output logic                                    fe_pc_gen_ready_o
    );
-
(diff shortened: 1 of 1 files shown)


black-parrot:
Very nice.

black-parrot:
Hi Farzam, can you clone the FE schematic,  assign ownership to me, update it, and put a link to it (with google docs version number) into the Header of the FE top source code? Please try to keep it clean like the original and obey the schematic rules in BSG SystemVerilog Coding doc.

black-parrot:
https://docs.google.com/presentation/d/1l0gdejJGftE4zZGD-kzK7WB7p4EeQBBNXIZc-Vjm94c

black-parrot:
Can you also give the numbers in the more standard improvement “x” notation of computer architecture, where > 1.0 is good. So for example runtime speedup is (old cycles / new cycles), area improvement is (old area / new area) and energy improvement is (old joules per instr / new joules per instr). Also can you give the old and new absolute numbers?

farzamgl:
Here are the PPA results(performance and power are averaged over the 4 tests):
Performance (million Inst / Sec): 	78.3	-> 225.8		Improvement: 2.89x
Power (nJ / Inst):			0.9809 -> 0.3739		Improvement: 2.62x
Total Area (µm2):				1355590 -> 1340831 	Improvement: 1.01x

black-parrot:
@dpetrisko @HelloKayT Very interesting results from Farzam. 

The results pre-BTB for Petrisko/Lim are:

Performance (million Inst / Sec):  -> 187.9 (Minstr/sec) Improvement: 2.2X    
Power (nJ/Inst): 1.35 nJ->0.556 nJ  Improvement: 2.43X 
Total Area (um2): 1.2 um^2 -> 0.968 um^2 Improvement: 1.25X area 

So using the absolute numbers versus Petrisko/Lim is 1.2X perf improvement, 1.48X energy, but 0.72X area improvement (1.38X worse). So you guys are definitely both still on the Pareto Curve -- Congrats! 

Is there some mutual learning from the two designs that can improve the designs in both direction?

farzamgl:
I think these two pull requests are mostly independent so we should be able to merge them to get a better PPA.

black-parrot:
That would be pretty neat -- perhaps you can guys can explore that? Farzam, a comprable schematic to Katie's would be helpful for everybody to communicate their relative designs.

https://docs.google.com/presentation/d/147kLmlUw3yNTlgcXGe3MajmAJPjpB9kOL_5kUlKSWLo/edit#slide=id.g5078081ba4_0_0

farzamgl:
Yup. Working on it.

HelloKayT:
Quick question about your metrics since our baselines are different: what was the minimum clock period you were able to synthesize at (both baseline and optimized)?

Did you fix the branch prediction, so the BTB/BHT is outputting predictions for the right PC? I'm curious if the BTB/BHT has a significant performance improvement over the static method. One place this might be the case is in towers, because we cannot predict jalr and towers is recursive (jalr is used at the end of every function call).

A couple things about the code: 
-I think you need to either flop the redirect request (not just the valid) or pull down on the ready signal to completely respect the ready-valid handshake. Currently the ready is just assigned to valid, so the sender thinks you have acknowledged the redirect immediately, which might not be true under an icache miss, and is then free to change the data. I believe due to higher level properties, (you'll only have 1 redirect pending) this actually works fine right now, but I don't know if we should rely on this behavior?
-I don't think the two pull requests will be independent, because we have two stages of PC pipelining whereas you have 3

farzamgl:
- Baseline: 2.5ns period + 0.3ns uncertainty, After: 2.6ns period + 0.3ns uncertainty
- We didn't touch the branch predictions, so the improvement you got must be because of that change.
- I think you are right on the ready/valid issue. I'll look into it.
- This code doesn't rely on the pc pipeline being 3 stages, as long as it can keep the icache miss pc in the last one. So if you can do that with 2 stages, it shouldn't be hard to combine them.

taylor-bsg:
If you guys wanted to compare prediction accuracy, you could compute MPKI
in the backend (mispredicts per kilo instruction committed), although I
would actually report mispredicts and instructions committed to make sure
you are on the same page. Also interesting would be branch taken penalty
cycles per kilo instruction.

HelloKayT:
If the branch prediction wasn't touched from the original, I don't think it is actually contributing to throughput. I will test later, but I think the testbench fix should improve our throughput. Our design will take a redirect whenever the redirect valid signal is high. My understanding from reading the code Farzam's design specifically tests and only takes redirects the first cycle the valid is high.

farzamgl:
It was fixing the issue that we had in the BE with long pc redirects, after this push the BE sends only single-cycle pc redirects and waits until it gets the results from the pc gen, so I guess this problem is fixed now.

black-parrot:
Hi Farzam, do we have updated stats after pulling the fixes from the backend tester? Or was your change mostly to compensate for the bugs in the BE tester?

farzamgl:
Hi. This change was fixing the pc redirect problem in the FE side, but
since the BE got fixed it's not necessary in the FE anymore. Basically the
baseline numbers moved to our numbers after the BE change. The was also the
change to the pc pipeline which also can be found in Dan/Katie's pull
request. So I think we can close this one.

taylor-bsg:
Thank you Farzam!

PR #61, opened 2019-03-07
Back End critical path fixes

dpetrisko:


black-parrot:
Hi Dan, can you update the BE schematic,   and put a link to it (with version number) into the Header of the be top source code, ideally prior to commit.  We can make that a regular practice.

PR #64, opened 2019-03-07
Add clock gating to largest RAMs

derekcom17:
# Adds clock gating to largest RAMs

Uses the new bsg_ip_cores branch that includes clock gated rams.
Clock gating added to L2 memory and CCE Directory.

PPA Analysis (post synth):
- Power: down 1.15 %
- Performance: down 1.90 %
- Area: down 0.03 %

Since the memory end will most likely not be the critical path, it will likely not be clocked at its max frequency.

### DIFF
--- a/bp_me/src/v/cce/bp_cce_dir.v
+++ b/bp_me/src/v/cce/bp_cce_dir.v
@@ -113,6 +113,7 @@ module bp_cce_dir
   bsg_mem_1rw_sync_mask_write_bit
     #(.width_p(way_group_width_lp)
       ,.els_p(num_way_groups_p)
+      ,.enable_clock_gating_p(1'b1) // Gating enabled! 
       )
     wg_ram
      (
--- a/bp_me/syn/flist.vcs
+++ b/bp_me/syn/flist.vcs
@@ -31,6 +31,8 @@ $BP_BE_DIR/src/include/bp_be_dcache/bp_be_dcache_pkg.vh
 $BP_ME_DIR/src/include/v/bp_cce_pkg.v
 $BP_ME_DIR/src/include/v/bp_me_network_pkg.v
 ## bsg_ip_cores files
+$BSG_IP_CORES_DIR/bsg_misc/bsg_dlatch.v
+$BSG_IP_CORES_DIR/bsg_misc/bsg_clkgate_optional.v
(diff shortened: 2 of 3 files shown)


dpetrisko:
This is interesting. Do you know why clock gating would cause an IPC decrease?

derekcom17:
IPC stayed the same.
The critical path became longer.

dpetrisko:
Oh sorry, I misread your last sentence as no critical path change. You mean that the critical path for the whole chip will most likely not be in memory end?

derekcom17:
Scott said it would most likely be in the back end. I don't actually know.

derekcom17:
## Update:
Maybe this should be a separate pull request, but I found another optimization that does not involve any changes to this code.

On the ee 477 bsg_ip_cores repo I have changed the bsg_mem_1r1w (asynchronous) to also have a clock gating option. 

1. Pull the branch clock_gating.
2. Go to bsg_mem/bsg_mem_1r1w.v and change the default parameter value of enable_clock_gating_p from 1'b0 to 1'b1. 
3. Do the same to bsg_mem_1r1w.v in the hard folder. 

This will make all of the asynchronous 1r1w memories clock gated (including all of the FIFOs in the network.)


**Results: (post synth, compared to baseline)**
- Power: down **79.9%** (i.e. uses 1/5 the power or baseline!)
- Performance: down 2.7%
- Area: up 0.1%

black-parrot:
Hahaha -- that's awesome. But rather than enable clock gating for everybody who uses bsg_ip_cores, can you identify the modules in the ME that instantiate bsg_mem_1r1w and then add the parameter there? It may require that you do pull requests to bsg_ip_cores to add enable_clock_gating_p flags to higher up modules that instantiate bsg_mem_1r1w.v

derekcom17:
I know for ME most of those memories are in bsg_two_fifo, which is in bsg_mesh_router_buffered which are in the network part of the ME. 

Not a bad idea. Do you think I should add that enable_clock_gating parameter to every bsg module that includes a memory?

taylor-bsg:
Maybe just the modules that are used by ME.

black-parrot:
You can also post the absolute numbers before and after?

black-parrot:
Closing; the original pull request neutral benefits, but derek's clock gating of 1r1w idea sounds great and he should do a pull request based on our discussion.

PR #67, opened 2019-03-07
Modified the bp_be_int_alu.v file to remove case dependent computation of 32 bit and 64 bit operations

richarao:
- The code is simpler (has fewer lines).
 - Power consumption by alu (bp_be_int_alu_0) was 4.21e+03uW, and now it is 3.46e+03uW. (17.8% improvement)
 - Area is now 32801.59um^2, which is around 10% improvement from baseline(~36600um^2).
 - Performance: Running at 2.5ns(400MHz)

 - The overall area and power has not improved much as alu contributes very little to area and power.
(Overall area reduced by ~0.2% and overall power is almost the same)

### DIFF
--- a/bp_be/src/v/bp_be_calculator/bp_be_int_alu.v
+++ b/bp_be/src/v/bp_be_calculator/bp_be_int_alu.v
@@ -55,38 +55,28 @@ module bp_be_int_alu
 // Intermediate connections
 // These are signed because we're doing math on them, most of which is signed
 logic signed [reg_data_width_lp-1:0] src1_sgn   , src2_sgn;
-logic signed [word_width_lp-1:0]     src1_w_sgn , src2_w_sgn;
+//logic signed [word_width_lp-1:0]     src1_w_sgn , src2_w_sgn;
 logic signed [reg_data_width_lp-1:0] result_sgn ;
-logic signed [word_width_lp-1:0]     resultw_sgn;
+//logic signed [word_width_lp-1:0]     resultw_sgn;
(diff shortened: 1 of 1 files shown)


dpetrisko:
Does this pass the word width tests? I would be surprised...

taylor-bsg:
Yeah it seems like some fixup would need to be done on the 32bit results.

richarao:
@dpetrisko @taylor-bsg  I tried the towers, vvadd and median benchmarks and it works for them. 
I should probably try multiply too.

black-parrot:
@dpetrisko @stdavids Are there instructions for how Richa can run the instruction tests.

dpetrisko:
I wouldn’t be surprised if the benchmarks were all dword instructions. Try rv64ui_p_addiw and friends

dpetrisko:
Make run ROM_NAME=rv64ui_p_addiw_rom

richarao:
I tested using the ROM you suggested and after few trace matches, the simulation fails. It passes for the towers, vvadd and median benchmarks but not for this.

dpetrisko:
The RISC-V isa defines operations on 32 bit operands as opposed to 64 bit operands (Think uint32_t vs uint64_t in C).   The ISA tests that end in 'w' check that these instructions run correctly.  The benchmark tests do not use these instructions because they compute based on 64 bit values exclusively.

stdavids:
Closing based on failure to pass rv64ui_p_addiw_rom test.

PR #69, opened 2019-03-08
bsg_dff_reset_end moved and updated in bsg_ip_cores

tommydcjung:


dpetrisko:
If everything is fine with the new bsg_ip_cores, good to go

PR #72, opened 2019-03-09
Ydch be clockgating

cyd187:
We tried mem hardening and clock-gating at same time but it seems slow down the design but just the clock-gating seems really beneficial to the design in terms of both power and timing.


dpetrisko:
Please make the pull request into pparrot, not master

PR #51, opened 2019-03-06
PParrot BE critical path fixes

dpetrisko:
Original: Critical paths through npc_check, then through nop logic, then cache->cache path which was unnecessary because of cache miss poison logic.

Frequency: 400 MHz
Area: 1283284.900015

Median
IPC: 466
MIPS: 186.4 
Power: 1.1842e+05 uW
EPI: 635.3 pJ

Multiply
IPC: 488
MIPS: 195.2
Power: 1.1769e+05 uW
EPI: 602.9 pJ

Towers
IPC: 505
MIPS: 202
Power: 1.2628e+05 uW
EPI: 625.1 pJ

Vvadd:
IPC: 453
MIPS: 181.2
Power: 1.1553e+05 uW
 EPI: 637.6 pJ

———————————————
With fixes: critical path now through dcache->bypass paths->dispatch.  Additionally, a parameter (default off) is added to disable this load->use forwarding. Expect an IPC decrease if this parameter is enabled.

Frequency: 513 MHz
Area: 1287764.798082

Median
IPC: 454
Power: 1.1736e+05
MIPS: 232.9
EPI: 503.9

Multiply
IPC: 460
Power: 1.1724e+05
MIPS: 236.0
EPI: 496.8

Towers
IPC: 486
Power: 1.2545e+05
MIPS: 249.3
EPI: 503.2

Vvadd:
IPC: 443
Power: 1.1548e+05
MIPS: 227.3
EPI: 508.1

———————————————
Note: All results post-synthesis

PR #76, opened 2019-03-09
Balanced Pipeline stage in BE

iihihiuh:
This is a pull request merges Dan's changes in backend and #51. We can run our clock with 2.2 ns constrain. Benchmark and ISA tested.

PA after place and route

Performance wise: 390 Mhz -> 460 Mhz
Cycle counts are unaffected.
Before our changes: 2.5601 ns
After our changes: 2.176 ns
Improvement: 15 %
The path we optimized got 16.0 % up

Area:
Before : 1283418.5
After : 1301663.4474 um^2
Improvement: - 1.4 %.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,0 +1,3 @@
+By submitting code to this repository, you agree that 
+it will be subject to the open source license selected
+by Michael Taylor for the project.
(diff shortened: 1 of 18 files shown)


iihihiuh:
@taylor-bsg This pull request merges Dan's and our changes. I also incorporated Dan's minor changes. I believe this pull request is ready to merge.

PR #79, opened 2019-03-10
clk gating on mems

cyd187:
We tried to do clock gating on mems. The results show ~8% more in area and ~20% less in power.

clock speed: 366 Mhz -> 380 Mhz
area:1.21 mm^2 -> 1.31 mm^2
power:1.46e+05 uW -> 1.16+e05 uW


dpetrisko:
Hey -- it seems like you're having trouble making a pull request.  Could you talk to @stdavids about this before opening and closing another request?

PR #103, opened 2019-03-15
Serialization of Cache Line Transmission

chaseemory:
**Serialize Cache line transfers through network**

**Question**: The network as a whole uses up 67% of the area and 28% of the total area of the memory end. This is because the cache lines in the system are transferred all at once using massive, almost 600 bit wide, busses. Spread throughout the network are also many fifos used for the routing of the caches lines through the network. All of these giant busses take up a ton of area just in wires to connect all of the LCEs and CCEs together, not to mention the area used up by the fifos. This causes a great deal of congestion in the design of the network, which forces us to use a sparseness value in placement of standard cells of .33 or lower in order for routing to be successful, as determined above. A previous attempt to fix some of this congestion involved limiting the size of the fifos in the network, which helped limit the power used as well as removing some of the offending wiring that was required, but it was really just a bandaid over the actual problem, the movement of 512bit cache lines in a single cycle congesting the entire design.

**Hypothesis**: Serializing the transfer of the cache lines through the network should alleviate many of the congestion issues in the design. Limiting the bus widths for the cache line transfers to ~64bits will dramatically limit the amount of connecting wires throughout the network. This will incur a multi-cycle overhead to the transfer of cache lines, but the regained improvements to routing speed as well as cell density should be worth it.

**Notes**: Many ideas were considered to “fix” the network and serialize the transfers. The solutions we considered were to freeze a path through the network until all packets from a single cache line were transferred, using either the network or updates to the micro-code, or to let the network remain “free” and reassemble the packets on the receiving end. This problem arises because while the network guarantees that individual packets from any given source will arrive at the destination in the order they were sent, it does not guarantee that they will arrive consecutively at the receiver. The packets can become interspersed with packets from other sources, and you must have a way to reassemble the original cache lines being sent in order to give them to the final destination. The idea of constricting the network in order to make that above guarantee, seems to lack efficiency and would create too much of a bottleneck in the system. Letting the network route the packets as it seems fit will give rise to the more efficient behavior, but then the packets must be reassembled into the original cache lines. To do this a small scratchpad RAM will be used. It will be cache line width, and contain as many entries as we have sources for a given destination. This is to guarantee a lack of deadlock incase packets from every source were to be in flight to the destination at the same time. I’m sure there would be a statistical likelihood that a smaller number of entries would work the vast majority of the time, but we need to guarantee correct operation of the network in all situations.

**Experiment**: Design a serializer and de-serializer module to be placed at every entrance and exit to the network in order to limit the bitwidth of the busses, and therefore the area being taken up by all of the FIFOs and wires within the network. These modules will be designed to interface with the existing modules in Black Parrot seamlessly. This will provide a proof of concept of the idea of serializing the network as well as remove the burden of redesigning significant portions of the processor to facilitate the serialization. 

**Results**: Our first attempt at the serialization involved packetizing all of the data busses down to 64bits which seemed like a solid first attempt at a compromise between area/power and performance.

Post Synth |   |  
-- | -- | --
Power p(J/I) | Performance M(I/s) | Area (cell only) (mm^2)
2,601.69 | 60.2 | 2.323
-9.07 | -2.54 | -5.94

Only post synthesis numbers are provided. These numbers are a bit of a letdown to the goals we had for power and are improvements. It turned out that the area we used for all of the scratch pads nearly balanced out the area that was previously used by the FIFOs inside the network. 

There are several reasons we see for this behavior. One could be the location we have available to us to put the serializers. The optimal location for them would be within the sources themselves to limit the places the massive ~600bit wires have to go. For us, in this project, we are limited to working on the Memory End of the processor, so even if we put them on the inputs, we still have multiple ~600bit busses coming in as ports on the design. Another is that because we are only instantiating a few cores, there are not enough sources in order to facilitate hardening the scratchpad RAMs, they remain synthesized memories, which causes quite a bit of wiring congestions on their own.

While I was a little let down by the small improvements to power and cell area of the synthesized design, I was sure we would still be able to makeup even more by being able to vastly increase the placement density of the design now with the dramatically smaller wiring footprint. This turned out to be not the case, even .35 density does not pass without DRCs, so we saw no improvement to density versus the baseline design.

**Results 2**: Our second attempt was to use 16bit packets over the network to first see what gains would be seen, and also to see if the 16bit busses would alleviate the routing problems that continued to exist with the 64bit versions above.

Post Synth |   |  
-- | -- | --
Power p(J/I) | Performance M(I/s) | Area (cell only) (mm^2)
2,517.15 | 59.2 | 2.278
-12.03 | -4.09 | -7.75

The power and area improvements do not seem to be worth it with the 16bit packets, versus the even further drop in performance that resulted. 

Another downside to the 16bit design is that the place and route problems we saw above were not alleviated. Further effort was put into finding the density required to actually place and route at 16bits, but even at .33, which was determined to be the max density supported by the baseline in this flow, the backend took ~9hrs and there were 17k DRCs in the finished chip, which is worse than the baseline as well.

**Takeaways**: The idea that we wanted to change the network to alleviate the routing issues in the baseline is a good one. This implementation has several flaws though. We underestimated the impact of the scratchpad memories on the size of the design. I also believe that the scratchpad memories are what is causing the routing problems in the “fixed” network. If the design were to be scaled up to 4 or more cores, the memories could be hardened, rather than synthesized from standard cells, which would help alleviate much of the routing congestion.
    
Another consideration is the PPA of the overall design. Power and are improvements are considerable, but as anticipated the performance dropped by serializing the data transfers of cache lines over the network. Another thing to consider is that many improvements were made to the front end throughout the final project. The baseline performance of the front end limits the mIPC to around 150, so our performance downgrade is small. The improvements to the front end increased the mIPC to somewhere around 800. I think the performance degradation of the serialized network could increase, even as a percentage, if the mIPC baseline were improved to that degree.

This implementation was also plagued because we were restricted to working within the Memory End module. It would be more efficient to place the serialization at the sources, to limit the number of locations that need the massive ~600bit wide busses. We still had the large busses as top level ports in our design requiring a great deal of routing just from there to the serializers.


chaseemory:
A change is also required on bsg_mem_1r1w_sync_mask_write_var to get this to work. I edited the input mask port to only be as wide as there are masking regions, rather than being as wide as the data input was. It is in the fixed-mask-var-mem branch of the bsg_ip_cores repo we are using for the lab

PR #105, opened 2019-03-16
BTB Memory Clock Gating

shadix888:
From a high level view, not every instruction requires us to update the branch related memory since not every instruction is a new branch, and the cycles where we don’t have to update branch related memory are cycles where we can save power. We think that we can reduce the power usage by implementing clock gating logic for the branch target buffer memory.

Here is our baseline

BASELINE INFO | Instr/S | J/Instr | Area (um^2)
-- | -- | -- | --
towers_rom | 1.7172e8 | 4.2744e-10 | 1333894.888021
vvadd_rom | 2.6003e8 | 2.7574-10 | 1333894.888021
multiply_rom | 1.6439e8 | 4.5441e-10 | 1333894.888021
median_rom | 2.2896e8 | 3.2102e-10 | 1333894.888021
average | 2.06275e8 | 3.696525e-10 | 1333894.888021

Here is our ppa after the implementation

BASELINEINFO | Instr/S | J/Instr | Area (um^2)
-- | -- | -- | --
towers_rom | 1.7627e8 | 3.9542e-10 | 1339819.217655
vvadd_rom | 2.6692e8 | 2.4801e-10 | 1339819.217655
multiply_rom | 1.6875e8 | 4.2012e-10 | 1339819.217655
median_rom | 2.3503e8 | 2.9230e-10 | 1339819.217655
average | 2.1174e8 | 3.3896e-10 | 1339819.217655

There is a 2.6494% increase in Instr/S
There is a 8.3031% decrease in J/Instr
The cost of this optimization is a 0.4441% increase in chip area, which is relatively negligible.
We think this optimization is pretty much free ppa gains and is worth pulling.

### DIFF
--- a/bp_fe/src/v/bp_fe_btb.v
+++ b/bp_fe/src/v/bp_fe_btb.v
@@ -49,6 +49,7 @@ bsg_mem_1r1w
  #(.width_p(eaddr_width_p)
    ,.els_p(2**bp_fe_pc_gen_btb_idx_width_lp)
    ,.addr_width_lp(bp_fe_pc_gen_btb_idx_width_lp)
+   ,.enable_clock_gating_p(1'b1)
    ) 
  bsg_mem_1rw_sync_synth_1 
   (.w_clk_i(clk_i)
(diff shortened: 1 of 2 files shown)


shadix888:
@stdavids for pull request review, can't add as a reviewer so mentioning in a comment

Issue #107, opened 2019-03-16
[CLARITY] Should not be forced to source setup_env.sh to run Makefile infrastructure


stdavids:
In general, we should avoid using sh scripts in favor of makefile to be consistent with the rest of our infrastructure code. But we should especially avoid requiring that a sh script be sourced for the makefiles to work. Common environment variables should be put in a common location for makefiles to include.

I personally find it cumbersome to have to source a script every time I login or open a new shell... but beyond that, sourcing is also shell specific unlike makefiles (ie. setup_env.sh only works with bash, it won't work with other common shells like csh). The other thing is that sourcing has this really bad persistence issue when developing infrastructure. When writing infrastructure code you basically want to make as few assumptions about the environmental state of the machine as you can. But when sourcing is part of your setup it become real tedious to get back to that near-0 assumption state (you need to close your shell, open a new one and re-source). If you don't then you can get into really silly issues... for example: if you want to change the name of a variable in setup_env.sh, if you don't shutdown your shell then when you re-source the script both the old and new variables will exist in your environment which can hide dependencies on the old name that were not updated.

dpetrisko:
Sounds good.  I originally did it because I couldn't figure out how to get flists to respect environment variables (but I guess you can just .export from Makefiles).  I'll refactor into a Makefile.common that sets the environment variables

dpetrisko:
(Probably should do this for the regression scripts, too, although I'm kind of waiting to see how CI will interact with them)

dpetrisko:
Fixed in 82dda501d5dd12a8d3d31cd1bc7302058693e950

PR #108, opened 2019-03-16
Reorganizing Makefiles

dpetrisko:
This PR targets #107 

Major changes are
1) Get rid of setup_env.sh and replace with Makefile.common at top level
2) Put Makefile.dc, Makefile.vcs, Makefile.verilator in bp_common/syn and referenced from other Ends
(Only BE is done for now)
3) Redirecting logs to bp_\<end\>/syn/logs/\<tool\>/\<tb\>_\<operation\>.log

Still needs to be done
1) Unify all End makefiles
2) change regress scripts to be makefile targets (need to think about how this interacts with CI)
3) Comment all of the makefiles
4) add a default make help target


dpetrisko:
There are also some synthesis fixes embedded in here, which I ran into when running check_synth with the new flow

bp_top/syn/Makefile
```diff
@@ -1,3 +1,8 @@
+## Set common environment variables
+TOP ?= $(shell git rev-parse --show-toplevel)
+
+include $(TOP)/Makefile.env
```
stdavids: Makefile.common?

stdavids:
Woah, cool git command to get the top level directory, very clever! That will likely be brought to our other infrastructure code :)

One line looked like a bug to me, other than that it looks good to me!

bp_top/syn/Makefile
dpetrisko: good catch!

dpetrisko:
@muwyse Could you take a look at this branch? I redid the ME makefiles a bit and I was able to build your bp_me_top_mesi TB, but it was erroring out when I ran it.

"Error: (E109) complete binding failed: port not bound: port 'DUT.port_12' (sc_out)
In file: communication/sc_port.cpp:235"

`make run.sc TB=bp_me_top_mesi` is the command


AFAICT, I kept all of the parameters and verilator flags the same between the builds.  I left your old Makefile and Makefile.verilator as Makefile.old and Makefile.verilator.old for reference

muwyse:
@dpetrisko This error is due to not having updated the ME systemc test bench. It is trying to bind a signal to a port that no longer exists. You can ignore this error for the moment and merge, then I will fixup.

dpetrisko:
Merged into dev instead of be_dev

PR #112, opened 2019-03-21
PC Gen PPA improvement

dpetrisko:
In EE477, we determined that using a tagged BTB resulted in 1.3X performance improvements, 2.5X energy efficiency improvement, 1.23X area decrease.  

For this class project, we remove the first stage of pc_gen, removed the BHT completely and inserted the BTB at the first stage of the pc_gen pipeline.  Pipeline changes can be seen here:

https://docs.google.com/presentation/d/147kLmlUw3yNTlgcXGe3MajmAJPjpB9kOL_5kUlKSWLo/edit#slide=id.g5078081ba4_0_0

This reduced the 2 cycle branch taken penalty to 1 cycle with BTB miss and 0 cycle with BTB hit.  Additionally, it reduced the branch misprediction penalty and cache miss penalties by 1. Last, we flush bad speculation on redirects so that we do not send known bad speculative instructions to the Back End.  More in-depth architectural analysis can be seen here:

https://docs.google.com/presentation/d/1Rid1otSmzvT3IQaati25I-MyeChhKwY0K6PHmCZTyHw/edit#slide=id.g544b39038c_2_2

Additional improvements that will need before merging:

1. Re-add BHT and use it to override BTB predictions e.g. predict not taken when target is in BTB
2. Write to BTB on branch taken only to save space. Currently, we write on not-taken branches as well.
3. BTB is only written on redirects at the moment. We should also add calculated branch targets on predicted taken branches which are not in the BTB. This may get complicated because of the 1RW BTB memory.  Hopefully, there is a solution which does not require dropping conflicting BTB writes.


HelloKayT:
Non-technical nit: there are backend changes included in this pull request. Is there a way to drop them?

dpetrisko:
That’s a good point. I would like to drop them since they’re mostly already incorporated, but I’m not sure if there’s a clean way. I feel like there might be through an interactive rebase

bp_fe/src/v/bp_fe_pc_gen.v
```diff
@@ -131,111 +155,137 @@ always_comb
...
+        pc_n = pc_f2;
     end
-  else if (prediction_on_p && predict && icache_pc_gen_v_i && (scan_instr.instr_scan_class == e_rvi_branch))
+    else if (btb_br_tgt_v_lo)
     begin
-      next_pc = icache_pc_gen.addr + scan_instr.imm; 
+        pc_n = btb_br_tgt_lo;
     end
-  else if (icache_pc_gen_v_i && (scan_instr.instr_scan_class == e_rvi_jal))
+    else if (predict_taken)
```
HelloKayT: What happens if we've already predicted this branch from the branch target buffer? Is it possible we accidentally kill the branch we just predicted?

PR #127, opened 2019-04-12
Adding VM support to BE dev

dpetrisko:


dpetrisko:
Can you do an interactive rebase to combine commits 1875344ec61165cda501aac2d7335b56031db1fa - fb5f07a4a59c6443d78ecc8e17c1df7c0b4ae217 into a few larger ones?

bp_be/src/include/bp_be_internal_if_defines.vh
```diff
@@ -104,8 +106,13 @@
     logic                                   mem3_v;                                                \
     logic [rv64_eaddr_width_gp-1:0]         mem3_pc;                                               \
     logic                                   mem3_cache_miss_v;                                     \
+    logic                                   mem2_tlb_miss_v;                                       \
```
dpetrisko: Don't need this -- there's no advantage to poisoning in mem2, but it increases complexity.  A single poison point at mem3 for memory exceptions is better

bp_be/src/include/bp_be_ctl_defines.vh
```diff
@@ -132,6 +134,9 @@ typedef struct packed
   logic load_fault_v;
   logic store_fault_v;
   logic cache_miss_v;
+  logic itlb_fill_v;
+  
+  logic [rv64_eaddr_width_gp-1:0] pc;
```
dpetrisko: PC is already passed in instr_metadata, take it out of the exception pipe

bp_be/src/include/bp_be_internal_if_defines.vh
```diff
@@ -104,8 +106,13 @@
...
+    logic                                   mem3_tlb_miss_v;                                       \
     logic                                   mem3_exception_v;                                      \
     logic                                   mem3_ret_v;                                            \
+    logic                                   mem3_itlb_fill_v;                                      \
```
dpetrisko: These should go directly from bp_be_mem to bp_be_director, not through the calculator

bp_be/src/include/bp_be_vm_defines.vh
```diff
@@ -4,31 +4,31 @@
 
 `define declare_bp_sv39_pte_s(pte_width_mp, ppn_width_mp) \
   typedef struct packed {                                 \
+    logic [pte_width_mp-10-ppn_width_mp-1:0] reserved;    \
```
dpetrisko: I believe this is defined in bp_common fe_be_if already

bp_be/src/v/bp_be_calculator/bp_be_calculator_top.v
```diff
@@ -405,8 +428,8 @@ bp_be_pipe_mem
    ,.rs2_i(dispatch_pkt_r.rs2)
    ,.imm_i(dispatch_pkt_r.imm)
 
-   ,.mmu_cmd_o(mmu_cmd_o)
-   ,.mmu_cmd_v_o(mmu_cmd_v_o)
+   ,.mmu_cmd_o(mmu_cmd)
```
dpetrisko: The itlb fill shouldn't come from the instr_decoder

bp_be/src/v/bp_be_calculator/bp_be_calculator_top.v
```diff
@@ -598,9 +624,16 @@ always_comb
     calc_status.mem3_pc           = calc_stage_r[2].instr_metadata.pc;
     // We don't want cache_miss itself to trigger the exception invalidation
     calc_status.mem3_cache_miss_v = cache_miss_mem3 & calc_stage_r[2].pipe_mem_v & ~exc_stage_r[2].poison_v; 
+    calc_status.mem2_tlb_miss_v   = exc_stage_n[2].tlb_miss_v & calc_stage_r[1].pipe_mem_v & ~exc_stage_n[2].poison_v;
+    calc_status.mem3_tlb_miss_v   = exc_stage_r[2].tlb_miss_v & calc_stage_r[2].pipe_mem_v & ~exc_stage_r[2].poison_v;
     calc_status.mem3_exception_v  = (illegal_csr_mem3 | exc_stage_r[2].illegal_instr_v) & calc_stage_r[2].pipe_mem_v & ~exc_stage_r[2].poison_v & ~exc_stage_r[2].roll_v;
     calc_status.mem3_ret_v        = exc_stage_r[2].ret_instr_v & ~exc_stage_r[2].poison_v;
     calc_status.instr_cmt_v       = calc_stage_r[2].instr_v & ~exc_stage_r[2].roll_v;
+
```
dpetrisko: Don't need

bp_be/src/v/bp_be_checker/bp_be_detector.v
```diff
@@ -178,29 +178,41 @@ always_comb
...
-assign chk_dispatch_v_o = ~(data_haz_v | struct_haz_v);
-assign chk_roll_o       = calc_status.mem3_cache_miss_v;
+assign chk_dispatch_v_o = ~(data_haz_v | struct_haz_v); 
+assign chk_roll_o       = calc_status.mem3_cache_miss_v
+                          | calc_status.mem3_tlb_miss_v;
+                          
 assign chk_poison_isd_o = reset_i
                           | calc_status.mem3_cache_miss_v
-                          | calc_status.mem3_exception_v;
+                          | calc_status.mem2_tlb_miss_v
```
dpetrisko: Poison just from mem3

bp_be/src/v/bp_be_checker/bp_be_director.v
```diff
@@ -290,6 +291,14 @@ always_comb
 
         fe_cmd_v = fe_cmd_ready_i & ~chk_roll_fe_o & ~redirect_pending;
       end
+    else if(calc_status.mem3_itlb_fill_v)
+      begin
+        fe_cmd.opcode = e_op_itlb_fill_response;
+        fe_cmd.operands.itlb_fill_response.vaddr = calc_status.mem3_itlb_fill_vaddr;
```
dpetrisko: Should come directly from bp_be_mem

bp_common/src/include/bp_common_aviary_pkg.vh
```diff
@@ -15,7 +15,7 @@ package bp_common_aviary_pkg;
       ,num_lce: 1
 
       ,vaddr_width: 39
-      ,paddr_width: 39
+      ,paddr_width: 56
```
dpetrisko: 39 is correct.  We only support 39 bits of physical address.  Please revert.

bp_be/src/v/bp_be_calculator/bp_be_instr_decoder.v
```diff
@@ -33,28 +33,47 @@
...
 
-   , output [decode_width_lp-1:0] decode_o
-   , output                       illegal_instr_o
-   , output                       ret_instr_o
-   , output                       csr_instr_o
+   , output [decode_width_lp-1:0]        decode_o
+   , output                              illegal_instr_o
+   , output                              ret_instr_o
+   , output                              csr_instr_o
+   , output                              itlb_fill_exc_o
```
dpetrisko: The exception logic should be external to instr_decoder.  If fe_exception (in instr_metadata), then insert a nop.  Look at how fe_nop, be_nop and friends are inserted.

bp_be/src/include/bp_be_internal_if_defines.vh
farzamgl: The tlb miss happens at mem2, which might cause problems if we do not poison the instruction at mem1. Because it can be a load/store instruction which will access the tlb and lower the tlb miss signal.

bp_common/src/include/bp_common_aviary_pkg.vh
farzamgl: Can't we just leave the address at 56 bits according to match the spec and just use the lower 39 bits of it?

bp_be/src/include/bp_be_vm_defines.vh
farzamgl: This is the pte as is saved in the memory. I didn't find the definition in common files. Can you point to the location if you find it?

bp_be/src/include/bp_be_internal_if_defines.vh
dpetrisko: I think there's a cleaner way to deal with this, but it's lower priority than the other fixes

bp_common/src/include/bp_common_aviary_pkg.vh
dpetrisko: No, this is the parameter which defines all of our internal structs, which are 39 bits.

dpetrisko: Anywhere you're reading from software (your PTW) should use the 56 bit rv64 definition and truncate to 39

bp_be/src/include/bp_be_vm_defines.vh
dpetrisko: Oh yeah, you're right

bp_be/src/include/bp_be_ctl_defines.vh
farzamgl: The mem_resp outputs the itlb fill results(aside from the tlb entry) in an exception struct. I added the pc section to pass the fill vaddr to FE

dpetrisko: You can still pass it to FE.  But it's already propagating down the pipe, we don't need to duplicate it

dpetrisko: Actually, we don't need to pass the vaddr in the itlb fill result struct at all, then

dpetrisko: Wait no, I'm mistaken.  This is the result of the fill, not the request.  But still, don't put it in the exception pipe. Directly output the fill result to the director

dpetrisko: The exception pipe is for exceptions that occur during the execution of an instruction. A TLB fill response is not an exception, but a request is

farzamgl: So just add separate output ports for the itlb fill response in the bp_be_mem_top module?

dpetrisko: Yep

bp_be/src/include/bp_be_internal_if_defines.vh
dpetrisko: I thought about it.  As long as the tlb_miss goes into the exception pipe at mem2, then it's fine.  The following instruction doesn't matter

dpetrisko: This way is better, because the TLB lookup is a long path, and if it goes to the poison logic it will most likely become the critical path.

bp_be/src/include/bp_be_internal_if_defines.vh
```diff
@@ -94,6 +94,8 @@
     logic                                    ex1_v;                                                \
     logic [rv64_eaddr_width_gp-1:0]          ex1_pc;                                               \
                                                                                                    \
+    logic                                    fe_exception_v;                                       \
```
dpetrisko: Don't need this either.  The line in detector should be 
    // Detect misprediction
    mispredict_v = (calc_status.ex1_pc != expected_npc_i);

farzamgl:
That would result in mispredict and poison signals being x on some cycles
that the instruction pc is not actually valid.

On Apr 12, 2019 9:41 PM, "Dan Petrisko" <notifications@github.com> wrote:

*@dpetrisko* commented on this pull request.
------------------------------

In bp_be/src/include/bp_be_internal_if_defines.vh
<https://github.com/black-parrot/pre-alpha-release/pull/127#discussion_r275104545>
:

> @@ -94,6 +94,8 @@
     logic                                    ex1_v;
                             \
     logic [rv64_eaddr_width_gp-1:0]          ex1_pc;
                             \

                             \
+    logic                                    fe_exception_v;
                             \

Don't need this either. The line in detector should be
// Detect misprediction
mispredict_v = (calc_status.ex1_pc != expected_npc_i);

—
You are receiving this because you were mentioned.
Reply to this email directly, view it on GitHub
<https://github.com/black-parrot/pre-alpha-release/pull/127#pullrequestreview-226338725>,
or mute the thread

dpetrisko:
It doesn't, because nops are also associated with PCs.  That's how the code is in my branch of be_dev and it works fine

farzamgl:
I remember seeing X valued PCs in the pipe, I'll look for the cause.

On Apr 12, 2019 10:08 PM, "Dan Petrisko" <notifications@github.com> wrote:

It doesn't, because nops are also associated with PCs. That's how the code
is in my branch of be_dev and it works fine

—
You are receiving this because you were mentioned.
Reply to this email directly, view it on GitHub
<https://github.com/black-parrot/pre-alpha-release/pull/127#issuecomment-482777096>,
or mute the thread

dpetrisko:
Also, please grep for tabs.  There are tabs in bp_be_vm_defines, so there are probably more

Issue #135, opened 2019-04-20
Add DRAMSim2 Support to our test memory


dpetrisko:
This will enable a more accurate performance regression. The most likely path forward is through SV-DPI calls into the libdramsim.so library. This will also allow us to support higher ranges than VCS, as the memory will be dynamically allocated in C++

dpetrisko:
Live in dev

PR #136, opened 2019-04-20
Adding DRAMSim2 support to test memory model

dpetrisko:
This PR adds DRAMSim2 support to our top level testbench, in support of issue #135 

This works by compiling DRAMSim2 as a static library and linking with the VCS build, using a C++ wrapper with SV-DPI calls to interface the two models. The simulated DRAM memory is dynamically allocated in C++, not Verilog. I tested with the 4 benchmarks we have and it passed. Furthermore, if I increased or decreased the clock period, I got a corresponding IPC decrease / increase.

There is still much work to be done on this PR, but I'd like to get some early feedback
1) Change the testing flow to generate hex memories instead of boot roms.
2) Support uncached requests.
3) Figure out the time conversion between the DRAM model and VCS clock period.
4) Support Verilator
5) General code cleanup


dpetrisko:
6. Add an id field so that we can have multiple mem modules interact with the dramsim model

dpetrisko:
> Variables should not end in _s. It’s just not right man!

I agree. What do you suggest for casted input structs? We have a few different approaches in BlackParrot and we should probably standardize
1) foo_i_s
2) foo (drop the port direction)
3) foo_cast_i

taylor-bsg:
3

dpetrisko:
> 3

We should definitely include in BlackParrot style guide since it's so common.  Possible not common enough to justify going in BSG style? (I'm thinking of BaseJump STL, although I don't know the manycore code)

> Very cool! Going to suggest something crazy here — dramsim2 interface should be genericized a little and put into BaseJump STL as a _nonsynth module!

Agreed! Getting the basic version up and running has priority, of course. One UX concern is that the building + linking the dramsim2 library / DPI headers is not intuitive (and also heavily simulator dependent).  We'll have to write an in-depth README tutorial.

As an aside, based on our struggles with readmemh ranges, elf2hex formatting problems, etc., just having the memory model in C++ is a win. I'm also not sure of the (simulation speed) performance impact of DRAMSim2.  Having a fixed latency variant might also be valuable (and avoid needing to build + link DRAMSim2).

> DRAMs usually support multiple channels and some degree of concurrency. Is anything lost in our interface?

I can imagine the enhancements necessary (I think just request id to sort out parallel requests and request size to split requests based on channel width).  DRAMSim2 has an extremely intuitive interface for parallel requests, so we would basically convert the current bp_mem_dramsim2.v to have multiple channels with very few changes to the .cpp wrapper

taylor-bsg:
Yep, I am curious how much slower it is when using dramsim2, even when not
accessing the ram, and whether execution time in verilator and vcs changes
significantly as dram timings are worsened. In generally, adding DRAM miss
latency means greater numbers of machine cycles, which means more work for
the simulator. But if the core is stalled and nothing is toggling, then
maybe it doesn't slow down so much.  I think we would want to have an
option to pass in smaller latencies for dramsim2 when we don't care about
timing accuracy. But the alternative of just using SystemVerilog
associative arrays in this use case is also appealing, although it does
through Verilator under the bus for now.

We will be using your code pretty soon in HammerBlade simulation  =),
that's partly why I suggest genericizing the interface (mostly so it does
not refer to CCE/LCE, and so width is parameterized.) The complexity of
integration does give me pause though about whether it should go in STL or
another library. I have been thinking about having something that is one
level up from STL and focuses more on microarchitectural widgets.

M

dpetrisko:
Merging to enable others to run regression. Will genericize interface for BaseJump STL offline.

PR #142, opened 2019-04-22
Be dev feature isavmtests

farzamgl:
This is a first attempt to verify the functionality of the Virtual Memory behavior. The riscv vm tests have been modified to consider the things that have not been implemented in our hardware yet (like S mode csr emulations and VM fence instructions).
Also for simulation, I'm relying on self-verification of the isa programs instead of trace replay because generating traces that will match with the actual hardware is going to be difficult if we take into account that some parts of the hardware are going to be emulated.


bp_be/test/common/bp_be_nonsynth_tracer.v
```diff
@@ -88,9 +88,6 @@ wire [reg_data_width_lp-1:0] unused1 = fwb_result_i;
        ,.data_o(iwb_br_tgt_r)
        );
 
-    integer file;
```
dpetrisko: Revert.  I want this to log to a file.

bp_common/syn/Makefile.regress
```diff
@@ -57,6 +109,7 @@ RV64_BENCHMARKS := \
                    vvadd
 
 RV64_ISA_TEST_ROMS   := $(addsuffix _rom, $(RV64_ISA_TESTS))
+RV64_ISA_V_TEST_ROMS := $(addsuffix _rom, $(RV64_V_ISA_TESTS))
 RV64_BENCHMARK_ROMS  := $(addsuffix _rom, $(RV64_BENCHMARKS))
-RV64_REGRESSION_ROMS := $(RV64_ISA_TEST_ROMS) $(RV64_BENCHMARK_ROMS) 
+RV64_REGRESSION_ROMS := $(RV64_ISA_TEST_ROMS) $(RV64_ISA_V_TEST_ROMS)
```
dpetrisko: We still need the benchmarks in regression

bp_top/test/rom/src/env/p/riscv_test.h
```diff
@@ -182,34 +182,11 @@ reset_vector:                                                           \
...
-        or a0, a0, a1;                                                  \
-        sw a0, 0(t1);                                                   \
-                                                                        \
-        fence;                                                          \
-1:      beqz TESTNUM, 1b;                                               \
-        sll TESTNUM, TESTNUM, 1;                                        \
-        or TESTNUM, TESTNUM, 1;                                         \
-        ecall
 
+#define RVTEST_PASS li t0, 0x0000beef; ecall
```
dpetrisko: Hmm, we don't have the infrastructure to support this ecall yet.  I would keep the old macros for now

bp_top/test/tb/bp_top_trace_demo/testbench.v
```diff
@@ -194,7 +197,7 @@ logic [num_cce_p-1:0] mem_data_cmd_v, mem_data_cmd_yumi;
             trace_replay
              (.clk_i(clk_i)
               ,.reset_i(reset_i)
-              ,.en_i(1'b1)
```
dpetrisko: We don't want to kill trace replay until the whole regression suite (including benchmarks) is supported in the new infastructure.

bp_be/src/include/bp_be_pkg.vh
```diff
@@ -9,7 +9,7 @@ package bp_be_pkg;
   `include "bp_be_internal_if_defines.vh"
   `include "bp_be_vm_defines.vh"
 
-  localparam bp_pc_entry_point_gp     = 39'h00_8000_0124;
+  localparam bp_pc_entry_point_gp     = 39'h00_8000_0000;
```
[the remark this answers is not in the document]
farzamgl: The old and new isa tests are working. Haven't worked on the benchmarks yet. I think we need to rewrite them based on the VM framework the we have for isa tests.

bp_be/src/v/bp_be_checker/bp_be_director.v
```diff
@@ -241,7 +241,7 @@ bsg_dff_reset_en
    );
 
 // Generate control signals
-assign expected_npc_o = npc_r;
+assign expected_npc_o = (trap_v_i)? npc_n : npc_r;
```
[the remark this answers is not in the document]
farzamgl: When trap_v_i is high the npc_r register is being written. On the next cycle, npc_r updates and trap_v_i goes down.

bp_be/src/v/bp_be_mem/bp_be_csr.v
```diff
@@ -470,7 +471,7 @@ always_comb
 assign mepc_o           = mepc_r;
 assign mtvec_o          = mtvec_r;
 assign satp_o           = satp_r;
-assign translation_en_o = (priv_mode_r < `RV64_PRIV_MODE_M) & (satp_r.mode == 4'h8);
+assign translation_en_o = (priv_mode_r < `RV64_PRIV_MODE_M) & (satp_r.mode == 1'b1);
```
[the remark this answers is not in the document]
farzamgl: I know but the bp struct you have defined has only 1 bit for mode.

bp_common/syn/Makefile.regress
farzamgl: I removed this for now because I wasn't testing the benchmarks yet, will revert.

bp_be/test/common/bp_be_nonsynth_tracer.v
farzamgl: I needed this to display in order to be picked up by the regression Makefile, all these testing files will be changed once we pinpoint how we want to do the simulation.

bp_top/test/rom/src/env/p/riscv_test.h
farzamgl: Yeah that call doesn't do anything, the tracegen stops the program on the register write

bp_be/src/v/bp_be_mem/bp_be_mem_top.v
```diff
@@ -176,9 +176,38 @@ logic                     csr_v_lo;
...
+logic dtlb_miss_r;
+logic [vaddr_width_p-1:0] vaddr_r, vaddr_r2, fault_vaddr;
+logic [dword_width_p-1:0] fault_pc;
+
+assign itlb_fill_cmd_v = mmu_cmd_v_i & (mmu_cmd.mem_op == e_ptw_i);
+assign dtlb_fill_cmd_v = mmu_cmd_v_i & (mmu_cmd.mem_op == e_ptw_l 
+                                        | mmu_cmd.mem_op == e_ptw_s);
+assign store_not_load = mmu_cmd_v_i & (mmu_cmd.mem_op == e_ptw_s);
+
+always_ff @(posedge clk_i) begin
```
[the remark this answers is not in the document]
farzamgl: This signal is to pass the address accessed at MEM1 to MEM3. Which is the input to the PTW.

bp_be/src/include/bp_be_pkg.vh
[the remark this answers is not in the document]
dpetrisko: we can leave them with translation off (for now), but they need to be part of the regression

bp_be/src/v/bp_be_mem/bp_be_mem_top.v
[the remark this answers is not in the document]
dpetrisko: Oh, I see. I got lost and thought I was in PTW

bp_be/test/common/bp_be_nonsynth_tracer.v
dpetrisko: Gotcha, add a $display("PAS"), but leave the rest

bp_be/src/v/bp_be_checker/bp_be_director.v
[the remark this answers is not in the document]
dpetrisko: The expected npc is used for poisoning though.  We're poisoning during a trap anyway, so we don't need this

bp_be/src/v/bp_be_checker/bp_be_director.v
```diff
@@ -288,7 +288,7 @@ always_comb
         fe_cmd_v = fe_cmd_ready_i;
       end
     // Redirect the pc if there's an NPC mismatch
-    else if(calc_status.ex1_instr_v & npc_mismatch_v) 
+    else if((calc_status.ex1_instr_v | trap_v_i) & npc_mismatch_v)
```
dpetrisko: (calc_status.ex1_instr_v & npc_mismatch_v) | trap_v_i.  We want to redirect on a trap even if we somehow match, due to potential privilege mode changes

bp_be/src/v/bp_be_mem/bp_be_csr.v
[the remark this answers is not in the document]
dpetrisko: Oh, good call. We should add a comment for that

bp_be/test/common/bp_be_nonsynth_tracer.v
dpetrisko: Really, the magic store addresses should be in MM ME-side IO enclave and not in this unit

bp_be/src/v/bp_be_checker/bp_be_director.v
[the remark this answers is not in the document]
farzamgl: It's also used as the redirect PC that's being sent to FE

bp_be/src/v/bp_be_calculator/bp_be_calculator_top.v
```diff
@@ -143,6 +143,7 @@ module bp_be_calculator_top
   // CSRs
   , output                               instret_o
   , output [vaddr_width_p-1:0]           exception_pc_o
+  , output [vaddr_width_p-1:0]           exception_vaddr_o
```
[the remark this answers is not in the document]
farzamgl: I think fault_vaddr is a good name for this one.
As for the vaddr field mem_resp, it cna be set to either PTW's fault vaddr, or simply the accessed vaddr at MEM3(for triggering the PTW).

bp_be/src/v/bp_be_checker/bp_be_director.v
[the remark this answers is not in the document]
dpetrisko: Oh, I see. Let’s do the mux down there at the redirect pc instead of the expected npc

bp_be/src/v/bp_be_calculator/bp_be_calculator_top.v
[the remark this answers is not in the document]
dpetrisko: Sounds good!

farzamgl:
@dpetrisko @zaazad @spikemar 
There are some changes in the FE here that have been made in order to pas the isa-v tests(mostly in pcgen). I think we should consider these changes when we are doing our rework of those FE modules.

bp_be/src/v/bp_be_checker/bp_be_director.v
```diff
@@ -222,7 +223,7 @@ bsg_dff_reset_en
  redirect_pending_reg
   (.clk_i(clk_i)
    ,.reset_i(reset_i)
-   ,.en_i(calc_status.ex1_instr_v | trap_v_i)
```
dpetrisko: Why is this now ex1_v? The sfence should be coming in at mem3

farzamgl: It's not related to sfence actually. 
Sometimes the last instruction in the FE queue could be a wrong exception (because of a wrong prediction on FE) instead of a wrong instruction.

farzamgl: There was a small change to the FE queue's dequeue signal which made this bug visible.
Before that change, we actually had extra elements in the queue.

dpetrisko:
Looks fine to me.  Small thing, but change RV64_TESTS and RV64_V_TESTS to be RV64_P_TESTS and RV64_V_TESTS

dpetrisko:
Yep, you’re right. I realized that right after commenting

PR #152, opened 2019-04-30
Minor synthesis fixes. Added warning for inferred latches to dc Makefile

dpetrisko:


stdavids:
Hey Dan, fixes looks good to me!

I just changed the warning checks by setting a flag in DC that will force elaboration to throw a warning specifically about whenever a latch is inferred and grepped for that warning instead. Might be slightly less prone to false passes.

Issue #123, opened 2019-04-10
reorg module hierarchy so bp_core includes network routers.


tommydcjung:

dpetrisko:
Which routers is this referring to?

tommydcjung:
talking about the routers inside the me networks.

dpetrisko:
What’s the motivation to move them to core?

tommydcjung:
we want to have BE+FE+CCE+routers as one hierarchy node in physical design, so we can stamp them out in tiled fashion. Just like we include router in manycore tile.

dpetrisko:
But that’s bp_top. What is in ME in this organization?

taylor-bsg:
bp_top can instantiate bp_tile; the network links will be the main
interface for bp_tile.

dpetrisko:
Wait, what’s bp_tile now?

taylor-bsg:
The replicated unit of the multicore; contains 1 CCE, 2 routers, 1 FE, 1 BE.

dpetrisko:
So bp_core becomes bp_tile? Or bp_tile instantiates bp_core and bp_me_top?

taylor-bsg:
bp_tile could instantiate two routers, one bp_core, and one CCE.

PR #153, opened 2019-05-03
Attempt at tiling BP

dpetrisko:
This is an attempt to reorganize BlackParrot as in #123 

Initially, I'm tiling with full networks and will replace with wormhole version of data_cmd and data_resp once this version passes RTL.

Currently, this is broken. There are Xs propagating. I will dig into it when I get a chance, but let me know if anything looks weird to you in bp_tile.v or bp_top_tiled.v


taylor-bsg:
Don’t forget to tie down the signals on the edges.

dpetrisko:
I thought I tied down everything on the edges...  The X's do seem to be coming from the E-W I$ D$ connection, so it's possible that I messed something up there

taylor-bsg:
Talk to Paul..

M

PR #156, opened 2019-05-09
Added fence_i instruction

farzamgl:
This commit adds the fence_i instruction which is used to make sure all the changes in D$ are reflected in the I$. This is implemented as a pipeline and FE queue flush.


bp_be/src/v/bp_be_calculator/bp_be_calculator_top.v
```diff
@@ -299,7 +299,7 @@ bsg_dff_reset_en
    ) 
  issue_reg
   (.clk_i(clk_i)
-   ,.reset_i(reset_i | chk_roll_i)
+   ,.reset_i(reset_i | chk_roll_i | chk_poison_isd_i)
```
farzamgl: @dpetrisko 
I'm not 100% sure about this change.
Basically, the problem is that the instruction in the issue register doesn't get poisoned at the pipe flush.

dpetrisko:
Ope, lost my comment

bp_be/src/v/bp_be_calculator/bp_be_calculator_top.v
dpetrisko: This will definitely break things.  We need a separate poison_iss_reg and chk_poison_iss_i signal (I think it will be the same as chk_poison_isd_i, but it's good to be explicit). Then, exc_n[0].poison = poison_iss_r | chk_poison_isd_i

dpetrisko: Good call though, we needed this for sfence.vma as well

bp_be/src/v/bp_be_calculator/bp_be_calculator_top.v
```diff
@@ -293,18 +296,31 @@ bp_be_regfile
...
 
-// Issued instruction registere
+// Issued instruction register
 bsg_dff_reset_en 
  #(.width_p(1+issue_pkt_width_lp)
    ) 
  issue_reg
   (.clk_i(clk_i)
-   ,.reset_i(reset_i | chk_roll_i | chk_poison_isd_i)
+   ,.reset_i(reset_i | chk_poison_iss_i)
```
dpetrisko: change this back to chk_roll_i

dpetrisko:
The fences probably shouldn't be executed in csr. For instance, we'll additionally need to support I/O fences which will go through the D$. However, they need to be executed at mem3. So I propose adding them to the MEM3 MMU command instead of CSR, like a tlb fill

farzamgl:
> The fences probably shouldn't be executed in csr. For instance, we'll additionally need to support I/O fences which will go through the D$. However, they need to be executed at mem3. So I propose adding them to the MEM3 MMU command instead of CSR, like a tlb fill

This will require additional bits in either `calc_stage_r` or `exc_stage_r` to keep track of the fence instructions though. Also for the `fence io`, the csr module can output the cmd to MMU at mem3.

Also, the `sfence.vma` should be inside the csr module to check for the current priv. level.

dpetrisko:
> This will require additional bits in either calc_stage_r or exc_stage_r to keep track of the fence instructions though.

Yeah, can add the few bits into the decode ctl struct 

> Also, the sfence.vma should be inside the csr module to check for the current priv. level.

We need to expose the privilege level to the MMU for other reasons (PMP checks, etc.)

farzamgl:
Moved the `fence` and `fence.i` out of the csr pipe. 
I've kept the `sfence.vma` in the csr module for now because all the other SYSTEM instructions go through the csr pipe and adding 1 exception might complicate the decoding.

bp_be/src/v/bp_be_calculator/bp_be_pipe_mem.v
```diff
@@ -126,16 +130,50 @@ bsg_shift_reg
    ,.data_o(csr_cmd_lo)
    );
 
+bsg_shift_reg
```
dpetrisko: Oof, that's an extra 200+ registers of redundant storage. Probably we should take decode_mem3_i as well and arbitrate based on that

farzamgl: There is no decode_mem3 in the calculator though

dpetrisko: calc_stage_r[2]

farzamgl: But I can reduce the size to the size of fu_op

bp_be/src/v/bp_be_calculator/bp_be_pipe_mem.v
```diff
@@ -126,16 +130,50 @@ bsg_shift_reg
...
+      mem3_cmd.mem_op = e_ptw_i;
+    end
+    else if(mem_resp.exception.dtlb_miss) begin
+      mem3_cmd.mem_op = mem3_store_not_load_i ? e_ptw_s : e_ptw_l;
+    end
+    else begin
+      mem3_cmd.mem_op = mem3_cmd_lo.mem_op;
+    end
+    mem3_cmd.vaddr = mem3_itlb_fill_v_i ? mem3_pc_i : mem3_vaddr_o;
+    mem3_cmd.data  = mem3_pc_i;
```
dpetrisko: Why don't we make data always pc and have the itlb_fill read from that instead of vaddr?

bp_be/src/v/bp_be_calculator/bp_be_pipe_mem.v
farzamgl: There is no decode in calc_stage_r, only general information about the instruction and the destination types, etc... 
I can add a fu_op to that but I think It doesn't really belong there

bp_be/src/v/bp_be_calculator/bp_be_pipe_mem.v
farzamgl: Because the fault_vaddr and some other signals are based on cmd.vaddr. Changing that would add some unnecessary ifs in the mem_top code

bp_be/src/v/bp_be_calculator/bp_be_pipe_mem.v
dpetrisko: Hmm, interesting. We already need to do this storage for the csr_cmd (because that one actually does need the address and data saved.  It probably makes sense to save this info in the stage reg so that it can be shared among all pipes that need it

dpetrisko: No, the shift reg saves fu_op, data and addr, that's 100 flops

dpetrisko: Leave it like this for now with a comment about this discussion, I'll refactor it in the future

farzamgl: I'll just reduce the size of the shift reg to only keep the fu_op for now

farzamgl: Actually, I'm not gonna do that, because later we might wanna pass the whole decode, or the other MMU fields.

dpetrisko:
Looks fine to me then. If it passes regression, go ahead and merge.  I'll take a look at refactoring to reduce the PPA impact

PR #161, opened 2019-05-13
Adding AOI as a supported bsg_cadenv machine.

stdavids:


dpetrisko:
This is expanding the scope a little bit, but what do you think about adding a list of variables that users will need to set in the ~(xor | aoi) case?

```
else
  BSG_CADENV = 0

  VCS ?= 
  DC   ?=
...
```

I'm not sure if there are other variables are set in our cadenv that are used in BP makefiles

stdavids:
Yeah, so for bsg_sv2v I do something similar which is I first “-include
cadenv.mk” and then underneath that have ?= for all of the cad tools
variables required (in that case just dc_shell). -include is just like
include but won’t error if file not found, so you might be able to get rid
of the BSG_CADENV variable.

taylor-bsg:
Picky me, but I don’t like how this is done. The point of cadenv is to
abstract away these details so we do not expose our machine names in our
public repos, and have a makefile that grows with the size of our cluster.
Instead we should have a cadenv public that has the unset variables and an
error message that says set these things. Then all of our cluster specific
cadenv stuff can go in the private repo.

stdavids:
https://github.com/bespoke-silicon-group/bsg_sv2v/blob/8b1e3a0ea4d19ffb7470a2b487af401101605362/Makefile#L12-L20

This is what I was referring to in bsg_sv2v. Doesn't expose the names of our machine, bsg_cadenv is private, and adding more machines only requires a change in bsg_cadenv -- no changes to bsg_sv2v. The only thing we are missing is the blank template. My only issue with the blank template is that the template would have slots for 20+ various cad tools (ASIC, FPGA, etc...) where bsg_sv2v only needs 1 (DC), and black_parrot only needs a couple (DC, VCS/Verilator). So then in black_parrot or bsg_sv2v, we need to say what parts of the template are required to fill in... at that point it is just easier to have the blank variable names that are required if and only if you don't have a bsg_cadenv setup.

dpetrisko:
That solution sounds good to me, with a nit that it should print a warning if the variables aren’t set

taylor-bsg:
Agree. Maybe rather than to give an example cadenv directory, we can have a
local file in any repo that is intended to be used standalone called
Makefile.cadenv.local that has these error checks and has some comments at
the top that say these variables need to be set. Makefile.cadenv.local is
included after bsg_ cadenv.  So then the sequence is, non bsg user runs,
gets error messages, then adds paths, then runs again successfully?

M

dpetrisko:
How about we have our Makefile.common be

```
BP_CADENV ?= Makefile.cadenv.local

include $(BP_CADENV)
```

Then we just ask bsg users to set BP_CADENV = /path/to/bsg_cadenv.mk?

dpetrisko:
This'll be the same flow of users running once, seeing proprietary tools missing, filling them in.  But it's even more bsg-agnostic

taylor-bsg:
No environment variables should be relied upon in BSG due to the
reproducibility problems they always entail -- just include the bsg_cadenv
file with a - to eliminate errors, and then if the BSG_CADENV is not
defined, pull in the other file.

M

dpetrisko:
Relying on a hardcoded path to a BSG repo is no more reproducible than an environment variable pointing to that path...

There's a purity argument that if BP is community owned, it shouldn't have references to any specific research group, bsg or otherwise

taylor-bsg:
On Tue, May 14, 2019 at 4:18 PM Dan Petrisko <notifications@github.com>
wrote:

> Relying on a hardcoded path to a BSG repo is no more reproducible than an
> environment variable pointing to that path...
>

It’s not exactly a hard-coded path — it says look in a sibling directory to
the current directory. This is a standard practice in our group — to put
repos in a common parent directory.

It is reproducible because it has two concrete options that are clearly
visible from the makefile. Picking up optionally-set environment variables
does not convey this intent as well...

There's a purity argument that if BP is community owned, it shouldn't have
> references to any specific research group, bsg or otherwise
>

When we have a few community maintained then we can phase it out. Until
then, I propose we optimize for the common case, while making it easy for
others to use.

dpetrisko:
> It is reproducible because it has two concrete options that are clearly
> visible from the makefile. Picking up optionally-set environment variables
> does not convey this intent as well...

Both approaches require both reading a Makefile comment and making an out of directory change (copying bsg_cadenv, setting an environment variable), conveying either intent equally well. They also both have a default and override behavior.

```
# If the machine you are working on is bsg_cadenv complient, then you do not
# need to setup the cad tools, simply put bsg_cadenv in the same root dir.
-include $(TOP_DIR)/../bsg_cadenv/cadenv.mk
```

```
# Users either can set their CAD variables in Makefile.cadenv.local or set BP_CADENV
# to point to their own Makefile.cadenv 
BP_CADENV ?= Makefile.cadenv.local

include $(BP_CADENV)
```

> When we have a few community maintained then we can phase it out. Until
> then, I propose we optimize for the common case, while making it easy for
> others to use.

> This is a standard practice in our group — to put
> repos in a common parent directory.

There's not a single point where we say "now we have a BP community". We're live now. There's more friction for potential community users if they come in and the default configurations are geared towards BSG -- it makes it seem like our research project that we're allowing access to, rather than a truly open project.

(This issue itself is a minor concern, but we should make Being Friendly a first class infrastructure consideration)

taylor-bsg:
Dan wins the bikeshedding award for the week.

1) Environment variables are terrible, implicit things in a global
namespace, and should only be defined in makefiles.
2) It is not that unfriendly to have references to code that are not used
by all users of the infrastructure. Not everybody uses all pieces of all
code.
3) I am also okay if you check in a bsg_cadenv into github that is public
but leaves the LICENSE_FILE and SYNOPSYS variables blank with an $(error ),
and then replicates whatever lines are needed for black_parrot. There are a
few others like, VCS_HOME that need to be set. Then we can have a common
understanding of tool makefile variable names across the broader bsg_ / hb_
/ bp_ space, which is makes everything more consistent.

dpetrisko:
🏆🚴‍♂️

Some variant of 3 works for me

stdavids:
Updated to match the bsg_sv2v style since it seems like we are going to do some variant of that. Also, added some clean up elsewhere in the infrastructure.

dpetrisko:
@stdavids I tested this and made a minor fix -- it's good to go now.  Do a rebase on the way in, though.

PR #162, opened 2019-05-13
Exposing harden_p parameter in bp_be_regfile.

stdavids:
Exposing harden_p parameter of bsg_mem to the bp_be_refile instantiation. Disabled hardening of floating point register file.


dpetrisko:
Looks good to me. I’m adding a parameter to remove FP entirely, which should also get merged in soon

PR #172, opened 2019-05-18
PC gen rework

dpetrisko:
This PR is intended to implement the pc_gen rework found at https://docs.google.com/document/d/1r18F7dxunUaRZTh07pzaYvLyV6VwMBJ5X5v9o_PvHIo/edit

It is not functional at the moment (hence the WIP DO NOT MERGE). Use it to give feedback and collaborate.

### DIFF
--- a/bp_fe/src/v/bp_fe_pc_gen.v
+++ b/bp_fe/src/v/bp_fe_pc_gen.v
@@ -84,14 +84,14 @@ bp_fe_exception_s        pc_gen_exception;
 bp_fe_instr_scan_s       scan_instr;
    
 // FSM Variables
-enum bit [1:0] {e_wait=2'd0, e_runs=2'd1, e_stall=2'd2} state_n, state_r;
+enum bit [1:0] {e_wait=2'd0, e_run=2'd1, e_stall=2'd2} state_n, state_r;
    
 // pipeline pc's
 logic [vaddr_width_p-1:0]       pc_f2;
 logic                           pc_v_f2;
 logic [vaddr_width_p-1:0]       pc_f1;
-logic [vaddr_width_p-1:0]       pc_resume;
-logic [vaddr_width_p-1:0]       nxt_pc_resume;
+logic [vaddr_width_p-1:0]       pc_resume_r;
+logic [vaddr_width_p-1:0]       pc_resume_n;
 // pc_v_f1 is not a pipeline register, because the squash needs to be 
 // done in the same cycle as when we know the instruction in f2 is
 // a branch
@@ -107,11 +107,9 @@ logic                           is_back_br;
 logic                           predict_taken;
 
 // icache miss recovery wires
-logic                           icache_miss_recover;
 logic                           icache_miss_prev;
 
 // tlb miss recovery wires
-logic                           itlb_miss_recover;
 logic                           itlb_stall, itlb_exception_sent, itlb_miss_r, itlb_miss_f1;
 logic                           itlb_miss_f2;
 
@@ -193,10 +191,16 @@ begin
     end
 end
 
-assign pc_v_f1 = ~((predict_taken & ~btb_pred_f2_r));
+always_ff @(posedge clk_i) begin
+   if (reset_i) begin
+      pc_v_f1 = 'd0;
+   end else begin
+      pc_v_f1 = pc_v_f1_n;
+   end
+end
 
 // Stall FSM
-always @(posedge clk_i) begin : state_r_seq
+always_ff @(posedge clk_i) begin
    if (reset_i) begin
       state_r <= e_wait;
    end else begin
@@ -204,11 +208,13 @@ always @(posedge clk_i) begin : state_r_seq
    end
 end
 
-always @* begin : state_r_comb
+always_comb begin
+   state_n = e_wait;
+
    case (state_r)
-     e_wait : begin : w8
+     e_wait : begin
         // In the wait state
-	if (fe_pc_gen_v_i && ~fe_pc_gen_cmd.attaboy_valid) begin
+	if (fe_pc_gen_v_i & ~fe_pc_gen_cmd.attaboy_valid) begin
 	   // FE command received
 	   state_n = e_stall;
 	end else begin
@@ -216,64 +222,48 @@ always @* begin : state_r_comb
 	   state_n = e_wait;
 	end
      end
-     e_runs : begin : run
+     e_run : begin
 	// In the run state
-	if (pc_v_f2 && itlb_miss_f2) begin
+	if (pc_v_f2 & itlb_miss_f2) begin
 	   state_n = e_wait;
 	end else begin
-	   if ((pc_v_f2 && !icache_pc_gen_v_i) || pc_gen_icache_ready_i) begin
+	   if ((pc_v_f2 & !icache_pc_gen_v_i) || ~pc_gen_icache_ready_i) begin
 	      state_n = e_stall;
 	   end else begin
-	      state_n = e_runs;
+	      state_n = e_run;
 	   end
 	end
      end
-     e_stall : begin : stall
+     e_stall : begin
 	// In the stall state
 	if (pc_v_f1_n) begin
-	   state_n = e_runs;
+	   state_n = e_run;
 	end else begin
 	   state_n = e_stall;
 	end
      end
-     default : begin : default_wait
-	// In the wait state
-	if (fe_pc_gen_v_i && ~fe_pc_gen_cmd.attaboy_valid) begin
-	   // FE command received
-	   state_n = e_stall;
-	end else begin
-	   // FE command not received
-	   state_n = e_wait;
-	end
-     end
+
    endcase // case (state_r)
 end
 
 assign pc_v_f1_n = (state_r != e_wait) & pc_gen_itlb_ready_i & pc_gen_fe_ready_i & pc_gen_icache_ready_i;
 
-always @(posedge clk_i) begin
-   if (reset_i) begin
-      pc_resume <= {vaddr_width_p{1'd0}};
-   end else begin
-      pc_resume <= nxt_pc_resume;
-   end
+always_ff @(posedge clk_i) begin
+  pc_resume_r <= pc_resume_n;
 end
 
-always @* begin
-   if (state_r == e_runs) begin
+always_comb begin
+   if (state_r == e_run) begin
       // Run
-      nxt_pc_resume = pc_f2;
+      pc_resume_n = pc_f2;
    end else begin
       if (fe_pc_gen_v_i) begin
-	 nxt_pc_resume = fe_pc_gen_cmd.pc;
+	 pc_resume_n = fe_pc_gen_cmd.pc;
       end else begin
-	 nxt_pc_resume = pc_resume;
+	 pc_resume_n = pc_resume_r;
       end
    end
 end
-
-assign icache_miss_recover = icache_miss_prev & (~icache_miss_i);
-assign itlb_miss_recover   = itlb_fill_v;
    
 // icache and  itlb miss recover logic
 always_ff @(posedge clk_i)
@@ -303,7 +293,7 @@ end
 always_ff @(posedge clk_i) begin
   if(reset_i)
     itlb_stall <= '0;
-  else if(state_r == e_runs)
+  else if(state_r == e_run)
     itlb_stall <= itlb_miss_f1;
   else if(fe_pc_gen_v_i)
     itlb_stall <= '0;
@@ -312,10 +302,8 @@ end
 always_ff @(posedge clk_i) begin
   if(reset_i)
     itlb_miss_r <= '0;
-  else if(((state_n == e_stall) & ~itlb_miss_exception) & itlb_miss_i)
+  else if(((state_n == e_stall) & ~itlb_miss_exception))
     itlb_miss_r <= 1'b1;
-  else if(state_n == e_runs)
-    itlb_miss_r <= '0;
 end
 
 assign itlb_miss_f1 = (itlb_miss_r | itlb_miss_i) & pc_v_f1;
@@ -338,19 +326,12 @@ begin
     if(state_reset_v) begin
         pc_n = fe_pc_gen_cmd.pc;
     end
+    else if (state_r == e_stall) begin
+	pc_n = pc_resume_r;
+    end
     // if we need to redirect
     else if (pc_redirect_v | icache_fence_v | itlb_fence_v) begin
         pc_n = fe_pc_gen_cmd.pc;
-    end
-    // if we've missed in the itlb
-    else if (itlb_miss_recover & pc_v_f2)
-    begin
-        pc_n = pc_f2; 
-    end
-    // if we've missed in the icache
-    else if (icache_miss_recover & pc_v_f2)
-    begin
-        pc_n = pc_f2;
     end
     else if (btb_br_tgt_v_lo)
     begin
@@ -366,14 +347,12 @@ begin
     end
 end
 
-always_ff @(posedge clk_i)
-begin
-    if (reset_i) 
-    begin
-        pc_f2 <= '0;
-        pc_v_f2 <= '0;
-        pc_f1 <= '0;
+always_ff @(posedge clk_i) begin
+  if (reset_i) begin
+    pc_f2 <= '0;
+    pc_f1 <= '0;
 
+<<<<<<< HEAD
         btb_pred_f1_r <= '0;
         btb_pred_f2_r <= '0;
     end
@@ -383,13 +362,29 @@ begin
         begin
             pc_f2 <= pc_f1;
             pc_v_f2 <= pc_v_f1 & ~pc_redirect_v & ~icache_fence_v & ~itlb_fence_v;
+=======
+    pc_v_f2 <= '0;
 
-            pc_f1 <= pc_n;
+    btb_pred_f1_r <= '0;
+    btb_pred_f2_r <= '0;
+  end else begin
+    pc_v_f2 <= pc_v_f1 & ~pc_redirect_v;
+>>>>>>> Updated based on peer review comments. median and rv64-add tests no longer pass.
 
-            btb_pred_f1_r <= btb_br_tgt_v_lo;
-            btb_pred_f2_r <= btb_pred_f1_r;
-        end
+    if (state_r == e_stall) begin
+      pc_f2 <= pc_f2;
+      pc_f1 <= pc_f1;
+
+      btb_pred_f1_r <= btb_pred_f1_r;
+      btb_pred_f2_r <= btb_pred_f2_r;
+    end else begin
+      pc_f2 <= pc_f1;
+      pc_f1 <= pc_n;
+
+      btb_pred_f1_r <= btb_br_tgt_v_lo;
+      btb_pred_f2_r <= btb_pred_f1_r;
     end
+  end
 end
 
 assign fe_queue_branch_metadata = '{btb_tag: pc_gen_fetch.pc[2+btb_idx_width_p+:btb_tag_width_p]
--- a/bp_fe/src/v/bp_fe_pc_gen.v
+++ b/bp_fe/src/v/bp_fe_pc_gen.v
@@ -49,7 +49,6 @@ module bp_fe_pc_gen
 
 // Suppress unused signal warnings
 wire unused0 = v_i;
-wire unused1 = pc_gen_itlb_ready_i;
 
 assign icache_pc_gen_ready_o = '0;
 assign pc_gen_itlb_v_o = pc_gen_icache_v_o;
@@ -147,7 +146,7 @@ assign itlb_fill_v   = fe_pc_gen_v_i & fe_pc_gen_cmd.itlb_fill_valid;
 assign icache_fence_v = fe_pc_gen_v_i & fe_pc_gen_cmd.icache_fence_valid;
 assign itlb_fence_v   = fe_pc_gen_v_i & fe_pc_gen_cmd.itlb_fence_valid;
 
-assign fe_instr_v         = pc_v_f2 & ~flush_f1;
+assign fe_instr_v         = pc_v_f2 & ~flush_f2;
 assign fe_exception_v     = misalign_exception | itlb_miss_exception;
 assign misalign_exception = pc_redirect_v 
                             & ~fe_pc_gen_cmd.pc[1:0] == 2'b00;
@@ -185,7 +184,7 @@ begin
   else 
     begin
       fe_pc_gen_ready_o = fe_pc_gen_v_i;
-      pc_gen_fe_v_o     = fe_instr_v | fe_exception_v;
+      pc_gen_fe_v_o     = (fe_instr_v | fe_exception_v) & pc_gen_fe_ready_i;
       pc_gen_icache_v_o = pc_v_f1_n;
     end
 end
@@ -294,8 +293,8 @@ begin
     end
 end
 
-assign flush_f1 = itlb_miss_f2 | icache_miss_i;   //CHECK: should it be "(itlb_miss_f2 | icache_miss_i) & ~pc_redirect_v"?
-assign flush_f2 = itlb_miss_f2 | icache_miss_i | pc_redirect_v;
+assign flush_f1 = (itlb_miss_f2 | icache_miss_i) & ~(pc_redirect_v | icache_fence_v | itlb_fence_v);
+assign flush_f2 = itlb_miss_f2 | icache_miss_i | pc_redirect_v | icache_fence_v | itlb_fence_v;
 assign pc_v_f1_n = pc_gen_itlb_ready_i & pc_gen_fe_ready_i & pc_gen_icache_ready_i;
 always_ff @(posedge clk_i) begin
   if (reset_i) begin
--- a/bp_fe/src/v/bp_fe_pc_gen.v
+++ b/bp_fe/src/v/bp_fe_pc_gen.v
@@ -85,15 +85,20 @@ bp_fe_instr_scan_s       scan_instr;
    
 // Logic for handling coming out of reset
 enum bit [0:0] {e_reset, e_run} state_n, state_r;
+
+enum bit [1:0] {e_wait=2'd0, e_runs=2'd1, e_stall=2'd2} stall_state, nxt_stall_state;
    
 // pipeline pc's
 logic [vaddr_width_p-1:0]       pc_f2;
 logic                           pc_v_f2;
 logic [vaddr_width_p-1:0]       pc_f1;
+logic [vaddr_width_p-1:0]       pc_resume;
+logic [vaddr_width_p-1:0]       nxt_pc_resume;
 // pc_v_f1 is not a pipeline register, because the squash needs to be 
 // done in the same cycle as when we know the instruction in f2 is
 // a branch
 logic                           pc_v_f1;
+logic                           pc_v_f1_n;
 logic [vaddr_width_p-1:0]       pc_n;
 
 logic                           stall;
@@ -112,7 +117,8 @@ logic                           icache_miss_prev;
 // tlb miss recovery wires
 logic                           itlb_miss_recover;
 logic                           itlb_stall, itlb_exception_sent, itlb_miss_r, itlb_miss_f1;
-   
+logic                           itlb_miss_f2;
+
 logic [vaddr_width_p-1:0]       btb_target;
 logic [instr_width_lp-1:0]      next_instr;
 logic [instr_width_lp-1:0]      instr;
@@ -131,7 +137,6 @@ logic                          fe_exception_v;
 logic                          misalign_exception;
 logic                          itlb_miss_exception;
 
-   
 bp_fe_branch_metadata_fwd_s fe_queue_branch_metadata, fe_queue_branch_metadata_r;
 
 logic btb_pred_f1_r, btb_pred_f2_r;
@@ -217,6 +222,84 @@ assign pc_v_f1 = ~((predict_taken & ~btb_pred_f2_r));
 // stall logic
 assign stall = (state_r == e_reset) | ~pc_gen_fe_ready_i | ~pc_gen_icache_ready_i | itlb_stall | icache_miss_i;
 
+
+// Stall FSM
+always @(posedge clk_i) begin : stall_state_seq
+   if (reset_i) begin
+      stall_state <= e_wait;
+   end else begin
+      stall_state <= nxt_stall_state;
+   end
+end
+
+always @* begin : stall_state_comb
+   case (stall_state)
+     e_wait : begin : w8
+        // In the wait state
+	if (fe_pc_gen_v_i && ~fe_pc_gen_cmd.attaboy_valid) begin
+	   // FE command received
+	   nxt_stall_state = e_stall;
+	end else begin
+	   // FE command not received
+	   nxt_stall_state = e_wait;
+	end
+     end
+     e_runs : begin : run
+	// In the run state
+	if (pc_v_f2 && itlb_miss_f2) begin
+	   nxt_stall_state = e_wait;
+	end else begin
+	   if ((pc_v_f2 && !icache_pc_gen_v_i) || pc_gen_icache_ready_i) begin
+	      nxt_stall_state = e_stall;
+	   end else begin
+	      nxt_stall_state = e_runs;
+	   end
+	end
+     end
+     e_stall : begin : stall
+	// In the stall state
+	if (pc_v_f1_n) begin
+	   nxt_stall_state = e_runs;
+	end else begin
+	   nxt_stall_state = e_stall;
+	end
+     end
+     default : begin : default_wait
+	// In the wait state
+	if (fe_pc_gen_v_i && ~fe_pc_gen_cmd.attaboy_valid) begin
+	   // FE command received
+	   nxt_stall_state = e_stall;
+	end else begin
+	   // FE command not received
+	   nxt_stall_state = e_wait;
+	end
+     end
+   endcase // case (stall_state)
+end
+
+assign pc_v_f1_n = (stall_state != 2'd0) & pc_gen_itlb_ready_i & pc_gen_fe_ready_i & pc_gen_icache_ready_i;
+
+always @(posedge clk_i) begin
+   if (reset_i) begin
+      pc_resume <= {vaddr_width_p{1'd0}};
+   end else begin
+      pc_resume <= nxt_pc_resume;
+   end
+end
+
+always @* begin
+   if (stall_state == 2'd1) begin
+      // Run
+      nxt_pc_resume = pc_f2;
+   end else begin
+      if (fe_pc_gen_v_i) begin
+	 nxt_pc_resume = fe_pc_gen_cmd.pc;
+      end else begin
+	 nxt_pc_resume = pc_resume;
+      end
+   end
+end
+
 assign icache_miss_recover = icache_miss_prev & (~icache_miss_i);
 assign itlb_miss_recover   = itlb_fill_v;
    
@@ -265,8 +348,16 @@ end
 
 assign itlb_miss_f1 = (itlb_miss_r | itlb_miss_i) & pc_v_f1;
 
+always_ff @(posedge clk_i) begin
+   if (reset_i) begin
+      itlb_miss_f2 <= 1'd0;
+   end else begin
+      itlb_miss_f2 <= itlb_miss_f1;
+   end
+end
+
 logic [vaddr_width_p-1:0] btb_br_tgt_lo;
-logic                     btb_br_tgt_v_lo;
+logic                     btb_br_tgt_v_lo=0;
 
 bp_fe_branch_metadata_fwd_s fe_cmd_branch_metadata;
 always_comb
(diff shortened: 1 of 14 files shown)


bp_fe/src/v/bp_fe_pc_gen.v
```diff
@@ -83,10 +83,8 @@ bp_fe_fetch_s            pc_gen_fetch;
 bp_fe_exception_s        pc_gen_exception;
 bp_fe_instr_scan_s       scan_instr;
    
-// Logic for handling coming out of reset
-enum bit [0:0] {e_reset, e_run} state_n, state_r;
-
-enum bit [1:0] {e_wait=2'd0, e_runs=2'd1, e_stall=2'd2} stall_state, nxt_stall_state;
```
dpetrisko: e_run, not runs.

bp_fe/src/v/bp_fe_pc_gen.v
```diff
@@ -83,10 +83,8 @@ bp_fe_fetch_s            pc_gen_fetch;
...
+// FSM Variables
+enum bit [1:0] {e_wait=2'd0, e_runs=2'd1, e_stall=2'd2} state_n, state_r;
```
dpetrisko: e_run, not runs

bp_fe/src/v/bp_fe_pc_gen.v
```diff
@@ -207,73 +183,69 @@ begin
     end 
   else 
     begin
-      fe_pc_gen_ready_o = ((stall_state == e_runs) | (nxt_stall_state == e_runs)) & fe_pc_gen_v_i;
+      fe_pc_gen_ready_o = ((state_r == e_runs) | (state_n == e_runs)) & fe_pc_gen_v_i;
```
dpetrisko: should just be state_n. If we're transitioning away from run, that means that we've missed in the tlb/cache.

bp_fe/src/v/bp_fe_pc_gen.v
```diff
@@ -207,73 +183,69 @@ begin
...
       pc_gen_fe_v_o     = pc_gen_fe_ready_i & pc_v_f2 & (icache_pc_gen_v_i | fe_exception_v) & ~pc_redirect_v;
-      pc_gen_icache_v_o = pc_gen_fe_ready_i & pc_gen_icache_ready_i & ((stall_state == e_runs) | (nxt_stall_state == e_runs));
+      pc_gen_icache_v_o = pc_gen_fe_ready_i & pc_gen_icache_ready_i & ((state_r == e_runs) | (state_n == e_runs));
```
dpetrisko: Same here

bp_fe/src/v/bp_fe_pc_gen.v
```diff
@@ -207,73 +183,69 @@ begin
...
     end
 end
 
 assign pc_v_f1 = ~((predict_taken & ~btb_pred_f2_r));
```
dpetrisko: Take out branch prediction logic for simplicity.  You use this as pc_v_f2 <= pc_v_f1 & ~pc_redirect_v;

bp_fe/src/v/bp_fe_pc_gen.v
```diff
@@ -207,73 +183,69 @@ begin
...
    end
 end
 
-always @* begin : stall_state_comb
-   case (stall_state)
+always @* begin : state_r_comb
+   case (state_r)
      e_wait : begin : w8
         // In the wait state
 	if (fe_pc_gen_v_i && ~fe_pc_gen_cmd.attaboy_valid) begin
```
dpetrisko: Use single & in conditionals

bp_fe/src/v/bp_fe_pc_gen.v
```diff
@@ -284,7 +256,7 @@ always @(posedge clk_i) begin
 end
 
 always @* begin
-   if (stall_state == e_runs) begin
+   if (state_r == e_runs) begin
       // Run
       nxt_pc_resume = pc_f2;
```
dpetrisko: pc_resume_n, pc_resume_r

bp_fe/src/v/bp_fe_pc_gen.v
```diff
@@ -327,7 +299,7 @@ end
 always_ff @(posedge clk_i) begin
```
dpetrisko: itlb_stall is not used

bp_fe/src/v/bp_fe_pc_gen.v
```diff
@@ -336,9 +308,9 @@ end
...
```
dpetrisko: itlb_miss should only be latched in f2 for a single cycle, and used for sending the fill request and transitioning to wait

bp_fe/src/v/bp_fe_pc_gen.v
```diff
@@ -403,7 +375,7 @@ begin
     end
     else 
     begin
-        if ((stall_state == e_runs) | (nxt_stall_state == e_runs))
+        if ((state_r == e_runs) | (state_n == e_runs))
```
dpetrisko: The valids should be latched regardless of state. Otherwise you could leave something valid permanently during a stall

bp_fe/src/v/bp_fe_pc_gen.v
```diff
@@ -207,73 +183,69 @@ begin
...
 	if (pc_v_f1_n) begin
-	   nxt_stall_state = e_runs;
+	   state_n = e_runs;
 	end else begin
-	   nxt_stall_state = e_stall;
+	   state_n = e_stall;
 	end
      end
      default : begin : default_wait
 	// In the wait state
```
dpetrisko: Just state_n = e_wait. We shouldn't ever use the default case, so don't waste logic on it

bp_fe/src/v/bp_fe_pc_gen.v
```diff
@@ -112,7 +113,8 @@ logic                           icache_miss_prev;
 // tlb miss recovery wires
 logic                           itlb_miss_recover;
```
dpetrisko: Get rid of these

bp_fe/src/v/bp_fe_pc_gen.v
```diff
@@ -202,16 +183,90 @@ begin
...
+	end
+     end
+   endcase // case (state_r)
+end
+
+assign pc_v_f1_n = (state_r != e_wait) & pc_gen_itlb_ready_i & pc_gen_fe_ready_i & pc_gen_icache_ready_i;
+
+always @(posedge clk_i) begin
+   if (reset_i) begin
+      pc_resume <= {vaddr_width_p{1'd0}};
```
dpetrisko: Don't need to reset pc_resume. It's not used until it's set by the reset cmd

bp_fe/src/v/bp_fe_pc_gen.v
```diff
@@ -202,16 +183,90 @@ begin
...
+      // Run
+      nxt_pc_resume = pc_f2;
+   end else begin
+      if (fe_pc_gen_v_i) begin
+	 nxt_pc_resume = fe_pc_gen_cmd.pc;
+      end else begin
+	 nxt_pc_resume = pc_resume;
+      end
+   end
+end
```
dpetrisko: Don't need miss_recover signals

bp_fe/src/v/bp_fe_pc_gen.v
```diff
@@ -253,16 +308,24 @@ end
...
     itlb_miss_r <= '0;
-  else if((stall & ~itlb_miss_exception) & itlb_miss_i)
+  else if(((state_n == e_stall) & ~itlb_miss_exception) & itlb_miss_i)
     itlb_miss_r <= 1'b1;
-  else if(~stall)
+  else if(state_n == e_runs)
     itlb_miss_r <= '0;
 end
 
 assign itlb_miss_f1 = (itlb_miss_r | itlb_miss_i) & pc_v_f1;
```
dpetrisko: Remove

bp_fe/src/v/bp_fe_pc_gen.v
```diff
@@ -253,16 +308,24 @@ end
...
+   if (reset_i) begin
+      itlb_miss_f2 <= 1'd0;
+   end else begin
+      itlb_miss_f2 <= itlb_miss_f1;
+   end
+end
+
 logic [vaddr_width_p-1:0] btb_br_tgt_lo;
-logic                     btb_br_tgt_v_lo;
+logic                     btb_br_tgt_v_lo=0;
```
dpetrisko: FPGA style resets do not synthesize and should never be used in BSG projects

bp_fe/src/v/bp_fe_pc_gen.v
```diff
@@ -202,16 +183,90 @@ begin
...
+	   // FE command not received
+	   state_n = e_wait;
+	end
+     end
+     e_runs : begin : run
+	// In the run state
+	if (pc_v_f2 && itlb_miss_f2) begin
+	   state_n = e_wait;
+	end else begin
+	   if ((pc_v_f2 && !icache_pc_gen_v_i) || pc_gen_icache_ready_i) begin
```
dpetrisko: Stall if icache is ready?

bp_fe/src/v/bp_fe_pc_gen.v
```diff
@@ -202,16 +183,90 @@ begin
...
+always @(posedge clk_i) begin : state_r_seq
+   if (reset_i) begin
+      state_r <= e_wait;
+   end else begin
+      state_r <= state_n;
+   end
+end
+
+always @* begin : state_r_comb
+   case (state_r)
```
dpetrisko: Set a default state_n = state_r at the top of this block and then override

bp_fe/src/v/bp_fe_pc_gen.v
```diff
@@ -202,16 +183,90 @@ begin
...
+// Stall FSM
+always @(posedge clk_i) begin : state_r_seq
+   if (reset_i) begin
+      state_r <= e_wait;
+   end else begin
+      state_r <= state_n;
+   end
+end
+
+always @* begin : state_r_comb
```
dpetrisko: Use always_comb and always_ff

taylor-bsg:
@dpetrisko thank you for providing Anthony feedback, Dan! He's going to be a pro soon enough!

farzamgl:
@zaazad @apducimo can you rebase this branch and fe_dev with dev?

dpetrisko:
You have to merge dev->fe_dev and then rebase this branch onto fe_dev

(Because fe_dev is protected)

bp_fe/src/v/bp_fe_pc_gen.v
farzamgl: shouldn't `pc_v_f1 ` be changed to a register and loaded from `pc_v_f1_n`?

bp_fe/src/v/bp_fe_pc_gen.v
```diff
@@ -83,21 +83,22 @@ bp_fe_fetch_s            pc_gen_fetch;
 bp_fe_exception_s        pc_gen_exception;
 bp_fe_instr_scan_s       scan_instr;
```
farzamgl: A lot of the "first level structs" are redundant in my opinion and the global BE-FE structs can be used instead of them.

bp_fe/src/v/bp_fe_pc_gen.v
```diff
@@ -202,16 +183,90 @@ begin
     end 
   else 
     begin
-      fe_pc_gen_ready_o = ~stall & fe_pc_gen_v_i;
+      fe_pc_gen_ready_o = ((state_r == e_runs) | (state_n == e_runs)) & fe_pc_gen_v_i;
       pc_gen_fe_v_o     = pc_gen_fe_ready_i & pc_v_f2 & (icache_pc_gen_v_i | fe_exception_v) & ~pc_redirect_v;
-      pc_gen_icache_v_o = pc_gen_fe_ready_i & pc_gen_icache_ready_i & ~stall;
+      pc_gen_icache_v_o = pc_gen_fe_ready_i & pc_gen_icache_ready_i & ((state_r == e_runs) | (state_n == e_runs));
```
farzamgl: I think these 3 control signals should be changed. Look at my notes in the document

bp_fe/src/v/bp_fe_pc_gen.v
farzamgl: Might be a good idea to use the cache miss_o signal instead of ~v_o

bp_fe/src/v/bp_fe_pc_gen.v
```diff
@@ -312,7 +375,7 @@ begin
     end
     else 
     begin
-        if (~stall)
+        if ((state_r == e_runs) | (state_n == e_runs))
         begin
             pc_f2 <= pc_f1;
             pc_v_f2 <= pc_v_f1 & ~pc_redirect_v;
```
farzamgl: I think these valid signals should be flushed on f2 miss or redirect.
Look at the document for details.

bp_fe/src/v/bp_fe_pc_gen.v
dpetrisko: They're not identical. There are reasons why the cache would not return valid that are not miss

farzamgl: What are those cases?
And also should we wait for icache ready in those cases?

dpetrisko: One example: a full write buffer. 

Yes, cache ready is the all-encompassing signal.

bp_fe/src/v/bp_fe_pc_gen.v
apducimo: Yup, I forgot to change that back when cleaning it up. I'll make sure to do that.

bp_fe/src/v/bp_fe_pc_gen.v
apducimo: I disagree: && is the logical AND and should therefore be used in conditionals. Lint  should pick up on the incorrect usage of & in conditional statements.

FWIW, Here's the message from Cadence's lint tool:
image

bp_fe/src/v/bp_fe_pc_gen.v
apducimo: Right, that was my temporary fix to remove branch prediction out of the mix temporarily. In hindsight I should have added a comment for clarity

bp_fe/src/v/bp_fe_pc_gen.v
apducimo: Good catch, that should have been inverted.

bp_fe/src/v/bp_fe_pc_gen.v
apducimo: There's no need if my case statement is complete.

bp_fe/src/v/bp_fe_pc_gen.v
apducimo: Sure. There are some simulation benefits and some additional checking. It all gets mapped correctly in FPGA and ASIC synthesis.

bp_fe/src/v/bp_fe_pc_gen.v
apducimo: Yup, I forgot to change that back when cleaning it up. I'll make sure to do that.

bp_fe/src/v/bp_fe_pc_gen.v
apducimo: Can you elaborate as to why this is no longer needed?

bp_fe/src/v/bp_fe_pc_gen.v
apducimo: Sure, thing. I missed this when trying to port over naming conventions

apducimo: FYI, a lot of us in industry reserve the use of the suffix 'n' and '_n' to denote active low signaling. Also most FPGA and ASIC synthesis and implementation tools add '_r' or '_reg' to the names of variables that get mapped to flops.

bp_fe/src/v/bp_fe_pc_gen.v
dpetrisko: Yep

bp_fe/src/v/bp_fe_pc_gen.v
apducimo: Okay. For some reason, the next state alone wasn't working for me. Most likely because of some of the bugs you found else where in the code.

bp_fe/src/v/bp_fe_pc_gen.v
dpetrisko: BSG style. 

We do things like
If (a[1:0] & b[:1:0]) where we want the result of 10 & 01 to be 0 more often than we would want it to be 1. 

In this case it doesn’t matter either way

bp_fe/src/v/bp_fe_pc_gen.v
apducimo: In the dev branch, itlb_miss is not used here either. I assume you're trying to say is that replacing
         ~stall
with
          (state_n == e_stall)
is not sufficient. This makes sense because itlb_miss was part of the combinatoric assignment of stall but is no longer part of combinatoric assignment of state_n (at least directly).

bp_fe/src/v/bp_fe_pc_gen.v
dpetrisko: http://www.sunburst-design.com/papers/CummingsSNUG2016SV_SVLogicProcs.pdf

https://www.verilogpro.com/systemverilog-always_comb-always_ff/

There’s a greater chance of pre and post synth mismatch with always vs always_x. They allow you to write from more than one process which is almost never what you want.

bp_fe/src/v/bp_fe_pc_gen.v
dpetrisko: Less verbose, less chance of inferred latches by forgetting a clause in a case

bp_fe/src/v/bp_fe_pc_gen.v
dpetrisko: In BSG we do not use active low signals, so this conflict doesn’t arise. We use _n, _r for flops

bp_fe/src/v/bp_fe_pc_gen.v
apducimo: Are you implying that using a default assignment at the top of an always_comb block generates less logic than having a fully populated case statement?

bp_fe/src/v/bp_fe_pc_gen.v
dpetrisko: I’m saying the line should be removed. Itlb miss should cause transition to the wait state, but there should not be any separate tracking of this event

bp_fe/src/v/bp_fe_pc_gen.v
dpetrisko: The TLB miss happens in f1 and should be latched in f2 for a single cycle. There should be no TLB miss state in f1

bp_fe/src/v/bp_fe_pc_gen.v
dpetrisko: Use a wire, which is synthesizable. Temporary fixes are fine, but non-synthesizable stuff is scary

bp_fe/src/v/bp_fe_pc_gen.v
farzamgl: That signal was added to keep track of a misses tlb access, but now with tge state machine it's not needed anymore

bp_fe/src/v/bp_fe_pc_gen.v
farzamgl: Yeah. That was the case before because there where cases when I needed to keep the miss in f1 because the pipe was stalled, otherwise it would have been lost. But now we flush the pipe before stalling it so there shouldn't be a problem

bp_fe/src/v/bp_fe_pc_gen.v
apducimo: True. I believe always_ff also safeguards against inferred latches.

bp_fe/src/v/bp_fe_pc_gen.v
apducimo: Not sure I understand the example. Regardless I will conform to the BSG style.

bp_fe/src/v/bp_fe_pc_gen.v
dpetrisko: No, the default statement here corresponds to a state outside of the defined enum (2'd3).  This should never come up in normal operation. In case some exceptional conditional comes up, we should just return to wait state to define the behavior. There's no need to maintain it as a second wait state.

bp_fe/src/v/bp_fe_pc_gen.v
apducimo: Fair enough.

bp_fe/src/v/bp_fe_pc_gen.v
apducimo: To clarify, replace this line:
          assign pc_v_f1 = ~((predict_taken & ~btb_pred_f2_r));
with an always_ff block, correct?

bp_fe/src/v/bp_fe_pc_gen.v
dpetrisko: 10 and 01 both correspond to true, so for a=10 b=01,

a&b == false
a && b == true

apducimo: I see what you're saying with your example. I would think a good coding style would disallow the usage of instances like:
          a && b

bp_fe/src/v/bp_fe_pc_gen.v
apducimo: Understood, but inst a logic a hybrid wire/reg? I guess some tools dont necessarily interpret that

bp_fe/src/v/bp_fe_pc_gen.v
apducimo: where the pc_v_f1 <= pc_v_f1_n if (!reset)

bp_fe/src/v/bp_fe_pc_gen.v
apducimo: where a,b are multi-bit vectors

bp_fe/src/v/bp_fe_pc_gen.v
apducimo: Oh sorry, I was looking at the wrong line of code. You meant to remove the whole block that registers itlb_stall. Sorry for the confiusion

apducimo: But itlb_stall is used in the assignment to itlb_miss_exception.

bp_fe/src/v/bp_fe_pc_gen.v
apducimo: Okay

bp_fe/src/v/bp_fe_pc_gen.v
apducimo: That doesn't align with what you said in your document:
pc_f2 <= pc_f2 if stall
Else pc_f1
pc_f1 <= pc_f1 if stall
Else pc_n

apducimo: Oh. Forgive me. I misinterpreted what you were saying.

apducimo: The valid should be latched regardless, not the data

bp_fe/src/v/bp_fe_pc_gen.v
apducimo: Okay

bp_fe/src/v/bp_fe_pc_gen.v
apducimo: okay

bp_fe/src/v/bp_fe_pc_gen.v
dpetrisko: Yep

bp_fe/src/v/bp_fe_pc_gen.v
```diff
@@ -202,16 +183,90 @@ begin
...
+      pc_gen_icache_v_o = pc_gen_fe_ready_i & pc_gen_icache_ready_i & ((state_r == e_runs) | (state_n == e_runs));
     end
 end
 
 assign pc_v_f1 = ~((predict_taken & ~btb_pred_f2_r));
 
-// stall logic
-assign stall = (state_r == e_reset) | ~pc_gen_fe_ready_i | ~pc_gen_icache_ready_i | itlb_stall | icache_miss_i;
+// Stall FSM
+always @(posedge clk_i) begin : state_r_seq
```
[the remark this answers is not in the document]
apducimo: okay

bp_fe/src/v/bp_fe_pc_gen.v
```diff
@@ -202,16 +183,90 @@ begin
...
+	   if ((pc_v_f2 && !icache_pc_gen_v_i) || pc_gen_icache_ready_i) begin
+	      state_n = e_stall;
+	   end else begin
+	      state_n = e_runs;
+	   end
+	end
+     end
+     e_stall : begin : stall
+	// In the stall state
+	if (pc_v_f1_n) begin
```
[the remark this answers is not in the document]
apducimo: okay

bp_fe/src/v/bp_fe_pc_gen.v
dpetrisko: logic a = 0; becomes reg a = 0, FPGA style reset
wire a = 0; is equivalent to logic a; assign a = 0;

bp_fe/src/v/bp_fe_pc_gen.v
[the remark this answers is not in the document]
dpetrisko: Moreover, prefer ternaries to single line, dual clause if-elses. The X-behavior more closely matches synthesis (and have the extra benefit of terseness)

bp_fe/src/v/bp_fe_pc_gen.v
dpetrisko: That's effectively what we're saying, as && and & are equivalent for single bit vectors. But this way we only use a single operator that is well understood. We're very into Principle of Least Astonishment.

I'm also unsure if X-behavior is identical for && and & in either the single bit or multi-bit case.

taylor-bsg: Great discussion. I did a bunch of experiments on edaplayground (https://www.edaplayground.com/x/2pGQ) and have updated the coding guidelines on logical statements:

https://docs.google.com/document/d/1xA5XUzBtz_D6aSyIBQUwFk_kSUdckrfxa2uzGjMgmCU/edit#heading=h.31hyy1y35p1g

I also added a section on if's, logicals, and X's.

https://docs.google.com/document/d/1xA5XUzBtz_D6aSyIBQUwFk_kSUdckrfxa2uzGjMgmCU/edit#heading=h.tcvgxwesxsuf

So in Dan's example, the correct expression would be if (|(a[1:0] & b[1:0]))

fe_pc_gen_v_i && ~fe_pc_gen_cmd.attaboy_valid

violates the purity rule; more compliant would be

fe_pc_gen_v_i  & ~ fe_pc_gen_cmd.attaboy_valid

also correct but stylistically discouraged since it implies we are working with word-level data, is 

fe_pc_gen_v_i  && !fe_pc_gen_cmd.attaboy_valid

bp_fe/src/v/bp_fe_pc_gen.v
```diff
@@ -83,21 +83,22 @@ bp_fe_fetch_s            pc_gen_fetch;
...
+// FSM Variables
+enum bit [1:0] {e_wait=2'd0, e_run=2'd1, e_stall=2'd2} state_n, state_r;
    
 // pipeline pc's
 logic [vaddr_width_p-1:0]       pc_f2;
 logic                           pc_v_f2;
 logic [vaddr_width_p-1:0]       pc_f1;
+logic [vaddr_width_p-1:0]       pc_resume_r;
+logic [vaddr_width_p-1:0]       pc_resume_n;
 // pc_v_f1 is not a pipeline register, because the squash needs to be
```
farzamgl: remove this comment

bp_fe/src/v/bp_fe_pc_gen.v
```diff
@@ -303,30 +293,37 @@ begin
...
-        begin
-            pc_f2 <= pc_f1;
-            pc_v_f2 <= pc_v_f1 & ~pc_redirect_v & ~icache_fence_v & ~itlb_fence_v;
-
-            pc_f1 <= pc_n;
-
-            btb_pred_f1_r <= btb_br_tgt_v_lo;
-            btb_pred_f2_r <= btb_pred_f1_r;
-        end
+assign flush_f1 = (itlb_miss_f2 | icache_miss_i) & ~(pc_redirect_v | icache_fence_v | itlb_fence_v);
```
dpetrisko: Why wouldn't we flush on icache_fence or itlb_fence?

bp_fe/src/v/bp_fe_pc_gen.v
```diff
@@ -206,67 +183,87 @@ begin
...
     end
-    else
-    begin
-        icache_miss_prev    <= icache_miss_i;
+    e_run : begin
+      // In the run state
+      if (pc_v_f2 & itlb_miss_f2) begin
+        state_n = e_wait;
+      end else begin
+        if ((pc_v_f2 & !icache_pc_gen_v_i) || ~pc_gen_icache_ready_i) begin //CHECK: should it be "if ((pc_v_f2 & !icache_pc_gen_v_i) || ~pc_v_f1_n)"?
```
dpetrisko: Remove, !icache_pc_gen_v_i is sufficent

`|| ~pc_gen_icache_ready_i) begin //CHECK: should it be "if ((pc_v_f2 & !icache_pc_gen_v_i) || ~pc_v_f1_n)"? `

We don't transition to stall on ~icache_ready, otherwise we'll

bp_fe/src/v/bp_fe_pc_gen.v
```diff
@@ -206,67 +183,87 @@ begin
...
-assign itlb_miss_recover   = itlb_fill_v;
-   
-// icache and  itlb miss recover logic
-always_ff @(posedge clk_i)
-begin
-    if (reset_i)
-    begin
-        icache_miss_prev    <= '0;
+always_comb begin
+  state_n = e_wait;
```
dpetrisko: state_n = state_r, then remove the corresponding null-state transitions.

bp_fe/src/v/bp_fe_pc_gen.v
```diff
@@ -206,67 +183,87 @@ begin
...
+        end
+      end
+    end
+    e_stall : begin
+      // In the stall state
+      if (pc_v_f1_n) begin
+        state_n = e_run;
+      end else begin
+        state_n = e_stall;
+      end
```
dpetrisko: default: state_n = e_wait;

bp_fe/src/v/bp_fe_pc_gen.v
farzamgl: Because pc_n is set to the fe_cmd's pc on fence, and so pc_f1 should not be flushed for the next cycle, even if there is a f2 miss

dpetrisko: Oh interesting. Well it depends if the fence is done in a single cycle. I believe that's true for icache fence because it's coherent, but is it true for itlb fence? Since it's doing a flush?

Basically, how does the itlb handle read-during-write?

bp_fe/src/v/bp_fe_pc_gen.v
dpetrisko: Not anymore :)

bp_fe/src/v/bp_fe_pc_gen.v
farzamgl: The fence is done in 1 cycle.
It doesn't handle it. ready_o is low during write, so you cannot read.

bp_fe/src/v/bp_fe_pc_gen.v
dpetrisko: Agreed, but let's open a separate issue for that

bp_fe/src/v/bp_fe_pc_gen.v
dpetrisko: Gotcha. In that case

assign pc_gen_itlb_v_o = pc_gen_icache_v_o; 

should be

assign pc_gen_itlb_v_o = pc_gen_itlb_ready_i & pc_gen_icache_v_o;

dpetrisko: Unrelated, but shouldn't this be 1?
assign icache_pc_gen_ready_o = '0;

dpetrisko: And last one,
pc_gen_icache_v_o = pc_v_f1_n;
should be
pc_gen_icache_v_o = pc_gen_icache_ready_i & pc_v_f1_n;

farzamgl: > Unrelated, but shouldn't this be 1? (I assume it's unused, but still)
> assign icache_pc_gen_ready_o = '0;

Oh, right good catch.
> assign pc_gen_itlb_v_o = pc_gen_itlb_ready_i & pc_gen_icache_v_o;

right now it's in `pc_v_f1_n` do you wanna remove it?
> pc_gen_icache_v_o = pc_gen_icache_ready_i & pc_v_f1_n;

also for this

dpetrisko: We definitely need it in pc_v_f1_n because we stall based on it.  I think having it in both is fine. I just like to see the handshake explicitly in the logic signal (and it should easily get optimized out)

bp_fe/src/v/bp_fe_pc_gen.v
```diff
@@ -303,30 +245,34 @@ begin
...
 end
 
-always_ff @(posedge clk_i)
-begin
-    if (reset_i) 
-    begin
-        pc_f2 <= '0;
-        pc_v_f2 <= '0;
-        pc_f1 <= '0;
+assign flush_f1 = (itlb_miss_f2 | icache_miss_i) & ~(pc_redirect_v | icache_fence_v | itlb_fence_v);
```
farzamgl: Should `(itlb_miss_f2 | icache_miss_i)` be replaced with `(pc_v_f2 & ~icache_pc_gen_v_i`?

bp_fe/src/v/bp_fe_pc_gen.v
```diff
@@ -196,79 +148,76 @@ assign pc_gen_icache.virt_addr          = pc_n;
...
+        else
+          state_n = e_stall;
+      end 
     end
-    else
-    begin
-        icache_miss_prev    <= icache_miss_i;
+    e_run : begin
+      // In the run state
+      if (pc_v_f2 & itlb_miss_f2)
```
farzamgl: I think we sould add another jump to `e_stall` for when fe_ready_i is low.
`if((fe_instr_v | fe_exception_v) & ~pc_gen_fe_ready_i) state_n = e_stall;`
Right now if `fe_ready_i` suddenly goes low at f2 we loose the f2 instr or exception.

dpetrisko: Yep, we need that. And the queue not being ready as a flush condition

dpetrisko:
@farzamgl , could you take a final pass over this? The branch prediction override stuff is still disabled for now, although I think it could potentially work with the BE fix and some RTL changes. But we should punt it for now to derisk for tapeout

bp_fe/src/v/bp_fe_pc_gen.v
```diff
@@ -47,228 +46,165 @@ module bp_fe_pc_gen
...
+        pc_v_if1_r      <= '0;
+        pc_v_if2_r      <= '0;
+        btb_v_if1_r     <= '0;
+        itlb_miss_if2_r <= '0;
+      end
+    else
+      begin
+        pc_v_if1_r      <= pc_v_if1_n;
+        pc_v_if2_r      <= pc_v_if2_n;
+        btb_v_if1_r     <= btb_br_tgt_v_lo;
```
farzamgl: > pc_v_if1_r      <= pc_v_if1_n;
        pc_v_if2_r      <= pc_v_if2_n;
        btb_v_if1_r     <= btb_br_tgt_v_lo;

should these be also under `if(state_n == e_run)` ?

bp_fe/src/v/bp_fe_pc_gen.v
```diff
@@ -311,11 +250,50 @@ instr_scan
...
+        pc_gen_queue.msg.exception.vaddr          = pc_if2_r; 
+        pc_gen_queue.msg.exception.exception_code = misalign_exception
+                                                    ? e_instr_misaligned
+                                                    : itlb_miss_exception
+                                                      ? e_itlb_miss
+                                                      : e_illegal_instr;
+      end
+    else 
+      begin
+        pc_gen_queue.msg.fetch.pc                  = icache_pc_gen.addr;
```
farzamgl: Also, set `pc_gen_queue.msg_type` here.

bp_fe/src/v/bp_fe_pc_gen.v
dpetrisko: Nope, otherwise when we stall we could end up with stages being always valid

bp_fe/src/v/bp_fe_pc_gen.v
dpetrisko: Good catch! I assume it works right now because the exception message type happens to be zero

bp_fe/src/v/bp_fe_pc_gen.v
farzamgl: You are right

PR #174, opened 2019-05-20
ME Directory Reorganization

muwyse:
This change reorganizes the coherence directory to improve physical design. Top and ME regressions all pass (except ME dir test, which was not updated and is known to fail).

Key changes are:

1. Directory RAM now stores way-groups across multiple rows; reading the way-group takes multiple cycles and tag comparisons are performed per RAM row to reduce the amount of comparison hardware required. CCE microcode, directory, and gad modules underwent significant changes (largely shifting functionality location, and adding multi-cycling of the way-group reads to directory).

2. Removed way-group and cache block registers from the CCE.

3. Fixed bug in bp_tile related to signal connections on the LCE Data Response network.

4. Fixed bug in Mock LCE related to LRU way use and tracking in the context of upgrade misses.


dpetrisko:
Horse has left the barn on this one, but in future before going into me_dev, rebase into 2-3 major commits, to keep history nicer

PR #175, opened 2019-05-21
Modify logging to support multicore

HelloKayT:
Small tweak in the back-end nonsynth tracer, so each core prints to a
separate debug log file

### DIFF
--- a/bp_be/test/common/bp_be_nonsynth_tracer.v
+++ b/bp_be/test/common/bp_be_nonsynth_tracer.v
@@ -7,7 +7,7 @@ module bp_be_nonsynth_tracer
    `declare_bp_proc_params(cfg_p)
 
    // Default parameters
-   , parameter debug_file_p = "debug.log"
+   , parameter debug_file_p = "debug"
 
    // Calculated parameters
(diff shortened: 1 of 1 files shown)


taylor-bsg:
Is there a way to tell the interleaving of events between the files m? This is often useful, but requires that you be pretty strict about formatting.

HelloKayT:
Unfortunately not. The single file as it is set up right now though is pretty much infeasible for debugging, because the two cores appear to be overwriting each other in the file. For example, the first instruction that was showing up as committed for core 0 was halfway through main.

dpetrisko:
The long-term way to support this would be to only instantiate one logger that probes into all cores and can order events by core.  Right now, the bind is per core.  I'll make an issue to change it, but this should let Katie debug in the meantime

taylor-bsg:
I’m all for making short term tweaks to make progress! In terms of the code, the previous code should work for single file tracing if you hoist the file open out of the bind statement and use the same file handle across the binds. We do this in the manycore and it works fine in VCS, I guess because it visits binds in the looping order of the generate statements.

dpetrisko:
Ohh, I think I see. Cool! I’ll take a look at that at some point

PR #177, opened 2019-05-21
Upgrade demos with atomics

HelloKayT:
Added a basic test (hello_world_atomic) where cores use amoadd to increment a shared counter and print out their ID. Also made a version of the queue demo (atomic_queue_demo) that uses amoswap to lock the shared queue. Both of these were tested up to 8 cores.

This PR also include modifications to the start code to properly support multicore as every core needs its own emulation stack as well as program stack. The trap vector code was also modified to save stack pointers appropriately to reflect this change

### DIFF
--- a/bp_common/test/Makefile.frag
+++ b/bp_common/test/Makefile.frag
@@ -1,13 +1,16 @@
 BP_DEMOS = \
-  hello_world        \
-  atomic_queue_demo  \
-  basic_demo         \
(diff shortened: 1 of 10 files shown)


dpetrisko:
Can we add the atomic_demo_2, atomic_demo_4, ... thing like we did for queue_demo? Also remove reloc_queue_demo from the makefile and add a comment in the test file that it's not support

PR #181, opened 2019-05-24
Implement Credit Flow Control in D$ for Uncached Stores

muwyse:
This change implements the credit flow control for uncached stores in the D$ LCE. It adds the Unached Store Done command from the CCE to the LCE to signal when the LCE can consider an uncached store to be completed.

CCE microcode is updated to reflect the new command and requirement to send a message to the LCE on uncached store completion.

Top and ME regressions pass; DC does not due to a width mismatch in bsg_flow_counter. See this pull request against BaseJump STL for a proposed fix (https://github.com/bespoke-silicon-group/basejump_stl/pull/55)


muwyse:
Note that only 8-byte uncached access is supported. The memory and LCE currently disagree on who is responsible for extracting the requested bytes from the 64-bit dword that the request resides in. Currently, both the backing memory and LCE do this extraction, so the LCE will get 0 data if it tries to access anything that is not 64-bit aligned.

I suggest modifying the LCE to blindly consume the data from the memory/CCE and have the LCE assume that a request of N bytes (N = 1, 2, 4, or 8) is returned in the least significant N bytes of the LCE data command.

dpetrisko:
To add to Mark's comment, there are misalignment implications for both approaches, but we need to decide and be consistent.

taylor-bsg:
I don’t know the context especially well, but what has worked well in the
past is specify a bit mask of bytes that should be read or written. Caches,
write buffers, dma engines and dram controllers all desire this format.

M

dpetrisko:
Looks good to me. A couple issues I raised based on this:

https://github.com/black-parrot/pre-alpha-release/issues/182
Generating binaries instead of ROMs so that we don't have these huge diffs for ucode changes

https://github.com/black-parrot/pre-alpha-release/issues/183
Getting the alignment stuff nailed down for uncached accesses

PR #194, opened 2019-05-29
Add uncached startup mode to D$

muwyse:


taylor-bsg:
General thoughts:

- seems like we should have two structs (a _sif interface) for the config packets?
- is it possible to have a modeless implementation, where the implementation just passes through un cached traffic before it starts getting reset messages from the CCE? Then we don't need to have the config logic at all.

dpetrisko:
> seems like we should have two structs (a _sif interface) for the config packets?

It'll only be a w_v_i, addr_i, data_i in the end, since bsg_tag is broadcast and unidirectional.

> is it possible to have a modeless implementation, where the implementation just passes through un cached traffic before it starts getting reset messages from the CCE? Then we don't need to have the config logic at all.

My opinion: It's possible, but I think it's better to treat the alt-modes as chicken bits.  We're only planning on using this functionality as a fallback if things go wrong, so they should be far away from our regular state machine. Another reason: what if the CCE has a bug in such a way that it spits out a reset packet on boot?  Now our uncached mode doesn't work.

taylor-bsg:
Config structs: No, I am saying everywhere in BlackParrot, we should use a _sif struct for the config structs -- this is just consistent with the rest of the design. Without a struct, it is a lot of unnecessary signal routing, and putting it in a struct allows us to change what signals we route in there; including whether outputs are supported or not.  And we should have a standard decoder module at the leaves so we don't have to have all of that replicated code.

bsg_tag versus config: we are actually routing config over the wormhole network, so it doesn't have the restrictions imposed by bsg_tag. 

Bugs:  Bugs appear in 1) unexercised code paths in gate-level simulation 2) clock domain crossings (which you can't test) 3) backend flow and 4) communication to the FPGA that we did not test. Chicken bits being the foremost example of things people don't test -- simulating the superposition of all chicken bits / mode bits over all benchmarks. 

A more unified code path way to do it would be to say, okay, as part of bootup code, the processor does some uncached memory operations to say "I'm alive", which the FPGA (in simulation or in emulation) sees, after which the FPGA sends the code to the directory, and tells it to enable caching.  The LCE will stall on a cached access if it encounters one until it receives a wakeup command from the CCE. Just brainstorming with you guys.

BUT: if you want it to act like a chicken bit, then I suggest that we use bsg_tag for this. It is purpose designed for chicken bits.

muwyse:
Testing in progress, do not merge yet

dpetrisko:
Just to sync -- we're doing start in uncached mode, transition to cached mode on sync command from CCE, now, right?

(This will require BE changes, too, namely the frozen signal routed to the director, preventing it from booting until it's low)

taylor-bsg:
I think it’s Mark W’s call as to whether this is difficult or not.

One issue we briefly mentioned is that it is not easy for the FE or BE to
detect whether an address is uncached, so it might not be possible to stall
on a cached access waiting for the LCE to be ready. I guess maybe it is the
LCE that has to stall, and the BE is just clueless?

What is the state of the FE doing uncached accesses?

M

dpetrisko:
> One issue we briefly mentioned is that it is not easy for the FE or BE to
> detect whether an address is uncached, so it might not be possible to stall
> on a cached access waiting for the LCE to be ready. I guess maybe it is the
> LCE that has to stall, and the BE is just clueless?

Yes, that's right. There are two solutions here, depending on how we're using the uncached mode.

1. Have the LCE stall on cached access until cache is enabled by sync
2. Raise an instr/load/store exception upon cached access in uncached mode.

I think 2 is cleaner, because it lets us delegate to software either to halt or communicate with the host to figure out what to do.

taylor-bsg:
Oh I see — in the later case, the request is never presented to the LCE to
stall on — it is more like a memory access exception.

muwyse:
> Just to sync -- we're doing start in uncached mode, transition to cached mode on sync command from CCE, now, right?

No, I think doing the live transition is too complicated to make happen right now. The I$, D$, and CCE all have a mode bit / register that will get set by the config link while freeze_i is high. Once freeze goes low, the system will be stuck in either an uncached only mode or a normal operation mode. The LCEs will not raise the ready_o signal to the FE/BE until after freeze goes low. In uncached mode, ready_o is raised immediately and all accesses need to be to uncached memory. In normal mode, ready_o will be raised after the CCEs/LCEs finish the init and sync sequence. As long as we can set the initial PC appropriately we should be okay.

> What is the state of the FE doing uncached accesses?

I'm adding uncached access support to the I$. I made most of the changes yesterday, but I need to test them. If we can compile the benchmarks to run entirely in uncached memory, then I think that would be sufficient.

> 2. Raise an instr/load/store exception upon cached access in uncached mode.
> I think 2 is cleaner, because it lets us delegate to software either to halt or communicate with the host to figure out what to do.

Dan, I might need to sync with you on this. But, if we handle the uncached/cached modes as I describe above, this also might not be a problem we need to worry about right now.

dpetrisko:
> No, I think doing the live transition is too complicated to make happen right now.

Gotcha.

>As long as we can set the initial PC appropriately we should be okay.
>But, if we handle the uncached/cached modes as I describe above, this also might not be a problem we need to worry about right now.

If we're doing the config link setup, then the whole program has to run in uncached mode, all instr/load/stores for the entirety of the program, since we can't adjust config link after frozen is lowered.  So we do need to fault on uncached accesses, since they're now illegal.  It shouldn't be too hard to add, though, and we'll want that support for when we eventually add the uncached->cached on sync.

muwyse:
> If we're doing the config link setup, then the whole program has to run in uncached mode, all instr/load/stores for the entirety of the program, since we can't adjust config link after frozen is lowered. So we do need to fault on uncached accesses, since they're now illegal. It shouldn't be too hard to add, though, and we'll want that support for when we eventually add the uncached->cached on sync.

Not sure I followed all of that. When freeze goes low, BP will be either in uncached only mode or normal mode. In uncached mode, yes, the entirety of the program needs to be in uncached memory. In this mode, do you mean we need to fault on _cached_ access?

Does the PC get set by the config link? If so, when we are configuring the LCEs and CCEs, the config link can also set the PC to a cached or uncached address appropriately, based on what mode we are going to put the processor in.

dpetrisko:
> In this mode, do you mean we need to fault on cached access?

Whoops, yep.

> Does the PC get set by the config link?

Yes

>If so, when we are configuring the LCEs and CCEs, the config link can also set the PC to a cached or uncached address appropriately, based on what mode we are going to put the processor in.

We can do that, but there's still a chance that it accesses instr/data outside of the uncached range once the program is running

taylor-bsg:
As long as we don't have speculative memory accesses, this should not
happen? Which brings us to -- does the i-cache speculate?

M

dpetrisko:
1) pc_gen is speculative. Could go into uncached territory, although unlikely if we do something like relocate to 0x0. 

2) d$ is not not speculative, so we shouldn’t do loads/stores in cached unless the program is buggy. 

3) .... but, we’re talking about user mode programs here, not machine mode code. So there’s no reason to trust the program to be correct. 

4) ... except if we’re saying that, for this tapeout, we’re going to control the program so that this isn’t something to worry about.

PR #202, opened 2019-06-12
increase buffering for LCE and Mem Response messages

muwyse:
This change increase the amount of buffering at each CCE for LCE and MEM Response message networks. This buffering is required to prevent the CCE / system from potentially deadlocking due to these networks filling up. Messages on these networks are sunk by the CCE in a delayed manner (exactly when depends on what the CCE microcode is doing).

The CCE may be unable to immediately sink a memory response because it is processing another request. The maximum number of outstanding memory requests is capped at the number of way-groups managed by the CCE (enforced by the pending bit).

The CCE may be unable to immediately sink a transfer or coherence ack (LCE Responses) because it is processing another request. The maximum number of these acks that may be outstanding prior to the CCE entering a state in which in can sink LCE Responses is the number of way-groups managed by the CCE.

This buffering is needed to safely allow concurrent request processing in the CCE.

### DIFF
--- a/bp_me/src/v/cce/bp_cce_top.v
+++ b/bp_me/src/v/cce/bp_cce_top.v
@@ -6,6 +6,20 @@
  * Description:
  *   This is the top level module for the CCE.
  *
+ * Notes:
+ *   All inputs from the LCE are buffered. LCE Responses require a FIFO that can hold up to
+ *   N responses where N is the number of way groups managed per CCE. Future optimizations could
+ *   lessen this requirement to a 1 or 2 element FIFO. All other inputs use 2-FIFOs.
+ *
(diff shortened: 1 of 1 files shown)


dpetrisko:
Looks fine to me. Roughly how big are the additional buffers (in bits)? We'll probably need to harden the data_resps

muwyse:
LCE Response message is 43 bits wide
Mem Response message is 151 bits wide

How many entries per the two FIFOs depends on number of cores/tiles/cce's:
8 core = 64/8 = 8 entries
- LCE Response = 8*43 = 344 bits
- Mem Response = 8*151 = 1,208 bits
16 core = 64/16 = 4 entries
- LCE Response = 4 *43 = 172 bits
- Mem Response = 4*151 = 604 bits

PR #205, opened 2019-06-15
CCE: fix lce data command for uncached data in normal execution mode

muwyse:
This change fixes a bug related to determining if the data message returning from memory is for a cached or uncached request. Basically, the current MSHR in the CCE is not guaranteed to be related to the returning memory data response so the CCE should instead examine the data response itself to form the proper LCE Data Command message.

### DIFF
--- a/bp_me/src/v/cce/bp_cce_msg.v
+++ b/bp_me/src/v/cce/bp_cce_msg.v
@@ -365,7 +365,7 @@ module bp_cce_msg
       // Data is copied directly from MemDataResp
       // For uncached responses, only the least significant 64-bits will be valid
       lce_data_cmd.data = mem_data_resp.data;
-      if (mshr.flags[e_flag_sel_ucf] == e_lce_req_non_cacheable) begin
+      if (mem_data_resp.non_cacheable == e_lce_req_non_cacheable) begin
         lce_data_cmd.msg_type = e_lce_data_cmd_non_cacheable;
         lce_data_cmd.way_id = '0;
       end else begin


dpetrisko:
Trying to understand how the MSHR wouldn't be related to the returning memory data. Is it because they can return out of order?

dpetrisko:
In that case, what metadata _is_ guaranteed to be valid in the MSHR?

muwyse:
Sorry, my comment wasn't clear, let me clarify. 

To answer your questions: There is no guarantee that any data in the CCE's MSHR is valid in the scope of the returning Mem Data Response. (The MSHR likely holds valid state belonging to a completely unrelated LCE Request, but it should be assumed the CCE processed other requests while waiting for the original request to go to memory). There is no assumption about the order memory returns data; the CCE doesn't care and assumes the L1 caches handle memory consistency ordering.

The MSHR is the CCE's internal state for the current LCE request it is currently processing locally. After the CCE sends a Mem Cmd to request a cache block from memory, the CCE potentially moves on to process a new LCE request, which would result in resetting and repopulating the MSHR for this new, unrelated request.

At some later point in time when the Mem Data Response returns to the CCE there is no guarantee that the CCE's current MSHR state is in any way related to the Mem Data Response that just arrived. Fortunately, the Mem Data Response contains all the information needed to forward the data block to the correct LCE. The Mem Data Response should be checked to determine if it is a response to a cached or uncached request to set the LCE Data Command fields properly.

dpetrisko:
Makes sense!

PR #206, opened 2019-06-18
Add mem_payload_width to bp_proc_params and fix CCE microcode to avoid deadlock in sync

muwyse:
This contains two changes:
1. Add mem_payload_width_p to bp_proc_params struct that defines the size of payload field in the CCE-MEM interface to hold the CCE's MSHR. This keeps the payload size and contents opaque. The CCE casts the bits into the MSHR format used internally.

2. A small fix to CCE microcode to avoid deadlock during startup/sync. Previously, the LCE Response buffer could fill up if the number of LCEs was greater than the number of way groups managed by each CCE. This results in deadlock during the sync routine between the CCEs and LCEs. The CCE will now wait for the sync ack after each sync command. This adds some latency to the sync sequence at startup (only time it happens), but avoids having to double or quadruple the size of the LCE Response buffer at each CCE. The buffer is instead sized for the maximum number of response messages that could arrive during normal operation (i.e., after startup/sync)

### DIFF
--- a/bp_common/src/include/bp_common_aviary_defines.vh
+++ b/bp_common/src/include/bp_common_aviary_defines.vh
@@ -66,6 +66,8 @@ typedef struct packed
   integer instr_width;
   integer reg_addr_width;
   integer page_offset_width;
+
+  integer mem_payload_width;
 }  bp_proc_param_s;
 
 `define declare_bp_proc_params(bp_cfg_e_mp) \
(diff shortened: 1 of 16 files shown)


dpetrisko:
Looks good to me.  Once I debug and merge in the new uncached I/O stuff I'll approve this.  There shouldn't be a conflict between these ucode changes and the ones for uncached stuff, right?

PR #211, opened 2019-06-22
Use generalized wormhole router in LCE-CCE Data networks

muwyse:
This change uses the bsg_wormhole_router_generalized module in the LCE-CCE Data Command and Response networks within the bp_tile. Routing is StrictYX, route Y first. This allows N & S ports to send messages to E & W, in addition to P, since the LCEs and CCEs acting as sources are connected to N & S.

This change also includes a bug fix to stub the LCE Data Response link_i connection at x=1 on the Proc port, which was previously left floating. This is required because the CCE only connects to the router at x=0. The lce_req_link_i_stitch[1][P] and lce_resp_link_i_stitch[1][P] are also stubbed to 0. Previously the ready_and_rev wire of these signals were tied to the CCE ready_lo signals, but the CCE does not exist at x=1, so the links should be fully stubbed.

regress.v and regress_beebs.v both pass

### DIFF
--- a/bp_top/src/v/bp_tile.v
+++ b/bp_top/src/v/bp_tile.v
@@ -12,6 +12,7 @@ module bp_tile
  import bp_cce_pkg::*;
  import bsg_noc_pkg::*;
  import bp_cfg_link_pkg::*;
+ import bsg_wormhole_router_pkg::StrictYX;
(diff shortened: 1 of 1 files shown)


taylor-bsg:
Cool!

PR #214, opened 2019-07-09
move wormhole conversion for CCE-MEM interface into bp_tile and bp_clint

muwyse:
regress.v passes,  currently running regress_beebs.v and atomic queue demos


dpetrisko:
looks good to me

muwyse:
atomic queue demos look good (2, 4, 8, and 16-core ); beebs regressions passing (a couple tests still running, but no failures yet, so expecting to pass)

PR #230, opened 2019-07-24
LCE-CCE and CCE-MEM Interface Update

muwyse:
This change is a major refactor of the coherence networks in BlackParrot to reduce the number of physical networks. Logically, and at a high level, the operation is nearly identical to the previous design.

LCE-CCE Interface is now 3 networks: Request, Command, Response
CCE-MEM Interface is now 2 networks: Command, Response

Currently, FSM CCE and Mock LCE have been refactored for these changes, along with the CLINT and ME Wormhole Master/Client.

Most of the modules in bp_top still need refactoring for these changes.

Do not merge. Work in progress.


dpetrisko:
Nice! I'll take a look tomorrow and see about making the top changes if everything makes sense to me

bp_common/src/include/bp_common_me_if.vh
```diff
@@ -19,52 +19,46 @@
...
+ *
+ *
+ * Clients should use the declare_bp_me_if() macro to declare all of the interface structs at once.
+ *
  */
 
+
+
+/* TODO list
+
```
dpetrisko: For the first point, do you mean requiring cache block aligned transfers? I think that's a reasonable requirement

Why would cache block size being a multiple of flits help in the LCE-CCE-MEM case? I would think that the 'extra' flit data would just be zeros

bp_common/src/include/bp_common_me_if.vh
```diff
@@ -149,21 +134,34 @@ typedef enum bit [1:0]
...
-    logic [`BSG_SAFE_CLOG2(lce_assoc_mp)-1:0]    lru_way_id;    \
-    bp_lce_cce_lru_dirty_e                       lru_dirty;     \
-    bp_lce_cce_req_non_cacheable_e               non_cacheable; \
-    bp_lce_cce_nc_req_size_e                     nc_size;       \
+  }  bp_lce_cce_req_uc_req_s
+
+`define declare_bp_lce_cce_req_s(num_cce_mp, num_lce_mp, msg_width_mp) \
+  typedef struct packed                                         \
+  {                                                             \
+    logic [msg_width_mp-1:0]                     msg;           \
```
dpetrisko: Can we make this a packed union?

It might be helpful to write this in the style of fe_be_if, where it's just a monolithic declare, if there's more than one parameterizable struct

bp_common/src/include/bp_common_me_if.vh
```diff
@@ -183,105 +181,123 @@ typedef enum bit [1:0]
...
+  //                          3'b101 - potentially dirty, not owned, shared (not exclusive)
+  ,e_COH_M                  = 3'b110 // Modified - potentially dirty, owned, not shared (exclusive)
+  ,e_COH_O                  = 3'b111 // Owned - potentially dirty, owned, shared (not exclusive)
+} bp_coh_states_e;
+
+`define bp_coh_shared_bit 0
+`define bp_coh_owned_bit 1
+`define bp_coh_dirty_bit 2
+
+`define bp_coh_bits $bits(bp_coh_states_e)
```
dpetrisko: Nit, but I don't think we need a macro for width on enums, since they can never be parameterizable

dpetrisko: (so $bits will always be sufficient)

bp_common/src/include/bp_common_me_if.vh
```diff
@@ -183,105 +181,123 @@ typedef enum bit [1:0]
...
+ *  msg_type:      is the command
+ *  addr:          specifies the memory address associated with the command
+ *  way_id:        is the way within the set that addr maps to in the LCE that should be used for the command
+ *  state:         specifies the Coherence State to be used for invalidate, set_tag, and set_tag_wakeup
+ *  target:        is the LCE that will receive a transfer for a transfer command
+ *  target_way_id: is the way within the target LCE's set (computed from addr) to fill the data in to
  */
-`define declare_bp_cce_lce_cmd_s(num_cce_mp, num_lce_mp, addr_width_mp, lce_assoc_mp) \
+`define declare_bp_lce_cmd_cmd_s(num_cce_mp, num_lce_mp, addr_width_mp, lce_assoc_mp) \
   typedef struct packed                                         \
```
dpetrisko: Don't we still need dst_id to determine which LCE to send the command to?

bp_common/src/include/bp_common_me_if.vh
```diff
@@ -183,105 +181,123 @@ typedef enum bit [1:0]
...
- *  bp_lce_data_cmd_s is used to send cache block data from CCE to LCE or from LCE to LCE.
+ *  bp_lce_cmd_data_s is used to send cache block data from CCE to LCE or from LCE to LCE.
  *
- *  dst_id:   LCE receiving the cache block data
- *  msg_type: indicates whether this is from CCE or from another LCE. LCE uses this field to decide
- *            what type of lce_resp to send. or whether this is non_cacheable data_cmd from CCE.
  *  way_id:   the way within the receiving LCE's target set to fill the data in to.
  *  data:     the cache block data
  */
+`define declare_bp_lce_cmd_data_s(data_width_mp) \
```
dpetrisko: I'd vote for just having a raw data field instead of a struct

bp_common/src/include/bp_common_me_if.vh
```diff
@@ -361,22 +366,64 @@ typedef enum logic
...
  *
  * CCE-Memory Interface
  *
  */
 
+/*
+ * bp_cce_mem_cmd_type_e specifies the memory command from the CCE
+ */
+typedef enum bit [2:0]
+{
```
dpetrisko: nit, no spaces ,e_cce_mem_rd

bp_common/src/include/bp_common_me_if.vh
```diff
@@ -361,22 +366,64 @@ typedef enum logic
...
+  } bp_lce_cce_resp_s
+
 
 /*
  *
  * CCE-Memory Interface
  *
  */
 
+/*
```
dpetrisko: Can we move this to ME? It's not technically a 'common' interface

bp_common/src/include/bp_common_me_if.vh
```diff
@@ -406,65 +453,16 @@ typedef enum logic
...
-    logic [addr_width_mp-1:0]                    addr;     \
-    bp_cce_mem_cmd_payload_s                     payload;  \
-    bp_lce_cce_req_non_cacheable_e               non_cacheable; \
-    bp_lce_cce_nc_req_size_e                     nc_size;       \
+`define declare_bp_cce_mem_cmd_s(addr_width_mp, data_width_mp)  \
+  typedef struct packed                                         \
+  {                                                             \
+    logic [data_width_mp-1:0]                    data;          \
+    bp_cce_mem_cmd_payload_s                     payload;       \
+    bp_lce_cce_uc_req_size_e                     uc_size;       \
```
dpetrisko: Two thoughts:
1) Let's not use any lce_cce fields in these structs
2) Possibly uc_size is the wrong concept? For example: If we end up doing hierarchical coherence, we may need to send a req size to the next level cache depending on the block size.  So just req_size?

bp_top/src/v/bp_clint.v
```diff
@@ -341,36 +334,18 @@ wire [dword_width_p-1:0] rdata_lo = plic_cmd_v
 
 // Possibly unnecessary, if we convert the WH adapter to buffer in both directions,
 //   rather than sending in 1 cycle and receiving in the next
```
dpetrisko: Can we definitely get rid of this?

bp_common/src/include/bp_common_me_if.vh
muwyse: I mean having the first bit of the data be the first bit of some flit. The size of flits doesn't matter a whole lot actually, but I like powers of two :)

What I think we don't want is having the first bit of data be, e.g., bit 5 in a flit. The headers for data messages between LCE-CCE and CCE-MEM are likely to be different, so we don't really want any flit to contain both header and data info. We want some header flit(s), then data flits, where we can easily strip the header flits, create a new outbound header, then pass the data flits along effectively without touching them.

bp_common/src/include/bp_common_me_if.vh
muwyse: I think that requires adding padding into the structs that get union-ed, correct, because all members of the union must be the same size?

muwyse: A monolithic declare would probably be a good thing to do. Most of the structs are parameterized.

bp_common/src/include/bp_common_me_if.vh
muwyse: The Command network carries both data commands and regular commands. We could name the field "data" if we want, but the regular commands would still use those bits as the struct bp_lce_cmd_cmd_s above. We could also leave the field named as msg, and forget about the two data only structs, using the msg bits directly.

bp_common/src/include/bp_common_me_if.vh
muwyse: yeah, it is in the bp_lce_cmd_s since it is needed for both regular and data commands.

bp_common/src/include/bp_common_me_if.vh
dpetrisko: Oh yep, definitely agree!  For ease of implementation, maybe standardize 1 header flit padded to data flit size (e.g. width of link). Do you think we need any headers > 32 bits?

bp_common/src/include/bp_common_me_if.vh
dpetrisko: Yeah, we would have to add padding fields to the declaration.  However, we don't have to transmit those empty bits, so it's essentially the same as we have now. The nice thing would be then we don't have to have the msg_widths as part of the struct declaration.  Just the adapters would have to know about the unpadded widths.

Maybe we can do the monolithic declare now and think more about packed unions?

bp_common/src/include/bp_common_me_if.vh
dpetrisko: Oh, I see. (Another thing that could be cleared up by a union?)

>  We could also leave the field named as msg, and forget about the two data only structs, using the msg bits directly.

Yeah, I think I like this better.

bp_top/src/v/bp_clint.v
muwyse: Yeah, I think so. Mem Cmd and Mem Data Cmd are both on one network, and Mem Resp and Mem Data Resp are both on one network. So only the single buffer is needed. We don't differentiate between the two logical message types here. We still should think about how to potentially eliminate buffering throughout LCE-CCE-MEM.

bp_common/src/include/bp_common_me_if.vh
muwyse: On the Command network we should be less than 32 bits. On the Response network, there is an address field, which is 39-bits by default so we probably need two 32-bit flits for the header. It might be possible to get rid of the address field though. For some messages the CCE might not care, and for others it might know what address the response is for based on its (CCE's) state.

dpetrisko: What could the address be used for? Would it make sense to encode that behavior as metadata?

bp_top/src/v/bp_clint.v
dpetrisko: Oh, I didn't see that you left the other buffer.  Yep, makes sense.  Agree about general buffer elimination

bp_me/src/include/v/bp_cce_pkg.v
```diff
@@ -38,12 +38,12 @@ package bp_cce_pkg;
     logic [`BSG_SAFE_CLOG2(lce_assoc_mp)-1:0]     tr_way_id;              \
     logic [`bp_coh_bits-1:0]                      next_coh_state;         \
     logic [`bp_cce_inst_num_flags-1:0]            flags;                  \
-    logic [`bp_lce_cce_uc_req_size_width-1:0]     uc_req_size;            \
+    logic [$bits(bp_lce_cce_uc_req_size_e)-1:0]   uc_req_size;            \
```
dpetrisko: Just use the enum type here

dpetrisko:
Looks good to me!

bp_common/src/include/bp_common_me_if.vh
```diff
@@ -344,14 +312,12 @@ typedef enum bit [1:0]
...
  */
-
 typedef enum logic
 {
-  e_lce_resp_wb              = 1'b0  // Normal Writeback Response (full data)
-  ,e_lce_resp_null_wb        = 1'b1  // Null Writeback Response (no data)
-} bp_lce_cce_resp_msg_type_e;
+  e_lce_cce_resp_wb              = 1'b0  // Normal Writeback Response (full data)
+  ,e_lce_cce_resp_null_wb        = 1'b1  // Null Writeback Response (no data)
+} bp_lce_cce_resp_data_msg_type_e;
```
dpetrisko: Can we collapse this into the resp_msg_type?  It won't require any additional bits, but I think it's confusing to have two msg type fields in one packet

dpetrisko: Same comment on cmd_msg_type

muwyse: agreed. I was just coming to the same conclusion while updating the CCE implementation.

PR #237, opened 2019-08-08
Adding generalized bp_mem

dpetrisko:
I consolidated our test memories into a more generalized form:
1) transducer, sits between mem_cmd/mem_resp and normal addr, data, w, v buses
2) storage, able to read from a hex file or not.  Handles dram offset, so we can just read in hex files
3) delay model, has fixed latency, random latency and dramsim2 models

~Current issue -- Verilator doesn't like the system task for substituting out hex base address, not sure what a good solution is for that.  I'd rather have it in the Verilog than the Makefile, because we need to refer to the parameterization to know what base address is~

Works for both VCS and Verilator


dpetrisko:
Appreciate a code review, and also if you could port your tests to use bp_mem instead of bp_mem_dramsim2, that would be super

PR #243, opened 2019-08-27
Bugs found during linux boot so far

farzamgl:
Here are some of the hardware bugs found so far in the Linux's booting process:

- incorrect sstatus read mask
- incorrect enable signal for issue posion register in calculator
- not trapping on illegal CSR instructions like rdtime
- flushing FE pipe on invalid itlb misses


farzamgl:
Couldn't that lead to missing the redirect command?

PR #246, opened 2019-09-04
Merge me_dev into dev

muwyse:
This PR updates the LCE-CCE Interface so all coherence state transitions are atomic (i.e., the occur at the LCE with the arrival of a single message). It also updates one of the ME tests and makes it work with the new fake memory.

The bp_cce test in the ME is updated for changes in top level and memory, and it is added to the regressions to prevent breaking with non-ME changes. The test is not relied on for functionality since it only executes 16 loads/stores total, but is rather used as a basic sanity check.


bp_me/src/include/c/bp_common_me_if.h
```diff
@@ -109,6 +109,7 @@ typedef enum {
```
dpetrisko: So one thing I was thinking about, that I saw somewhere else.  It should be totally possible to generate the C headers from the Verilog headers, just with a grep.  Or a python script if it turns out to be difficult
#define -> `define, etc.

bp_me/src/include/v/bp_me_cce_mem_if.vh
```diff
@@ -113,7 +114,7 @@ typedef enum bit [2:0]
 
 // CCE-MEM Interface
 `define bp_cce_mem_cmd_payload_width(num_lce_mp, lce_assoc_mp) \
-  (`BSG_SAFE_CLOG2(num_lce_mp)+`BSG_SAFE_CLOG2(lce_assoc_mp))
+  (`BSG_SAFE_CLOG2(num_lce_mp)+`BSG_SAFE_CLOG2(lce_assoc_mp)+`bp_coh_bits)
```
dpetrisko: bikeshed, but I think $bits(bp_coh_states_e) everywhere is better.  Up to you, though

bp_me/src/v/cce/bp_cce_fsm.v
```diff
@@ -889,13 +899,17 @@ module bp_cce_fsm
         lce_cmd_cmd.addr = mshr_r.paddr;
         lce_cmd_cmd.target = mshr_r.lce_id;
         lce_cmd_cmd.target_way_id = mshr_r.lru_way_id;
+        lce_cmd_cmd.state = mshr_r.next_coh_state;
 
         // Assign Command subtype to msg field
         lce_cmd.msg.cmd = lce_cmd_cmd;
 
-        state_n = (lce_cmd_ready_i) ? TRANSFER_ST_CMD : TRANSFER_CMD;
+        //state_n = (lce_cmd_ready_i) ? TRANSFER_ST_CMD : TRANSFER_CMD;
```
dpetrisko: Get rid of commented code if it no longer works.  Add explanation if it might be used in future

bp_me/src/v/cce/bp_cce_fsm.v
```diff
@@ -889,13 +899,17 @@ module bp_cce_fsm
...
+        state_n = (lce_cmd_ready_i) ? TRANSFER_WB_CMD : TRANSFER_CMD;
+        end
       end
       TRANSFER_ST_CMD: begin
+        $error("TRANSFER_ST_CMD");
```
dpetrisko: ?

bp_me/src/v/cce/bp_cce_fsm.v
```diff
@@ -987,15 +1005,18 @@ module bp_cce_fsm
           mem_cmd.size = e_mem_size_64;
           mem_cmd.payload.lce_id = mshr_r.lce_id;
           mem_cmd.payload.way_id = mshr_r.lru_way_id;
+          mem_cmd.payload.state = mshr_r.next_coh_state;
 
-          state_n = (mem_cmd_ready_i) ? SEND_SET_TAG : READ_MEM;
+          //state_n = (mem_cmd_ready_i) ? SEND_SET_TAG : READ_MEM;
```
dpetrisko: Same as above

bp_me/src/v/cce/bp_cce_top.v
```diff
@@ -31,6 +31,8 @@ module bp_cce_top
   #(parameter bp_cfg_e cfg_p = e_bp_inv_cfg
     `declare_bp_proc_params(cfg_p)
 
+    , parameter cce_trace_p = 0
```
dpetrisko: Shouldn't need this if the bind is in testbench and parameter is surrounding the statement (https://github.com/black-parrot/pre-alpha-release/blob/tapeout_bp_rc1/bp_top/test/tb/bp_top_trace_demo/testbench.v#L213)

Should be able to do that in your own testbenches.  Otherwise it can cause confusion for the backend flow, propagating parameters which can make the module nonsynth

bp_me/test/common/bp_mem_storage_sync.v
```diff
@@ -44,6 +44,8 @@ import "DPI-C" context function string rebase_hexfile(input string memfile_name
 string hex_file;
 always_ff @(posedge clk_i)
   begin
+    // TODO: need a mechanism to zero-init memory. This can be done using a "zero.mem" file with
```
dpetrisko: Yep, I'll add this

bp_me/test/tb/bp_cce/testbench.v
```diff
@@ -28,34 +28,38 @@ module testbench
...
    , parameter lce_perf_trace_p = 0
 
-   // Number of elements in the fake BlackParrot memory
-   , parameter clock_period_in_ps_p = 1000
-   , parameter prog_name_p = "prog.mem"
-   , parameter dram_cfg_p  = "dram_ch.ini"
-   , parameter dram_sys_cfg_p = "dram_sys.ini"
-   , parameter dram_capacity_p = 16384
+   // TODO: must supply a valid initial memory file, e.g., a file of all zeros for 1 MB of storage,
+   // or, trace replay script must only load data that was previously stored
```
dpetrisko: Does the test depend on things being zero? Or is that XXXX causes problems?

bp_common/src/include/bp_common_me_if.vh
```diff
@@ -117,6 +117,13 @@
     logic [`BSG_SAFE_CLOG2(num_cce_mp)-1:0]      src_id;                                                 \
   }  bp_lce_cmd_cmd_s;                                                                                   \
                                                                                                          \
+  typedef struct packed                                                                                  \
```
dpetrisko: Comment explaining?

dpetrisko:
Is bp_cce_dual deprecated?

dpetrisko:
I added the zero option to memory and took the zero_mem script out of me, your regression test passes

bp_me/test/tb/bp_cce/testbench.v
muwyse: The CCE tests depend on memory having an initial value of 0 for all entries. This makes writing the trace replay tests much easier, and it is also a requirement for the AXE consistency tool.

bp_me/test/common/bp_mem_storage_sync.v
muwyse: Thanks!

bp_me/src/v/cce/bp_cce_top.v
muwyse: Okay, I'll take a look. I thought I had run into an issue when it wasn't being passed down, but I'll try to replicate this, which is better.

bp_me/src/include/c/bp_common_me_if.h
muwyse: I've tried this before, but not everything is straightforward / 1:1 mapping. There are some oddities that make it more complicated than just sed or grep. So, for now I've been maintaining manually. It would be nice to auto-generate though; I'll add it to my future To Do list.

bp_fe/src/v/bp_fe_lce_cmd.v
```diff
@@ -163,6 +163,7 @@ module bp_fe_lce_cmd
           lce_cmd_yumi_lo   = lce_resp_yumi_i;
 
         end else if (lce_cmd_li.msg_type == e_lce_cmd_set_tag) begin
+          $error("set tag received in icache");
```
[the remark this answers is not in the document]
muwyse: see above.

bp_common/test/Makefile
```diff
@@ -16,6 +16,8 @@ MKLFS = $(TEST_DIR)/bin/bsg_newlib_mklfs
 
 NC ?= 1
 
+.EXPORT_ALL_VARIABLES:
```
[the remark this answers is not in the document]
muwyse: Done. The demos sources need TEST_DIR, changing Makefile to export only that.

bp_me/src/v/cce/bp_cce_top.v
dpetrisko: As caution, it won't work if your tracer module itself needs the parameter, since the parameter doesn't exist at the scope of the bound tracer

bp_be/src/v/bp_be_mem/bp_be_dcache/bp_be_dcache_lce_cmd.v
```diff
@@ -313,6 +313,7 @@ module bp_be_dcache_lce_cmd
             //  <set tag>
             //  set the tag and coherency state of given index/way.
             e_lce_cmd_set_tag: begin
+              $error("set tag received in dcache");
```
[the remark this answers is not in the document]
muwyse: With this change, the LCEs receive either a Set Tag Wakeup or Data (and Tag) command. This makes the state transition "atomic" in the LCE. Basically, the LCE receives a single message to transition states, instead of multiple messages. This may give a very minor performance boost, but it is mostly to update our implementation to the description of the protocol that we will give in papers.

bp_me/src/include/c/bp_common_me_if.h
dpetrisko: Sounds good!

bp_be/src/v/bp_be_mem/bp_be_dcache/bp_be_dcache_lce_cmd.v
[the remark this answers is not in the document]
muwyse: Both the old way and this new way are correct, and I expect performance change to be minimal. It is more about our explanation and reasoning about the protocol.

bp_me/src/include/v/bp_me_cce_mem_if.vh
muwyse: I will fix in a separate change.

bp_me/src/v/cce/bp_cce_fsm.v
muwyse: Will do.

bp_be/src/v/bp_be_mem/bp_be_dcache/bp_be_dcache_lce_cmd.v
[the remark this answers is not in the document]
dpetrisko: Ah, I see.  With wormhole routing, I don't believe there's an advantage to having both set_tag and set_data_and_tag.  But without wormhole routing, there is, since we would send an empty payload.  So a couple thoughts then:

1) We continue to support both set_tag and set_data_and_tag in the LCEs, get rid of the $error
2) We remove support for set_tag in the LCE, and pull the $error into a separate assertion at the bottom.

bp_me/src/v/cce/bp_cce_fsm.v
muwyse: Related to interface/protocol change above. The CCE will no longer send a set tag command when initiating a transfer. Instead, it sends the transfer command to the LCE that will send the block, and that command includes the information required for the sending LCE to send both data and tag in the data command (LCE to LCE transfer).

bp_common/src/include/bp_common_me_if.vh
muwyse: added.

bp_be/src/v/bp_be_mem/bp_be_dcache/bp_be_dcache_lce_cmd.v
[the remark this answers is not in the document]
muwyse: I will remove the $error statement from the LCEs, I think that makes more sense. There may be a situation in the future when we want to send a set tag command (that doesn't include data, and isn't also a wakeup command).

The ucode CCE will be able to support both the old and new protocols. The FSM CCE will need to be hardcoded to one of these, which will be the new protocol.

bp_me/src/v/cce/bp_cce_fsm.v
muwyse: removed.

bp_me/src/v/cce/bp_cce_fsm.v
muwyse: FSM CCE will only implement the new protocol. I'll remove this entire state.

bp_me/src/v/cce/bp_cce_top.v
muwyse: It works. I was using the parameter as a check inside the tracer to only trace if the parameter was set, but if we only instantiate based on the parameter, the check isn't needed!

bp_be/src/v/bp_be_mem/bp_be_dcache/bp_be_dcache_lce_cmd.v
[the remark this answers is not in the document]
dpetrisko: Works for me!

dpetrisko:
Cool, will merge once CI passes!

PR #251, opened 2019-09-11
Add BHT overriding BTB functionality

dpetrisko:
- Adding a pipeline reg struct in pc_gen
- Rather than always taking BTB entry, we now check BHT whether to 'override' this decision
- Moving riscv defines into bp_common instead of bp_be, so that other Ends can use it


dpetrisko:
No major changes.  Performance actually stays fairly similar since a 64 entry BTB is huge, probably something like SPEC would show improvement

taylor-bsg:
Maybe you can run on the specs that we have working and see the benefit?
When adding a performance optimization, part of the verification process is
confirming the anticipated benefit.

M

dpetrisko:
Yep, can do

dpetrisko:
Changes moving to larger changeset

PR #254, opened 2019-09-13
CSR cmd and response fixes

dpetrisko:
Adding immediate extension for CSR cmds. Also fixing pipeline to flush on illegal CSR instructions.  Simplifying flush logic a bit


dpetrisko:
Merging, larger refactor in progress.

PR #257, opened 2019-09-20
Add more sophisticated CI testing

dpetrisko:
How's this look? I'd like to make it flexible so we can copy for other BSG repos.

You can check out the pipeline here: https://gitlab.com/black-parrot/pre-alpha-release/pipelines/83526423 once it's finished.

This regression does riscv_tests, beebs (for both VCS/Verilator) some ME unit tests and Scott's check_synthesis script.


ci/cache.sh
```diff
@@ -0,0 +1,23 @@
+#!/bin/bash
```
stdavids: Some comments about what is going on here. Also, this feels like a place where some high-level abstractions might exist that will cross bsg-repos? Or maybe a template?

dpetrisko: I'm actually going to use the gitlab ci cache, which seems to be able to handle all these operations.  So this file will just go away :)

bp_me/syn/Makefile
```diff
@@ -90,19 +90,19 @@ lint.me:
 	@echo "ME lint not supported"
 
 regress.me.sc: dirs.sc
-	$(MAKE) build.sc sim.sc TB=bp_cce NUM_INSTR_P=$(NUM_INSTR_P)
-	$(MAKE) build.sc sim.sc TB=bp_cce_uc
-	$(MAKE) build.sc sim.sc TB=bp_cce_fsm NUM_INSTR_P=$(NUM_INSTR_P)
-	$(MAKE) build.sc sim.sc TB=bp_cce_fsm_uc
+	$(MAKE) -j 1 build.sc sim.sc TB=bp_cce NUM_INSTR_P=$(NUM_INSTR_P)
```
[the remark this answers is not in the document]
dpetrisko: It basically just disables the job server.  If there's another way to force a job to run with only 1 thread (in the recipe), I'd prefer to do it that way.  I hesitate to just to 'make', because I'd still like users to be able to specify their MAKE program if they so choose.

PR #265, opened 2019-09-27
Making itlb writes unconditional

dpetrisko:
while having fetches depend on icache_ready.  Because the fe_mem will conditionally accept the command, it needs to be valid-yumi instead of ready-valid.  This incurs no additional hardware cost, because pc_gen is already helpful, due to the wait state.

Give this a shot in Linux run before we merge it in, but it's passing regression locally


dpetrisko:
There might be a regression with beebs, I'm looking into it. I'll hold off merging, but feel free to use this in your Linux branch in the meantime

PR #268, opened 2019-10-02
Removing redundant CCE buffers from bp_tile 

dpetrisko:
These are unnecessary because the network is buffered by the wormhole adapters.  Each adapter has a two-fifo (buffer two packets), so it can connect directly to the cce.  I renamed cce_top to cce_buffered to indicate the change.


bp_top/test/tb/bp_top_trace_demo/testbench.v
```diff
@@ -153,7 +153,7 @@ wrapper
       (.clk_i(clk_i & (testbench.cmt_trace_p == 1))
        ,.reset_i(reset_i)
 
-       ,.mhartid_i('0)
+       ,.mhartid_i(proc_cfg.core_id)
```
muwyse: Should this be be_calculator.proc_cfg.core_id, similar to the bind below? Or does it not matter?

muwyse: If this is being fixed, should we also fix the filename issue inside the tracer to prevent multiple cores from opening the same file for writing?

dpetrisko:
As for the symlinks, I don’t know why they constantly break. Maybe the old version of git we have on our machine doesn’t truly support them? Is there a solution without symlink? My spitball is using the same TB and passing a parameter to determine the test, similar to PROG

bp_top/test/tb/bp_top_trace_demo/testbench.v
dpetrisko: Yeah, good call

dpetrisko: Yep, good call.  It works by coincidence, since I use the same name

dpetrisko: fixed

dpetrisko: fixed

dpetrisko:
Review addressed.  I think we should revisit the symlinks in a separate issue.  The current build system (having the test logic essentially in bp_common) is a little unwieldy

taylor-bsg:
Let’s remove symlinks. They are always unnecessary and usually result in
unhappiness.

dpetrisko:
This is going to merge in through the tapeout branch, since we wanted it there earlier.

PR #272, opened 2019-10-15
CSR related fixes

farzamgl:
Enabled SUM & MXR bits in mstatus
Fixed the global interrupt enable for higher privileges
sepc & stval sign-extension


dpetrisko:
Great finds!

Why were some of the width macros deleted?

farzamgl:
They were unused. Also we need to parameterize the csr.vh file so I was
removing the excess code.

dpetrisko:
We might use them in the future, though, and when we parameterize it the width macros will stick around. I’d say revert that bit and then go ahead and merge

PR #275, opened 2019-10-17
Fixing interrupt behaviour

farzamgl:
Fixed the latching of physical interrupt lines
Fixed CSR read/write masks
Fixed interrupt delegation condition
Interrupts have higher priority than sync exceptions

### DIFF
--- a/bp_be/src/v/bp_be_mem/bp_be_csr.v
+++ b/bp_be/src/v/bp_be_mem/bp_be_csr.v
@@ -43,7 +43,6 @@ module bp_be_csr
     , input                          timer_irq_i
     , input                          software_irq_i
     , input                          external_irq_i
-    , input [vaddr_width_p-1:0]      interrupt_pc_i
 
     , output [trap_pkt_width_lp-1:0] trap_pkt_o
 
@@ -469,8 +468,9 @@ always_comb
     mip_li.msip = software_irq_i;
     mip_li.meip = external_irq_i;
 
-    if(m_interrupt_icode_v_li | s_interrupt_icode_v_li) begin
-      if(m_interrupt_icode_v_li)
+    // Only take interrupt during nop
+    if (~exception_v_i & (m_interrupt_icode_v_li | s_interrupt_icode_v_li))
+      if (m_interrupt_icode_v_li)
         begin
           priv_mode_n          = `PRIV_MODE_M;
 
@@ -478,7 +478,7 @@ always_comb
           mstatus_li.mpie      = mstatus_lo.mie;
           mstatus_li.mie       = 1'b0;
 
-          mepc_li              = paddr_width_p'($signed(interrupt_pc_i));
+          mepc_li              = exception_pc_i;
           mtval_li             = '0;
           mcause_li._interrupt = 1'b1;
           mcause_li.ecode      = m_interrupt_icode_li;
@@ -495,7 +495,7 @@ always_comb
           mstatus_li.spie      = mstatus_lo.sie;
           mstatus_li.sie       = 1'b0;
 
-          sepc_li              = paddr_width_p'($signed(interrupt_pc_i));
+          sepc_li              = exception_pc_i;
           stval_li             = '0;
           scause_li._interrupt = 1'b1;
           scause_li.ecode      = s_interrupt_icode_li;
@@ -504,8 +504,7 @@ always_comb
           interrupt_v_o        = 1'b1;
           ret_v_o              = 1'b0;
         end
-    end
-    else if(exception_v_i & exception_ecode_v_li) begin
+    else if (exception_v_i & exception_ecode_v_li)
       if (medeleg_lo[exception_ecode_li] & ~is_m_mode)
         begin
           priv_mode_n          = `PRIV_MODE_S;
@@ -546,7 +545,6 @@ always_comb
           interrupt_v_o        = 1'b0;
           ret_v_o              = 1'b0;
         end
-    end
   end
 
 // CSR slow paths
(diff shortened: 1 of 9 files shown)


dpetrisko:
It actually had interrupts higher priority before, since it's all in combinational block and interrupts occur later.  There's a slight X-pessimism preference to keep it that way rather than using if else

dpetrisko:
Unrelated: could you change
bp_params_priv_data_o to cfg_priv_data_o? It was a sed error that snuck by me

bp_be/src/v/bp_be_mem/bp_be_csr.v
```diff
@@ -130,17 +134,26 @@ bsg_priority_encode
...
-  (.i(exception_icode_dec_li)
-   ,.addr_o(exception_icode_li)
-   ,.v_o(exception_icode_v_li)
+ m_interrupt_enc
+  (.i(interrupt_icode_dec_li & ~mideleg_lo & 16'($signed(mgie)))
+   ,.addr_o(m_interrupt_icode_li)
+   ,.v_o(m_interrupt_icode_v_li)
+   );
+
+bsg_priority_encode
```
dpetrisko: Good change

bp_be/src/v/bp_be_mem/bp_be_csr.v
```diff
@@ -130,17 +134,26 @@ bsg_priority_encode
    ,.v_o(exception_ecode_v_li)
    );
 
-// TODO: This priority encoder needs to be swizzled, right now it is non-compliant with the spec...
```
dpetrisko: Keep this comment, it is still non-compliant

bp_be/src/v/bp_be_mem/bp_be_csr.v
```diff
@@ -220,38 +233,40 @@ bsg_dff_reset
...
-                            ,msie: mideleg_lo.msi, ssie: 1'b1
+
+// sip and sie masks
+assign sie_rwmask_li    = '{meie: 1'b0, seie: 1'b1
+                            ,mtie: 1'b0, stie: 1'b1
+                            ,msie: 1'b0, ssie: 1'b1
                             ,default: '0
                             };
-assign sip_wmask_li     = '{meip: 1'b0, seip: 1'b1
+assign sip_rwmask_li    = '{meip: 1'b0, seip: 1'b1
```
dpetrisko: Maybe just set the s* and keep m as default 0? Not important though

bp_be/src/v/bp_be_mem/bp_be_csr.v
```diff
@@ -502,46 +546,7 @@ always_comb
...
-
-    if (exception_icode_v_li)
-      if (mideleg_lo[exception_icode_li] & ~is_m_mode)
-        begin
-          priv_mode_n          = `PRIV_MODE_S;
-
-          mstatus_li.spp       = priv_mode_r;
-          mstatus_li.spie      = mstatus_lo.sie;
-          mstatus_li.sie       = 1'b0;
-
```
dpetrisko: So I think we still need this (or something like it) right now.  Basically, if there's an interrupt while an exception is  committing, we'll return to the instruction after the exception and not take the second trap.

dpetrisko: I think the right thing to do is to add the alternate interrupt mechanism to this PR.  I'll try to add it tonight so you can look it over, it shouldn't change too much in this module

bp_common/src/include/bp_common_csr_defines.vh
```diff
@@ -485,35 +485,20 @@ typedef struct packed
 
 typedef struct packed
 {
-  logic mei;
```
dpetrisko: Shouldn't this be only m_* and not s_*? setting sti means delegating s-mode timer interrupts to u-mode, no?

farzamgl:
> It actually had interrupts higher priority before, since it's all in combinational block and interrupts occur later. There's a slight X-pessimism preference to keep it that way rather than using if else

I don't think there is any X-pessimism problem right now. All the signals are initialized at the beginning of the always block. Also this way is more explicit in showing the priority.

dpetrisko:
```
if (interrupt_i)

else
```

If interrupt is X, then else is executed (basically treating it as 0).  In this case, it's actually fine because we do if else if, so the if (exception_v_i) should gate anything weird happening, but generally the other way is preferred

bp_be/src/v/bp_be_mem/bp_be_csr.v
farzamgl: It's still here
https://github.com/black-parrot/pre-alpha-release/blob/18802c1046054e2c5cf235baee32849fbd14568c/bp_be/src/v/bp_be_mem/bp_be_csr.v#L153

dpetrisko: oh, okay!

bp_common/src/include/bp_common_csr_defines.vh
farzamgl: No. It means sei, sti, ssi will trap to S mode instead of M mode.
We don't need the m* ones because we don't want to delegate mei, mti, msi to S mode.

farzamgl: The X in Xei, Xti, Xsi doesn't have to do anything with the mode we are running that. It simply means that `Xip & Xie == 1`

farzamgl: For delegating to U, we need the sideleg CSR.

PR #276, opened 2019-10-22
Add multicore sanity check demo to regressions

muwyse:
Adding a multicore test to regressions.


bp_common/syn/Makefile.vcs
```diff
@@ -128,6 +128,16 @@ regress_beebs.v: $(BEEBS_REGRESSION_V)
...
 	$(MAKE) sim.v PROG=$(basename $@)
 
+LCFG_2=e_bp_dual_core_cfg
+LCFG_4=e_bp_quad_core_cfg
+LCFG_8=e_bp_oct_core_cfg
+LCFG_16=e_bp_sexta_core_cfg
+
+regress_mc.v: $(MC_REGRESSION_V)
+$(MC_REGRESSION_V):
+	$(eval CFG_N := $(shell echo $(basename $@) | sed -r "s/(^.*_)//g"))
```
dpetrisko: What's this doing exactly?

dpetrisko:
write_demo probably isn't the right name here.  Maybe mc_sanity?

bp_common/syn/Makefile.vcs
muwyse: It extracts the number from the file name. For example, if the file is demo_2.v, it sets CFG_N = 2. Then this number gets combined with the text "LCFG_" to set the CFG parameter passed to "make sim.tool CFG=xxx"

Like I said, it is a bit of a hack. But we need to set the CFG variable for the simulation properly since these tests don't use the default CFG (single core) defined in the bp_top/syn Makefile

.gitlab-ci.yml
```diff
@@ -127,6 +127,18 @@ top-beebs-verilator:
     - make -C bp_top/syn clean build.sc
     - $CI_PROJECT_DIR/ci/regress.sh regress_beebs.sc bp_top
 
+top-mc-verilator:
```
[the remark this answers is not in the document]
muwyse: They aren't very long. I didn't know what all the CI test stages were and it is set to 'long' because I copied that from one of the other tests

bp_common/syn/Makefile.vcs
dpetrisko: Ohh, I get it now.  Could we override it in the gitlab yaml instead?

.gitlab-ci.yml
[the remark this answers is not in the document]
muwyse: We can also change the parameter N in the source code (pick a new default) or pass in a value to the compiler somehow when the program gets compiled.

bp_common/syn/Makefile.vcs
muwyse: I made an adjustment. But does the new way I have it written allow the multicore tests to run concurrently?

.gitlab-ci.yml
```diff
@@ -162,3 +179,20 @@ top-beebs-vcs:
...
+    - make -C bp_top/syn clean build.v CFG=e_bp_single_core_cfg
+    - make -C bp_top/syn build.v CFG=e_bp_dual_core_cfg
+    - make -C bp_top/syn build.v CFG=e_bp_quad_core_cfg
+    - make -C bp_top/syn build.v CFG=e_bp_oct_core_cfg
+    - make -C bp_top/syn build.v CFG=e_bp_sexta_core_cfg
+    - $CI_PROJECT_DIR/ci/regress.sh "mc_sanity_1.v CFG=e_bp_single_core_cfg" bp_top
+    - $CI_PROJECT_DIR/ci/regress.sh "mc_sanity_2.v CFG=e_bp_dual_core_cfg" bp_top
+    - $CI_PROJECT_DIR/ci/regress.sh "mc_sanity_4.v CFG=e_bp_quad_core_cfg" bp_top
+    - $CI_PROJECT_DIR/ci/regress.sh "mc_sanity_8.v CFG=e_bp_oct_core_cfg" bp_top
+    - $CI_PROJECT_DIR/ci/regress.sh "mc_sanity_16.v CFG=e_bp_sexta_core_cfg" bp_top
```
muwyse: @dpetrisko Will these run concurrently during regressions if there are cores/threads available?

dpetrisko: nope, those should be sequential

.gitlab-ci.yml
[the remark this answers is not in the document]
dpetrisko: Yeah, we can just keep it as is for now, add more variants later

dpetrisko:
LGTM! Merge if ready

PR #279, opened 2019-11-06
[WIP] Distribute CLINT and generalize SoC memory maps

dpetrisko:
This is still broken, but opening the PR for early feedback.  Basically, there are three main problems this PR solves.  

1) The CLINT isn't great for PD, with wires routing throughout chips.  To solve this, we distribute CLINT slices into tiles.  This will help with interrupt latency as a bonus.

2) Another is that the tag programmed coordinates are extra wires threaded throughout the chip, and don't take advantage of the known topology of the chip (e.g. chip 1 doesn't know for sure the chip next to it is at x+1).  This PR couples the coordinates with the topology (2D mesh) to maintain a simple memory map

3) Our previous implementation of ASIC chaining requires overprovisioning coordinates in order to globally address each chip.  For example, a 16 core BP connected to 4 single core BP asics requires (4+4)*log(4) bits for coordinate, despite only using a fraction of them.  Now, we use two-stage routing to solve this.  There's a document we'll include in the uarch spec as to how this currently works, but in general we route off-chip requests to the top of the chip, then down the chain, then according to a local, per chip address map.  This means that each chip is free to interpret the incoming request as it pleases.


dpetrisko:
Update -- passes regression now :)

dpetrisko:
One side effect is that performance numbers / regression runtime gets worse.  There are extra serializations for going off-chip now.  The performance regression is a red herring, since it represents an FPGA crossing, which should be much higher latency than we model, anyway.  The extra runtime is unfortunate, but should be mitigated when we get optimize out the ser/des in the RDP

taylor-bsg:
Can we change chip ID to domain ID to make it less specific to asic cloud?
Or does that not make sense?

In terms of going off-chip, to replicate the existing functionality, we can
toss the domain ID before transmitting, and rematerialize on the other
size?

M

dpetrisko:
> Can we change chip ID to domain ID to make it less specific to asic cloud?

My reasoning here is that there are two kinds of inter-domain crossings.  1 is going between two chips for BP, which is the current.  The other is going between BP-MC (or more generally BP-coprocessor).  For the latter, since it's still on the same chip, the current thinking is that the coprocessor space will be embedded in the 'local' address space of BP, (chip id = 0).  That way, there's no need to go to the I/O router

dpetrisko:
> In terms of going off-chip, to replicate the existing functionality, we can toss the domain ID before transmitting, and rematerialize on the other size?

Not sure I understand this proposal. All off chip requests go to the io node, then the coordinate substitution happens, so you can’t get around it. What we could do is have another version of the node that just always shoves off-chip. That would get rid of the serialization. The issue that comes with that is 1) two testbenches 2) we lose the ability to have multiple coordinates on the host side, since there’s no remapping being done

PR #282, opened 2019-11-10
Adding locking logic to d$ to guarantee (eventual) atomicity for LR/S…

dpetrisko:
…C and D$ misses

This should improve performance for multi-core LR/SC without introducing live/deadlock conditions.  Additionally prevents that the unlikely case of cache miss starvation in the BE

### DIFF
--- a/bp_be/src/v/bp_be_mem/bp_be_dcache/bp_be_dcache.v
+++ b/bp_be/src/v/bp_be_mem/bp_be_dcache/bp_be_dcache.v
@@ -81,6 +81,7 @@ module bp_be_dcache
  #(parameter bp_params_e bp_params_p = e_bp_inv_cfg
    `declare_bp_proc_params(bp_params_p)
    
+    , parameter lock_max_limit_p=8
     , parameter debug_p=0 
 
     , localparam cfg_bus_width_lp= `bp_cfg_bus_width(vaddr_width_p, num_core_p, num_cce_p, num_lce_p, cce_pc_width_p, cce_instr_width_p)
@@ -638,6 +639,7 @@ module bp_be_dcache
   logic lce_stat_mem_pkt_v;
   logic lce_stat_mem_pkt_yumi;
  
+  logic lce_cmd_ready_lo, lce_cmd_lock_lo;
   bp_be_dcache_lce
     #(.bp_params_p(bp_params_p))
     lce
@@ -685,7 +687,7 @@ module bp_be_dcache
 
       ,.lce_cmd_i(lce_cmd_i)
       ,.lce_cmd_v_i(lce_cmd_v_i)
-      ,.lce_cmd_ready_o(lce_cmd_ready_o)
+      ,.lce_cmd_ready_o(lce_cmd_ready_lo)
 
       ,.lce_cmd_o(lce_cmd_o)
       ,.lce_cmd_v_o(lce_cmd_v_o)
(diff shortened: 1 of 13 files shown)


bp_be/src/v/bp_be_mem/bp_be_dcache/bp_be_dcache.v
```diff
@@ -1061,6 +1063,38 @@ module bp_be_dcache
...
+  //      responsiveness
+  // TODO: Extra into bsg_edge_detector
+  logic cache_miss_r;
+  always_ff @(posedge clk_i)
+    cache_miss_r <= cache_miss_o;
+  wire cache_miss_resolved = cache_miss_r & ~cache_miss_o;
+
+  logic [`BSG_SAFE_CLOG2(lock_max_limit_p+1)-1:0] lock_cnt_r;
+  wire lock_clr = v_o || (lock_cnt_r == lock_max_limit_p);
+  wire lock_inc = ~lock_clr & (cache_miss_resolved || lr_op_tv_r || (lock_cnt_r > 0));
```
muwyse: lr_op_tv_r will be high regardless of the LR hitting or missing in the cache. We don't want to start the lock when an LR operation misses, because this will block lce command processing while the LR miss is being resolved, until the timeout triggers. We only want to lock starting from the time any miss is resolved or the LR succeeds.

I think the lr_op_tv_r condition should be (lr_op_tv_r & ~lr_miss_tv). This will start the lock increment for an LR hit only.

bp_be/src/v/bp_be_mem/bp_be_dcache/bp_be_dcache.v
```diff
@@ -1061,6 +1063,38 @@ module bp_be_dcache
...
+   lock_counter
+    (.clk_i(clk_i)
+     ,.reset_i(reset_i)
+
+     ,.clear_i(lock_clr)
+     ,.up_i(lock_inc)
+     ,.count_o(lock_cnt_r)
+     );
+  assign lce_cmd_lock_lo = (lock_cnt_r != '0);
+  assign lce_cmd_ready_o = lce_cmd_ready_lo & ~lce_cmd_lock_lo;
```
muwyse: I think we want to combine the lock signal with the valid_o signal from the LCE Command buffer inside bp_be_dcache_lce_cmd. That way, the LCE Cmd unit won't see any valid commands while the lock is set. Combining with lce_cmd_ready_lo still allows commands already in the LCE Command buffer to be processed. I think this could allow a pathological case where we still have livelock if the command in the buffer is an invalidation for the LR/SC block.

bp_be/src/v/bp_be_mem/bp_be_dcache/bp_be_dcache.v
dpetrisko: Yep, definitely right

bp_be/src/v/bp_be_mem/bp_be_dcache/bp_be_dcache.v
dpetrisko: oh, yeah I think you're right.  I'm used to ready->valid, which means that this would suppress it.  But these links are ready-valid&, right?

muwyse: I believe so, yes. The lce_cmd_ready_o signal goes directly to the network. Since the LCE command module assumes valid->yumi handshaking we had to insert a 2-fifo for commands inside the LCE.

dpetrisko: Ah, I see.  I actually think we can remove that buffer as well.  But another day, another PR :).  I'll change it to mask valid instead

dpetrisko:
Review addressed

bp_be/src/v/bp_be_mem/bp_be_dcache/bp_be_dcache.v
```diff
@@ -460,14 +461,15 @@ module bp_be_dcache
...
-  logic lr_miss_tv;
+  logic lr_hit_tv, lr_miss_tv;
   logic sc_success;
   logic sc_fail;
   logic [ptag_width_lp-1:0]  load_reserved_tag_r;
   logic [index_width_lp-1:0] load_reserved_index_r;
   logic load_reserved_v_r;
 
   // Upgrade if a load reserved and we don't have the line in exclusive state
+  assign lr_hit_tv = v_tv_r & lr_op_tv_r & load_hit;
```
muwyse: load_hit should be store_hit to ensure that the LR hit is high only when the block is in E or M state.

bp_be/src/v/bp_be_mem/bp_be_dcache/bp_be_dcache.v
```diff
@@ -1061,6 +1064,40 @@ module bp_be_dcache
...
+     ,.reset_i(reset_i)
+
+     ,.clear_i(lock_clr)
+     ,.up_i(lock_inc)
+     ,.count_o(lock_cnt_r)
+     );
+  // We could actually be more clever here.  We only need to block invalidations to this
+  //   specific line.  However, being extra safe is easier to implement for now.
+  assign lce_cmd_lock_lo = (lock_cnt_r != '0);
+  assign lce_cmd_v_li = lce_cmd_v_i & ~lce_cmd_lock_lo;
```
muwyse: This still has the effect of gating new messages into the LCE Command buffer, not valid messages coming out of the buffer for processing. The lock signal needs to be fed all the way into bp_be_dcache_lce_cmd.v and combined with the current lce_cmd_v_li signal in that module to prevent the LCE Command module from processing any valid commands that might be in the buffer.

bp_be/src/v/bp_be_mem/bp_be_dcache/bp_be_dcache.v
dpetrisko: oh, makes sense.  Should the next line be 
assign lr_miss_tv = v_tv_r & lr_op_tv_r & ~store_hit;
instead of
assign lr_miss_tv = v_tv_r & lr_op_tv_r & load_hit & ~store_hit;
?

bp_be/src/v/bp_be_mem/bp_be_dcache/bp_be_dcache.v
dpetrisko: Oh, I see what you mean.  I misunderstood the issue.  It's not handshaking, it's the residual stuff in the buffer.  Yeah, I think removing the buffer is the easiest fix (and better for PPA)

dpetrisko:
Addressed comments round two

PR #285, opened 2019-11-25
RISCV-DV and bugfixes associated

dpetrisko:
This is a large commit with several unfortunately unrelated pieces

1) Add skeleton riscv-dv program generation to flow/CI
2) Add debug mode support including DCSR
3) Add program buffer support to debug mode
* 2 and 3 are tested based on configuration in mmio config loader, not every test
4) Various bugfixes found while getting riscv-dv working

  a) Strobing mtime to slow down "RTC"
  b) having fencei poison issue, but not the rest of the pipe
  c) adding tvm, tw, and tsr support
  d) clearing mprv on xret to less privileged mode
  e) clearing $ uncached valids on poison + changing the mux selection based on current request instead of last request (necessary for interrupted uncached requests)
  f) Only take exception if instruction valid at commit
  g) Only update branch metadata on branch instructions
  h) only send mispredictions on dispatch, preventing multiple redirects from being sent


bp_be/src/v/bp_be_checker/bp_be_director.v
```diff
@@ -240,13 +240,16 @@ always_comb
...
+    // TODO: Needs to happen at mem3, not mem1.  Currently, we flush which is okay
+    //         because we also drain the pipeline
     else if (calc_status.mem1_fencei_v)
       begin
         fe_cmd.opcode = e_op_icache_fence;
         fe_cmd.vaddr  = expected_npc_o;
 
         fe_cmd_v = fe_cmd_ready_i;
+
+        flush_o = 1'b1;
```
farzamgl: Why don't we do the flush at mem3 right now?

farzamgl:
So we are sending the fence_i FE cmd at isd and just poisoning the issue stage?
Why aren't we doing it at mem3?

bp_common/test/Makefile.frag
```diff
@@ -160,7 +177,27 @@ RV64_V_TESTS = \
   rv64ui-v-subw    \
   rv64ui-v-sw      \
   rv64ui-v-xor     \
-  rv64ui-v-xori
+  rv64ui-v-xori    \
+  \
+  #rv64ua-v-amoadd_d  \
```
farzamgl: Why are these tests commented out?

bp_me/test/common/bp_cce_mmio_cfg_loader.v
```diff
@@ -82,7 +82,11 @@ module bp_cce_mmio_cfg_loader
     ,SEND_PC
     ,SEND_IRF
     ,RECV_IRF
+    ,BP_ENTER_DEBUG
     ,BP_FREEZE_CLR
+    ,BP_NINSTR_RD
+    ,BP_NINSTR_LD
+    ,BP_EXIT_DEBUG
```
farzamgl: What are some of these unused states? `BP_NINSTR_LD` for instance.

dpetrisko: oh yeah, debug stuff.

"Next Instruction write to RD", "Next Instruction write to LD"

bp_common/test/Makefile.frag
dpetrisko: Oh right, I forgot.  A couple of them fail in spike, probably worth looking into since they're VM+emulation tests

dpetrisko:
> So we are sending the fence_i FE cmd at isd and just poisoning the issue stage?
> Why aren't we doing it at mem3?

Yeah, we should do it at mem3, but it's a more invasive change because I wasn't quite sure where to put it.  I'll change it as part of this PR though

PR #296, opened 2020-01-15
Add NPOT functionality to coherence system

muwyse:
These changes enable non-power-of-two setups (CCE and LCE). Number of CCE's is constrained by bsg_hash_bank, so not all possible configurations are supported. Number of LCE's is unconstrained.

These changes also include a few minor bug fixes for the D$ and I$ that were discovered during testing.


dpetrisko:
What are the constraints for CCE bsg_hash bank? I think we should put that in bp_common_aviary_pkg

dpetrisko:
Or better yet, an assert in bp_nonsynth_if_verif

muwyse:
bsg_hash_bank supports (2^n) * (2^(m-1)), n = 0, 1, 2, ... and m = 1, 2, 3, ...
= 1, 2, 3, 4, 6, 7, 8, 12, 14, 15, 16, 24, 28, 30, 31, 32, etc.

Since that is what we rely on to hash cache blocks to CCE, the number of CCEs is our number of banks.

I can add a check in if_verif to check up to 32 CCE for now. I doubt we would go larger than that for the foreseeable future.

dpetrisko:
Sounds good thanks!  That's pretty good coverage, and I think it wouldn't be terrible to extend the module if really necessary

bp_me/src/v/cce/bp_cce.v
```diff
@@ -71,13 +67,15 @@ module bp_cce
    , input                                             mem_cmd_ready_i
   );
 
+  /*
```
dpetrisko: Remove?

bp_me/src/v/cce/bp_cce_dir.v
```diff
@@ -22,161 +22,206 @@
...
+  initial begin
+    assert(tag_sets_per_row_lp == 2) else
+      $error("Unsupported configuration: number of sets per row must equal 2");
+  end
 
-  // Directory
+  // address offset table
+  // TODO: how expensive is this?
+  // bits is O(rows_per_set), computation should all be static since it is generate/params
+  // lookup: lce_i[1+:], assuming tag_sets_per_row_lp == 2
```
dpetrisko: Generally static addition is very cheap, but I can keep an eye out in the tapeout

bp_me/src/v/cce/bp_cce_dir.v
```diff
@@ -22,161 +22,206 @@
...
   assign sharers_coh_states_o = sharers_coh_states_r;
 
-  logic [dir_tag_sets_per_row_lp-1:0]                                 sharers_hits;
-  logic [dir_tag_sets_per_row_lp-1:0][lg_lce_assoc_lp-1:0]            sharers_ways;
-  logic [dir_tag_sets_per_row_lp-1:0][`bp_coh_bits-1:0]               sharers_coh_states;
+  logic [tag_sets_per_row_lp-1:0]                                 sharers_hits;
+  logic [tag_sets_per_row_lp-1:0][lg_lce_assoc_lp-1:0]            sharers_ways;
+  logic [tag_sets_per_row_lp-1:0][$bits(bp_coh_states_e)-1:0]     sharers_coh_states;
+
+  // TODO: debug, remove
```
dpetrisko: remove :)

Or add to a tracer

bp_me/src/v/cce/bp_cce_dir.v
muwyse: Thanks. This is supposed to turn into a simple lookup table so there should be no arithmetic hardware introduced if the tool gets it right.

PR #302, opened 2020-01-23
Combine I/O and Memory Messages

dpetrisko:
This PR simplifies the networks. Originally, we had the memory and I/O messages and networks the same. The problem with this was having a 2D network, when we really needed a 1D I/O network and a 0.5D Memory network.

Next, we separated out the memory and I/O messages and networks. This was better for PPA, but lead to duplicate code for encoding packets, adapters, formats, etc.

This PR makes leaves the memory and I/O networks separate but keeps the same format for memory and I/O messages. This allows us to reduce networking code as well as keep the same interface for io_cce's, MMIO components and CCE connected components. Lastly, forward thinking, this should allow much simpler coherent I/O components.


dpetrisko:
Oh, another benefit to this is it should make fixing the ME testbenches much easier, since we can just tie the _loaders to the configuration registers

muwyse:
Did you run the other multicore tests (mc_rand_walk and mc_work_share_sort)? Only mc_sanity is checked in regressions. Also would be good to check 3, 6, and maybe 12 core setups, which aren't in regressions.

bp_top/test/tb/bp_top_trace_demo/testbench.v
```diff
@@ -319,41 +318,53 @@ bind bp_be_top
...
-bp_me_cce_to_io_link_bidir
- #(.bp_params_p(bp_params_p))
- host_cfg_link
+bp_me_cce_to_mem_link_bidir
+ #(.bp_params_p(bp_params_p)
+   ,.num_outstanding_req_p(io_noc_max_credits_p)
+   ,.flit_width_p(io_noc_flit_width_p)
+   ,.cord_width_p(io_noc_cord_width_p)
+   ,.cid_width_p(io_noc_cid_width_p)
+   ,.len_width_p(io_noc_len_width_p)
```
muwyse: Same question as above.

bp_top/test/tb/bp_top_trace_demo/testbench.v
```diff
@@ -319,41 +318,53 @@ bind bp_be_top
...
    );
 
 bp_me_cce_to_mem_link_client
- #(.bp_params_p(bp_params_p))
+ #(.bp_params_p(bp_params_p)
+   ,.num_outstanding_req_p(mem_noc_max_credits_p)
+   ,.flit_width_p(mem_noc_flit_width_p)
+   ,.cord_width_p(mem_noc_cord_width_p)
+   ,.cid_width_p(mem_noc_cid_width_p)
+   ,.len_width_p(mem_noc_len_width_p)
```
muwyse: Same comment as above.

bp_top/src/v/bp_tile.v
```diff
@@ -522,7 +522,12 @@ for (genvar i = 0; i < 2; i++)
   localparam dram_y_cord_lp = ic_y_dim_p + cc_y_dim_p + mc_y_dim_p;
   wire [mem_noc_cord_width_p-1:0] dst_cord_li = dram_y_cord_lp;
   bp_me_cce_to_mem_link_master
-   #(.bp_params_p(bp_params_p))
+   #(.bp_params_p(bp_params_p)
+     ,.flit_width_p(mem_noc_flit_width_p)
+     ,.cord_width_p(mem_noc_cord_width_p)
+     ,.cid_width_p(mem_noc_cid_width_p)
+     ,.len_width_p(mem_noc_len_width_p)
```
muwyse: Are these parameters not part of bp_params_p? What is the reason for having them be separate parameters into bp_me_cce_to_mem_link_master? Is there a case where we would want to instantiate that module and override those params?

dpetrisko: We're reusing the bp_me_cce_to_mem_link_master for both I/O network and memory networks, so we need to override the network parameters.  It's a similar case to how we only have a bp_tlb.v and override els_p with itlb_els_p or dtlb_els_p

dpetrisko:
> Did you run the other multicore tests (mc_rand_walk and mc_work_share_sort)? Only mc_sanity is checked in regressions. 
I'll add the other mc tests to regression, good call.  

> Also would be good to check 3, 6, and maybe 12 core setups, which aren't in regressions.
I'll add those in a separate PR

bp_top/src/v/bp_tile.v
dpetrisko: Do you have an idea for how to do this better? This was just the most clear way for me

muwyse: I don't have a better idea, that seems good to me. I didn't catch they were actually being overridden, but I see that now. Just making sure we weren't duplicating for no reason.

muwyse:
> > Did you run the other multicore tests (mc_rand_walk and mc_work_share_sort)? Only mc_sanity is checked in regressions.
> 
> I'll add the other mc tests to regression, good call.

Sounds good. We can scale the input sizes if needed for regressions. The larger core counts probably take in the handful of minutes range to run.

> > Also would be good to check 3, 6, and maybe 12 core setups, which aren't in regressions.
> 
> I'll add those in a separate PR

👍

PR #303, opened 2020-01-29
Simplify FE-BE interface

dpetrisko:
This PR is to get rid of the fence logic which sits between FE and BE, making it simpler to test them in isolation (it also decouples them further, reducing risk of inter-module bugs). The interface contract on the FE side is a standard FIFO.  On the BE side, we go into a fence state when we send a non-attaboy command and then suppress issue until all commands have been processed.  Last, it gets rid of the "fence fifo" since it's an unnecessary wrapper.

This is purely a clarity change (other than removing a flop).  Critical path for queue/cmd messages should be unaffected (or possibly improve, very slightly).

It also bumps basejump.


dpetrisko:
@farzamgl thoughts on this?

bp_be/src/v/bp_be_checker/bp_be_director.v
```diff
@@ -199,6 +204,8 @@ always_ff @(posedge clk_i)
       state_r <= state_n;
     end
 
+assign suppress_iss_o = (state_r == e_fence) & fe_cmd_fence_i;
```
farzamgl: Why not just `(state_r == e_fence)`?

bp_top/src/v/bp_core.v
```diff
@@ -124,9 +115,11 @@ module bp_core
      ,.v_o(fe_cmd_v_lo)
      ,.yumi_i(fe_cmd_yumi_li)
      );
+  wire fe_cmd_empty_lo = ~fe_cmd_v_lo;
+  wire fe_cmd_full_lo  = ~fe_cmd_ready_lo;
```
farzamgl: These 2 signals are not used.

bp_be/src/v/bp_be_checker/bp_be_director.v
dpetrisko: That would create a bubble when fe_cmd_fence_i goes low.  Could just do (state_n == e_fence), I think

farzamgl: Oh yeah, you're right.

PR #305, opened 2020-01-30
Preventing host addresses from punching through cached address space

dpetrisko:


dpetrisko:
This bug causes any cached address matching 0xXXXX_XX1X_XXXX to get routed to I/O routers instead of a CCE.  Our non-Linux benchmarks didn't catch this because they're too small / not located in the "lucky" space.

@taylor-bsg 🙏the cosim flow.

taylor-bsg:
Oh yeah!!!!!!! Hell yeah! Dromajo in the Haus!

PR #307, opened 2020-01-31
Dispatch packets should be poisoned instead of flushing to be dequeued  from FE queue

farzamgl:
When we take a trap all the instructions read from FE queue should be dequeued, but flushing them zeros their `queue_v` field which causes them not to dequeued from the pipe and making problems when rolling back in the future.
Instead, we keep its `decode` field and just poison it in the calculator.

### DIFF
--- a/bp_be/src/v/bp_be_calculator/bp_be_calculator_top.v
+++ b/bp_be/src/v/bp_be_calculator/bp_be_calculator_top.v
@@ -207,7 +207,8 @@ bp_be_dispatch_pkt_s reservation_n, reservation_r;
 always_comb
   begin
     reservation_n        = dispatch_pkt_i;
-    reservation_n.decode = (~flush_i & dispatch_pkt.v) ? dispatch_pkt.decode : '0;
+    reservation_n.decode = dispatch_pkt.v ? dispatch_pkt.decode : '0;
+    reservation_n.poison = (~flush_i & dispatch_pkt.v) ? dispatch_pkt.poison : 1'b1;
     reservation_n.rs1    = bypass_rs1;
     reservation_n.rs2    = bypass_rs2;
   end
@@ -433,7 +434,7 @@ always_comb
         exc_stage_n[2].roll_v          = exc_stage_r[1].roll_v   | pipe_mem_miss_v_lo;
         exc_stage_n[3].roll_v          = exc_stage_r[2].roll_v   | pipe_mem_miss_v_lo;
 
-        exc_stage_n[0].poison_v        = dispatch_pkt.poison     | flush_i;
+        exc_stage_n[0].poison_v        = reservation_n.poison    | flush_i;
         exc_stage_n[1].poison_v        = exc_stage_r[0].poison_v | flush_i;
         exc_stage_n[2].poison_v        = exc_stage_r[1].poison_v | flush_i;
         exc_stage_n[3].poison_v        = exc_stage_r[2].poison_v | pipe_mem_miss_v_lo | pipe_mem_exc_v_lo;
--- a/bp_be/src/v/bp_be_calculator/bp_be_calculator_top.v
+++ b/bp_be/src/v/bp_be_calculator/bp_be_calculator_top.v
@@ -203,12 +203,11 @@ bp_be_bypass
    ,.bypass_rs2_o(bypass_irs2)
    );
 
+// Override operands with bypass data
 bp_be_dispatch_pkt_s reservation_n, reservation_r;
 always_comb
   begin
     reservation_n        = dispatch_pkt_i;
-    reservation_n.decode = dispatch_pkt.v ? dispatch_pkt.decode : '0;
-    reservation_n.poison = (~flush_i & dispatch_pkt.v) ? dispatch_pkt.poison : 1'b1;
     reservation_n.rs1    = bypass_rs1;
     reservation_n.rs2    = bypass_rs2;
   end
--- a/bp_be/src/v/bp_be_calculator/bp_be_calculator_top.v
+++ b/bp_be/src/v/bp_be_calculator/bp_be_calculator_top.v
@@ -440,10 +440,9 @@ always_comb
   end
 
 assign commit_pkt.v          = calc_stage_r[2].v & ~exc_stage_r[2].poison_v;
-assign commit_pkt.queue_v    = calc_stage_r[2].queue_v & ~exc_stage_r[2].roll_v;
-assign commit_pkt.bubble_v   = ~calc_stage_r[2].queue_v & ~exc_stage_r[2].poison_v;
-assign commit_pkt.instret    = calc_stage_r[2].instr_v & ~exc_stage_n[3].poison_v;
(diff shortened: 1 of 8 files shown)


farzamgl:
The problem isn't fully solved yet. I'm still seeing cases where some of the deqs are skipped.

dpetrisko:
why is this a different case than an exception?

dpetrisko:
queue_v is a field in the bp_be_pipe_stage_reg_s, not dispatch pkt.  Could you outline the exact scenario where you see this problem?

dpetrisko:
I’m not sure what problems you’re seeing, but I think this is a good change overall.
Added a small patch for clarity

farzamgl:
Update:
The interrupts are masking the issued packets in the scheduler which causes them to never be dequeued from the FE FIFO. We have to inject the interrupt packets into the FE FIFO from FE to avoid this masking. So basically any issued packet from FE FIFO at scheduler should get to mem3(even if poisoned) so it can be dequeued from the FIFO.

dpetrisko:
Hold off on this until interrupt rework.

taylor-bsg:
Hi, can we have small test case checked in to our regression? We don’t want
to have to hunt for this one in the wild to reproduce!

M

dpetrisko:
I think a triggering torture sequence would be:
set up mtvec to X

for N = 1000
set msip

Check I

X:
Load Y*i
i+=64
epc+=4
mret

dpetrisko:
Could use the same logic for stress testing exceptions

dpetrisko:
@farzamgl we still need a regression test for this, but I've confirmed that interrupts are triggering in RISC-V tests based on this changeset.  I also brought in your mtime_r/w_v change here.  Last, I removed the out-of-band debug mode instructions.  It looks like, with the rolly fifo, injecting into the pipeline is a bad idea, so we'll need another way to enter debug mode (eventually, we're not using that feature for anything right now)

farzamgl:
I think it's not that easy to create a regression for this, because it's really tied to the timing of the program. But we can actually use assertions to check if the fifo packet that's being dequeued has the same pc/instr as the one in mem2.

dpetrisko:
That’s an interesting idea. I think that’s a good assertion. I still think we need a directed stress test for interrupts, cache misses and exceptions. It doesn’t have to be super clean, just blast all of those conditions and and see what happens

bp_be/src/v/bp_be_mem/bp_be_csr.v
```diff
@@ -361,7 +361,43 @@ always_comb
     instr_misaligned_o    = '0;
     ebreak_o              = '0;
 
-    if (csr_cmd_v_i)
+    // TODO: Need to force a bubble every N-cycles with pending interrupt
+    //   Otherwise a halt loop with a pending interrupt will hang without interrupting
+    if (~is_debug_mode & bubble_i & m_interrupt_icode_v_li)
```
farzamgl: Doesn't this reintroduce the problem that we might take an interrupt during a PTW?

dpetrisko: Yes, good point. This is still wip, since I don’t want to commit without solving the livelock issue. 

Remind me what the problem with taking an interrupt during a PTW is?

dpetrisko:
Okay, I think this is good to go.  The assertion is pretty tricky to get right, but if you want to add it it's probably a good idea

bp_be/src/v/bp_be_mem/bp_be_csr.v
farzamgl: Changes the priv level during the walk which messes up with some access checks.

farzamgl:
@dpetrisko I don't think the correct epc is being loaded when we take an interrupt.

bp_be/src/v/bp_be_calculator/bp_be_calculator_top.v
```diff
@@ -435,11 +435,12 @@ always_comb
         exc_stage_n[0].poison_v        = reservation_n.poison    | flush_i;
         exc_stage_n[1].poison_v        = exc_stage_r[0].poison_v | flush_i;
         exc_stage_n[2].poison_v        = exc_stage_r[1].poison_v | flush_i;
-        exc_stage_n[3].poison_v        = exc_stage_r[2].poison_v | pipe_mem_miss_v_lo | pipe_mem_exc_v_lo;
+        exc_stage_n[3].poison_v        = exc_stage_r[2].poison_v | flush_i;
```
farzamgl: I don't think this change is correct. Now we are poisoning all the instructions in mem3 that raise the flush signal i.e. eret, sfence, ifence.

dpetrisko: yeah you're right.  I should start running with dromajo to verify that things like instret match, since that doesn't show up on the functional tests.  

Anyway, I got feedback that this approach may introduce a security hole.  I think the way to go is to inject the interrupt into the decode stage, but attach it to a valid instruction when decoded.  I think that should solve all issues, since:
1) It'll be attached to a queue entry so it can be rolled back
2) It will propagate down the pipe, so it will be serialized
3) It can be flushed from the pipe normally if an exception happens, and it'll just get reinjected if the interrupt is still valid

This is actually pretty close to where we started on this PR... which is good!

farzamgl: I think even the old way to generate interrupt packets work. You just have to make sure you keep the instruction in the issue stage and do not issue any more instructions. I tried this change locally on top of `14055e1` and I think it worked.

```suggestion
--- a/bp_be/src/v/bp_be_checker/bp_be_scheduler.v
+++ b/bp_be/src/v/bp_be_checker/bp_be_scheduler.v
@@ -90,7 +90,7 @@ bsg_dff_reset_en
  issue_pkt_reg
   (.clk_i(clk_i)
    ,.reset_i(reset_i | cache_miss_v_i)
-   ,.en_i(issue_v | dispatch_v_i)
+   ,.en_i(issue_v | (dispatch_v_i & ~accept_irq_i))

    ,.data_i({issue_v, issue_pkt})
    ,.data_o({issue_pkt_v_r, issue_pkt_r})
@@ -102,7 +102,7 @@ bsg_dff_reset_en
  issue_status_reg
   (.clk_i(clk_i)
    ,.reset_i(reset_i)
-   ,.en_i(issue_v | dispatch_v_i | poison_iss_i | npc_mismatch)
+   ,.en_i(issue_v | (dispatch_v_i & ~accept_irq_i) | poison_iss_i | npc_mismatch)

    ,.data_i(poison_iss_i | npc_mismatch)
    ,.data_o(poison_iss_r)
@@ -189,7 +189,7 @@ always_comb
   end

 // Interface handshakes
-assign fe_queue_yumi_o = ~debug_mode_i & fe_queue_v_i & (dispatch_v_i | ~issue_pkt_v_r);
+assign fe_queue_yumi_o = ~debug_mode_i & fe_queue_v_i & ((dispatch_v_i & ~accept_irq_i) | ~issue_pkt_v_r);
```

PR #308, opened 2020-01-31
Fixing mtime address and adding mtime write signal

farzamgl:


dpetrisko:
Do we need this? Writing mtime is fundamentally broken in the SBI without patching, because of our distributed Clint. 

The address change is fine

dpetrisko:
I believe this will also break all of our current tests

farzamgl:
I believe the tests are going to be fine, they use the `0x030bff8` address.
We need to write mtime, if we want to do the checkpointing. Also, Linux reads the mtime frequently. 
I suggest having the changes until we figure out how to decentralize the timer.

dpetrisko:
>We need to write mtime, if we want to do the checkpointing(the code doesn't write to timer).

Good point, write to mtime for checkpointing is fine.  

>I believe the tests are going to be fine, they use the 0x030bff8 address
Are you saying that interrupt-based tests are all currently broken?

>Linux reads the mtime frequently.
The address should come in from the SBI, which we can patch to our address map.  My position is that partial compliance where things almost work is worse than explicit non-compliance.

farzamgl:
> Are you saying that interrupt-based tests are all currently broken?

Kind of broken. The mtime reads in those tests wasn't that critical.

> The address should come in from the SBI, which we can patch to our address map. My position is that partial compliance where things almost work is worse than explicit non-compliance.

The address comes from the device tree parsing in the initial bootloader code. I don't see the point of keeping it at a wrong address though. It's just gonna read zero instead of the timer value. It's still a silent problem. At least now single core works with the SBI.

dpetrisko:
>I don't see the point of keeping it at a wrong address though
>The address comes from the device tree parsing in the initial bootloader code.

Oh yeah definitely not.  I'm saying to change this in software (I guess the mechanism is to modify the device tree?) rather than changing the hardware addresses.

farzamgl:
But I still don't know what's the point of changing the timer address from the standard one in RTL, tests, and the simulator programs? The functionality is going to stay the same but we're just using the wrong address.

dpetrisko:
There’s no “right” or “wrong” one. There is the SiFive choice, which is a unified system-wide mtime at 8ff0. BlackParrot has distributed mtime, which has different functionality and which we can put at wherever we want. Our other clint addresses are also different than the SiFive clint, taking into account both the base and the per-core addresses. 

Firmware which uses the SiFive mtime offset will work for single but fail in multicore for BlackParrot in mysterious ways. Since we’re already patching the memory maps, I don’t see the advantage of partial compliance, only downsides.

farzamgl:
Merged with #307

PR #314, opened 2020-02-05
A number of unrelated fixes to FE and BE

farzamgl:
- Fixed not checking access faults on SC instructions
- Enabled FS and SD bits in mstatus CSR which are used by Linux to manage float context switching
- Poison I$ on access and page faults so the I$ doesn't freeze indefinitely

### DIFF
--- a/bp_fe/src/v/bp_fe_mem.v
+++ b/bp_fe/src/v/bp_fe_mem.v
@@ -136,8 +136,10 @@ wire is_uncached_mode = (cfg_bus_cast_i.icache_mode == e_lce_mode_uncached);
 wire mode_fault_v = (is_uncached_mode & ~uncached_li);
 // TODO: Enable other domains by setting enabled dids with cfg_bus
 wire did_fault_v = (ptag_li[ptag_width_p-1-:io_noc_did_width_p] != '0);
+// Don't allow speculative access to local tile memory
+wire local_fault_v = (ptag_li < (dram_base_addr_gp >> page_offset_width_p));
 
-assign instr_access_fault_v = fetch_v_r & (mode_fault_v | did_fault_v);
+assign instr_access_fault_v = fetch_v_r & (mode_fault_v | did_fault_v | local_fault_v);
 assign instr_page_fault_v   = fetch_v_r & itlb_r_v_lo & mem_translation_en_i & (instr_priv_page_fault | instr_exe_page_fault);
 
 assign mem_cmd_yumi_o = itlb_fence_v | itlb_fill_v | fetch_v;
--- a/bp_fe/src/v/bp_fe_mem.v
+++ b/bp_fe/src/v/bp_fe_mem.v
@@ -54,11 +54,13 @@ bp_cfg_bus_s cfg_bus_cast_i;
 bp_fe_mem_cmd_s  mem_cmd_cast_i;
 bp_fe_mem_resp_s mem_resp_cast_o;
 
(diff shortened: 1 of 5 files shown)


bp_fe/src/v/bp_fe_mem.v
```diff
@@ -161,24 +123,24 @@ always_ff @(posedge clk_i)
...
+wire instr_priv_page_fault = ((mem_priv_i == `PRIV_MODE_S) & itlb_r_entry.u)
                                | ((mem_priv_i == `PRIV_MODE_U) & ~itlb_r_entry.u);
-wire instr_exe_access_fault = ~itlb_r_entry.x;
+wire instr_exe_page_fault = ~itlb_r_entry.x;
 
 // Fault if in uncached mode but access is not for an uncached address
 wire is_uncached_mode = (cfg_bus_cast_i.icache_mode == e_lce_mode_uncached);
 wire mode_fault_v = (is_uncached_mode & ~uncached_li);
 // TODO: Enable other domains by setting enabled dids with cfg_bus
 wire did_fault_v = (ptag_li[ptag_width_p-1-:io_noc_did_width_p] != '0);
```
farzamgl: @dpetrisko I think we have to make this check more strict so it doesn't freeze the I$ on low addresses below the DRAM base.

dpetrisko: Uncached icache requests are not confirmed to be working at all right now, I think the prudent thing is to just fault on uncached.

dpetrisko: Long term, it's reasonable to have local device in the non-idempotent region of the PMA, so that there are no speculative accesses (i.e. FE accesses)

farzamgl: I'm not talking about uncached, but the problem where a vaddr is passed as paddr by btb.

bp_fe/src/v/bp_fe_mem.v
```diff
@@ -188,4 +150,42 @@ assign mem_resp_cast_o = '{instr_access_fault: instr_access_fault_r
...
+   ,.cfg_bus_i(cfg_bus_i)
+
+   ,.vaddr_i(mem_cmd_cast_i.operands.fetch.vaddr)
+   ,.vaddr_v_i(fetch_v)
+   ,.vaddr_ready_o(icache_ready_lo)
+
+   ,.uncached_i(uncached_li)
+   ,.ptag_i(ptag_li)
+   ,.ptag_v_i(ptag_v_li)
+   ,.poison_i(mem_poison_i | instr_access_fault_v | instr_page_fault_v)
```
dpetrisko: Don't understand the need here: in D$ we only poison during the exception, not the speculative exception

farzamgl: If there is a page-fault or access fault, the instruction cannot be fetched. The pc is speculative, but the fault is deterministic for that PC. Same reason in D$ we poison on faults.

dpetrisko: Oh, I didn't know that we did that in the D$.  Looks good then!

bp_fe/src/v/bp_fe_mem.v
dpetrisko: That is uncached

dpetrisko: Ope, well it should be

PR #315, opened 2020-02-05
[WIP] Adding streaming accelerator complex skeleton

dpetrisko:
This change brings back in the streaming accelerator complex skeleton so that we can add that next.

The BPAI branch is currently broken for single core runs, so I cannot test this, hence WIP


bp_top/src/v/bp_accelerator_complex.v
```diff
@@ -18,74 +18,82 @@ module bp_accelerator_complex
```
zaazad: So we need to change the file name to bp_coherent_accelerator_complex and have bp_streaming_accelerator_complex at the left side of the core_complex?

PR #316, opened 2020-02-05
Check-pointing and Cosimulation with Dromajo

farzamgl:
This PR adds the capability to run a benchmark on Dromajo(as an ISA simulator) until for a certain number of instructions and then resume with cosimulation in RTL.

To run on Dromajo and generate the checkpoint files run:
`cd bp_common/test`
`make <test>.dromajo MAXINSN=<n> MEMSIZE=<k in MB>`
This generates test files under the name `<test>.dromajo.<n>`

And then to load the state in RTL and run with cosim enabled run:
`cd bp_top/syn`
`make build.v sim.v PROG=<test>.dromajo.<n> PRELOAD_MEM_P=1 LOAD_NBF_P=1 COSIM_P=1`
This command preloads the mainram with the generated binary, and restores the processor state through the generated nbf files.

### DIFF
--- a/bp_be/src/include/bp_be_internal_if_defines.vh
+++ b/bp_be/src/include/bp_be_internal_if_defines.vh
@@ -123,6 +123,7 @@
   /* TODO: make opcode */                                                                          \
   typedef struct packed                                                                            \
   {                                                                                                \
+    logic [dword_width_p-1:0]       cause;                                                         \
     logic [vaddr_width_p-1:0]       epc;                                                           \
     logic [vaddr_width_p-1:0]       tvec;                                                          \
     logic [rv64_priv_width_gp-1:0]  priv_n;                                                        \
@@ -188,7 +189,7 @@
    )
  
 `define bp_be_trap_pkt_width(vaddr_width_mp) \
-  (2 * vaddr_width_mp + rv64_priv_width_gp + 4)
+  (2 * vaddr_width_mp + rv64_priv_width_gp + dword_width_p + 4)
 
 `define bp_be_wb_pkt_width(vaddr_width_mp) \
   (1                                                                                               \
--- a/bp_common/test/src/demos/Makefile
+++ b/bp_common/test/src/demos/Makefile
@@ -1,8 +1,8 @@
 
 include Makefile.frag
 
-RISCV_GCC  = riscv64-unknown-elf-gcc --static -nostdlib -nostartfiles -fPIC -march=rv64ia -mabi=lp64 -mcmodel=medany -I$(TEST_DIR)/include
-RISCV_LINK = -static -nostdlib -nostartfiles -L$(TEST_DIR)/lib -T src/riscv.ld 
+RISCV_GCC  = riscv64-unknown-elf-gcc --static -nostartfiles -fPIC -march=rv64ia -mabi=lp64 -mcmodel=medany -I$(TEST_DIR)/include
+RISCV_LINK = -static -nostartfiles -L$(TEST_DIR)/lib -T src/riscv.ld 
 
 .PHONY: all bp-demo-riscv bp-demo-s
 
--- a/bp_common/test/src/perch/start.S
+++ b/bp_common/test/src/perch/start.S
@@ -5,7 +5,7 @@ _start:
 /* setup stack pointers. Stacks start at 0x8FFF_CFFF */
 /* We then subtract off 8K*coreID. The top 4K is for the core emulation stack
  * the lower 4K is for the program TODO: Maybe larger stack value? */
-    li   sp, 0x8FFFCFF0
+    li   sp, 0x80010000
     csrr x1, mhartid 
     slli x1, x1, 13
     sub  sp, sp, x1
--- a/bp_top/test/tb/bp_top_trace_demo/testbench.v
+++ b/bp_top/test/tb/bp_top_trace_demo/testbench.v
@@ -145,29 +145,28 @@ wrapper
        ,.rd_data_i(be_calculator.wb_pkt.rd_data)
        );
 
-  if(cosim_p)
-    bind bp_be_top
-      bp_nonsynth_cosim
-       #(.bp_params_p(bp_params_p)
-         ,.config_file_p("prog.cfg"))
-        cosim
-        (.clk_i(clk_i)
-         ,.reset_i(reset_i)
-         ,.freeze_i(be_checker.scheduler.int_regfile.cfg_bus.freeze)
+  bind bp_be_top
+    bp_nonsynth_cosim
+     #(.bp_params_p(bp_params_p)
+       ,.config_file_p("prog.cfg"))
+      cosim
+      (.clk_i(clk_i & (testbench.cosim_p == 1))
+       ,.reset_i(reset_i)
+       ,.freeze_i(be_checker.scheduler.int_regfile.cfg_bus.freeze)
 
-         ,.mhartid_i(be_checker.scheduler.int_regfile.cfg_bus.core_id)
+       ,.mhartid_i(be_checker.scheduler.int_regfile.cfg_bus.core_id)
 
-         ,.commit_v_i(be_calculator.commit_pkt.instret)
-         ,.commit_pc_i(be_calculator.commit_pkt.pc)
-         ,.commit_instr_i(be_calculator.commit_pkt.instr)
+       ,.commit_v_i(be_calculator.commit_pkt.instret)
+       ,.commit_pc_i(be_calculator.commit_pkt.pc)
+       ,.commit_instr_i(be_calculator.commit_pkt.instr)
 
-         ,.rd_w_v_i(be_calculator.wb_pkt.rd_w_v)
-         ,.rd_addr_i(be_calculator.wb_pkt.rd_addr)
-         ,.rd_data_i(be_calculator.wb_pkt.rd_data)
+       ,.rd_w_v_i(be_calculator.wb_pkt.rd_w_v)
+       ,.rd_addr_i(be_calculator.wb_pkt.rd_addr)
+       ,.rd_data_i(be_calculator.wb_pkt.rd_data)
 
-         ,.interrupt_v_i(be_mem.csr.trap_pkt_cast_o._interrupt)
-         ,.cause_i(be_mem.csr.trap_pkt_cast_o.cause)
-         );
+       ,.interrupt_v_i(be_mem.csr.trap_pkt_cast_o._interrupt)
+       ,.cause_i(be_mem.csr.trap_pkt_cast_o.cause)
+       );
 
   bind bp_be_director
     bp_be_nonsynth_npc_tracer
--- a/bp_top/test/tb/bp_top_trace_demo/testbench.v
+++ b/bp_top/test/tb/bp_top_trace_demo/testbench.v
@@ -467,7 +467,7 @@ bp_cce_mmio_cfg_loader
     ,.inst_ram_addr_width_p(cce_instr_ram_addr_width_lp)
     ,.inst_ram_els_p(num_cce_instr_ram_els_p)
     ,.skip_ram_init_p(skip_init_p)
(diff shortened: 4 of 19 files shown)


dpetrisko:
Is there a way we could do this in one command?  i.e. from make sim.v COSIM_P=1, first generate the dromajo checkpoint and then do the cosim? I get that we don't want to premake all the checkpoints because they're big

bp_be/src/include/bp_be_internal_if_defines.vh
```diff
@@ -124,6 +124,7 @@
   /* TODO: make opcode */                                                                          \
   typedef struct packed                                                                            \
   {                                                                                                \
+    logic [dword_width_p-1:0]       cause;                                                         \
```
dpetrisko: Is this just used for validating against dromajo? (no hardware use?)

bp_be/src/include/bp_be_internal_if_defines.vh
```diff
@@ -189,7 +190,7 @@
    )
  
 `define bp_be_trap_pkt_width(vaddr_width_mp) \
-  (2 * vaddr_width_mp + rv64_priv_width_gp + 4)
+  (2 * vaddr_width_mp + rv64_priv_width_gp + dword_width_p + 4)
```
dpetrisko: Can we make this smaller than a full word?

bp_common/test/src/demos/Makefile
```diff
@@ -1,8 +1,8 @@
 
 include Makefile.frag
```
dpetrisko: So this means we're using bsg_newlib right?

bp_common/test/src/perch/start.S
```diff
@@ -5,7 +5,7 @@ _start:
 /* setup stack pointers. Stacks start at 0x8FFF_CFFF */
 /* We then subtract off 8K*coreID. The top 4K is for the core emulation stack
  * the lower 4K is for the program TODO: Maybe larger stack value? */
-    li   sp, 0x8FFFCFF0
+    li   sp, 0x80010000
```
dpetrisko: Seems like a pretty small stack.  Have you tried running e.g. SPEC VPR with this?

Maybe we should make it configurable?

bp_top/test/tb/bp_top_trace_demo/testbench.v
```diff
@@ -143,6 +145,30 @@ wrapper
        ,.rd_data_i(be_calculator.wb_pkt.rd_data)
        );
 
+  if(cosim_p)
```
dpetrisko: Can you do the parameter-based clockgating of the other tracers? The reason is that this causes a mismatch in Verilator and VCS where it'll sometimes build in one and not the other due to how they elaborate

taylor-bsg:
I think at least a rule that is like:
make PROG=<test>.dromajo-resume.<n> so the user does not need to specify
all the extra stuff that is always the same?

How big are the checkpoints? We can generate them on a non-backed up shadow
directory if they are big. And/or we can pipe directly to and from
Facebook’s z standard compression algorithm. That might be an interesting
feature to open source to Dromajo if they don’t already have it.

M
————


cd bp_top/syn
make build.v sim.v PROG=<test>.dromajo.<n> PRELOAD_MEM_P=1 LOAD_NBF_P=1
COSIM_P=1

bp_be/src/include/bp_be_internal_if_defines.vh
farzamgl: Yes

bp_be/src/include/bp_be_internal_if_defines.vh
farzamgl: I don't think so, because the high bit indicates whether it's an interrupt.

dpetrisko:
Interesting idea! The checkpoint scales linearly with expanded memory size. So say, 2 MB per test. At several hundred tests, it gets a little cumbersome. 

I think it makes more sense to compile the test, the dromajo checkpoint, etc right before simulation.

1) space efficient 
2) faster (users don’t have to pre-compile all tests before running them)
3) can compile with different options per simulation. Useful for trying tests with different linker scripts, parameters etc.

bp_common/test/src/demos/Makefile
farzamgl: I think it works with either standard or bsg_newlib. We need that to import some printf functionality from the C library.

bp_common/test/src/perch/start.S
farzamgl: No, I don't have access to that repo. How does the configuration work though? Because I think it should be recompiled every time we change the stack size.

bp_top/test/tb/bp_top_trace_demo/testbench.v
farzamgl: Yeah, you're right. I'll change that.

farzamgl:
> Is there a way we could do this in one command? i.e. from make sim.v COSIM_P=1, first generate the dromajo checkpoint and then do the cosim? I get that we don't want to premake all the checkpoints because they're big

In my view, these are 2 separate jobs. One generates a test and the other one runs a simulation. Also, in the future, I might add the capability to run the cosim from the beginning of the test without any checkpointing(for small regressions maybe). So I think it makes sense for them to be separated.

bp_be/src/include/bp_be_internal_if_defines.vh
farzamgl: Actually, since we only notify Dromajo on interrupts(not sync exceptions), I can only send the low bits to the nonsynth_cosim  module and recreate the whole 64-bit over there. But what's the concern here? Because it's gonna be optimized out in any synthesis since it's not used, right?

taylor-bsg:
Two thoughts:

a. I think it's still useful to make it so that file size is not a big
concern for Dromajo by making compressed files the default. In particular,
we don't have to worry about guessing the required memory size that much,
and across a large number of tests and users, we get a large savings, even
if we don't drop that many checkpoints per test...

I downloaded zstd and tried the following test on a 24M file full of 0's:

mbt@MTs-MacBook-Pro zstd % dd if=/dev/zero of=output.dat  bs=1m  count=24

24+0 records in

24+0 records out

25165824 bytes transferred in 0.006383 secs (3942668320 bytes/sec)


mbt@MTs-MacBook-Pro zstd % time ./zstd output.dat

output.dat           :  0.00%   (*25165824* =>    *805 bytes*,
output.dat.zst)

./zstd output.dat  0.02s user 0.01s system 2% cpu 1.188 total


The compression ratio is 29000X, and it was almost instantaneous. So if we
have memory files that are mostly sized for safety and are zeros, the
compressed storage and compute requirements are very small.

Maybe Farzam can take a look and see how compressible the files are for the
snapshots?

b. The concern with running in Dromajo to a checkpoint location, and then
switching over to RTL is that for long-running programs, you are redoing a
bunch of Dromajo simulation. So for Linux or spec, which runs a few billion
cycles, it adds some overhead. A way to get around this would be to do a
Dromajo-to-Dromajo checkpoint? I.e. checkpoint every 15 seconds of Dromajo
(=15*12 M instructions = ~  200M instructions). So you have a layer that
runs dromajo checkpoints, and then you can use Dan's method of fast
forwarding to one checkpoint with only 15 seconds of wasted time.

bp_be/src/include/bp_be_internal_if_defines.vh
dpetrisko: It's a good question.  @taylor-bsg and I had a discussion on this a ways back.  Basically, how much should we rely on synthesis tools? 

One school of thought is to forbid wires only intended for nonsynth use, to make the hardware intent explicit.

Another is that, like you say, those wires should be ripped out.  However, different synth settings could lead to extra buffer insertion, i.e. if this packet is on a feedthrough path, without flattening enabled.

The other question is if we ever think mcause will be used in the trap packet, for example to send different types of messages to the front end.

bp_common/test/src/perch/start.S
dpetrisko: What do you mean by the configuration?

But yes, this way we need to recompile libperch for each different stack size.  I think some way of the firmware querying the stack size of the program is ideal.  I guess it should use the same mechanism as getargs? (which we haven't implemented yet)

dpetrisko:
Certainly attaching program build / checkpoint to simulation run is not a showstopper, we can discuss in a separate PR

bp_be/src/include/bp_be_internal_if_defines.vh
farzamgl: I don't think it'll be used in the logic unless they change smth in the riscv spec.

farzamgl:
The big checkpoint file which is the mainram binary is quite compressible because usually large partitions of the memory are zero, and the compress/decompress seems to be pretty quick. But what's the point of compressing them or saving them in the repo? Those dumps should be generated based on the test and the checkpoint that the user wants, and the Dromajo run is usually pretty fast.

Also, doing multiple checkpoints every N instructions on Dromajo is on my todo list, but needs some non-minor changes to Dromajo, so I prefer finalizing this first iteration before jumping to that.

dpetrisko:
I would like to see cosim without checkpoint before we merge this in.  Otherwise, it's not going to help the issue of silent failures

farzamgl:
I agree, that way we can make that the default option for regressions too. I'll check to see if Dromajo supports cosim without a checkpoint. I might have to add that functionality.

taylor-bsg:
if you break Linux or Spec up into 1 M instruction pieces to run, Dromajo
fast forwarding from the beginning to each individual piece repeats a bunch
of computation turning it into a N^2 computation.

It’s quite common to have a bunch of these files lying around after a run,
so that’s why compression is helpful.


On Thu, Feb 6, 2020 at 4:32 PM Farzam Gilani <notifications@github.com>
wrote:

> The big checkpoint file which is the mainram binary is quite compressible
> because usually large partitions of the memory are zero, and the
> compress/decompress seems to be pretty quick. But what's the point of
> compressing them or saving them in the repo? Those dumps should be
> generated based on the test and the checkpoint that the user wants, and the
> Dromajo run is usually pretty fast.
>
> Also, doing multiple checkpoints every N instructions on Dromajo is on my
> todo list, but needs some non-minor changes to Dromajo, so I prefer
> finalizing this first iteration before jumping to that.
>

Makes sense!

dpetrisko:
> Dromajo supports cosim without a checkpoint.

Could you just make a checkpoint before the first instruction?

> if you break Linux or Spec up into 1 M instruction pieces to run, Dromajo
fast forwarding from the beginning to each individual piece repeats a bunch
of computation turning it into a N^2 computation.

K * N where K is number of checkpoints, and N is number of instructions.  And these K can be run in parallel, same as the cosims.

I think we're all on the same page that it's a good feature to have and we should raise an issue for it, but I don't think it's essential right now.

farzamgl:
> Could you just make a checkpoint before the first instruction?

Either I've to do that or just tell Dromajo to start at the beginning without a bootram like the simulation process. I think the second way is cleaner but I'll let you know when I take a look at the C code.

farzamgl:
Okay so cosim now works for without checkpointing. Turns out we just had to give Dromajo the elf file instead of the json file which has the checkpoint info.
`make build.v sim.v PROG=<test> PRELOAD_MEM_P=1 LOAD_NBF_P=0 COSIM_P=1`

dpetrisko:
Awesome!

bp_be/src/include/bp_be_internal_if_defines.vh
dpetrisko: How about just using mcause_li?

dpetrisko: It's a little less clean in the cosim module, but I'd prefer it over adding unnecessary wires

dpetrisko: Actually, forget it.  This might actually make another change I'm planning easier

PR #319, opened 2020-02-11
Refactoring Cache Miss Interface

Shashank-Vijay:
- Added standard cache miss interface.

- Miss tracking enabled within cache.


bp_be/src/include/bp_be_dcache/bp_be_dcache_lce_pkt.vh
```diff
@@ -10,47 +10,47 @@
 `define BP_BE_DCACHE_LCE_PKT_VH
 
 `include "bp_common_me_if.vh"
-
+/*
```
dpetrisko: Delete this file if unused

bp_fe/src/v/bp_fe_icache.v
```diff
@@ -270,22 +322,75 @@ module bp_fe_icache
...
+  bp_cache_data_mem_pkt_s lce_data_mem_pkt;
   assign lce_data_mem_pkt = lce_data_mem_pkt_i;
-  bp_fe_icache_lce_tag_mem_pkt_s tag_mem_pkt;
+  bp_cache_tag_mem_pkt_s tag_mem_pkt;
   assign tag_mem_pkt = lce_tag_mem_pkt_i;
-  bp_fe_icache_lce_stat_mem_pkt_s stat_mem_pkt;
+  bp_cache_stat_mem_pkt_s stat_mem_pkt;
   assign stat_mem_pkt = lce_stat_mem_pkt_i;
 
+  enum logic [1:0] { A, B, C } state_n, state_r;
```
dpetrisko: Need to simplify

bp_be/src/v/bp_be_mem/bp_be_dcache/bp_be_dcache_lce_req.v
```diff
@@ -159,15 +170,19 @@ module bp_be_dcache_lce_req
     lce_resp.addr = miss_addr_r;
     lce_resp.data = '0;
 
+    cache_miss_ready_o = 1'b1;
```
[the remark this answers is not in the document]
Shashank-Vijay: They were essentially the same thing. We have removed one of them.

bp_be/src/v/bp_be_mem/bp_be_dcache/bp_be_dcache.v
```diff
@@ -130,28 +140,35 @@ module bp_be_dcache
...
     // for lock logic inside LCE
     , output logic lr_hit_tv_o
+    // for logic inside LCE
+     , output logic store_hit_o
+
 
     // data_mem
     , input lce_data_mem_pkt_v_i
     , input [dcache_lce_data_mem_pkt_width_lp-1:0] lce_data_mem_pkt_i
+    // TODO: Keep the data_mem_data_o. Reflect this in icache
```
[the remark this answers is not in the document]
Shashank-Vijay: This signal is used in transfers so we have kept it intact. The interface handles the data for writebacks and stores.

bp_be/src/v/bp_be_mem/bp_be_dcache/bp_be_dcache.v
```diff
@@ -116,6 +123,9 @@ module bp_be_dcache
 
     , input [dcache_pkt_width_lp-1:0] dcache_pkt_i
     , input v_i
+
+    // TODO: What ready_i signal is this?
+    // LCE ready signal
```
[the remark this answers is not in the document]
dpetrisko: Never mind, this is a new signal because we can no longer rely on the lce_ready signal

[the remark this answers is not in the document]
Shashank-Vijay: But the ready_i signal is connected to the lce_ready. So it is the same.

bp_be/src/v/bp_be_mem/bp_be_dcache/bp_be_dcache_lce_req.v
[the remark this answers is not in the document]
Shashank-Vijay: We saw that having only one of them makes the system go on a combinational loop. We can resolve it by having either the ready_o signal as before and the cache_miss_ready_o, or we can change the nature of the interface and put in a FIFO.

bp_be/src/v/bp_be_mem/bp_be_dcache/bp_be_dcache.v
```diff
@@ -665,20 +688,167 @@ module bp_be_dcache
...
+	cache_miss = 1'b0;
+      end
+      else begin
+	cache_miss = 1'b1;
+      end 
+    end
+  end
+*/
+
+  enum logic [1:0] { A, B, C } state_n, state_r;
```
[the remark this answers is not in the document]
Shashank-Vijay: This FSM is tracking the current miss based on v_o and resolves the miss based on the tag and data packets if it is cached or if the data packet returned is for an uncached request.

We considered using the lce_ready directly, but that signal is also based on the timeout of the LCE and we thought there might be a case where the masking would fail. 

No, we think cache_miss is only used to mask the v_o.

bp_be/src/v/bp_be_mem/bp_be_dcache/bp_be_dcache.v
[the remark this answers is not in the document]
dpetrisko: Transfers should be same as writeback from the perspective of the cache, no?

bp_be/src/v/bp_be_mem/bp_be_dcache/bp_be_dcache.v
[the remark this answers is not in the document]
dpetrisko: This should be at most a flop, possibly just a logical expression

bp_be/src/v/bp_be_mem/bp_be_dcache/bp_be_dcache_lce_req.v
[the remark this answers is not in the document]
dpetrisko: No fifo, but where’s the timing loop from? If you run with vcs it should print a loop report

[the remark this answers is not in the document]
Shashank-Vijay: So, the signal cache_miss_o in the lce_req is being generated based on the cache_miss_v_i signal. The ready_o signal is based on the cache_miss_o signal (from lce_req). If we use this as the ready signal from the LCE, then we get a timing loop because cache_miss_v_o (in the cache) is generated based on the ready from the LCE.

bp_be/src/v/bp_be_mem/bp_be_dcache/bp_be_dcache.v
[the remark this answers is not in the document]
Shashank-Vijay: We tried using the flop before, but we were getting a one cycle delay which messed everything up in terms of the data going back to the core. I think, we came to you with this problem earlier.

bp_be/src/v/bp_be_mem/bp_be_dcache/bp_be_dcache_lce_req.v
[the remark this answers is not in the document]
dpetrisko: Didn't we decide not to have cache_miss logic in the lce? That's why we moved it to the caches?

bp_be/src/v/bp_be_mem/bp_be_dcache/bp_be_dcache.v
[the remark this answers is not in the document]
dpetrisko: That doesn't make sense to me.  Shouldn't that be cache_miss_ready_i?

bp_be/src/v/bp_be_mem/bp_be_dcache/bp_be_dcache.v
```diff
@@ -111,7 +123,10 @@ module bp_be_dcache
 
     , input [dcache_pkt_width_lp-1:0] dcache_pkt_i
     , input v_i
-    , output logic ready_o
```
dpetrisko: This should be what the core side looks at to see if cache is ready, rather than lce_ready

bp_be/src/v/bp_be_mem/bp_be_dcache/bp_be_dcache.v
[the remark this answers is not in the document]
dpetrisko: If the problem is that there's a one cycle delay, the solution is to use (cache_miss | cache_miss_r)

bp_be/src/v/bp_be_mem/bp_be_dcache/bp_be_dcache_lce.v
```diff
@@ -126,25 +132,64 @@ module bp_be_dcache_lce
     , output logic lce_cmd_v_o
     , input lce_cmd_ready_i
 
-    , output credits_full_o
-    , output credits_empty_o
+    , output logic credits_full_o
+    , output logic credits_empty_o
+
+    // for locking logic
```
dpetrisko: The locking logic should live in the cache, per the interface spec

bp_be/src/v/bp_be_mem/bp_be_dcache/bp_be_dcache.v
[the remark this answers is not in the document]
dpetrisko: @muwyse I'm not entirely sure here, please correct me.

You are right that there are cases here: basically, which packets we need to wait for for the miss to be complete.  We detect a cache miss initially based on store_miss | load_miss | etc....  
Then we clear the miss:
1) when we ack a set_tag_and_wakeup and a stat_pkt
2) when we ack any other tag command and a stat_pkt and a data_pkt
Because the LCE waits until all packets are acked before moving on, it's enough to & the yumi signals, don't need any additional state

bp_be/src/v/bp_be_mem/bp_be_dcache/bp_be_dcache_lce_req.v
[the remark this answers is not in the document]
sripathi-muralitharan: Yes, the cache miss logic is in the cache. The LCE is using the cache_miss_o signal (you can actually rename it) to be able to tell whether it is ready to receive another request or not. The resolution of the miss is still happening within the cache.

bp_be/src/v/bp_be_mem/bp_be_dcache/bp_be_dcache_lce.v
Shashank-Vijay: Yes, we will shift this logic to the cache.

bp_be/src/v/bp_be_mem/bp_be_dcache/bp_be_dcache.v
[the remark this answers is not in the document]
Shashank-Vijay: In the bp_be_mem_top module, the lce_ready_i signal is connected to this ready_i signal in the dcache module. The dcache_pkt_i and v_i are controlled by the ptw_busy signal and the ready_i in ptw is connected to the lce_ready_i. So it looks like there is no ready signal going from cache to processor. 

Right now we're ending up with two different ready signals from the LCE, one which controls the tl_we and is dependent on the cache_miss_signal in LCE and a timeout, and a second one which is used by the cache miss packet interface. We tried to combine this functionality into one, but that resulted in a combinational loop because the ready (in LCE) is using cache_miss_v_o (into the LCE) to determine its value while the cache_miss_v_o (in the cache) is using the ready (from LCE) to determine its value.

bp_be/src/v/bp_be_mem/bp_be_dcache/bp_be_dcache.v
```diff
@@ -975,7 +1133,9 @@ module bp_be_dcache
 
   // write buffer
   //
-  assign wbuf_v_li = v_tv_r & store_op_tv_r & store_hit & ~sc_fail & ~uncached_tv_r;
+  // disallow write buffer write on store hit that cannot be processed by LCE
+  // to avoid multiple wbuf entries when the store replays
+  assign wbuf_v_li = v_tv_r & store_op_tv_r & store_hit & ~sc_fail & ~uncached_tv_r & ~cache_miss;
```
dpetrisko: If there's a store hit, then there's not a cache miss.  So the LCE state should be irrelevant, no?

bp_be/src/v/bp_be_mem/bp_be_dcache/bp_be_dcache.v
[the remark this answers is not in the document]
dpetrisko: The LCE should not depend on the miss state of the cache.  All it cares about is that it is in a state which is able to accept requests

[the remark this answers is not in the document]
dpetrisko: The cache_pkt_v_i should depend on the cache's ready_o, not the LCE.  One of the points of this refactor is the logical separation between LCE and core -- there should be _zero_ connections between them

bp_be/src/v/bp_be_mem/bp_be_dcache/bp_be_dcache_lce_req.v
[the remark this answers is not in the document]
dpetrisko: That's what I mean though.  cache_miss_ready_o is the helpful interface to the cache.  It should not depend on the miss state of the cache.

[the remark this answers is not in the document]
sripathi-muralitharan: I get the point. So, in the end, are we going to have just one signal called cache_miss_ready_o from the LCE and not two signals - cache_miss_ready_o & lce_ready_o (currently called ready_o)??

bp_be/src/v/bp_be_mem/bp_be_dcache/bp_be_dcache.v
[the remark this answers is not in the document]
sripathi-muralitharan: For the cache that is responding to the transfer command, it is not the same. 

For example:
Cache 1 asks for some data. The CCE realises that this data is present in Cache 2. It sends a command to LCE 2 to transfer that data to LCE 1 (and hence Cache 1). LCE 2 then has to request this data from Cache 2. This is not a miss from Cache 2's perspective, right? If that's the case, sending the data through the cache miss interface will raise a false miss and also potentially tamper the miss resolution logic.

bp_be/src/include/bp_be_dcache/bp_be_dcache_pkg.vh
```diff
@@ -54,8 +55,24 @@ package bp_be_dcache_pkg;
     
   } bp_be_dcache_lce_data_mem_opcode_e;
 
+
+  typedef enum logic [1:0] {
```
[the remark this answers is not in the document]
muwyse: Perhaps this whole package/header should be consolidated into a single cache package. The I$ would then, naturally, support only a subset of operations (i.e., it would not support stores).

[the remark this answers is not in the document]
dpetrisko: agreed

bp_be/src/v/bp_be_mem/bp_be_dcache/bp_be_dcache_lce.v
sripathi-muralitharan: In case of the lock logic, currently the LCE contains the logic to block the processing of LCE cmds right? If we move this to the cache, then is it right in assuming that we now need to block the processing of the dcache_pkts? As far as the requirements are concerned, I think the current logic already meets all of them and since, we have the appropriate signals in the cache, we can directly shift it there. However, I don't understand the minor tweaks that are required.

bp_be/src/v/bp_be_mem/bp_be_dcache/bp_be_dcache.v
```diff
@@ -858,24 +1030,32 @@ module bp_be_dcache
...
       end
-      e_dcache_lce_tag_mem_set_tag: begin
+      e_cache_tag_mem_set_tag: begin
         tag_mem_data_li = {lce_assoc_p{lce_tag_mem_pkt.state, lce_tag_mem_pkt.tag}};
         for (integer i = 0; i < lce_assoc_p; i++) begin
           tag_mem_mask_li[i].coh_state = {`bp_coh_bits{lce_tag_mem_way_one_hot[i]}};
           tag_mem_mask_li[i].tag = {ptag_width_lp{lce_tag_mem_way_one_hot[i]}};
         end
       end
+      /*e_cache_tag_mem_set_tag_wakeup: begin
```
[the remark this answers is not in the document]
sripathi-muralitharan: We realised that there was no need for this command and the reasoning is as follows. Please correct me if I am wrong here - 

Misses are resolved in the following cases - 
1. Uncached loads - if the data_mem_opcode is uncached
2. Uncached stores - nothing special is done. upon receiving the uc_st_done command, only the credit counters get decremented. (All this is already happening in the LCE)
3. Cached loads - Need to check if valid tag and data mem packets are set. tag packet needs to use set_tag opcode and data_mem needs to use data_mem_write opcode.
4. Cached stores - Data to be stored waits in the write buffer while the block is brought to the cache. This is again done through the data command and hence the conditions are the same as for cached loads.

bp_be/src/v/bp_be_mem/bp_be_dcache/bp_be_dcache_lce.v
dpetrisko: Yes, blocking the pkts will block the command, since the transaction is only completed when a lce_resp is sent, which depends on pkts, which are generated by lce_cmd

dpetrisko:
@sripathi-muralitharan @Shashank-Vijay 
There are some changes to the interface which will be required for the UCE implementation.  I added them to the service interface doc and highlighted the diff.  Should be fairly minimal changes

bp_be/src/v/bp_be_mem/bp_be_dcache/bp_be_dcache_lce_req.v
```diff
@@ -130,7 +134,7 @@ module bp_be_dcache_lce_req
      );
 
   always_comb begin
-    cache_miss_o = 1'b0;
+    cache_req_complete_o = 1'b1;
```
dpetrisko: cache_req_complete_o should only be high for 1 cycle on the transition when a transaction is complete

bp_be/src/v/bp_be_mem/bp_be_dcache/bp_be_dcache_lce.v
```diff
@@ -350,8 +304,8 @@ module bp_be_dcache_lce
   // when the timer reaches max, it deasserts ready_o of dcache for one cycle, allowing it to access mem
   // by creating a free slot.
   logic [`BSG_SAFE_CLOG2(timeout_max_limit_p+1)-1:0] timeout_cnt_r;
-  wire coherence_blocked = (data_mem_pkt_v_o | tag_mem_pkt_v_o | stat_mem_pkt_v_o)
-                           & (~data_mem_pkt_yumi_i & ~tag_mem_pkt_yumi_i & ~stat_mem_pkt_yumi_i);
+  wire coherence_blocked = (lce_cmd_v_i & (lce_cmd_in.msg_type != e_lce_cmd_uc_st_done))
```
dpetrisko: wire coherence_blocked = lce_cmd_v_i & ~lce_cmd_yumi_o

bp_be/src/v/bp_be_mem/bp_be_dcache/bp_be_dcache.v
```diff
@@ -116,7 +119,7 @@ module bp_be_dcache
 
     , input [dcache_pkt_width_lp-1:0] dcache_pkt_i
     , input v_i
-    , input ready_i
```
dpetrisko: Add ready_o, remove dcache_miss_o (combine the conditions)

bp_be/src/v/bp_be_mem/bp_be_dcache/bp_be_dcache.v
```diff
@@ -688,21 +757,54 @@ module bp_be_dcache
...
+  always_ff @(posedge clk_i)
+    cache_miss_r <= cache_miss;
+  wire cache_miss_resolved = cache_miss_r & ~cache_miss;
+
+  logic [`BSG_SAFE_CLOG2(lock_max_limit_p+1)-1:0] lock_cnt_r;
+  wire lock_clr = v_o || (lock_cnt_r == lock_max_limit_p);
+  wire lock_inc = ~lock_clr & (cache_miss_resolved || lr_hit_tv || (lock_cnt_r > 0));
+
+  logic lock_reset, cache_reset;
+  assign cache_reset = (lce_tag_mem_pkt_v_i && lce_stat_mem_pkt_v_i) &&  (lce_stat_mem_pkt.opcode == e_cache_stat_mem_set_clear && lce_tag_mem_pkt.opcode == e_cache_tag_mem_set_clear);
```
dpetrisko: why do we need this?

bp_be/src/v/bp_be_mem/bp_be_dcache/bp_be_dcache.v
```diff
@@ -688,21 +757,54 @@ module bp_be_dcache
...
+    (.clk_i(clk_i)
+     ,.reset_i(lock_reset)
+
+     ,.clear_i(lock_clr)
+     ,.up_i(lock_inc)
+     ,.count_o(lock_cnt_r)
+     );
+  
+  assign cache_lock = (lock_cnt_r != '0);
+  assign lce_data_mem_pkt_v = lce_data_mem_pkt_v_i & ~cache_lock;
```
dpetrisko: Should block readies, not valids.

bp_fe/src/v/bp_fe_lce.v
```diff
@@ -228,8 +241,8 @@ module bp_fe_lce
       timeout_cnt_n = '0;
     end
     else begin
-      if (data_mem_pkt_v_o | tag_mem_pkt_v_o | stat_mem_pkt_v_o) begin
-        timeout_cnt_n = ~(data_mem_pkt_yumi_i | tag_mem_pkt_yumi_i | stat_mem_pkt_yumi_i)
+      if (lce_cmd_v_i) begin
+        timeout_cnt_n = ~(data_mem_pkt_ready_i | tag_mem_pkt_ready_i | stat_mem_pkt_ready_i)
```
dpetrisko: similar to dcache: lce_cmd_v_i & ~lce_cmd_yumi_o

bp_fe/src/v/bp_fe_lce.v
```diff
@@ -249,6 +262,6 @@ module bp_fe_lce
   end
 
   wire lce_ready = (cfg_bus_cast_i.icache_mode == e_lce_mode_uncached) ? 1'b1 : lce_ready_lo;
-  assign ready_o = lce_ready & ~timeout & ~cache_miss_o;
+  assign lce_ready_o = lce_ready & ~timeout & cache_req_complete_o ;
```
dpetrisko: We want there to only be 1 ready signal -- the cache_req_ready_o.  Can we have the timeout logic in lce_req (pipe in a coherence_blocked wire) and modify the ready signal in there?

bp_fe/src/v/bp_fe_icache.v
```diff
@@ -270,22 +287,64 @@ module bp_fe_icache
...
+    #(.width_p(1)
+     ,.reset_val_p(0))
+     cache_miss_tracker
+     (.clk_i(clk_i)
+     ,.reset_i(reset_i || miss_tracker_reset_li)
+     ,.en_i(miss_tracker_en_li)
+     ,.data_i(cache_req_v_o)
+     ,.data_o(miss_tracker_lo)
+     );
+
```
dpetrisko: logic miss_tracker_r;
bsg_dff_reset_en
    #(.width_p(1)
     ,.reset_val_p(0))
     cache_miss_tracker
     (.clk_i(clk_i)
     ,.reset_i(reset_i)
     ,.en_i(cache_req_v_o | cache_req_complete_i)
     ,.data_i(cache_req_v_o)
     ,.data_o(miss_tracker_r)
     );

This en_i variant will synthesize better.  _r complies more with the coding standard.  Shouldn't be functionally different

bp_fe/src/v/bp_fe_icache.v
```diff
@@ -270,22 +287,64 @@ module bp_fe_icache
...
+  bp_cache_data_mem_pkt_s lce_data_mem_pkt;
   assign lce_data_mem_pkt = lce_data_mem_pkt_i;
-  bp_fe_icache_lce_tag_mem_pkt_s tag_mem_pkt;
+  bp_cache_tag_mem_pkt_s tag_mem_pkt;
   assign tag_mem_pkt = lce_tag_mem_pkt_i;
-  bp_fe_icache_lce_stat_mem_pkt_s stat_mem_pkt;
+  bp_cache_stat_mem_pkt_s stat_mem_pkt;
   assign stat_mem_pkt = lce_stat_mem_pkt_i;
 
+  always_comb begin
```
dpetrisko: What happens when cache_req is not ready but there is a miss in tv?

bp_common/src/include/bp_common_cache_service.vh
```diff
@@ -0,0 +1,148 @@
...
+  typedef struct packed {                                                 \
+    logic [`BSG_SAFE_CLOG2(sets_mp)-1:0]    index;                        \
+    logic [`BSG_SAFE_CLOG2(ways_mp)-1:0]    way_id;                       \
+    bp_cache_stat_mem_opcode_e              opcode;                       \
+  } bp_cache_stat_mem_pkt_s;
+
+`define bp_cache_stat_mem_pkt_width(sets_mp, ways_mp) \
+  (`BSG_SAFE_CLOG2(sets_mp)+`BSG_SAFE_CLOG2(ways_mp)+`bp_cache_stat_mem_opcode_width)
+
+`define declare_bp_cache_if_widths(ways_mp, sets_mp, tag_width_mp, lce_data_width_mp)                                \
```
dpetrisko: Include declare_bp_cache_req_widths here

sripathi-muralitharan: If you want cache_req_widths to create localparams for the data, tag and stat men packets, then we are calling it cache_if_widths. cache_req_widths is associated with cache_req_o

sripathi-muralitharan: If you want cache_req_widths to create localparams for the data, tag and stat men packets, then we are calling it cache_if_widths. cache_req_widths is associated with cache_req_o

dpetrisko: Sorry, I think I wasn't clear.  I meant

`declare_bp_cache_if_widths
, localparam bp_cache_req_width_lp = `bp_cache_req_width(data_width_mp, ways_mp, paddr_width_mp)
, localparam bp_cache_data_mem_pkt_width_lp = 
....
`

dpetrisko: Additionally, drop the bp_ prefix for the width names to follow convention

bp_be/src/v/bp_be_mem/bp_be_dcache/bp_be_dcache.v
Shashank-Vijay: We wanted to move the lock logic from the LCE to the cache right?

bp_be/src/v/bp_be_mem/bp_be_dcache/bp_be_dcache.v
Shashank-Vijay: What ready_o is this?

bp_be/src/v/bp_be_mem/bp_be_dcache/bp_be_dcache_lce.v
Shashank-Vijay: What about the conditions based on the readies for the coherence_blocked signal?

bp_be/src/v/bp_be_mem/bp_be_dcache/bp_be_dcache.v
dpetrisko: dcache_pkt_i
v_i
ready_o

bp_be/src/v/bp_be_mem/bp_be_dcache/bp_be_dcache.v
dpetrisko: why do we need a special cache_reset condition?

bp_be/src/v/bp_be_mem/bp_be_dcache/bp_be_dcache_lce.v
dpetrisko: It requires more redundant logic.  Why do you think this is insufficient?

bp_fe/src/v/bp_fe_icache.v
dpetrisko: To clarify, I believe it should force a miss.  That way, the core will replay the miss once the CE is ready

bp_be/src/v/bp_be_mem/bp_be_dcache/bp_be_dcache_lce.v
Shashank-Vijay: Yeah, it's redundant. I'll remove it.

bp_fe/src/v/bp_fe_icache.v
dpetrisko: This has not been fixed

bp_fe/src/v/bp_fe_icache.v
```diff
@@ -464,10 +521,21 @@ module bp_fe_icache
...
+      lce_tag_mem_pkt_way_r <= tag_mem_pkt.way_id;
+    end
+  end
+
+  assign lce_tag_mem_o = tag_mem_data_lo[lce_tag_mem_pkt_way_r][0+:tag_width_lp];
+  assign lce_tag_mem_pkt_ready_o = ~tl_we;
 
   // LCE: stat_mem
-  assign lce_stat_mem_pkt_yumi_o = ~(v_tv_r & ~uncached_tv_r) & lce_stat_mem_pkt_v_i;
+  assign lce_stat_mem_o = 1'b0;
```
dpetrisko: ? read from stat mem?

sripathi-muralitharan: This is the I-Cache. Why would we wanna read from the stat mem? Anyways, nothing is dirty. stat_mem_read is not used for anything other looking at the dirty bits right?

bp_fe/src/v/bp_fe_icache.v
sripathi-muralitharan: Yes, with the current commit on the PR it is not fixed. We haven't committed to the branch with the new changes.

bp_fe/src/v/bp_fe_icache.v
dpetrisko: stat_mem has both lru and dirty bits

bp_fe/src/v/bp_fe_icache.v
dpetrisko: please don't resolve comments until the issue is fixed and committed

bp_be/src/v/bp_be_mem/bp_be_dcache/bp_be_dcache.v
Shashank-Vijay: We removed it now.

bp_fe/src/v/bp_fe_icache.v
sripathi-muralitharan: I totally forgot you needed both components for the UCE. I'll add that right away.

dpetrisko: It's actually not important for the UCE, but it's just good for symmetry

bp_fe/src/v/bp_fe_icache.v
```diff
@@ -270,22 +287,59 @@ module bp_fe_icache
...
+     ,.en_i(cache_req_v_o | cache_req_complete_i)
+     ,.data_i(cache_req_v_o)
+     ,.data_o(miss_tracker_r)
+     );
+
+  assign cache_miss = cache_req_v_o || miss_tracker_r;
+
   // Fault if in uncached mode but access is not for an uncached address
-  assign data_v_o = v_tv_r & ((uncached_tv_r & uncached_load_data_v_r) | ~cache_miss_i);
+  assign data_v_o = v_tv_r & ((uncached_tv_r & uncached_load_data_v_r) | ~cache_miss);
```
dpetrisko: same comment as dcache, should depend on tag comparison, not cache_miss

bp_be/src/v/bp_be_mem/bp_be_dcache/bp_be_dcache_lce_req.v
```diff
@@ -23,36 +23,35 @@
...
-    , input uncached_load_req_i
-    , input uncached_store_req_i
-    , input [dword_width_p-1:0] store_data_i
-    , input [1:0] size_op_i
+    , input [bp_cache_req_width_lp-1:0] cache_req_i
+    , input cache_req_v_i
+    , output logic cache_req_ready_o
 
-    , output logic cache_miss_o
+    , output logic cache_req_complete_o
```
dpetrisko: Cache req complete should come from lce_cmd, not lce_req.  The request is completed once the pkts are set, not when the state changes to begin making changes.

bp_be/src/v/bp_be_mem/bp_be_dcache/bp_be_dcache_lce.v
```diff
@@ -369,6 +325,6 @@ module bp_be_dcache_lce
 
   // LCE Ready Signal
   wire lce_ready = lce_ready_lo;
-  assign ready_o = lce_ready & ~timeout & ~cache_miss_o; 
+  assign cache_req_ready_o = lce_ready & ~timeout & cache_req_ready_lo;
```
dpetrisko: Would rather move this timeout logic into the lce_req module.  Overriding intermediate ready signals is a recipe for confusion

bp_be/src/v/bp_be_mem/bp_be_dcache/bp_be_dcache.v
```diff
@@ -130,59 +133,66 @@ module bp_be_dcache
     , output logic store_op_tl_o
 
     // ctrl
-    , input cache_miss_i
+    , output dcache_miss_o // Used for mem connections (ptw and MMU resp connections)
```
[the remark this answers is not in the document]
dpetrisko: Never mind!

PR #320, opened 2020-02-14
Changing interface handshake on cce cmd outputs

dpetrisko:
These used to be connected for fifo, so rv& was appropriate.  Now, it's connecting to cache / clint, so r->v has clearer intent.


bp_me/src/v/cce/bp_cce.v
```diff
@@ -58,7 +58,7 @@ module bp_cce
 
    // CCE-MEM Interface
    // inbound: valid->ready (a.k.a., valid->yumi), demanding consumer (connects to FIFO)
-   // outbound: ready&valid (connects to FIFO)
+   // outbound: ready->valid
```
muwyse: make same comment change a few lines up for lce-cce IF comment.

PR #321, opened 2020-02-15
Filter out core id when address matching in clint slice

farzamgl:


farzamgl:
The register address parameters are clint_base + offset, but the mem_cmd address field also contains the core offset. So it was only working on core 0. Now it works on all cores.

dpetrisko:
Oh, I misread dev for cce.  Nice!

farzamgl:
Oh you're actually right. The commit msg should be "Filter out cce id".

PR #324, opened 2020-02-20
Bump dromajo

dpetrisko:
The commit tag for dromajo in dev is not in dromajo/blackparrot_mods.  Was dromajo rebased?


farzamgl:
Yes. The one in sw_dev is valid.

dpetrisko:
Cool, let's just bump it for now to unblock anyone trying to clone to repo

PR #333, opened 2020-02-24
Adding autotrigger for CI

dpetrisko:
This should get of those false negative failures where you update a tool / test and it passes locally but fails in CI.


muwyse:
LGTM

PR #336, opened 2020-02-25
Fixes for verilator cosim build

dpetrisko:


dpetrisko:
It is bits.  Good call, I'll make it a bit bigger

dpetrisko:
Actually, parameter size is [31:0], so we'll run into that before anything else.  I'll leave it for now

PR #337, opened 2020-02-25
Support FPGA Testbench

gaozihou:
Side effects:
1. Update basejump_stl version
2. Tiny syntax fix in bp_cce_dir.v to be compatible with Vivado


bp_me/src/v/cache/bp_me_cce_to_cache_dma.v
```diff
@@ -0,0 +1,238 @@
+// 
+// bp_me_cce_to_cache_dma.v
```
dpetrisko: What's this used for?

bp_top/test/tb/bp_top_fpga/README
```diff
@@ -0,0 +1,30 @@
+
+/* Steps for opening Vivado project: */
+
+cd pre-alpha-release
```
dpetrisko: update to black-parrot

bp_top/test/tb/bp_top_fpga/README
```diff
@@ -0,0 +1,30 @@
...
+
+cd pre-alpha-release/bp_top/test/tb/bp_top_fpga
+vivado -mode tcl
+source bp_fpga.tcl
+start_gui
+
+
+/* Steps for updating project tcl: */
+
+cd pre-alpha-release/bp_top/test/tb/bp_top_fpga
```
dpetrisko: Can we make a makefile target for this to be run in syn?

bp_top/test/tb/bp_top_fpga/README
```diff
@@ -0,0 +1,30 @@
...
+vivado -mode tcl
+open_poroject ./artix_7_xdma/artix_7_xdma.xpr
```
dpetrisko: project*

bp_top/test/tb/bp_top_fpga/README
```diff
@@ -0,0 +1,30 @@
...
+write_project_tcl -force bp_fpga.tcl
+exit
+source modify_tcl.sh
+
+
+/* Switching between single-core and dual-core settings */
+
+In system_wrapper.v:
+
+Single Core: parameter bp_params_e bp_params_p = e_bp_single_core_cfg
```
dpetrisko: Can we use the CFG_P= approach like we do for the other targets?

bp_top/test/tb/bp_top_fpga/bp_fpga.tcl
```diff
@@ -0,0 +1,2599 @@
...
+#*****************************************************************************************
+# NOTE: In order to use this script for source control purposes, please make sure that the
+#       following files are added to the source control system:-
+#
+# 1. This project restoration tcl script (bp_fpga.tcl) that was generated.
+#
+# 2. The following source(s) files that were local or imported into the original project.
+#    (Please see the '$orig_proj_dir' and '$origin_dir' variable setting below at the start of the script)
+#
+#    "/home/paul/artix_7_test/artix_7_xdma_dram_host_bp_cfg_new_exported/artix_7_xdma.srcs/pre-alpha-release/external/basejump_stl/bsg_misc/bsg_defines.v"
```
dpetrisko: I don't really know much about vivado.  Is it a problem that these paths are your local paths?

bp_top/test/tb/bp_top_fpga/bp_fpga.tcl
```diff
@@ -0,0 +1,2599 @@
+#*****************************************************************************************
```
dpetrisko: Is there something more minimal that we can use to recreate the project? It's just that this is a pretty big file and I'm worried that when we regenerate it, it will create enormous diffs every time.

bp_top/test/tb/bp_top_fpga/bp_stream_host.v
```diff
@@ -0,0 +1,111 @@
...
+  ,output [stream_data_width_p-1:0]             stream_data_o
+  ,input                                        stream_ready_i
+  );
+  
+  `declare_bp_me_if(paddr_width_p, cce_block_width_p, lce_id_width_p, lce_assoc_p);
+  
+  // Stream address map
+  logic nbf_v_li, mmio_v_li;
+  logic nbf_ready_lo, mmio_ready_lo;;
+
```
dpetrisko: Comment what these addresses are?

bp_top/test/tb/bp_top_fpga/bp_stream_mmio.v
```diff
@@ -0,0 +1,195 @@
...
+  ,.ready_o(io_resp_ready_lo)
+  ,.data_o (io_resp_o)
+  ,.v_o    (io_resp_v_o)
+  ,.yumi_i (io_resp_ready_i & io_resp_v_o)
+  );
+  
+  logic sipo_v_lo, sipo_yumi_li;
+  logic [1:0][stream_data_width_p-1:0] sipo_data_lo;;
+  
+  bsg_serial_in_parallel_out_full
```
dpetrisko: Shouldn't this be dynamic? The io cmd could be != 64 bits

bp_me/src/v/cache/bp_me_cce_to_cache_dma.v
dpetrisko: Is it just to reuse the bsg_cache_to_axi module?

bp_top/test/tb/bp_top_fpga/system_wrapper.v
```diff
@@ -0,0 +1,714 @@
...
+        nbf_resp_v_li = nbf_resp_ready_lo & load_resp_v_li;
+    
+        cfg_resp_li = '0;
+        cfg_resp_v_li = 1'b0;
+    
+        load_resp_yumi_lo = nbf_resp_v_li;
+      end
+  
+  // Logic Analyzer
+  design_1 design_1_i
```
dpetrisko: More descriptive names?

bp_top/test/tb/bp_top_fpga/system_wrapper.v
```diff
@@ -0,0 +1,714 @@
...
+  ,.cmd_link_o      (proc_cmd_link_li)
+  ,.resp_link_i     (proc_resp_link_lo)
+  ,.resp_link_o     (proc_resp_link_li)
+  );
+   
+  // pcie stream host (NBF and MMIO)
+  logic nbf_done_lo;
+  assign led[3] = nbf_done_lo;
+  
+  bp_stream_host
```
dpetrisko: So I don't fully understand AXI (does anyone? :)), but could you explain what's going on here?  What is the stream_nbf loader doing? Where are the writes from the axi bus to the stream_nbf coming from?

dpetrisko: I get that the host receives I/O commands which it then puts on the bus to blink an led or something, that's fine.  It's mostly the nbf loader I'm confused about here.

bp_me/src/v/cache/bp_me_cce_to_cache_dma.v
gaozihou: Yes it is for reusing cache_to_axi. It looks weird for now, but after we have bsg_wormhole_streamer implemented this will look more natural (going directly from mem_link to cache_dma with streamer, no more cce in between).

bp_top/test/tb/bp_top_fpga/bp_fpga.tcl
gaozihou: These are comments in the tcl file (beginning with "#"), vivado will ignore all of them.

bp_top/test/tb/bp_top_fpga/bp_fpga.tcl
gaozihou: The only time we would like to regenerate it is:

bp_top/test/tb/bp_top_fpga/bp_stream_mmio.v
gaozihou: If I understand correctly, io_cmd data width is at most 64-btis, so I hard-coded here for simplicity. 

Or to make it more "dynamic", we could make els_p = (dword_width_p / stream_data_width_p), although dword_width_p never changes...

bp_top/test/tb/bp_top_fpga/system_wrapper.v
gaozihou: The whole chain is like: bp_processor <---> bp_stream_host <---> bsg_m_axi_lite_to_fifo <---> axi_lite_port. 

bp_stream_host consists of two parts: mmio_host and nbf_loader:
1. mmio_host receives ready/write requests from processor then forwards to the PC host. For read request it also receives responses from PC host then forwards to the processor. 
2. nbf_loader receives write requests from PC host, then forwards to the processor.

bsg_m_axi_lite_to_fifo is an adapter between bsg_stream_host and axi_lite_port. PC->processor traffic is written directly into the buffer fifo. Processor->PC traffic is handled with polling, if buffer fifo is not empty then PC initiates axil_read to the adapter and gets the data back.

bp_top/test/tb/bp_top_fpga/bp_stream_mmio.v
dpetrisko: Ah, IO cmd can now be up to 512 bits so that it’s possible to use the same module with mem and I/O. I think the els_p is a good idea

bp_top/test/tb/bp_top_fpga/system_wrapper.v
dpetrisko: I see, thanks!

dpetrisko:
My concern is that this setup is really coupled to our BSG FPGA dev environment, which is an uncommon setup.  I'm not sure it belongs it in the main repo.

My thought is to merge in cce_to_cache_dma, but move the testbench and other infrastructure to [Our ASIC/FPGA example organization](https://github.com/black-parrot-examples).  That way, you could run this just like ASIC testing.

Do you think that the \*\_stream\_\* modules are useful for other things than this testbench? If so, we could merge them into bp_top/ as well

bp_top/test/tb/bp_top_fpga/bp_stream_mmio.v
gaozihou: I see, thanks Dan!
Does black-parrot actually generate mmio_host request that is wider than 64-bits? Because this bp_stream_mmio module only handles mmio_host requests...

dpetrisko: It only does 64 bits now, but may do a full cache line in the future.  Farzam noticed that the nbf loader can take a long time in simulation and this may help (although not de-serializing may have a larger impact...)

gaozihou: Got it, thanks Dan!

gaozihou:
Yeah I agree that it is better to move the testbench and other infrastructure to ASIC/FPGA example organization.

dpetrisko:
Merging this: main development moved to black-parrot-examples

Issue #239, opened 2019-08-15
Separate out Proc params from Coh params from SoC params


dpetrisko:
All of these parameter sets are orthogonal.  It'll prevent factorial expansion of parameters where most of the parameters remain the same

PR #338, opened 2020-02-25
Adding Accelerator Complexes to BP SoC

zaazad:


bp_common/src/include/bp_common_aviary_pkg.vh
```diff
@@ -424,7 +432,424 @@ package bp_common_aviary_pkg;
...
+
+      ,async_io_clk         : 1
+      ,io_noc_did_width     : 3
+      ,io_noc_max_credits   : 16
+      ,io_noc_flit_width    : 64
+      ,io_noc_cid_width     : 1
+      ,io_noc_len_width     : 4
+      };
+
+  localparam bp_proc_param_s bp_accelerator_single_core_cfg_p =
```
dpetrisko: Have all these been tested? Since it's just a demo, I would suggest maybe only add 1 single core and 1 multicore config

bp_common/src/v/bp_pma.v
```diff
@@ -13,8 +13,8 @@ module bp_pma
    );
 
   wire is_local_addr = (ptag_i < (dram_base_addr_gp >> page_offset_width_p));
-  wire is_io_addr    = (ptag_i[ptag_width_p-1-:io_noc_did_width_p] != '0);
-
+  wire is_io_addr    = (/*ptag_i[ptag_width_p-1-:io_noc_did_width_p]*/ptag_i[27:24] != '0);
```
dpetrisko: This should be right, delete the test wire

bp_common/test/src/demos/src/accel_dp.c
```diff
@@ -0,0 +1,82 @@
+#include <stdint.h>
```
dpetrisko: Can we condense to 1 .c test by using #defines  ?

dpetrisko: 1 test per type that is.

bp_common/test/src/perch/bp_utils.h
```diff
@@ -18,4 +18,43 @@ void bp_finish(uint8_t code);
 
 #define BP_CFG_BASE_ADDR ((char *)(0x00200000))
```
dpetrisko: This might be nitty, but can we put the accelerator specific stuff in a vdp.h instead of the main perch?

bp_common/test/src/perch/bp_utils.h
```diff
@@ -18,4 +18,43 @@ void bp_finish(uint8_t code);
...
+
+struct VDP_CSR
+{ 
+  uint64_t *input_a_ptr;
+  uint64_t *input_b_ptr;
+  uint64_t input_length;
+  uint64_t *resp_ptr;
+};
+
+void bp_set_CSR(uint64_t *accel_base_address, uint8_t csr_idx, uint64_t csr_value);
```
dpetrisko: bp_set_mmio_csr/bp_get_mmio_csr

Or something similar.  If I were a user, I'd assume this were BP RISC-V CSRs

bp_common/test/src/perch/bp_utils.h
```diff
@@ -18,4 +18,43 @@ void bp_finish(uint8_t code);
...
+uint64_t bp_get_CSR(uint64_t *accel_base_address, uint8_t csr_idx); 
+void mem_cpy(uint64_t *src, uint64_t *dest, uint64_t length);
```
dpetrisko: Can't we use the normal memcpy from bsg_newlib? If that's too heavyweight, maybe rename to something closer to dma

bp_me/src/v/cache/bp_me_cache_dma_to_cce.v
```diff
@@ -320,10 +320,9 @@ module bp_me_cache_dma_to_cce
         if (resp_count_r == block_size_in_words_lp-1)
```
dpetrisko: Revert random formatting changes

bp_me/src/v/wormhole/bp_me_addr_to_cce_id.v
```diff
@@ -47,20 +48,22 @@ bsg_hash_bank
...
     // Split uncached I/O region by max 128 cores
     cce_id_o = local_addr_li.cce;
   else if (dram_addr_v_li)
     // Stripe by cache line
     cce_id_o[0+:lg_num_cce_lp] = cce_dst_id_lo;
   else
-    // TODO: Coprocessor address space, figure out.  Probably striped by CCEs within the AC
-    cce_id_o = '0;
+     cce_id_o = (num_sacc_p > 1)  
+                ? paddr_i[bp_sv39_vaddr_width_gp-io_noc_did_width_p-1: bp_sv39_vaddr_width_gp-io_noc_did_width_p-1-`BSG_SAFE_CLOG2(num_sacc_p)]
```
dpetrisko: I think this should slice down from paddr_width_p? You're missing the top bit

bp_me/src/v/wormhole/bp_me_addr_to_cce_id.v
```diff
@@ -47,20 +48,22 @@ bsg_hash_bank
 
 always_comb begin
   cce_id_o = '0;
-  if (external_io_v_li || (local_addr_v_li && (local_addr_li.dev == host_dev_gp)))
+  if (external_io_v_li || (local_addr_v_li && (local_addr_li.dev == host_dev_gp) && local_addr_li.cce < num_core_p))
```
dpetrisko: wire core_local_addr_li = local_addr_v_li & local_addr_li.cce < num_core_p

if (external_io_v_li || (core_local_addr_li && (local_addr_li.dev == host_dev_gp))

bp_top/test/common/bp_nonsynth_host.v
```diff
@@ -47,7 +47,7 @@ end
 // Host I/O mappings (arbitrarily decided for now)
 //   Overall host controls 32'h0300_0000-32'h03FF_FFFF
 
-localparam getchar_base_addr_gp = paddr_width_p'(64'h0010_0000);
+localparam getchar_base_addr_gp = paddr_width_p'(64'h0010_0???);
```
dpetrisko: Shouldn't be sending to anything other than _0000, I don't think?

bp_top/test/common/bp_nonsynth_host.v
```diff
@@ -126,6 +138,13 @@ always_ff @(negedge clk_i)
       pop();
     for (integer i = 0; i < num_core_p; i++)
       begin
+        if (putchar_data_cmd_v & io_cmd_yumi_o) begin
+          $write("%c", io_cmd_cast_i.data[0+:8]);
+          $fflush(32'h8000_0001);
+        end
+        if (getchar_w_v_li[i] & getchar_data_cmd_v & io_cmd_yumi_o)
+          //pop();
+          $display("[CORE%0x PRT] %x", i, io_cmd_cast_i.data[0+:8]);
```
dpetrisko: Use putchar to print, not getchar

bp_top/test/common/bp_nonsynth_if_verif.v
```diff
@@ -55,9 +55,9 @@ initial
     $fatal("Error: Must have exactly 1 row of I/O routers");
   if (mc_y_dim_p != 0)
     $fatal("Error: L2 expansion nodes not yet supported, MC must have 0 rows");
-  if (ac_x_dim_p != 0)
+/*  if (ac_x_dim_p != 0)
```
dpetrisko: Add check if cac_x_dim and sac_x_dim <= 1

dpetrisko:
For the sockets, I'm not sure the level of abstraction is right.  I had pictured (for example)

bp_tile_node (input [S:W] coh_link_i, output [S:W] coh_link_o
  - bp_tile (core+cce+adapters) (input coh_link_i, output coh_link_o ...)
  - bp_nd_socket #(coh_network) (input [S:P] coh_link_i, output [S:P] coh_link_o)
  - bp_nd_socket #(mem_network) ...
  - bp_nd_socket #(io_network) ...

Reason being that we're moving towards a model where the full messages in the network never get deserialized, so having full messages between the tile and socket will need to be rewritten.  Plus, some modules will serialize and some won't, so we would have to maintain both full messages and just flits.  If the connection between the socket and tiles is just flits, it's unlikely to change.

Now, to completely contradict myself.  What might be nice is having in the tile, a set of modules 'endpoints' These modules would have the interface:

```
bp_coherent_endpoint_master
 (input lce_req
   input lce_cmd
   output lce_cmd
   input lce_resp

   input coh_link_i
   output coh_link_o
...
)

bp_coherent_endpoint_client
 (output lce_req
   output lce_cmd
   input lce_cmd
   output lce_resp

   output coh_link_i
   input coh_link_o
...
)

bp_stream_endpoint_master
 (input lce_req
  output lce_cmd

   input coh_link_i
   output coh_link_o
...
  )

bp_stream_endpoint_client
 (output lce_req
  input lce_cmd

   output coh_link_i
   input coh_link_o
...
  )
```

etc.

Unfortunately, this is a ton of wrapper code.  I have a vague idea of how we could generate these with verilog macros or python, since they are very similar.  However, I'm not sure what the most maintainable solution is.  Python is already a dependency of the repo, so that's not a huge deal.  We could also use the basejump stl solution of checking in the generated files as well.

@taylor-bsg What are your thoughts here?

Basically, the problem is that there are too many potential attach combinations (coherent tile w/w/o memory w/w/o IO, streaming tile w/w/o memory w/w/o I/O, etc.) to do just an "endpoint standard" approach like bsg_manycore.  We would ideally like to have developers be able to write an accelerator tile without dealing with the boilerplate of adapters and concentrators, but we don't want to prevent people from using non-deserialization approaches or, conversely, to enforce them.

bp_common/src/include/bp_common_aviary_pkg.vh
dpetrisko: Another thing is that we probably want to move a model with separate core parameters and soc parameters (#239), to avoid this cross-product issue

zaazad: Not yet, I've just tested single, dual, and quad configs with accelerator. Agree, single and quad should be enough for demo purpose.

bp_common/test/src/demos/src/accel_dp.c
zaazad: Good idea! Will do that.

zaazad: Good idea! Will do that.

bp_common/test/src/perch/bp_utils.h
zaazad: Agree!

bp_me/src/v/wormhole/bp_me_addr_to_cce_id.v
zaazad: we want paddr_i[35-:log(num_sacc_p)], right?

zaazad: we want paddr_i[35-:log(num_sacc_p)], right?

bp_top/test/common/bp_nonsynth_host.v
zaazad: In bp_cprint function we have *(CPRINT_BASE_ADDR+core_id*8) = ch, but in the host we don't check the core_id. Don't we want to print core_id as well?

bp_top/test/common/bp_nonsynth_if_verif.v
zaazad: Good point, will do.

bp_common/src/include/bp_common_aviary_pkg.vh
black-parrot: Quad core works right? 

Yeah, delete the other configs for now and we'll revisit it during #239

bp_me/src/v/wormhole/bp_me_addr_to_cce_id.v
dpetrisko: bp_sv39_vaddr_width_gp should be bp_sv39_paddr_width_gp.  You're right that we want to slice after the domain but that's just paddr_width_p-io_noc_did_width_p-1

bp_top/test/common/bp_nonsynth_host.v
dpetrisko: Oh no, that changed to a more serial-like interface where everything just prints directly to screen.  We needed to change so that Linux printfs look reasonable.  So there's only the single getchar/putchar now

dpetrisko:
Okay, I would like to get this in, but we need more discussion on the socket abstraction.

Can we revert to dee2a53 (or thereabouts, right after n-d socket but before the specialized sockets), backport the PR review changes and merge that in?  Then we can iterate on the socket stuff on a new branch

bp_common/src/include/bp_common_aviary_pkg.vh
zaazad: yes, it works

PR #339, opened 2020-02-25
Added Dromajo section in Software Guide

farzamgl:
Added Checkpointing and Cosimulation instructions in the software development guide.

### DIFF
--- a/docs/software_guide.md
+++ b/docs/software_guide.md
@@ -19,6 +19,22 @@ BlackParrot test sources live in `bp_common/test/src` and are compiled with `mak
   * The file format is (size(in 2^N bytes)_address(40 bit)_data(size bits)
     * 03_0080000000_ffd1011b00090137 = 8 bytes, address 0x8000_0000, data ffd1011b00090137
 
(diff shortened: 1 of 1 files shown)


dpetrisko:
Thanks Farzam! Could you add a bit on how cosim works (which dpi calls happen when and where, interrupt synchronization, etc.).  And specify that it's done during runtime, not offline? 

Also, we do only support 1 checkpoint right now, right?

farzamgl:
> Thanks Farzam! Could you add a bit on how cosim works (which dpi calls happen when and where, interrupt synchronization, etc.). And specify that it's done during runtime, not offline?
> 
> Also, we do only support 1 checkpoint right now, right?

Yeah, only 1 checkpoint.
Should I list the DPI calls? There are 3(initialization, stepping, and trapping).

dpetrisko:
> Should I list the DPI calls? There are 3(initialization, stepping, and trapping). 

Yeah, list with a sentence is probably fine.  Trapping maybe a bit more because that's kind of a unique feature

PR #341, opened 2020-02-25
Updating submodules and adding timing loop check to DC

dpetrisko:
I updated spike to latest + our changes. Most everything works, except for the spike monitor causes the run to hang.  I disabled it since we're not planning on using it right now, but we should probably raise an issue about it.

The timing loop check I think will be very valuable.  This goes through FreePDK45 elaboration and mapping and checks for timing loops.  This should prevent us from inserting timing loops in RTL and then discovering them months later during tapeout

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -81,7 +81,6 @@ update-cache-auto:
         - external/*
 
 hotfix_job:
-  <<: *job_definition
   only:
     - /^.*hotfix.*$/
   stage: check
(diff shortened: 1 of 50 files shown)


farzamgl:
@dpetrisko can you rebase our old commits on top of the new Spike changes?
So we can keep track of them? Right now I can't figure out which BP changes are on that branch.

dpetrisko:
Unfortunately, no.  That will break the build for every branch's submodule.

What I do is:

```
git checkout blackparrot_mods
git diff origin/master
```

Works pretty well, since the diff is small

farzamgl:
> Unfortunately, no. That will break the build for every branch's submodule.
> 
> What I do is:
> 
> ```
> git checkout blackparrot_mods
> git diff origin/master
> ```
> 
> Works pretty well, since the diff is small

What's the last merge commit? Is that all the new changes on the origin repo since we last forked?

dpetrisko:
Yeah it’s just merge upstream/master

dpetrisko:
@stdavids Btw, take a look at this timing loop check.  It's pretty nifty, takes less than 5 minutes total because it's checking post-elab.  Uses FreePDK45 for the target library

dpetrisko:
Anything blocking this?

PR #347, opened 2020-02-26
UW ee477 pparrot wi20 regression list

stdavids:
Adding a regression list target as well as a baseline list file.


dpetrisko:
LGTM!

PR #349, opened 2020-02-27
update interface spec for new LCE-CCE interface

muwyse:
This updates the LCE-CCE IF portion of the interface spec to represent the LCE-CCE IF refactor that is on dev. The refactor has not been merged to uw_ee477 branch, though.


dpetrisko:
Thanks Mark!

PR #355, opened 2020-02-29
Uw ee477 pparrot wi20

tanglingshu:
wave.v: opens a VPD file with GTKWAVE
wave.sc: opens a VCD file with GTKWAVE

dve.v: opens a VPD file with DVE
dve.sc: opens a VCD file with DVE


taylor-bsg:
Nice!

PR #356, opened 2020-02-29
removed underscore in front of DVE

tanglingshu:
removing the underscore in front of DVE due to it may not be setting the environment variable $DVE correctly

### DIFF
--- a/Makefile.common
+++ b/Makefile.common
@@ -27,7 +27,7 @@ export URG       ?= urg
 export VERILATOR ?= verilator
 export DC_SHELL  ?= dc_shell
 # Some environments have issues with setting DVE as an environment variable
-export _DVE      ?= dve
+export DVE	 ?= dve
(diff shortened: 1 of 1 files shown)


dpetrisko:
No, this breaks DVE on our local machines

taylor-bsg:
Why is the underscore necessary?

dpetrisko:
Setting DVE causes dve to break in some environments (including ours) with a “environment variable DVE is set; please unset it” error

taylor-bsg:
Oh I see — yeah we usually use longer variable names to try to avoid
conflicts e.g. DVE-BIN, VCS-BIN

stdavids:
You can set the variable DVE in the makefile but cannot export it. Even still, I tend to still use $(VCS_BIN)/dve which works will with our cadenv or $(VCS_HOME)/bin/dve since VCS_HOME must be set to for anyone to use vcs.

PR #358, opened 2020-03-02
Uw ee477 pparrot wi20 new testing corners

stdavids:
Most of what I was trying to do here was add rtl-hard, post-synth, post-pnr-functional and post-pnr-timing simulations to the bp_top/syn testing infrastructure. Most of this should not be pushed upstream to mainline branches.


dpetrisko:
Looks fine to me in the context of the class.  I can merge this as is.

A couple thoughts for long term.  It would be nice to reduce the friction between ASIC and RTL builds, which may want to be done out of tree (as we do in our bsg_designs).  How about doing something like this:

```
# Override simv with post-synth/post-apr simv
SIMV ?= $(RESULTS_DIR)/vcs/simv

# Override OUT_DIR to testing/rtl/out
OUT_DIR ?= $(SYN_DIR)
REPORTS_DIR ?= $(OUT_DIR)/reports
...
```

Another pain point is having to duplicate filelists (including TCL flist) from our flist.vcs.  I'm not sure the best way to fix that one is.  Possibly having a Makefile list variable which generates the flists? Not sure how the TCL latches that though.

Last, does this extend well to simulating multiple corners?

dpetrisko:
The bind thing is unfortunate.  Could we `ifdef MONITOR the binds, so we can reuse the same testbench?

stdavids:
Pushed change to consolidate bp_softcore_no_bind and bp_softcore testbenches. I used `ifndef NO_MONITOR rather than `ifdef MONITOR. The double negative is a bit of a bummer but this means that monitoring happens by default.

dpetrisko:
Nice, works for me. We can discuss other stuff offline, not blocking this PR

stdavids:
Thanks Dan! Yeah, let's talk soon about the RTL/ASIC stuff, I have an idea that I'd like to try out once ee477 is out the door.

PR #361, opened 2020-03-04
Merging top dev

dpetrisko:
Main changes:
- Refactoring Cache Service Interface
- Separating out $ and LCE
- Adding UCE
- Adding single core testbench
- Updating documentation
- Minor backend fixes

Thanks Sripathi and Shashank!

@farzamgl Can you take a quick pass over this?  It's a huge changeset, but it's had several eyes on it so just flag anything that looks very wrong


farzamgl:
@dpetrisko Wow that's a lot of commits.
Can you point me to the BE fixes?

dpetrisko:
Biggest changes in the backend are in the dcache.  The interface change means that miss tracking is now in the cache instead of the LCE.  Additionally, the tag/data/stat_packets are ready-valid instead of valid-yumi. Last, fencei is now a cache instruction instead of a CSR instruction.

dpetrisko:
Oh, I mean backend like CAD fixes :), which were timing loop related

bp_be/src/v/bp_be_mem/bp_be_mem_top.v
```diff
@@ -67,24 +71,33 @@ module bp_be_mem_top
...
-   , output                                  credits_empty_o
+   // D$-LCE Interface
+   // signals to LCE
+   , output logic [cache_req_width_lp-1:0]          cache_req_o
+   , output logic                                   cache_req_v_o
+   , input                                          cache_req_ready_i
+   , output logic [cache_req_metadata_width_lp-1:0] cache_req_metadata_o
+   , output logic                                   cache_req_metadata_v_o
+
+   , input cache_req_complete_i
```
farzamgl: What's the interface on this signal?

bp_be/test/common/bp_be_nonsynth_dcache_tracer.v
```diff
@@ -34,12 +34,8 @@ module bp_be_nonsynth_dcache_tracer
 integer file;
 string file_name;
 
-logic freeze_r;
-always_ff @(posedge clk_i)
-  freeze_r <= freeze_i;
-
 always_ff @(negedge clk_i)
```
farzamgl: Maybe also `negedge (reset_i | freeze_i)` here?

dpetrisko: good catch!

bp_be/src/v/bp_be_mem/bp_be_mem_top.v
dpetrisko: described in the interface specification -- basically just goes high when a request has been serviced so the cache knows to wakeup

PR #363, opened 2020-03-05
[WIP] Adding floating point unit

dpetrisko:
This adds a single/double precision floating point unit to the backend.  Main changes
- The unit itself.  It's a sp/dp unit based on Berkeley Hardfloat.  It relies on retiming to pipeline.  It's possible we want this in basejump stl.
- Delay writeback a cycle to align fp and int writeback.
- Change bypass to do x0 zeroing instead of the regfile.  This gets rid of some unnecessary comparisons and muxes, since we need to compare in the bypass unit anyway.

Things to do:
- General code cleanup, extract the FPU-specific package variables to an fpu package.  This will make it easier if we want to include in basejump.
- Add an emulation routine for FDIV and FSQRT.  This should let us pass all of the riscv-tests.
- Add parameter to include/exclude fpu

Future directions:
- Store recoded format in register file, which shortens critical path in FPU.
- Analyse if an early out mechanism is worth it for any of the FP ops.

I've run this through the backend and it's able to meet 4 cycles.  We may want to expand to 5 cycles so that our iterative multiplier can be 4 cycles.  Not a huge change if we decide to go this route

------------------------------------------------------------
TODO:
- [ ] general code cleanup
- [ ] adjust forwarding paths to removing unnecessary bypass
- [ ] Add FDIV and FSQRT to long latency pipe (dependent on #398)
- [ ] Expand to 5 cycles to help timing


taylor-bsg:
Neat!

Some feedback:

“Unification” of FPU and integer pipe (e.g. combined registered file and/or
pipeline) sounds nice but generally is inferior to separating them.  It is
better to have two smaller register files than one big one for energy and
cycle time reasons; and integer does not need 3R1W. And of course we want
to enable recoding. Hazard detection is also simplified. Dual issue int/fp
is relatively easy to enable if you are separate as well. In Raw, we went
unified, but in manycore we are separate. I believe Krste did a similar
thing with his micro architectures as well.

I think Waterman said FDIV and FSQRT in hardfloat are tiny? In Raw when we
studied this, and Stuart Oberman’s PhD thesis said you definitely need
FDIV, and only FSQRT is optional, except there is a lot of overlap with
FDIV in terms of hardware.

Do we have a parameter to disable the FP?

taylor-bsg:
Similar thing: started out with unified, realized it was worse, and
switched to separate

dpetrisko:
>“Unification” of FPU and integer pipe (e.g. combined registered file and/or
pipeline)

Oh yes.  I meant purely from a code perspective i.e. having a parameter for number of read ports, not physically having a unified regfile.  It's a little difficult / possibly not worth it because we need to instantiate a 2r1w versus 3r1w, so the code savings are automatically limited.

dpetrisko:
> I think Waterman said FDIV and FSQRT in hardfloat are tiny? In Raw when we
> studied this, and Stuart Oberman’s PhD thesis said you definitely need
> FDIV, and only FSQRT is optional, except there is a lot of overlap with
> FDIV in terms of hardware.

It's definitely on the list (and there are hardfloat modules for them).  The difficulty is that they are long latency, non-pipelined ops and we haven't decided on how to implement those, so I'm just punting for now.

> Do we have a parameter to disable the FP?
No, but good idea.  Added to TODO list

dpetrisko:
superceded by upcoming PR

PR #364, opened 2020-03-05
Bump master to dev

dpetrisko:
Notably, this version has tons of extra documentation, Cache Service Interface, and the first version of a single-core-optimized version of blackparrot.

See #311 for more details on the changelog


dpetrisko:
Wait for #367

PR #370, opened 2020-03-08
Matching dram_clock_period_in_ps_p to the core clock

stdavids:


dpetrisko:
Can you add this to the bp_processor testbench, too?

dpetrisko:
Oh, also: BP sim clk needs to be set here in test_bp.cpp for Verilator.  test_bp.v is only used for VCS runs.

https://github.com/black-parrot/black-parrot/blob/uw_ee477_pparrot_wi20/bp_top/test/tb/bp_softcore/test_bp.cpp

dpetrisko:
Fixed up!

taylor-bsg:
Very nice!

PR #374, opened 2020-03-10
[WIP] Simplifying interrupt handling by passing through FE

dpetrisko:
Need to test with Linux since it's possible this breaks things.  However, I think it's much simpler overall.  There's a slight penalty in interrupt latency (~2 cycles), but there are power savings in the FE, so I'd call it a wash.

-----------------------------------
TODO:
- [ ] Rebase on most recent be_dev
- [ ] Test on Linux (dependent on #399)


dpetrisko:
Closing in favor of new interrupt scheme

PR #387, opened 2020-03-13
Nonsynth Host Enhancement

gaozihou:
**Not fully tested**

Originally tests PASS when LSB of returned finish packet is 1'b0, which may not be a good way of checking. In this PR, PASS means returned value in finish packet is 64'b0.


dpetrisko:
LGTM! I'll run in through regression

dpetrisko:
So these are actually byte registers in the host, so I changed it to be byte sized.  Otherwise, passed regression!

PR #389, opened 2020-03-13
Fixing accidental Verilator pragma

dpetrisko:


muwyse:
For background, it appears that Verilator transforms all comments, even those starting with "//", to "/* */" style comments including removing whitespace between the "//" and the first word of the comment. Further, a comment with form "/*verilator ... */" is a special tool specific command comment, similar to the //synopsys translate_on/off comments for VCS/Synopsys.

Basically, any comment in the code shouldn't start with the word "verilator"

dpetrisko:
I'll add this to the style guide

PR #390, opened 2020-03-15
[WIP] Experiment with removing issue stage

dpetrisko:
This branch is testing replacing the synchronous with an asynchronous regfile (either ff or latch based).  In advanced nodes, memories scale much more poorly than logic, so it's possible the area benefit of hardening for a relatively small (2k) RAM is not that much.  However, there may be a routing overhead.

The advantages of this are:
1) Reduced branch mispredict penalty (1 cycle).
2) O(200-500) flops by removing issue stage.
3) Reduced code complexity for compressed instructions (issue stage complicates this by being outside of the rolly fifo).

Disadvantages are:
1) Less common to have hardened asynchronous 1r1w memories
2) Possibly a timing impact (although, we send our regfile read at the beginning of the cycle, so there should be no timing difference between hardened / not hardened).
3) Timing path complexity if latch-based regfile.
4) Possible area impact

Need to push this through CAD flow to see area / timing impact

TODO:
- [ ] Lots of code cleanup
- [ ] Area and power analysis
- [ ] Adjust core profiler

### DIFF
--- a/bp_be/src/v/bp_be_calculator/bp_be_pipe_long.v
+++ b/bp_be/src/v/bp_be_calculator/bp_be_pipe_long.v
@@ -29,7 +29,7 @@ module bp_be_pipe_long
 
   `declare_bp_be_internal_if_structs(vaddr_width_p, paddr_width_p, asid_width_p, branch_metadata_fwd_width_p);
   bp_be_decode_s decode;
-  rv64_instr_s instr;
+  rv64_instr_rtype_s instr;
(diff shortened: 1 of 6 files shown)


dpetrisko:
Branch name is now wrong, this is an experiment with pre-decoding register addresses to support synchronous RAMs without an issue stage.  Same benefits, slight area / complexity overhead (but still less than with issue)

dpetrisko:
Superceded by local changes [coming soon]

PR #398, opened 2020-03-19
Adding M extension to BlackParrot

dpetrisko:
This PR adds the M extension to BlackParrot.  
- 4-cycle pipelined 64x64->64 multiplier.  This saves significant area over a 64x64->128 multiplier
- We trap and emulate MULH, MULHSU, MULHU (super infrequent instructions)
- iterative divider.  Haven't done the area analysis, but known to be very small
- Adjusted spike and dromajo to handle these cases, including emulation.

The iterative divide means that writeback happens significantly after commit.  @farzamgl this means that writeback is not a fixed latency after commit anymore.  Can you adjust the cosim to handle this?

TODO:
- [x] cosim support for idiv
- [x] Add to platform documentation for emulated instructions


taylor-bsg:
How is the iterative divide write back and WAW / RAW handled?

dpetrisko:
For both — stall until idiv is done. We discussed scoreboard for the future, but that requires more analysis.

taylor-bsg:
Good.

dpetrisko:
@farzamgl can you take a look? I think this is good to merge, since the cosim change we discussed works

bp_be/src/v/bp_be_calculator/bp_be_calculator_top.v
```diff
@@ -448,9 +452,11 @@ assign commit_pkt.pc         = calc_stage_r[2].pc;
 assign commit_pkt.npc        = calc_stage_r[1].pc;
 assign commit_pkt.instr      = calc_stage_r[2].instr;
 
-assign wb_pkt.rd_w_v  = calc_stage_r[3].irf_w_v & ~exc_stage_r[3].poison_v;
-assign wb_pkt.rd_addr = calc_stage_r[3].instr.fields.rtype.rd_addr;
-assign wb_pkt.rd_data = comp_stage_r[3];
+assign calc_wb_pkt.rd_w_v  = calc_stage_r[4].irf_w_v & ~exc_stage_r[4].poison_v;
```
farzamgl: Why is the default calc wb point delayed by 1 cycle? Because of the multiplier?
Shouldn't the commit_pkt stage also move in that case?

bp_be/src/v/bp_be_calculator/bp_be_regfile.v
```diff
@@ -117,9 +115,9 @@ always_comb
     rs2_reread_addr = rs2_r_v_i ? rs2_addr_i : rs2_addr_r;
 end
 
-// RISC-V defines x0 as 0. Else, forward if we read/wrote, else pass out the register data
-assign rs1_data_o = zero_rs1_r ? '0 : fwd_rs1_r ? rd_data_r : rs1_reg_data;
-assign rs2_data_o = zero_rs2_r ? '0 : fwd_rs2_r ? rd_data_r : rs2_reg_data;
+// Forward if we read/wrote, else pass out the register data
```
farzamgl: Why are we removing the x0 condition?

bp_be/src/v/bp_be_calculator/bp_be_pipe_mul.v
```diff
@@ -6,31 +6,17 @@
  * Description:
  *   Pipeline for RISC-V float instructions. Handles float and double computation.
```
farzamgl: Fix file description.

bp_common/test/Makefile
```diff
@@ -84,7 +85,7 @@ opensbi_elf:
 
 bbl_elf:
 	cd $(TEST_DIR)/src/riscv-pk/ && mkdir -p build && cd build \
-	&& ../configure --host=riscv64-unknown-elf --with-arch=rv64ia --with-abi=lp64 --includedir=$(TEST_DIR)/src/riscv-pk/machine/
+	&& ../configure --host=riscv64-unknown-elf --with-arch=rv64ima --with-abi=lp64 --includedir=$(TEST_DIR)/src/riscv-pk/machine/
```
farzamgl: Can we just remove the bbl target from this Makefile?
We generate it from the other repo.

bp_common/test/src/perch/emulation.c
```diff
@@ -83,8 +90,12 @@ void decode_illegal(uint64_t *regs, uint64_t mcause, uint64_t instr)
 
   if (opcode == RISCV_OPCODE_ATOMIC) {
     regs[rd_addr] = amo_jt[funct3][funct5](rs1_data, rs2_data);
+  } else if (opcode == RISCV_OPCODE_OP) {
+    // We just assume the funct7 wis correct
```
farzamgl: is :)

bp_top/test/common/bp_nonsynth_cosim.v
```diff
@@ -37,33 +44,47 @@ always_ff @(negedge reset_i)
...
-     );
-     
+  logic                     commit_rd_w_v_r;
   logic                     interrupt_v_r;
   logic [dword_width_p-1:0] cause_r;
-  bsg_dff
-   #(.width_p(1+dword_width_p))
-   trap__reg
+  logic commit_fifo_v_lo, commit_fifo_yumi_li;
+  wire commit_rd_w_v_li = decode_r.irf_w_v | decode_r.pipe_long_v;
```
farzamgl: Doesn't `decode_r.irf_w_v` also include the long pipe instructions?

bp_top/test/common/bp_nonsynth_commit_tracer.v
```diff
@@ -32,9 +32,9 @@ always_ff @(negedge delay_li)
     file      = $fopen(file_name, "w");
   end
```
farzamgl: I think the same changes from `bp_nonsynth_cosim` should be applied here.

bp_common/test/Makefile.frag
```diff
@@ -147,7 +147,20 @@ RV64_P_TESTS = \
...
+  rv64um-p-mulhu     \
+  rv64um-p-mulw      \
+  rv64um-p-div       \
+  rv64um-p-divu      \
+  rv64um-p-divuw     \
+  rv64um-p-divw      \
+  rv64um-p-rem       \
+  rv64um-p-remu      \
+  rv64um-p-remuw     \
+  rv64um-p-remw
```
farzamgl: These should also be added to the regression.

bp_be/src/v/bp_be_calculator/bp_be_calculator_top.v
dpetrisko: yes, multiply is 4 cycle latency.

No, commit point is still at mem3.  The commit point is simply when the instruction is non-speculative, which is why we take traps there.  Arithmetic instructions cannot raise traps so it's safe to writeback afterwards

bp_be/src/v/bp_be_calculator/bp_be_regfile.v
dpetrisko: Moved to the bypass unit to save a mux

bp_common/test/Makefile
dpetrisko: yep, also will remove opensbi which will be similar

bp_top/test/common/bp_nonsynth_cosim.v
dpetrisko: No, irf_w_v only refers to the writebacks from the normally pipe.  Otherwise, we would automatically writeback the incomplete result before the correct one in the case of a long latency op

bp_top/test/common/bp_nonsynth_commit_tracer.v
dpetrisko: Good catch

PR #403, opened 2020-03-22
Adding LFS to our program flow

farzamgl:
- Changed the riscv-newlib branch to `dramfs`
- Updated the toolchain installation code
- Added LFS functionality to libperch
- Updated the littlefs_test code
- Updated our progs to receive their CC target from the Makefile in `bp_common/test` and changed the target to `riscv64-unknown-elf-dramfs` (we no longer build `riscv64-unknown-elf`)


bp_common/test/src/perch/start.S
```diff
@@ -54,6 +54,10 @@ _start:
     addi x30, x0, 0
     addi x31, x0, 0
 
+.option push
```
dpetrisko: Could you explain / comment this?

bp_common/test/Makefile
```diff
@@ -45,24 +47,23 @@ demos: demos_mem demos_nbf demos_dump
...
 #opensbi: opensbi_mem opensbi_dump
-#bbl: bbl_mem bbl_dump
 # We don't run demos on spike because a lot of them don't terminate
 
 # Spec is separate because it is proprietary
 spec: spec_mem spec_dump spec_spike
 
-# mtest is separate because there's not a good default
-mtest: mtest_mem mtest_dump mtest_spike
+# mtest is outdated
```
dpetrisko: Let’s remove it from the Makefile until we fix it up

external/Makefile.tools
```diff
@@ -58,15 +58,15 @@ axe:
 	cp $(AXE_DIR)/axe $(BIN_DIR)
 
 gnu:
-	rm -rf $(GNU_DIR)/riscv-newlib
 	@cd $(TOP); git submodule update --init --recursive $(GNU_DIR)
 	cd $(GNU_DIR); \
-		rm -rf riscv-newlib; \
-		git clone -b bsg_manycore_newlib https://github.com/bespoke-silicon-group/bsg_newlib_dramfs.git riscv-newlib; \
```
dpetrisko: Briefly, can you explain how our Linux gnu toolchain gets built and used?

bp_common/test/src/perch/start.S
farzamgl: gp is used in the LFS test, so it should be initialized in the start code.

external/Makefile.tools
farzamgl: We clone the toolchain, apply the `newlib.patch` which changes the url and branch of `riscv-newlib` and applies a small Makefile change in the toolchain. Then we sync the submodule and build.

dpetrisko: Yeah, that’s the newlib chain. I’m asking about the Linux chain, that’s separate, no?

bp_common/test/src/perch/start.S
dpetrisko: I see. And the global pointers variable is a generic RISCV gcc concept?

external/Makefile.tools
farzamgl: I don't think we build that chain here.

bp_common/test/src/perch/start.S
farzamgl: Yeah, apparently it points to the symbols at the start of the .bss section.
https://www.sifive.com/blog/all-aboard-part-3-linker-relaxation-in-riscv-toolchain

dpetrisko: Interesting!

PR #407, opened 2020-03-24
prevent bp_mem_transducer from overwriting current request

muwyse:
In bp_mem, the transducer may accept a new request when it is already processing a request. This causes the initial request to be overwritten, and bp_mem to send back an incorrect response to the first request.

### DIFF
--- a/bp_me/test/common/bp_mem_transducer.v
+++ b/bp_me/test/common/bp_mem_transducer.v
@@ -75,7 +75,7 @@ module bp_mem_transducer
   wire [cce_block_width_p-1:0]    rd_bit_shift = rd_word_offset*dword_width_p; // We rely on receiver to adjust bits
   wire [cce_block_width_p-1:0]   rd_byte_shift = rd_word_offset*num_word_bytes_lp;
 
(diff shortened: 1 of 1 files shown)


muwyse:
I am still investigating this. It only manifested in the bp_cce testbench in the ME, I think because it was able to issue multiple memory requests fast enough that bp_mem was already occupied when the second one arrived.

Perhaps a better solution here is to insert a fifo in bp_mem for inbound memory commands, since bp_mem is a simulation component?

dpetrisko:
mem_cmd_ready_o depends on ready_i, so mem_cmd_v_i should not be high (r->v)

dpetrisko:
(As in whatever module is connecting has a handshake bug)

muwyse:
Hmm, okay. I missed that when I looked at the code lol. I'll add a comment to bp_mem (unless I missed that, too) and check my TB.

dpetrisko:
The memory is definitely not representative of actual dram because the latency model only processes 1 request at a time, as you noted. Really, there should be a fifo, an array of n latency models and a rr arbiter.

Issue #408, opened 2020-03-24
COSIM terminates too early


dpetrisko:
When running with COSIM_P=1 DUMP=1, cosim terminates the simulation from CPP. However, because $finish has not been called, the vcd has not necessarily been dumped. This ends up with the unfortunate side effect of losing the few cycles before the cosim mismatch. 

What we really want is the cosim mismatch + 1/2 cycles, so that the user can look at what's going on during the failure.

taylor-bsg:
Maybe there is a way for cosim to call through to the finish function first.

dpetrisko:
Yep, that's what Farzam and I discussed.  Adding another DPI call from dromajo->verilog to call finish

farzamgl:
Yeah. Actually doesn't even need a new DPI call. We can just add a return code to the existing `step` call and call $finish based on that.

kabylkas:
I just thought this might be useful. To see the activity post-mismatch, I did this dirty work around in the `dromajo_step()` DPI. This basically commits a couple of more instructions after the mismatch (depending on the value of the `counter`).
```
 54   exit_code = dromajo_cosim_step(dromajo_pointer, hart_id, pc, insn, wdata, 0, check);
...
 60   if (exit_code != 0) {
 61     kill_soon = true;
 62   }
 63
 64   if (kill_soon) {
 65     if (counter == 0) {
 66       std::cout << "oops!" << std::endl;
 67       abort();
 68     }
 69     counter--;
 70   }
```

dpetrisko:
Thanks @kabylkas ! Definitely a reasonable workaround.  It's better for us long term to call $finish because simulators e.g. VCS do some bookkeeping after finish such as flushing buffers, closing files, etc. It's possible that the counter value needs to be tool, system or even system load dependent

PR #409, opened 2020-03-25
Call the cosim $finish from Verilog

farzamgl:
Fixes the #408 Issue.

### DIFF
--- a/bp_top/test/common/bp_nonsynth_commit_tracer.v
+++ b/bp_top/test/common/bp_nonsynth_commit_tracer.v
@@ -34,7 +34,7 @@ always_ff @(negedge delay_li)
 
   logic [30:0] itag_cnt;
   bsg_counter_clear_up
-   #(.max_val_p(2**31-1)
+   #(.max_val_p(2**30)
      ,.init_val_p(0)
      )
    itag_reg
--- a/bp_top/test/common/bp_nonsynth_cosim.v
+++ b/bp_top/test/common/bp_nonsynth_cosim.v
@@ -21,6 +21,8 @@ module bp_nonsynth_cosim
     , input [dword_width_p-1:0]               rd_data_i
     , input                                   interrupt_v_i
     , input [dword_width_p-1:0]               cause_i
+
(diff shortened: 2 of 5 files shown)


dpetrisko:
Maybe print out COSIM_PASS / COSIM_FAIL?

farzamgl:
> Maybe print out COSIM_PASS / COSIM_FAIL?
Well, there is not really a cosim pass. If you get to the end of the simulation without any errors I guess it's a pass. Also on a mismatch, we already print the error msg in Dromajo. Don't know how much the extra `COSIM_FAIL` line will benefit us.

dpetrisko:
So we usually grep for PASS/FAIL.  For full tests, we add in this print by finishing the program.  For partial checkpoints, we want to know if the sample code PASSed or FAILed

farzamgl:
> So we usually grep for PASS/FAIL. For full tests, we add in this print by finishing the program. For partial checkpoints, we want to know if the sample code PASSed or FAILed

Okay, so I'll add the FAIL keyword to the Dromajo fail. Is that enough. Since we still grep for the host PASS if there is no cosim error.

dpetrisko:
There is no host pass for a partial program trace

farzamgl:
Well there is no cosim pass condition either. At least until we add the "Run N instructions from checkpoint feature"

dpetrisko:
Agreed, it's more useful when there's a partial program traces.  It would also be useful for running programs that we have not modified with host commands.  I'm not sure I understand the resistance here.  Are you saying that executing an entire program with cosim is not a passing condition?

farzamgl:
I just don't understand what's a cosim PASS condition if we are running the entire program. If there are no cosim errors we are gonna get the host pass/fail at the end. There is no merit in, for example, printing a cosim pass when the host calls $finish with a fail.

dpetrisko:
> There is no merit in, for example, printing a cosim pass when the host calls $finish with a fail.

That would tell us that there is a software bug, not a hardware bug.

Issue #293, opened 2020-01-06
Critical Path: ME decode + branch


dpetrisko:
The path is cce_Instr_ram->cce_instr_ram.  More specifically, cce_instr_ram_data->decode->(branch)->alu->(next pc)->cce_instr_ram_addr.

Possible solutions:
1) take another cycle for branches.  This complicates the control flow slightly.
2) Add a decode pipeline stage.  This would solve future critical paths for other instructions.  We could also consider making the CCE regfile synchronous
3) Minor help, make gprs < 64 bits.  Orthogonal to other solutions

taylor-bsg:
Also something we can apply the standard BSG front end tricks to.

M

muwyse:
This should be addressed by commit 3a0a0f8fbc5e9e3a5edd00bd3d020d17538bf3df to dev.

Issue #378, opened 2020-03-11
Investigate RV64E as BedRock base ISA


dpetrisko:
This would let us reuse the RISCV assembler (not compiler out-of-box, because we'd still only have 8 registers).  It would also reduce the barrier of entry for new BedRock programmers.  Since we're not using compressed instructions, a simple encoding for BedRock extensions would be:

[<31 bits> | 0] -> BedRock extension
[<31 bits> | 1]  -> RV64E

The CCE may also support only a subset of instructions, depending on PPA. (i.e. probably not supporting subword instructions, although I don't know what the current ISA looks like)

muwyse:
I will investigate. It certainly would be great to reuse the RISCV assembler and not require a custom one.

The microcode encoding is fairly tight for a couple of the instruction types (e.g., push/send message and branch conditional based on flag registers).

muwyse:
Current status: RV64E is not a defined ISA within RISC-V. I have heard some discussion within the community expressing desire for it to exist, but it does not seem to have a strong enough backing to actually make it happen. There may be a way to "force" the compiler to generate RV64I code that only uses the registers 0-15 with compiler flags, but I am not sure that those flags are guaranteed to avoid accessing registers 16-31. Existing compilers would also need to be coerced to not access memory when generating assembly from C code.

An alternative approach would be programming in RV64I assembly without using registers 16-31, adding custom instructions to the assembler for coherence operations (e.g., read directory, process directory), and then use the assembler to transform the RV64E-like assembly into machine code executable by the coherence engine.

dpetrisko:
Nice find!

PR #411, opened 2020-03-30
CCE v3 architecture update

muwyse:
This change updates the CCE architecture. The main updates are separating instruction fetch from decoded+execute to break a critical path (Issue #293 ) and re-designing the coherence directory to support different I$, D$, and accelerator $ organizations.

ME tests are cleaned up, working, and added back to regressions.


dpetrisko:
This is super beefy.  I'll look through for style things, but I think I'll have to trust you for architecture changes.

Few high level questions:

>The main updates are separating instruction fetch from decoded+execute to break a critical path (Issue #293 ) 

Nice! Will run this through the backend.  Are there performance (IPC) implications?

>re-designing the coherence directory to support different I$, D$, and accelerator $ organizations.

We should probably have a test configuration for hetero cache-configs.  (Although you might need @Shashank-Vijay @sripathi-muralitharan changes to have the caches actually change size correctly)

34 bit instructions are good from an area perspective.  Could you post a list of the most crowded instruction encodings? Of course, the dream is #378, but that may not be realistic without 48 bit extensions for bedrock instructions.

muwyse:
> Are there performance (IPC) implications?
Likely, yes. There is a static branch prediction bit (that can be specified as predict taken / not taken by the ucode programmer). If the prediction is correct there is no penalty. Mispredict is a 1 cycle penalty.

> We should probably have a test configuration for hetero cache-configs
Yes, this will definitely require their changes. I only added the CCE support.

> 34 bit instructions are good from an area perspective. Could you post a list of the most crowded instruction encodings? Of course, the dream is #378, but that may not be realistic without 48 bit extensions for bedrock instructions.
I will cleanup the docs I have and upload. I think the tightest encodings are the special branch flag ops, push queue ops, and directory read/write ops. I still need to investigate RV64E to see what could be done with it.

dpetrisko:
>I will cleanup the docs I have and upload. I think the tightest encodings are the special branch flag ops, push queue ops, and directory read/write ops. I still need to investigate RV64E to see what could be done with it.

I'm definitely not going to hold this PR up for it, but 32 bits would be nice for 
- eventual RV64E
- Byte-sized RAMs are often easier/more efficient to generate from a CAD tool perspective

We can discuss once the docs are up, but some ideas to loosen up bits would be:
1) fewer GPRs (not sure how many you use now)
2) incomplete branch target space (say, +-32 instead of full range, fairly common for embedded code)
3) Move queue selection partially into opcode space.

Another way to handle it is to have longer instructions for CCE-special extensions.  i.e. 48 bit instructions, which are supported by the RV64E isa.  The downsides to this are the wider RAM and slightly more complicated fetch logic.

muwyse:
Regressions passed: https://gitlab.com/black-parrot/black-parrot/pipelines/131515979

I also verified the FSM CCE is up to date and working so @sripathi-muralitharan and @Shashank-Vijay can use it for their work.

dpetrisko:
LGTM!

PR #416, opened 2020-04-02
Cache Parameterization and Cache Testbench

sripathi-muralitharan:
- Added support for 2-way, 4-way and 8-way I$ and D$

- Added a testbench for D$ and I$ with some preliminary tests


bp_be/src/v/bp_be_mem/bp_be_dcache/bp_be_dcache.v
```diff
@@ -80,27 +80,30 @@ module bp_be_dcache
   import bp_common_aviary_pkg::*;
  #(parameter bp_params_e bp_params_p = e_bp_inv_cfg
    `declare_bp_proc_params(bp_params_p)
-   `declare_bp_cache_service_if_widths(paddr_width_p, ptag_width_p, lce_sets_p, lce_assoc_p, dword_width_p, cce_block_width_p)
```
dpetrisko, line 83: Have dcache_sets_p, dcache_block_width_p as well.  If the configuration is not supported, add a check in bp_nonsynth_if_verif (like dcache_block_width_p == cce_block_width_p, for example)

bp_be/src/v/bp_be_mem/bp_be_dcache/bp_be_dcache.v
```diff
@@ -80,27 +80,30 @@ module bp_be_dcache
...
     , parameter lock_max_limit_p=8
 
     , localparam cfg_bus_width_lp= `bp_cfg_bus_width(vaddr_width_p, core_id_width_p, cce_id_width_p, lce_id_width_p, cce_pc_width_p, cce_instr_width_p)
-    , localparam block_size_in_words_lp=lce_assoc_p
-    , localparam data_mask_width_lp=(dword_width_p>>3)
-    , localparam byte_offset_width_lp=`BSG_SAFE_CLOG2(dword_width_p>>3)
-    , localparam word_offset_width_lp=`BSG_SAFE_CLOG2(block_size_in_words_lp)
+    , localparam block_size_in_words_lp = dcache_assoc_p
+    , localparam cache_block_width_multiplier_lp = 2**(3 - `BSG_SAFE_CLOG2(dcache_assoc_p)) // Need to change '3' if assoc_p can be > 8
+    , localparam cache_block_width_lp = dword_width_p * cache_block_width_multiplier_lp
```
dpetrisko, line 92: Should start with block size derive bank size

bp_be/src/v/bp_be_mem/bp_be_dcache/bp_be_dcache.v
```diff
@@ -80,27 +80,30 @@ module bp_be_dcache
...
 
     , localparam cfg_bus_width_lp= `bp_cfg_bus_width(vaddr_width_p, core_id_width_p, cce_id_width_p, lce_id_width_p, cce_pc_width_p, cce_instr_width_p)
-    , localparam block_size_in_words_lp=lce_assoc_p
-    , localparam data_mask_width_lp=(dword_width_p>>3)
-    , localparam byte_offset_width_lp=`BSG_SAFE_CLOG2(dword_width_p>>3)
-    , localparam word_offset_width_lp=`BSG_SAFE_CLOG2(block_size_in_words_lp)
+    , localparam block_size_in_words_lp = dcache_assoc_p
+    , localparam cache_block_width_multiplier_lp = 2**(3 - `BSG_SAFE_CLOG2(dcache_assoc_p)) // Need to change '3' if assoc_p can be > 8
+    , localparam cache_block_width_lp = dword_width_p * cache_block_width_multiplier_lp 
+    , localparam bypass_data_mask_width_lp = (dword_width_p >> 3) // Need to change '3' if dword_width_p changes
```
dpetrisko, line 93: dword won't change in the dcache. The I$ fetches 32 bits, but there's no mask

bp_be/src/v/bp_be_mem/bp_be_dcache/bp_be_dcache.v
```diff
@@ -801,24 +804,40 @@ module bp_be_dcache
...
-    ,.els_p(lce_assoc_p)
+    .width_p(cache_block_width_lp)
+    ,.els_p(dcache_assoc_p)
   ) ld_data_set_select_mux (
     .data_i(ld_data_tv_r)
     ,.sel_i(load_hit_way ^ addr_word_offset_tv)
     ,.data_o(ld_data_way_picked)
   );
 
+  if(dcache_assoc_p == 8) begin
```
dpetrisko, line 821: The assertion here is really that bank_width_p == dword_width_p

bp_be/src/v/bp_be_mem/bp_be_dcache/bp_be_dcache.v
```diff
@@ -801,24 +804,40 @@ module bp_be_dcache
...
+    assign ld_data_dword_picked = ld_data_way_picked;
+  end
+  else begin  
+    bsg_mux
+      #(.width_p(dword_width_p)
+      ,.els_p(cache_block_width_multiplier_lp)
```
dpetrisko, line 827: Actually, the mux should work with 1 element, so I think you don't need this if-statement

bp_be/src/v/bp_be_mem/bp_be_dcache/bp_be_dcache.v
```diff
@@ -905,34 +924,54 @@ module bp_be_dcache
...
-      : {lce_assoc_p{lce_data_mem_v}});
+      : {dcache_assoc_p{lce_data_mem_v}});
 
   assign data_mem_w_li = wbuf_yumi_li
     | (data_mem_pkt_v_i & data_mem_pkt.opcode == e_cache_data_mem_write);
 
-  logic [lce_assoc_p-1:0][dword_width_p-1:0] lce_data_mem_write_data;
+  logic [dcache_assoc_p-1:0][cache_block_width_lp-1:0] lce_data_mem_write_data;
 
-  for (genvar i = 0; i < lce_assoc_p; i++) begin
```
dpetrisko, line 918: use bsg_expand_mask

bp_be/test/common/bp_be_nonsynth_dcache_tracer.v
```diff
@@ -6,65 +6,156 @@ module bp_be_nonsynth_dcache_tracer
...
+  always_ff @(posedge clk_i) begin
+ 
+      if(v_tl_r) 
+        $fwrite(file, "[%t] tag_lookup: %x \n", $time, v_tl_r);
+      
+      if(v_tv_r)
+        $fwrite(file, "[%t] tag_verify: %x \n", $time, v_tv_r);
+      
+      if(sc_success)
+        $fwrite(file, "SC SUCEESS! \n");
```
dpetrisko, line 131: spelling :)

bp_be/test/common/bp_be_nonsynth_dcache_tracer.v
```diff
@@ -6,65 +6,156 @@ module bp_be_nonsynth_dcache_tracer
  import bp_common_rv64_pkg::*;
```
dpetrisko, line 6: Nice!

dpetrisko: What would be needed to reuse this for I$?

bp_be/test/tb/bp_be_dcache/py/bsg_ascii_to_rom.py
```diff
@@ -0,0 +1,64 @@
+#!/usr/bin/python
```
dpetrisko, line 1: Use the one in basejump_stl

bp_be/test/tb/bp_be_dcache/py/trace_script.py
```diff
@@ -0,0 +1,134 @@
+#!/bin/usr/python
```
dpetrisko, line 1: Like the bsg_cache testbench, separate out each test into a script to generate a rom (or have 1 script that generates multiple roms, your choice)

bp_be/test/tb/bp_be_dcache/testbench.v
```diff
@@ -0,0 +1,340 @@
...
+
+  // Setting up the config bus
+  logic switch_cce_mode;
+  always_comb begin
+    cfg_bus_cast_li = '0;
+    cfg_bus_cast_li.freeze = '0;
+    cfg_bus_cast_li.core_id = '0;
+    cfg_bus_cast_li.dcache_id = '0;
+    cfg_bus_cast_li.dcache_mode = e_lce_mode_normal;
+    cfg_bus_cast_li.cce_mode = switch_cce_mode ? e_cce_mode_normal : e_cce_mode_uncached;
```
dpetrisko, line 90: I think you should be able to just start in cce normal mode.  Am I wrong?

bp_be/test/tb/bp_be_dcache/testbench.v
```diff
@@ -0,0 +1,340 @@
...
+  end
+ 
+  logic [6:0] count_lo;
+  localparam counter_max_val_lp = lce_sets_p + 1;
+
+  bsg_counter_clear_up
```
dpetrisko, line 96: Instead of this, you can add wait cycle command to the trace_rom

bp_be/test/tb/bp_be_dcache/testbench.v
```diff
@@ -0,0 +1,340 @@
...
+  assign ptag_li = trace_data_lo[dcache_pkt_width_lp+:ptag_width_lp];
+  assign uncached_li = trace_data_lo[dcache_pkt_width_lp+ptag_width_lp+:1];
+
+  // Output FIFO
+  logic fifo_yumi_li;
+  logic [dword_width_p-1:0] fifo_data_lo;
+  assign fifo_yumi_li = trace_v_li & trace_ready_lo;
+  assign trace_data_li = {'0, fifo_data_lo};
+
+  bsg_one_fifo
```
dpetrisko, line 169: Should be 2-fifo.  Otherwise it can't enqueue and dequeue in same cycle.  Are your tests running back2back ops?

bp_be/test/tb/bp_be_dcache/wrapper.v
```diff
@@ -0,0 +1,292 @@
...
+
+   ,.data_o(data_o)
+   ,.v_o(v_o)
+   ,.fencei_v_o()
+
+   ,.tlb_miss_i(1'b0)
+   ,.ptag_i(rolly_ptag_r)
+   ,.uncached_i(uncached_i)
+
+   ,.poison_i(1'b0)
```
dpetrisko, line 163: Should we be poisoning on cache misses or does the dcache do this?

bp_be/test/tb/bp_be_dcache/wrapper.v
```diff
@@ -0,0 +1,292 @@
...
+
+   ,.lce_cmd_o()
+   ,.lce_cmd_v_o()
+   ,.lce_cmd_ready_i()
+
+   ,.credits_full_o(credits_full_lo)
+   ,.credits_empty_o(credits_empty_lo)
+   );
+
+   // We need this to break the deadlock for the lce_cmd
```
dpetrisko, line 244: Can you describe this?

bp_be/test/tb/bp_be_dcache_broken/py/trace_gen.py
```diff
@@ -0,0 +1,98 @@
+#
```
dpetrisko, line 1: Make sure to delete all these deprecated files before merging

bp_common/src/include/bp_common_aviary_pkg.vh
```diff
@@ -91,9 +90,9 @@ package bp_common_aviary_pkg;
       ,dtlb_els             : 8
 
       ,dcache_sets          : 64
-      ,dcache_assoc         : 8
+      ,dcache_assoc         : 2
```
dpetrisko, line 93: Add new single core configs for the different associativities

bp_fe/src/include/bp_fe_icache_stat_info.vh
```diff
@@ -0,0 +1,21 @@
...
+ *    bp_fe_icache_stat_info.vh
+ *
+ *  Description:
+ *    stat_mem entry format.
+ */
+
+`ifndef BP_FE_ICACHE_STAT_INFO_VH
+`define BP_FE_ICACHE_STAT_INFO_VH
+
+`define declare_bp_fe_icache_stat_info_s(ways_mp)  \
```
dpetrisko, line 12: Isn't this part of the cache_service_if?

bp_fe/src/v/bp_fe_icache.v
```diff
@@ -22,20 +22,22 @@ module bp_fe_icache
...
     , localparam index_width_lp=`BSG_SAFE_CLOG2(lce_sets_p)                        
     , localparam block_offset_width_lp=(word_offset_width_lp+byte_offset_width_lp) 
-    , localparam tag_width_lp=(paddr_width_p-block_offset_width_lp-index_width_lp) 
+    , localparam ptag_width_lp=(paddr_width_p-bp_page_offset_width_gp) 
     
-    `declare_bp_icache_widths(vaddr_width_p, tag_width_lp, lce_assoc_p) 
+    `declare_bp_icache_widths(vaddr_width_p, ptag_width_lp, icache_assoc_p) 
 
-    , localparam bp_be_dcache_stat_width_lp = `bp_be_dcache_stat_info_width(lce_assoc_p)
+    , localparam stat_width_lp = `bp_be_dcache_stat_info_width(icache_assoc_p)
```
dpetrisko, line 40: Should be a common stat info?

bp_fe/src/v/bp_fe_icache.v
```diff
@@ -368,30 +370,46 @@ module bp_fe_icache
...
-    ,.els_p(lce_assoc_p)
+    .width_p(cache_block_width_lp)
+    ,.els_p(icache_assoc_p)
   ) data_set_select_mux (
     .data_i(ld_data_tv_r)
     ,.sel_i(hit_index ^ addr_word_offset_tv)
     ,.data_o(ld_data_way_picked)
   );
 
+  logic [dword_width_p-1:0] ld_data_dword_picked;
```
dpetrisko, line 384: same comments as dcache

bp_fe/src/v/bp_fe_icache.v
```diff
@@ -368,30 +370,46 @@ module bp_fe_icache
...
-    .data_i({uncached_load_data_r, ld_data_way_picked})
+    .data_i({uncached_load_data_r, ld_data_dword_picked})
     ,.sel_i(uncached_tv_r)
     ,.data_o(final_data)
   );
 
   logic lower_upper_sel;
 
-  assign lower_upper_sel             = addr_tv_r[byte_offset_width_lp-1];
+  assign lower_upper_sel             = addr_tv_r[2]; // We always need to pick the third bit, so it has been hardcoded here.
```
dpetrisko, line 412: How about 
// Select upper/lower 32 bits?

bp_fe/test/common/include/bp_fe_nonsynth_icache_tracer.v
```diff
@@ -0,0 +1,153 @@
...
+
+		if(v_tv_r)
+			begin
+				$fwrite(file, "Tag Verify stage activated at time [%t]\n", $time);
+			end
+
+
+		if(cache_req_v_o)
+			begin
+				$fwrite(file, "Cache request sent at time [%t]\n", $time);
```
dpetrisko, line 105: Indentation is funky here

bp_fe/test/tb/bp_fe_icache/py/bsg_ascii_to_rom.py
```diff
@@ -0,0 +1,64 @@
+#!/usr/bin/python
```
dpetrisko, line 1: Same comment, use basejump_stl

bp_me/src/v/cce/bp_uce.v
```diff
@@ -378,7 +386,7 @@ module bp_uce
               mem_cmd_cast_o.header.msg_type       = miss_load_v_li ? e_cce_mem_rd : e_cce_mem_wr;
               mem_cmd_cast_o.header.addr           = cache_req_r.addr;
               mem_cmd_cast_o.header.size           = e_mem_size_64;
-              mem_cmd_cast_o.header.payload.way_id = cache_req_metadata_r.repl_way;
+              mem_cmd_cast_o.header.payload.way_id = {'0, cache_req_metadata_r.repl_way};
```
dpetrisko, line 389: I think the Verilog semantics for concatenation larger than the lvalue are ill-defined here.  Use lce_assoc_p'(cache_req_metadata_r.repl_way), which will also zero extend.

dpetrisko:
^ Also, rebase on dev again.

And as a small note, we probably would prefer doing the D$ and I$ in separate PRs to fe_dev and be_dev in the future, to make it more manageable and also run through CI on the _dev branches

bp_common/src/include/bp_common_aviary_defines.vh
```diff
@@ -207,18 +208,23 @@ typedef struct packed
...
+  , localparam acache_block_width_p       = proc_param_lp.acache_block_width                       \
   , localparam lce_assoc_p                = `BSG_MAX(dcache_assoc_p,                               \
                                                      `BSG_MAX(icache_assoc_p, acache_assoc_p))     \
   , localparam lce_assoc_width_p          = `BSG_SAFE_CLOG2(lce_assoc_p)                           \
   , localparam lce_sets_p                 = `BSG_MAX(dcache_sets_p,                                \
                                                      `BSG_MAX(icache_sets_p, acache_sets_p))       \
   , localparam lce_sets_width_p           = `BSG_SAFE_CLOG2(lce_sets_p)                            \
                                                                                                    \
-  , localparam cce_block_width_p          = proc_param_lp.cce_block_width                          \
+  , localparam cce_block_width_p          = `BSG_MAX(dcache_block_width_p,                         \
```
muwyse, line 225: For now, there should be an assertion somewhere (nonsynth_if_verif?) that this is 512-bits (64-bytes), and that icache, dcache, and acache all have the same cache block width. I think allowing all the caches to use a different block size, such as 128 or 256-bits, should work (as long as they all have the same block size). Allowing different sizes per cache likely will not work without additional changes to the LCE/CCE.

bp_fe/test/common/include/bp_fe_nonsynth_icache_tracer.v
muwyse: looks like tabs? make sure indents are spaces.

bp_be/test/tb/bp_be_dcache/testbench.v
sripathi-muralitharan: Well, the CCE has to go to a state in the FSM called uncached_only to send the sync. Otherwise it jumps into the ready state where the sync is not performed. Without sync, the LCE doesn't get out of the uncached_only state. That's why we had this work around.

dpetrisko: @muwyse Seems like weird behavior.  Can we make it so CCE comes out of reset and immediately goes into cached mode?

bp_be/test/tb/bp_be_dcache/wrapper.v
sripathi-muralitharan: The poison input here is to signify load_access/store_access faults or load_page/store_page faults. If there is a poison, the cache doesn't send out a miss at all.

dpetrisko: Sorry, I'm asking if a miss in TV will automatically poison TL.  If not, a second hit/miss in TL may go out speculatively

Shashank-Vijay: We thought of two options for this - 

- poison_i is connected to (cache_req_v_o & cache_req_ready_i)

- poison_i is connected to the cache miss signal coming from the cache miss tracker

We think the second option might be safer since the poison_i will be high until the cache_req_complete_i signal goes high. Which would you suggest?

On a related note, does the PTW control this behavior of poison_i in the processor?

bp_be/src/v/bp_be_mem/bp_be_dcache/bp_be_dcache.v
```diff
@@ -80,27 +80,30 @@ module bp_be_dcache
...
+    , localparam byte_offset_width_lp = `BSG_SAFE_CLOG2(bank_width_lp>>3) 
+    , localparam word_offset_width_lp = `BSG_SAFE_CLOG2(block_size_in_words_lp)
     , localparam block_offset_width_lp=(word_offset_width_lp+byte_offset_width_lp)
-    , localparam index_width_lp=`BSG_SAFE_CLOG2(lce_sets_p)
+    , localparam index_width_lp=`BSG_SAFE_CLOG2(dcache_sets_p)
     , localparam ptag_width_lp=(paddr_width_p-bp_page_offset_width_gp)
-    , localparam way_id_width_lp=`BSG_SAFE_CLOG2(lce_assoc_p)
+    , localparam way_id_width_lp=`BSG_SAFE_CLOG2(dcache_assoc_p)
   
     , localparam lce_data_width_lp=(lce_assoc_p*dword_width_p)
```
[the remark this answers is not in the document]
Shashank-Vijay: If you're referring to lce_data_width_lp, it's not used anywhere in the file. So we're removing it in the next commit.

bp_fe/test/common/include/bp_fe_nonsynth_icache_tracer.v
Shashank-Vijay: Yeah they're tabs. Will make it spaces

bp_be/test/tb/bp_be_dcache/wrapper.v
dpetrisko: Option 1 doesn't work because of uncached stores.
I think option 2 probably works, since you won't be sending a new request until the miss is resolved.  The other option is to emulate pipeline behavior and just poison if ```dcache_v_rr & ~v_o```, where dcache_v_rr is just the input signal delayed by a bsg_dff_chain

> On a related note, does the PTW control this behavior of poison_i in the processor?

Not directly.  Poison_i is raised by the checker iff there's a cache miss or exception.  Exceptions are (sometimes) caused by PTW.

bp_be/test/tb/bp_be_dcache/wrapper.v
Shashank-Vijay: The lce_cmd interface from the CCE is a ready-valid interface (demanding producer) whereas the lce_cmd interface into the LCE is a valid-yumi interface (demanding consumer). That's why we added the FIFO. We actually didn't have this at the start but then noticed that both the FSMs were stuck in the DVE. That's when we realized this and added it.

bp_be/test/tb/bp_be_dcache/testbench.v
Shashank-Vijay: This is right. Right now our tests have a nop in between each test, so I guess we aren't stretching it. We'll make it two_fifo.

bp_be/test/tb/bp_be_dcache/wrapper.v
dpetrisko: got it, yeah this is common.  In general, the red flag is if a yumi port has a ready signal connected.

// lce_cmd demanding->demanding handshake conversion

Also, should be a two fifo

bp_be/test/tb/bp_be_dcache/testbench.v
Shashank-Vijay: Yeah, that's right. We'll change this.

bp_be/test/tb/bp_be_dcache/py/trace_script.py
Shashank-Vijay: Yeah, we're planning to have the same script but generate different traces for each block of tests.

bp_be/test/common/bp_be_nonsynth_dcache_tracer.v
Shashank-Vijay: We're going to add all the signals that are relevant to both caches in the I/Os, connect only the ones that are required (and connect the others such that they don't write to the file ever) and make the name cache_tracer. We can instantiate one each for the two caches and give the corresponding associativity and block width. Is this fine?

bp_be/test/tb/bp_be_dcache/py/bsg_ascii_to_rom.py
Shashank-Vijay: We copied the file from basejump_stl, changed it so that it writes to a file (it was printing to the screen).

dpetrisko: In the Makefile use > .txt

bp_be/test/common/bp_be_nonsynth_dcache_tracer.v
dpetrisko: Sounds good! How are you going to handle the filename difference?

bp_be/test/tb/bp_be_dcache/testbench.v
Shashank-Vijay: Actually, this counter's output is used to switch the cce_mode to normal. So I guess we would still need it?

bp_be/test/common/bp_be_nonsynth_dcache_tracer.v
Shashank-Vijay: We've made the filename a parameter and are placing this in bp_top/test/common (making the appropriate file path changes).

bp_be/test/tb/bp_be_dcache/wrapper.v
Shashank-Vijay: Done

bp_fe/test/common/include/bp_fe_nonsynth_icache_tracer.v
Shashank-Vijay: We've changed the nature of the cache tracer such that it will be used for both caches. So we're deleting this file.

bp_be/test/common/bp_be_nonsynth_dcache_tracer.v
dpetrisko: Have you tried compiling in verilator? There’s an issue with binding parameters which are defined at the toplevel

sripathi-muralitharan: Okay. Will try it out and let you know.

sripathi-muralitharan: Tried it out on verilator with the testbench we created for the $ (coz verilator was throwing some width parameter missing for bp_be_pipe_long in the processor testbench) and it seems to be working fine. I can see the traces being printed out correctly.

Shashank-Vijay:
We now have a single cache tracer that is parameterized by the file name. We have created Makefiles in each $ testbench to generate different ROMs for different tests. Some other comments have also been resolved.

The things remaining are getting the processor testbench to work with both cache block widths as 256 and 128, and adding configs in the aviary pkg. We think we can get this done by tomorrow, and also run the full regressions to confirm everything's working.

bp_be/src/v/bp_be_mem/bp_be_dcache/bp_be_dcache.v
sripathi-muralitharan: Right now, basejump_stl supports bsg_expand_bitmask, which takes in a number and expands each bit of the number expand_p number of times. The functionality that we are looking for is to place the mask at different locations rather than expand it. I don't think this behaviour is possible with the expand_bitmask or rather, I am unsure how it can be applied here.

dpetrisko: Oh, I see, you're right.

How about:
```
if (num_dwords_per_bank_lp == 1) begin : passthrough_wbuf_mask
  assign wbuf_mask = wbuf_entry_out.mask;
else begin : shift_wbuf_mask
  wbuf_mask = wbuf_entry_out.mask >> wbuf_entry_out.paddr[byte_offset_width_lp+:`BSG_SAFE_CLOG2(num_dwords_per_bank_lp)];
```

The else clause here is really an error condition, right?  Remove it and add an assertion unless it's covered somewhere else.

bp_fe/src/include/bp_fe_icache_stat_info.vh
Shashank-Vijay: It wasn't, but now I have added it.

dpetrisko: Nice!

bp_be/src/v/bp_be_mem/bp_be_dcache/bp_be_dcache.v
sripathi-muralitharan: Made a slight correction to your solution and updated the file. 
```
if (num_dwords_per_bank_lp == 1) begin : passthrough_wbuf_mask
  assign wbuf_mask = wbuf_entry_out.mask;
end else begin : shift_wbuf_mask
  wbuf_mask = wbuf_entry_out.mask << (wbuf_entry_out.paddr[3+:`BSG_SAFE_CLOG2(num_dwords_per_bank_lp)] << 3);
end
```

dpetrisko: Nice!

bp_be/test/common/bp_be_nonsynth_dcache_tracer.v
dpetrisko: Cool, long as it works!  It would be a compile error if you were running into the issue I was thinking of

bp_be/test/tb/bp_be_dcache/testbench.v
dpetrisko: Yeah, I guess keep the counter and wait for Mark's response to immediately switching to cached mode.

bp_common/src/include/bp_common_aviary_pkg.vh
sripathi-muralitharan: Followed your suggestion for now and added softcore_l1_small, softcore_l1_medium, single_core_l1_small, single_core_l1_medium. But we do support caches with different associativities. So we definitely need to come up with more scalable names for the different configurations.

dpetrisko: Good point, can add softcore_l1_hetero : I$ 4-way, D$ 8-way

dpetrisko: This is a good usecase for https://github.com/black-parrot/black-parrot/issues/422

bp_be/src/v/bp_be_mem/bp_be_dcache/bp_be_dcache.v
Shashank-Vijay: We tried to combine the two left shifts into a single statement. Turns out that doesn't give the expected result. We've now split it into two statements with one left shift each.

bp_common/src/include/bp_common_aviary_pkg.vh
```diff
@@ -320,12 +334,15 @@ package bp_common_aviary_pkg;
       ,dtlb_els             : 8
 
       ,dcache_sets          : 64
-      ,dcache_assoc         : 8
+      ,dcache_assoc         : 4
+      ,dcache_block_width   : 512
       ,icache_sets          : 64
-      ,icache_assoc         : 8
+      ,icache_assoc         : 2
```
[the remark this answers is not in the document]
Shashank-Vijay: We've added different configs and have the default values for the default config now.

bp_be/src/v/bp_be_mem/bp_be_dcache/bp_be_dcache.v
dpetrisko: Oh yeah, I think you probably needed parentheses.  Two statements is more clear though

bp_be/test/tb/bp_be_dcache/testbench.v
dpetrisko: Made an issue for this, not blocking this PR for it

bp_be/test/tb/bp_be_dcache/testbench.v
dpetrisko: Made an issue for it, not holding up this PR

bp_be/test/tb/bp_be_dcache/testbench.v
muwyse: fix is on dev

Issue #417, opened 2020-04-02
Enable easy selection of FSM or Microcode CCE in design


muwyse:
Currently, we must apply a change/patch to substitute in the FSM-based CCE for the microcode-based CCE. It would be ideal to enable this substitution without having to apply a code change. A couple possibilities (please suggest others!):

- Add a use_fsm_cce_p param to bp_params_p, and do conditional instantiation in the modules that instantiate the CCE or its nonsynth tracer (testbenches). The pro here is the code compiles as is using whatever the default is. A downside is it potentially doubles the number of configs we have (e.g., e_bp_single_core_cfg and e_bp_single_core_fsm_cce_cfg)

- Use a tick define or some other keyword and do a text substitution via the Makefile, similar to BP_CFG_FLOWVAR. Pro is we don't need the double configs, con is the design would require this preprocessing step before compilation. Something like this works fine for testbenches, but is probably not ideal for portability/packaging of the design source.

- others?

dpetrisko:
We could do 2 without the sed step. In aviary defines, we could do:

‘ifndef BP_CCE
‘define BP_CCE bp_cce
‘endif

And then compile with +define+BP_CCE=bp_cce_fsm to switch it out. 

The downside to this is that it globally defines the CCE as fsm or ucode, so it’s impossible to mix and match. Is that important? Who’s to say

muwyse:
ah, that could work well.

I don't think there is a need for mix and match at the moment. Since memory is striped across the CCE's there probably isn't benefit to having only some of them be programmable. The ucode CCE wouldn't be able to do a different coherence protocol on the fly unless the FSM CCE also supported runtime switching of the protocol, which is not planned.

taylor-bsg:
I think it depends on whether we expect others to use the flag. If it is a
supported feature of BP, it should be a config param. Otherwise it could be
preprocessor hackery, but usually we reserve that for when we are trying to
factor something in a way not supported by SV built in meta programming
support, but it seems like SV’s meta programming support is sufficient here?

dpetrisko:
Yeah that’s reasonable. I think there’s a use case for the FSM CCE as an FPGA build (since there is less motivation for post-synth configurability).  I’m sold on it being a parameter. 

From the perspective of code reduction, we could have a CCE wrapper which instantiates whichever module we want? It’s not a super big savings though, so the if statement could be fine

muwyse:
[...]

PR #421, opened 2020-04-03
Add configs that use FSM CCE, parameterized CCE instantiation

muwyse:
This change adds the fsm_cce_p parameter to bp_params_p, which is used to determine which CCE variant is instantiated in the design. A new bp_cce_wrapper module is added that does the conditional instantiation of the CCE type. Using a wrapper encapsulates the conditional instantiation and provides a common module for nonsynth tracers to bind to.

New configs are added that use the FSM CCE instead of microcode CCE.

PR addresses Issue #417

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -93,6 +93,8 @@ hotfix_job:
     - $CI_PROJECT_DIR/ci/regress.sh "check_design.syn TB=bp_processor CFG=e_bp_single_core_cfg" bp_top &
     - $CI_PROJECT_DIR/ci/regress.sh "check_design.syn TB=bp_processor CFG=e_bp_quad_core_cfg" bp_top &
     - $CI_PROJECT_DIR/ci/regress.sh "check_design.syn TB=bp_softcore" bp_top &
+    - $CI_PROJECT_DIR/ci/regress.sh "check_design.syn TB=bp_processor CFG=e_bp_single_core_ucode_cce_cfg" bp_top &
+    - $CI_PROJECT_DIR/ci/regress.sh "check_design.syn TB=bp_processor CFG=e_bp_quad_core_ucode_cce_cfg" bp_top &
     - wait
   cache:
     key: "dev"
@@ -132,6 +134,8 @@ check-design:
     - $CI_PROJECT_DIR/ci/regress.sh "check_design.syn TB=bp_processor CFG=e_bp_single_core_cfg" bp_top &
     - $CI_PROJECT_DIR/ci/regress.sh "check_design.syn TB=bp_processor CFG=e_bp_quad_core_cfg" bp_top &
     - $CI_PROJECT_DIR/ci/regress.sh "check_design.syn TB=bp_softcore" bp_top &
+    - $CI_PROJECT_DIR/ci/regress.sh "check_design.syn TB=bp_processor CFG=e_bp_single_core_ucode_cce_cfg" bp_top &
+    - $CI_PROJECT_DIR/ci/regress.sh "check_design.syn TB=bp_processor CFG=e_bp_quad_core_ucode_cce_cfg" bp_top &
     - wait
       
 lint-verilator:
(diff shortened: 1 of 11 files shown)


dpetrisko:
LGTM.  Only minor comment, maybe ucode_cce as the parameter instead of fsm? Ucode is kind of the "unusual condition" option.

Do we want FSM or ucode to be default?

muwyse:
I agree, ucode_cce for the parameter. FSM CCE will be default. We just need to make sure both are tested thoroughly in regressions.

.gitlab-ci.yml
```diff
@@ -93,6 +93,8 @@ hotfix_job:
     - $CI_PROJECT_DIR/ci/regress.sh "check_design.syn TB=bp_processor CFG=e_bp_single_core_cfg" bp_top &
     - $CI_PROJECT_DIR/ci/regress.sh "check_design.syn TB=bp_processor CFG=e_bp_quad_core_cfg" bp_top &
     - $CI_PROJECT_DIR/ci/regress.sh "check_design.syn TB=bp_softcore" bp_top &
+    - $CI_PROJECT_DIR/ci/regress.sh "check_design.syn TB=bp_processor CFG=e_bp_single_core_ucode_cce_cfg" bp_top &
+    - $CI_PROJECT_DIR/ci/regress.sh "check_design.syn TB=bp_processor CFG=e_bp_quad_core_ucode_cce_cfg" bp_top &
```
muwyse, line 97: Similar to checking the design with both FSM and ucode CCE, we likely need to enforce regressions being run against both designs, otherwise the non-default (ucode CCE) will almost certainly encounter bugs due to not being in the regression, even though we want to maintain it as supported. The question is do we create separate jobs for testing configs with FSM or ucode CCE, or do we test both configs within the same job. Properly testing both designs will effectively double our regressions, but the other option is to insufficiently test.

dpetrisko: How about just doubling the check_design targets and mc_* regressions?  I doubt we'll find any issues in the single core tests that we don't find there

dpetrisko: I think separate jobs is fine.  I'm going to refactor our CI system when I get some time in the next few weeks, so it won't be a high overhead

bp_me/syn/Makefile
```diff
@@ -85,19 +85,19 @@ lint.me:
...
 
 regress.me.sc: dirs.sc
-	$(MAKE) -j 1 build.sc sim.sc TB=bp_cce NUM_INSTR_P=$(NUM_INSTR_P)
-	$(MAKE) -j 1 build.sc sim.sc TB=bp_cce NUM_INSTR_P=$(NUM_INSTR_P) SKIP_INIT_P=1
-	$(MAKE) -j 1 build.sc sim.sc TB=bp_cce_fsm NUM_INSTR_P=$(NUM_INSTR_P)
-	$(MAKE) -j 1 build.sc sim.sc TB=bp_cce_fsm NUM_INSTR_P=$(NUM_INSTR_P) SKIP_INIT_P=1
+	$(MAKE) -j 1 build.sc sim.sc TB=bp_cce NUM_INSTR_P=$(NUM_INSTR_P) PROG=normal
+	$(MAKE) -j 1 build.sc sim.sc TB=bp_cce NUM_INSTR_P=$(NUM_INSTR_P) SKIP_INIT_P=1 PROG=uncached
+	$(MAKE) -j 1 build.sc sim.sc TB=bp_cce_fsm NUM_INSTR_P=$(NUM_INSTR_P) PROG=normal
+	$(MAKE) -j 1 build.sc sim.sc TB=bp_cce_fsm NUM_INSTR_P=$(NUM_INSTR_P) SKIP_INIT_P=1 PROG=uncached
```
muwyse, line 91: FYI, setting the program here to some dummy values so that unique logs are generated for all tests. If PROG is not specified, then the SKIP_INIT_P=1 tests overwrite the other one for the same config, potentially hiding a failure.

Since we use CFG and PROG to generate unique test result paths, and don't use the value of any specific params, setting PROG is the easiest way to get a unique name per test.

dpetrisko: Doesn’t TB affect the path though?

muwyse: yes, but the ME testbenches have a parameter that determines if they run in cached/normal mode or uncached mode. (This param gets passed to the trace rom generator to generate either cached accesses or uncached accesses). But this param doesn't affect path.

So the second test (for the same TB setting) overwrites the first, unless PROG is set to differentiate.

dpetrisko: gotcha!

bp_me/src/v/cce/bp_cce_wrapper.v
```diff
@@ -0,0 +1,79 @@
...
+  );
+
+  // Config Interface
+  `declare_bp_cfg_bus_s(vaddr_width_p, core_id_width_p, cce_id_width_p, lce_id_width_p, cce_pc_width_p, cce_instr_width_p);
+
+  // Config bus casting
+  bp_cfg_bus_s cfg_bus_cast_i;
+  assign cfg_bus_cast_i = cfg_bus_i;
+
+  if (ucode_cce_p == 1) begin
```
dpetrisko, line 67: Can you add begin : cce labels to this statement? Helps with backend

dpetrisko:
1 small fix requested otherwise is this good to go?

muwyse:
Yes. If you can do it and commit that works for me. Otherwise I can try to
do it later tonight or early tomorrow.

PR #428, opened 2020-04-06
Merge sw dev

dpetrisko:
- Updated software flow (Farzam)
- Extra spec benchmarks (Farzam)
- Updated checkpoint cosim capabilities
- Massively cleaning up tool flow
  - For this, all we need to do is run ```make prep -j N```  The Makefile is now smart enough to detect if the submodule has changed since the last build.  This makes CI and life so much easier.
- Adding an incoming I/O port to bp_softcore
- Adding config loader to bp_softcore
- Adding nbf loader to bp_softcore


dpetrisko:
Alright, CI is (finally) fixed.  This is good to go after we resolve the yaml conflicts

PR #431, opened 2020-04-08
Adding profilers for core and pc

dpetrisko:
- Adds bloodgraph.py
- Adds pc profiler
- Adds stall profiler

command is
```make build.v sim.v blood.v CORE_PROFILE_P=1```
Does not work for multiple cores at the moment


taylor-bsg:
Groovy!

PR #436, opened 2020-04-10
Fixing reserved keyword and prereq requirements

dpetrisko:


Makefile
```diff
@@ -59,7 +59,8 @@ $(TARGET_DIRS):
 libs: $(TARGET_DIRS)
 	$(MAKE) basejump
 	$(MAKE) dramsim2
-	$(MAKE) dramsim3
+	#$(MAKE) dramsim3
```
farzamgl, line 62: Why comment out dramsim3?

dpetrisko: we don't use it for anything, I just put it in there to see if it would build for the future

PR #438, opened 2020-04-12
fully specify TB and CFG in CI file for tests

muwyse:
This change fully specifies the TB and CFG params for the Makefiles for all tests in the gitlab CI file. Without fully specifying both parameters, some tests were running with TB=bp_processor and CFG=e_bp_softcore_cfg. I don't believe this is intended to be a supported TB/CFG pair.

Notably, the TB/CFG pair of bp_processor & e_bp_softcore_cfg results in the regress_riscv.v target failing some tests, but this does not appear to be detected by CI. Specifically, the rv64ui-v-* tests and rv64ui-p-fence_i tests report failures when I run them locally. Changing CFG to e_bp_single_core[_ucode_cce]_cfg results in all tests passing.

This pipeline: https://gitlab.com/black-parrot/black-parrot/pipelines/135024242 shows the CI tests reporting pass at a commit where there are failing tests due to the TB/CFG mismatch. Commit is 22579efd (current head of dev as of 4/12/20).

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -117,7 +117,7 @@ hotfix_job:
     - make -C bp_top/syn clean
     - $CI_PROJECT_DIR/ci/regress.sh "check_design.syn TB=bp_processor CFG=e_bp_single_core_cfg" bp_top &
     - $CI_PROJECT_DIR/ci/regress.sh "check_design.syn TB=bp_processor CFG=e_bp_quad_core_cfg" bp_top &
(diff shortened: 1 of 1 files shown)


muwyse:
NOTE: this does _**NOT**_ address the underlying issue of CI not detecting failing test cases from regress_riscv.v target and reporting the job as a failure.

muwyse:
To repro:
- check out "dev" at 22579efd
- cd bp_top/syn
- make build.v TB=bp_processor
- make regress_riscv.v TB=bp_processor -j 16

bp_top/syn/reports/vcs will hold the report files indicating tests are failing. This is not caught by CI though as mentioned above.

PR #440, opened 2020-04-13
Top dev

dpetrisko:
- Removing lots of unnecessary buffers for modules
- Adding TAGFL operation to cache, setting template for adding additional cache csrs
- Adding parameter for number of outstanding L2 reqs (should be low for softcore, high for high performance coherence)


bp_me/src/v/cache/bp_me_cce_to_cache.v
```diff
@@ -214,17 +213,27 @@ module bp_me_cce_to_cache
             endcase
           default: cache_pkt.opcode = LB;
         endcase
-        cache_pkt.data = cmd_data[cmd_counter_r];
-        cache_pkt.addr = cmd_addr + cmd_counter_r*data_mask_width_lp;
-        cache_pkt.mask = '1;
+
+        if ((mem_cmd_lo.header.addr < 32'h8000_0000) && (mem_cmd_lo.header.addr[0+:20] == cache_tagfl_base_addr_gp))
```
muwyse, line 217: Is there a define somewhere for this? dram_base_addr_gp? And I think the mem_cmd address is paddr_width so probable width mismatch here.

bp_top/src/v/bp_clint_slice.v
```diff
@@ -11,17 +11,20 @@ module bp_clint_slice
  #(parameter bp_params_e bp_params_p = e_bp_inv_cfg
    `declare_bp_proc_params(bp_params_p)
    `declare_bp_me_if_widths(paddr_width_p, cce_block_width_p, lce_id_width_p, lce_assoc_p)
+
+   // TODO: Should I be a global param?
+   , localparam clint_max_outstanding_p = 2
```
muwyse, line 16: Any decision here? Or leave as TODO?

dpetrisko: I think leave as todo for now, I couldn't think of a reason why we'd need more (at least not enough to go through the trouble of adding to aviary

bp_me/src/v/cache/bp_me_cce_to_cache.v
dpetrisko: good call, fixed throughout (although we should really refactor the global address map into its own module)

PR #448, opened 2020-04-22
Brings PC Histogram to Black Parrot

bornaehsani:
Adds Black Parrot PC Histogram profiler to parse the operation trace and generate basic block execution count for the entire network as well as for each bp core separately.


dpetrisko:
Thanks!

taylor-bsg:
Very cool Borna!

PR #451, opened 2020-04-22
Add instructions for build on docker

mpatnode:
Wayne is currently testing this for me.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,33 @@
+# Dockerfile References: https://docs.docker.com/engine/reference/builder/
+
+FROM ubuntu:18.04
+
+RUN apt-get update
+
+RUN DEBIAN_FRONTEND="noninteractive" apt-get install -y apt-utils tzdata git vim
+
+RUN apt-get install -y bash autoconf automake autotools-dev curl libmpc-dev libmpfr-dev libgmp-dev gawk
+
+RUN apt-get install -y build-essential bison flex texinfo gperf libtool patchutils bc zlib1g-dev libexpat-dev
+
+RUN apt-get install -y wget byacc device-tree-compiler python gtkwave python-yaml pkg-config swig python3-dev
+
+RUN apt-get clean
+
+ARG USER_ID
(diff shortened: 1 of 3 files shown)


dpetrisko:
Thanks for the PR! A few questions:

- What here is Mac specific? The exact docker command? Can we make this platform agnostic, assuming the user has docker installed
- This should work for tire kick, but we'll need more dependencies for doing useful things, like compiling programs.  We could have 2 docker images, but I think just having the more complete image is better, since users will probably want to run the tire kick, see it pass, and then move on to the next steps.  What do you think?

mpatnode:
- I'm not sure there's much value outside of a Mac, unless folks are seeing other Linux environments where the build is failing.   I didn't install any special versions of anything, just the standard ubuntu components.   Personally, I did all the work inside a WSL container on Windows (actually ran Docker inside of WSL!) 

- Ah, let me try some deeper exploration and see what else is missing.

dpetrisko:
>Personally, I did all the work inside a WSL container on Windows (actually ran Docker inside of WSL!)

That's awesome! 

> I'm not sure there's much value outside of a Mac, unless folks are seeing other Linux environments where the build is failing.

The use case I ran into a lot in university is with systems that you don't have full control over (campus compute clusters, ECE lab computers, research group servers, etc)

taylor-bsg:
Welcome Mike!

mpatnode:
So I've tested this on my Mac, and it seems to be working (still building after 4 hours).

taylor-bsg:
Nice — are you running with any parallel threads on multicore? That is a
long time to wait but we do tend to run with a lot of cores and threads.
Apparently can you rebuild gcc in 30 seconds on a 72-core machine..

mpatnode:
Hmm...  didn't try the parallel build.   Let me look into the Docker CPU allocation.

mpatnode:
"make -j" definitely helps.   Think this is ready to go.

dpetrisko:
Thanks!

PR #452, opened 2020-04-23
D$ Writethrough and Testbench Regressions

Shashank-Vijay:
Changes made:

- Updated mem_pkt interface to valid->yumi

- Made D$ optionally writethrough

- Added more tests to D$ testbench

- Created regressions for D$ and I$ testbenches

### DIFF
--- a/bp_be/src/v/bp_be_mem/bp_be_dcache/bp_be_dcache.v
+++ b/bp_be/src/v/bp_be_mem/bp_be_dcache/bp_be_dcache.v
@@ -672,7 +672,7 @@ module bp_be_dcache
     cache_req_cast_o = '0;
 
     // Assigning sizes to cache miss packet
-    if (uncached_tv_r)
+    if (uncached_tv_r | (wbuf_v_li & (writethrough_p == 1)))
       unique case (size_op_tv_r)
         e_dword: cache_req_cast_o.size = e_size_8B;
         e_word: cache_req_cast_o.size = e_size_4B;
@@ -693,6 +693,10 @@ module bp_be_dcache
     end
     else if(uncached_load_req) begin
       cache_req_cast_o.msg_type = e_uc_load;
+      cache_req_v_o = cache_req_ready_i;
+    end
+    else if(wbuf_v_li & (writethrough_p == 1)) begin
+      cache_req_cast_o.msg_type = e_wt_store;
       cache_req_v_o = cache_req_ready_i;
     end
     else if(uncached_store_req) begin
@@ -726,7 +730,7 @@ module bp_be_dcache
   // output stage
   // Cache Miss Tracking logic
   logic cache_miss_r;
-  wire miss_tracker_en_li = cache_req_v_o & ~uncached_store_req & ~fencei_req;
+  wire miss_tracker_en_li = cache_req_v_o & ~uncached_store_req & ~fencei_req & ~(wbuf_v_li & (writethrough_p == 1));
   bsg_dff_reset_en
    #(.width_p(1))
    cache_miss_tracker
@@ -772,7 +776,7 @@ module bp_be_dcache
    (.clk_i(clk_i)
     ,.reset_i(reset_i)
 
-    ,.en_i(wbuf_v_li | flush_req)
+    ,.en_i((wbuf_v_li & (writethrough_p == 0)) | flush_req)
     ,.data_i(wbuf_v_li)
 
     ,.data_o(gdirty_r)
@@ -1043,7 +1047,7 @@ module bp_be_dcache
     if (v_tv_r) begin
       lru_decode_way_li = store_op_tv_r ? store_hit_way : load_hit_way;
       dirty_mask_way_li = store_hit_way;
-      dirty_mask_v_li = store_op_tv_r;
+      dirty_mask_v_li = store_op_tv_r & (writethrough_p == 0);
       
       stat_mem_data_li.lru = lru_decode_data_lo;
       stat_mem_data_li.dirty = {dcache_assoc_p{1'b1}};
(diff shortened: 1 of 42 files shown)


dpetrisko:
Let's make writethrough_p a parameter in aviary

bp_be/src/v/bp_be_mem/bp_be_dcache/bp_be_dcache.v
```diff
@@ -1052,7 +1064,7 @@ module bp_be_dcache
     else begin
       lru_decode_way_li = stat_mem_pkt.way_id;
       dirty_mask_way_li = stat_mem_pkt.way_id;
-      dirty_mask_v_li = 1'b1;
+      dirty_mask_v_li = 1'b1 & (writethrough_p == 0);
```
dpetrisko, line 1067: Don't need this line.  If the cache engine wants to clear dirty bits for some reason, let it :)

bp_be/src/v/bp_be_mem/bp_be_dcache/bp_be_dcache.v
```diff
@@ -1043,7 +1055,7 @@ module bp_be_dcache
     if (v_tv_r) begin
       lru_decode_way_li = store_op_tv_r ? store_hit_way : load_hit_way;
       dirty_mask_way_li = store_hit_way;
-      dirty_mask_v_li = store_op_tv_r;
+      dirty_mask_v_li = store_op_tv_r & (writethrough_p == 0);
```
dpetrisko, line 1058: // Blocks are never dirty in a writethrough cache

bp_be/src/v/bp_be_mem/bp_be_dcache/bp_be_dcache.v
```diff
@@ -1098,7 +1112,7 @@ module bp_be_dcache
     );
   
   assign data_mem_o = lce_data_mem_data_li; 
-  assign data_mem_pkt_ready = ~(load_op & tl_we) & ~wbuf_v_lo;
+  assign data_mem_pkt_yumi_o = (data_mem_pkt.opcode == e_cache_data_mem_uncached) ? data_mem_pkt_v : ~(load_op & tl_we) & ~wbuf_v_lo & ~lce_snoop_match_lo & data_mem_pkt_v;
```
dpetrisko, line 1115: Comment explaining

bp_be/src/v/bp_be_mem/bp_be_dcache/bp_be_dcache.v
```diff
@@ -772,7 +784,7 @@ module bp_be_dcache
    (.clk_i(clk_i)
     ,.reset_i(reset_i)
 
-    ,.en_i(wbuf_v_li | flush_req)
+    ,.en_i((wbuf_v_li & (writethrough_p == 0)) | flush_req)
```
dpetrisko, line 787: let's do if (writethrough_p)
gdirty_r = '0;
else
gdirty_reg, etc.

sripathi-muralitharan:
Haven't yet made writethrough_p as part of aviary. Will do it as part of the next commit.

sripathi-muralitharan:
While running the regressions once again, I seemed to have stumbled upon a bug. There were some updates made to the L2 recently and while things work well with the writeback L1, they seem to break with the writethrough L1. 

The following is the situation I observed when I ran rv64ui-p-sd in bp_softcore - 
The core stalls for some reason and after some investigation and tracking signals down I find that one of the memory commands to the I$ is lost and hence the I$ is in an eternal state of a miss. 
What caused this situation is that when the I$ happens to send its command to the L2 for a read, a wt_store closely follows it. Although the L2 does read the data and gets the block back, the response is not registered because of the way the response FSM is setup in bp_me_cce_to_cache. 
I think the culprit is the bsg_dff_en in that part of the code which gets enabled on a mem_cmd_yumi. The way this module is structured is that the cache doesn't dequeue a cmd from the cmd FIFO until it has completely serviced it. Once a command is completely serviced, the mem_cmd_yumi_li signal goes high, which enables the dff to store the command (which in turn becomes the response header). However, when a wt_store closely follows, the mem_cmd_yumi_li is raised immediately when the store is accepted by the cache (L2) which overwrites the contents of the DFF. It so happens that when the FSM is the RESP_SEND state to send the data back to the L1 I$, the resp_header is overwritten by the resp_header for the wt_store which causes the loss of data causing the I$ and hence the core to stall. 

Proposed solution : We probably might need a fifo to handle this situation.

bp_be/src/v/bp_be_mem/bp_be_dcache/bp_be_dcache.v
```diff
@@ -726,7 +738,9 @@ module bp_be_dcache
   // output stage
   // Cache Miss Tracking logic
   logic cache_miss_r;
-  wire miss_tracker_en_li = cache_req_v_o & ~uncached_store_req & ~fencei_req;
+  logic wt_req;
```
dpetrisko, line 741: minor:
wire wt_req =

bp_me/src/v/cache/bp_me_cce_to_cache.v
```diff
@@ -168,7 +170,7 @@ module bp_me_cce_to_cache
       end
       READY: begin
         // Technically possible to bypass and save a cycle
-        if (mem_cmd_v_lo)
+        if (mem_cmd_v_lo & is_resp_ready)
```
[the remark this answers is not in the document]
sripathi-muralitharan: Do you mean to get rid of the READY state altogether and do everything in the SEND state?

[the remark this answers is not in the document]
dpetrisko: No, I mean move (... & is_resp_ready) to the SEND state, instead of the READY state.

This way, it waits for the resp to complete, moves to send, then starts the message.  No reason not to wait in send

dpetrisko:
Awesome. Looks good to me, just need to add writethrough to the aviary. Just 1 config for softcore is fine for now. And an assertion in if_verif to disallow writethrough and coherent_l1.

Issue #404, opened 2020-03-23
LCE req should use data length field


dpetrisko:
Uncached requests are assumed to be 64 bits max.  While we still need the uc_req_size field in order to perform subword uncached accesses, we also want to support up to cache line size uncached transfers, for host interfacing or DMA.

https://github.com/black-parrot/black-parrot/blob/top_dev/bp_top/src/v/bp_io_link_to_lce.v#L90

muwyse:
I will take a look at adding support for uncached cache block size load/store after the CCE v3 is tested and in (in the coming days). I think this will be relatively straightforward to support.

muwyse:
Merge #453 adds functionality to enable LCE Requests up to cache block size. More changes are needed to modify the LCE-CCE networks and endpoints (LCEs and CCEs) to fully handle this case. For example, the LCE Request network is currently instantiated system wide to only support up to 64-bits of data. This will need to change at the CCEs and the LCEs that want to send uncached stores up to cache block size.

It may make sense to roll this together with fixes for #260 to avoid unnecessarily wide connections. Or, we would want wormhole adapters that can be customized for the endpoint. For example, if an LCE will never do an uncached store larger than 64-bits, it doesn't need to connect to its wormhole adapter with an LCE Req struct setup to handle up to 512-bits.

dpetrisko:
> For example, the LCE Request network is currently instantiated system wide to only support up to 64-bits of data. This will need to change at the CCEs and the LCEs that want to send uncached stores up to cache block size.

Could you point to how this is handled currently?

> Or, we would want wormhole adapters that can be customized for the endpoint.

That's easy enough, just need to change the payload width to correspond to the max data size

https://github.com/black-parrot/basejump_stl/blob/c04be76f99eee6c1f1a75e6a22f6c5de34707132/bsg_noc/bsg_wormhole_router_adapter_in.v#L11

muwyse:
Here is my initial plan:
1. change "data_width_mp" param name in LCE-CCE IF macro to "lce_req_max_data_width_mp" to clarify naming. Add comments.

2. modify the encoder files in bp_me/src/v/wormhole to properly compute the wormhole length for all messages.
2a. Uncached LCE store request message size needs to be based on the size field of the message
2b. Uncached LCE Command data message size needs to be based on the size field of the message

[...]

Issue #427, opened 2020-04-05
Fixed-width CCE message sizes


sripathi-muralitharan:
The CCE (maybe in other parts of the code as well, but noticed it here) seems to have a few spots where the block widths and the corresponding message sizes are hardcoded to 512 bits. Ideally we would want this to be based on `cce_block_width_p` which makes it easier to support variable cache block sizes. @muwyse could you please have a look at this?

dpetrisko:
https://github.com/black-parrot/black-parrot/blob/dev/bp_me/src/v/cce/bp_cce_msg.v#L725

Here is one of the lines in question

muwyse:
Closed by merge #453

Issue #329, opened 2020-02-21
Unnecessary message types in memory interface


dpetrisko:
The memory interface should only define read and write commands.  Information about uncached / prefetch is only relevant as part of the return type and therefore should be part of the payload, not the main message.

https://github.com/black-parrot/pre-alpha-release/blob/master/bp_me/src/include/v/bp_me_cce_mem_if.vh#L22

Another advantage is that lets peripherals be a little bit leaner by defining smaller payloads

muwyse:
I'm not convinced we want it to be that restrictive. At a minimum I think we want read, write, prefetch, and an uncached bit to be visible to the memory-side controllers (i.e., not hidden in the payload). For example, if the CCE connects to an L2, the L2 may be able to read only a subset of a cache block (saving power, and maybe cycles) for an uncached request (this also depends on what we mean by uncached; our current design has the L2 being a victim cache on the memory-side, so we can imagine an uncached access from a CPU being served from the L2). Additionally, a prefetch could be ignored by the L2 (with the L2 simply responding immediately to the CCE and not prefetching the block) if, for example, the current load on the L2 or memory controller is high, where adding additional commands may further congest the networks or delay non-prefetch (required) accesses.

Further, it is likely (from a research standpoint) that we may want to support custom message types / commands that are visible to the memory controllers (i.e., not hidden in the payload). The memory controller could then issue simple read/write style commands, but it might be useful for it to accept more complicated commands and then implement their functionality.

dpetrisko:
You've convinced me on the prefetch bit, but I think the uncached bit is confusing as currently implemented.  It currently indicates uncached in L1, but it is sent to L2 / I/O devices.  It's caused bugs where I/O devices send back uncached responses to cached requests (when it should be perfectly valid to cache I/O in some cases).  

I think maybe what would solve my confusion is to have the uncached-in-L1 bit live in the payload, but the uncached-in-L2 bit live in the message type.

>  the L2 may be able to read only a subset of a cache block (saving power, and maybe cycles) for an uncached request
I think this is solved by the size field

> Further, it is likely (from a research standpoint) that we may want to support custom message types / commands that are visible to the memory controllers (i.e., not hidden in the payload).
There's no reason a "researchy" device couldn't inspect the payload. I think it makes sense to hide it by default unless supported.

Nit: do we need the speculative bit in the payload, if it's a prefetch request shouldn't it return a prefetch response?

dpetrisko:
[...]

Issue #260, opened 2019-09-25
Prevent serializing and deserializing in the CCE


dpetrisko:
Currently on a memory request, the LCE sends a request to the CCE through the wormhole router which deserializes it, reads the header, and then serializes it again.  This is wasteful because of both latency and area.  We should be able to only look at the header of the packet and forward the request/response to the correct FIFO.  This will require the module in https://github.com/bespoke-silicon-group/basejump_stl/issues/158.

1) Need to change packet format so that head is on LSB (transmitted first).
2) Is a constraint that the entire header is within a single flit reasonable? (I think yes)
3) CCE needs interface changes to accept a link and forward to correct LCEs

 One thing to be aware of is that this will break our decoupling of network and logic for the CCE.  It should be a large enough PPA boost that the abstraction is too costly to maintain.  The CCE interface will most likely have to change from having full packets to wormhole links.

dpetrisko:
The SIPOs and PISOs are 21% of non-RAM area

PR #453, opened 2020-04-24
Merge me_dev to dev - LCE-CCE and CCE-MEM Interface updates

muwyse:
This PR refines the LCE-CCE and CCE-MEM Interfaces to resolve the following issues: #427 and #329. 

It lays the groundwork for resolving #404, which will require modifying the system to allow LCE request channel to carry up to a cache block of data.

A future change will implement full support for any valid subset of the MOESIF protocol. The changes and newly added message types in the LCE-CCE Interface enable this.


dpetrisko:
Wow, knocking out those issues :)

muwyse:
> So we're committing to the smallest granularity of message for mem_msg_size_e being 64 bits for all of BedRock/Mem/IO? I think this is probably a reasonable restriction, just clarifying.

The interface supports message size down to 1 byte, which are used for uncached load/store. Cache block size is restricted to 64-bits at the smallest since we have a 64-bit processor / architecture.

muwyse:
In order to resolve issue #404 , more changes will be required such that the LCE Req networks are capable of carrying up to a cache block of data. Connections at I$ and D$ LCEs likely only need to allocate structs allowing up to 64-bits of data, but I/O LCE's will need to be able to send/receive up to cache block size data. The CCE will also need to be able to handle up to cache block size requests on the LCE Request network.

It likely makes sense to combine fixing this with issue #260 to avoid the serdes cost increases

PR #457, opened 2020-04-28
Fix verilator lint target

muwyse:
This change fixes the Verilator lint target so it should run correctly on CI. Note that Verilator reports failure due to mem_cap_in_bytes_p being set larger that Verilator likes in the bp_softcore testbench, as it is set to 2**29. The setting in bp_processor of 2**25 seems to be okay.

A small typo is also fixed in the bp_fe_bht.v file in a debug print statement, which removes an old format string item.


muwyse:
To be clear: merging will cause the Verilator lint job to fail due to the memory size of 2^29 set in bp_softcore testbench.v

PR #458, opened 2020-04-28
Disable cosim for multicore builds

muwyse:
This change instantiates the nonsynth cosim module in bp_processor testbench only if the core count is equal to one, since it is not expected to work for multicore.

### DIFF
--- a/bp_me/software/scripts/run.sh
+++ b/bp_me/software/scripts/run.sh
@@ -37,12 +37,12 @@ do
     # collect results
     for i in "${!nums[@]}"
     do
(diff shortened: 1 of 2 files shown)


muwyse:
See this pipeline on me_dev for regressions status of change: https://gitlab.com/black-parrot/black-parrot/pipelines/140838838

dpetrisko:
Nice fix.  Could you add ```assert (cosim_p == 0) else $error("cosim_p not supported for num_core_p > 1")``` to the else clause?

muwyse:
> Nice fix. Could you add `assert (cosim_p == 0) else $error("cosim_p not supported for num_core_p > 1")` to the else clause?

added in an initial block; VCS didn't like an assertion inside the instantiation else clause.

PR #465, opened 2020-04-30
enable verilator CI for multicore, beebs, and coremark

muwyse:
Re-enable CI runs of multicore tests and coremark using Verilator


muwyse:
Beebs is not re-enabled due to existing comment that it may take too long to run with Verilator. If we can confirm that is no longer an issue, we can re-enable now, too.

dpetrisko:
Beebs is good to go as well

muwyse:
Beebs enabled

PR #467, opened 2020-05-01
Revamp testing infrastructure

dpetrisko:
So this is a fairly big PR, but here are the highlights:

- bp_common/syn/<tools> are now truly generic, and only have flow steps that are applicable to all testbenches
- test-specific collateral is defined in tb/<testbench>/Makefile.*
- tests are only rebuilt when they change (like the tool build), or when you type e.g. make beebs_manual
- Tests are separated into "suites" for better organzation.  bp_tests is now a submodule containing our local tests
- Test derivatives such as nbf and spike are generated at sim time as opposed to pre-generated, to save space and maintain freshness
- regressions are run via "testlists"
- a new make report target summarizes results

For example, the toplevel testbench defines collateral as dynamically generated flists, cce ucode, etc.  This approach is much more flexible in the longterm.

1) Can you all take a look to see if this looks reasonable to you, try out the flow, etc?
2) Porting your testbenches to this new flow would be much appreciated. Or at least advising me on which collateral needs to be generated
3) I need to update the .gitlab scripts to use these new targets

Thanks!


bp_common/test/Makefile.frag
```diff
@@ -465,20 +468,21 @@ BEEBS_TESTS = \
   ud \
   whetstone \
   wikisort
-  # qsort works, but there's a name conflict with riscv-tests. We should fix this
+  # qsort works, but there's a name conflict with riscv_tests. We should fix this
   #   by putting each test suite in its own mem directory
```
muwyse, line 472: Does this PR fix this name conflict, or is it still present?

bp_fe/src/v/bp_fe_btb.v
```diff
@@ -49,6 +49,7 @@ logic                       tag_mem_v_lo;
 
 logic [vaddr_width_p-1:0]   tgt_mem_li, tgt_mem_lo;
 logic [btb_idx_width_p-1:0] tgt_mem_addr_li;
+logic                       tag_jmp_lo;
```
muwyse, line 52: This is a duplicate declaration of line 58 or 59 below. One of them needs to be removed.

bp_common/test/Makefile.frag
dpetrisko: fixed! I'll remove the comment

bp_fe/src/v/bp_fe_btb.v
dpetrisko: Good catch.  It snuck in because I implemented your fix manually and then rebased

bp_top/syn/Makefile
```diff
@@ -5,51 +5,24 @@ include $(TOP)/Makefile.common
...
-CCE_MEM_PATH = $(BP_ME_DIR)/src/asm/roms
-CCE_MEM      = $(COH_PROTO).mem
-
-DRAMSIM_CH_CFG  ?= DDR3_micron_32M_8B_x8_sg25E.ini
-DRAMSIM_SYS_CFG ?= system.ini
+export TB    ?= bp_tethered
+export CFG   ?= e_bp_softcore_cfg
+export SUITE ?= bp_tests
+export PROG  ?= hello_world
+export TAG   ?= none
```
sripathi-muralitharan, line 18: What is the intended functionality of TAG?

bp_common/syn/Makefile.common
```diff
@@ -8,7 +8,22 @@ include $(TOP)/Makefile.common
...
-	$(MAKE) clean.v clean.sc clean.syn
-	rm -rf logs
-	rm -rf reports
-	
+	@rm -rf results/
+	@rm -rf logs/
+	@rm -rf reports/
+	@rm -rf touchfiles/
+
+report:
```
sripathi-muralitharan, line 16: This is great!

bp_top/syn/Makefile
dpetrisko: Good question! It's not fully implemented yet.  It lets users tag their tests based on something other than the configuration and program -- for example, which testing corner you're using in a VLSI flow

bp_common/syn/Makefile.common
Shashank-Vijay: Yeah this is very helpful!

PR #471, opened 2020-05-04
Add partial fill support to Cache Service Interface

Songchun-Li:
### Cache and UCE Interface modification

1. fill mask added in data_mem_pkt to indicate the banks being filled
2. Data field in data_mem_pkt is changed to fill width from data_width
3. cache_req_critical signal is added
4. In-progress registers indicating the info of the only one outstanding missed cacheline are added
5. Corresponding logic like generating writemask to SRAM based on fill mask and checking hitted bank is filled or not are added

### DIFF
--- a/bp_common/src/include/bp_common_aviary_defines.vh
+++ b/bp_common/src/include/bp_common_aviary_defines.vh
@@ -133,6 +133,7 @@ typedef struct packed
   integer acache_sets;
   integer acache_assoc;
   integer acache_block_width;
+  integer acache_fill_width;
   integer cce_pc_width;
   integer ucode_cce;
 
@@ -240,6 +241,7 @@ typedef struct packed
   , localparam acache_sets_p              = proc_param_lp.acache_sets                              \
   , localparam acache_assoc_p             = proc_param_lp.acache_assoc                             \
   , localparam acache_block_width_p       = proc_param_lp.acache_block_width                       \
+  , localparam acache_fill_width_p        = proc_param_lp.acache_fill_width                        \
   , localparam lce_assoc_p                = `BSG_MAX(dcache_assoc_p,                               \
                                                      `BSG_MAX(icache_assoc_p, acache_assoc_p))     \
   , localparam lce_assoc_width_p          = `BSG_SAFE_CLOG2(lce_assoc_p)                           \
@@ -250,7 +252,6 @@ typedef struct packed
   , localparam cce_block_width_p          =  `BSG_MAX(dcache_block_width_p,                        \
                                                      `BSG_MAX(icache_block_width_p,                \
                                                        acache_block_width_p))                      \
-  , localparam lce_fill_width_p           =  `BSG_MAX(dcache_fill_width_p, icache_fill_width_p)    \
                                                                                                    \
                                                                                                    \
   , localparam cce_pc_width_p             = proc_param_lp.cce_pc_width                             \
--- a/bp_common/src/include/bp_common_aviary_defines.vh
+++ b/bp_common/src/include/bp_common_aviary_defines.vh
@@ -7,7 +7,7 @@
 `ifndef BP_COMMON_AVIARY_DEFINES_VH
 `define BP_COMMON_AVIARY_DEFINES_VH
 
-// Thoughts: 
+// Thoughts:
 // Hardcoding hartid and lceid width limits us to 8 cores for our standard configurations,
 //   but would allow the hierachical flow to reuse a single BP core for both dual-core and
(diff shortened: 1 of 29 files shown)


bp_common/src/include/bp_common_aviary_defines.vh
```diff
@@ -229,9 +231,11 @@ typedef struct packed
   , localparam dcache_sets_p              = proc_param_lp.dcache_sets                              \
   , localparam dcache_assoc_p             = proc_param_lp.dcache_assoc                             \
   , localparam dcache_block_width_p       = proc_param_lp.dcache_block_width                       \
+  , localparam dcache_fill_width_p        = proc_param_lp.dcache_fill_width                        \
   , localparam icache_sets_p              = proc_param_lp.icache_sets                              \
   , localparam icache_assoc_p             = proc_param_lp.icache_assoc                             \
   , localparam icache_block_width_p       = proc_param_lp.icache_block_width                       \
+  , localparam icache_fill_width_p        = proc_param_lp.icache_fill_width                        \
```
dpetrisko, line 238: Good, also add an assertion to bp_nonsynth_if_verif that fill_width == block_width

bp_fe/src/v/bp_fe_icache.v
```diff
@@ -191,21 +194,26 @@ module bp_fe_icache
...
   logic [icache_assoc_p-1:0][ptag_width_lp-1:0]          tag_tv_r;
   logic [icache_assoc_p-1:0][$bits(bp_coh_states_e)-1:0] state_tv_r;
   logic [icache_assoc_p-1:0][bank_width_lp-1:0]          ld_data_tv_r;
   logic [ptag_width_lp-1:0]                     addr_tag_tv;
   logic [index_width_lp-1:0]                    addr_index_tv;
-  logic [word_offset_width_lp-1:0]              addr_word_offset_tv;
+  logic [bank_offset_width_lp-1:0]              addr_bank_offset_tv;
+
   logic                                         fencei_op_tv_r;
+  // Partial fill in-progress register
```
dpetrisko, line 213: We don't want this as part of this PR -- remove for now

bp_fe/src/v/bp_fe_icache.v
```diff
@@ -410,8 +449,6 @@ module bp_fe_icache
     : final_data[instr_width_p-1:0];
 
   // data mem
-
-  logic data_mem_v;
```
dpetrisko, line 414: I don't think this should be removed?

dpetrisko:
We also want to implement these changes in the LCE (with block_width == fill_width)

bp_fe/src/v/bp_fe_icache.v
Songchun-Li: The declaration is moved to line 277 for updating the in-progress register. Will be recovered while removing all in-progress reg

Songchun-Li:
In-progress register and relevant logic are removed.
Assertion for fill width != block width is added.
IF modification on LCE are added.

dpetrisko:
LGTM.

1) Can you rebase on me_dev?
2) Has this been tested with bank_width != dword_width.  For example, 2-way 512 bit lines?

Songchun-Li:
> LGTM.
> 
> 1. Can you rebase on me_dev?
> 2. Has this been tested with bank_width != dword_width.  For example, 2-way 512 bit lines?

It is tested using bp_softcore and bp_processor(single core) on 8-way, 4-way, and 2-way with 512 bit lines.

bp_common/src/include/bp_common_aviary_defines.vh
```diff
@@ -245,6 +249,7 @@ typedef struct packed
   , localparam cce_block_width_p          =  `BSG_MAX(dcache_block_width_p,                        \
                                                      `BSG_MAX(icache_block_width_p,                \
                                                        acache_block_width_p))                      \
+  , localparam lce_fill_width_p           =  `BSG_MAX(dcache_fill_width_p, icache_fill_width_p)    \
```
dpetrisko, line 252: We want to support different fill widths for I$ and D$ for LCE

dpetrisko: So that’s similar to the UCE having a generic fill_width parameter, just in the LCE/LCE submodules

Songchun-Li:
1. The branch is now rebased on me_dev
2. LCE now receives set, assoc, block width, fill width parameters when it is instantiated in bp_core.
3. The lce_fill_width_p is removed and acache_fill_width_p is added.

PR #481, opened 2020-05-05
Dev hotfix vivado synthesis

dpetrisko:
- Fixes bad case statement comparison
- Fences go to mem pipe, not control pipe
- //synopsys translate_off unguarded nonsynth block

### DIFF
--- a/bp_be/src/v/bp_be_calculator/bp_be_pipe_long.v
+++ b/bp_be/src/v/bp_be_calculator/bp_be_pipe_long.v
@@ -42,21 +42,16 @@ module bp_be_pipe_long
   logic [dword_width_p-1:0] op_a, op_b;
   always_comb
     begin
-      op_a = rs1_i;
-      op_b = rs2_i;
-      unique case (decode.fu_op)
-        `RV64_DIVW, `RV64_REMW:
-          begin
-            op_a = $signed(rs1_i[0+:word_width_lp]);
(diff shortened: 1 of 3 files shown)


bp_be/src/v/bp_be_calculator/bp_be_pipe_long.v
```diff
@@ -42,21 +42,16 @@ module bp_be_pipe_long
...
-          begin
-            op_a = rs1_i[0+:word_width_lp];
-            op_b = rs2_i[0+:word_width_lp];
-          end
-        default : begin end
-      endcase
+      op_a = decode.opw_v 
+             ? signed_div_li
+               ? dword_width_p'($signed(rs1_i[0+:word_width_lp]))
+               : rs1_i[0+:word_width_lp]
```
farzamgl, line 48: Why not use the whole dword?

dpetrisko: What do you mean?

farzamgl: Why are we using the lower 32 bits in that case?

farzamgl: Oh nevermind I got it.

dpetrisko: yeah word op

Issue #402, opened 2020-03-20
BR + JMP target should be half-word aligned


dpetrisko:
https://github.com/black-parrot/black-parrot/blob/master/bp_be/src/v/bp_be_calculator/bp_be_pipe_int.v#L81

As per the spec.

(I believe this will fix #400)

PR #489, opened 2020-05-10
Fixing LSB of branch target to always be 0

### LINKED ISSUE #400: Cosim mismatch when running random instruction test [riscv_jump_stress_test_0]
I was running several tests that are randomly generated with [riscv-dv](https://github.com/google/riscv-dv) tool from Google. Several of them are failing. This is one of them.

Dromajo error:
```
[error] EMU PC 00000000800072b8, DUT PC 00000000800072b9 <<-- PC is off
[error] EMU INSN 011d6bb3, DUT INSN 011d6bb3
[error] EMU MSTATUS a000c0080, DUT MSTATUS 00000000
[error] DUT pending exception -1 pending interrupt -1
```

Here is the binary and generated assembly: [riscv_jump_stress_test_0.zip](https://github.com/black-parrot/black-parrot/files/4357647/riscv_jump_stress_test_0.zip)

Failing test stresses jump instructions. To generate this test I used default riscv-dv's configuration for this test:
```
 71 - test: riscv_jump_stress_test
 72   description: >
 73     Stress back-to-back jump instruction test
 74   iterations: 2
 75   gen_test: riscv_instr_base_test
 76   gen_opts: >
 77     +disable_compressed_instr=1
 78     +instr_cnt=5000
 79     +num_of_sub_program=5
 80     +directed_instr_1=riscv_jal_instr,20
 81   rtl_test: core_base_test
```

dpetrisko:
RISC-V spec defines JALR to zero LSB to 0, to prevent misalignment

Fixes #400 #402

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -74,7 +74,7 @@ tools: | $(TARGET_DIRS)
 
 progs: tools
 	git submodule update --init --recursive $(BP_COMMON_DIR)/test
-	$(MAKE) -C $(BP_COMMON_DIR)/test perch bp_tests riscv_tests beebs
+	$(MAKE) -C $(BP_COMMON_DIR)/test perch bp_tests riscv_tests beebs coremark
 
 ucode: | basejump
 	$(MAKE) -C $(BP_ME_DIR)/src/asm roms
--- a/bp_be/src/v/bp_be_calculator/bp_be_pipe_ctrl.v
+++ b/bp_be/src/v/bp_be_calculator/bp_be_pipe_ctrl.v
@@ -70,7 +70,7 @@ assign taken_tgt = baddr + imm_i;
 assign ntaken_tgt = pc_i + 4'd4;
 
 assign data_o   = vaddr_width_p'($signed(ntaken_tgt));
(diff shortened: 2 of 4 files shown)


farzamgl:
Is that only for JALR or all branching instructions?

dpetrisko:
It's required for JALR only, but it's 0 by construction for BR and JAL (since the immediates are 2-bit aligned and the PC is 2-bit aligned)

farzamgl:
I thought when the branch target is misaligned, we'll get an exception instead of masking it.

dpetrisko:
misaligned is defined when br_tgt[3:0] != 0000 for RV64I or br_tgt[1:0] != 00 for RV64C.  br_tgt[0] = 1 is impossible to construct out of PC+IMM (BR/JAL).  It's only possible for JALR to have br_tgt[0] because it comes from a register instead of being generated from the instruction

farzamgl:
Why is it impossible? What if the immediate field has garbage value?

dpetrisko:
because the immediates are defined as multiples of 2 bytes (and the actual PC is obviously aligned to 2 bytes)
image
image

PR #490, opened 2020-05-10
Fix poison condition for long latency instructions

dpetrisko:
This caused issues where things wouldn't quite be flushed in some situations, leading to a spurious long instruction writing back.  Interestingly, cosim had a hard time finding it, because the writeback happens out of band, so the commits looked fine.


dpetrisko:
Nope, that's why this is a subtle bug.  The commit was getting poisoned (so it doesn't show up in cosim), but the writeback was not getting poisoned.  So what would happen is the pipeline would flush on exception, then at some point in the future, the division would complete and write back garbage into the regfile

farzamgl:
Should we maybe latch the commit_v signal and writeback only if it's high?

dpetrisko:
Yeah, probably a good idea

taylor-bsg:
Hi, for each bug fix, we should also see a commit of a regression test.

dpetrisko:
I'll add the riscvdv tests that raised these bugs

PR #491, opened 2020-05-12
fix CCE and LCE tracing in bp_top and bp_me testbenches

muwyse:
This change adds back the CCE tracer to the bp_top testbench, instantiating only if CCE_TRACE_P=1 is passed on the command line.

Setting cce_trace_p to 1 for invalid configs (e.g., softcores) results in a build error since the CCE does not exist and therefore the bind statement fails. I think this is acceptable, as the alternative is to guard the tracer instantiation by checking the provided config, which is a burden to maintain as configs are added/changed/removed.

This change also fixes up the ME bp_cce testbench to re-enable the CCE and LCE tracers.

### DIFF
--- a/bp_me/syn/Makefile
+++ b/bp_me/syn/Makefile
@@ -26,24 +26,6 @@ include $(BP_COMMON_DIR)/syn/Makefile.dc
 include $(BP_COMMON_DIR)/syn/Makefile.verilator
 include $(BP_COMMON_DIR)/syn/Makefile.vcs
 
-regress.me.sc: dirs.sc
-	$(MAKE) -j 1 build.sc sim.sc TB=bp_cce NUM_INSTR_P=$(NUM_INSTR_P) PROG=normal
-	$(MAKE) -j 1 build.sc sim.sc TB=bp_cce NUM_INSTR_P=$(NUM_INSTR_P) SKIP_INIT_P=1 PROG=uncached
-	$(MAKE) -j 1 build.sc sim.sc TB=bp_cce_fsm NUM_INSTR_P=$(NUM_INSTR_P) PROG=normal
-	$(MAKE) -j 1 build.sc sim.sc TB=bp_cce_fsm NUM_INSTR_P=$(NUM_INSTR_P) SKIP_INIT_P=1 PROG=uncached
-
-regress.me.v: dirs.v
-	$(MAKE) -j 1 build.v sim.v TB=bp_cce NUM_INSTR_P=$(NUM_INSTR_P) PROG=normal
-	$(MAKE) -j 1 build.v sim.v TB=bp_cce NUM_INSTR_P=$(NUM_INSTR_P) SKIP_INIT_P=1 PROG=uncached
-	$(MAKE) -j 1 build.v sim.v TB=bp_cce_fsm NUM_INSTR_P=$(NUM_INSTR_P) PROG=normal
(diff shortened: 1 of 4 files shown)


muwyse:
This change also adds sim_dump.sc target to enable simulation waveform dumps when using Verilator.

dpetrisko:
Does the conditional instantiation work in Verilator now? The whole reason we do that weird clock gating solution was because Verilator had some problem with it

muwyse:
> Does the conditional instantiation work in Verilator now? The whole reason we do that weird clock gating solution was because Verilator had some problem with it

It appears to work fine, yes. make build.sc sim.sc PROG=hello_world CFG=e_bp_single_core_cfg works with both CCE_TRACE_P=0/1

dpetrisko:
And softcore?

muwyse:
It works.

Oddly enough, Verilator ignores the invalid bind if CFG=e_bp_softcore_cfg and CCE_TRACE_P=1, whereas VCS complains that the bind is invalid.

bp_common/syn/Makefile.verilator
```diff
@@ -84,6 +84,11 @@ $(SIM_DIR)/run_simsc: $(SIM_COLLATERAL)
 	-@grep "PASS" $(SIM_LOG) || echo "FAILED" > $(SIM_ERROR)
 	-@grep "STATS" -A 3 $(SIM_LOG) > $(SIM_REPORT)
 
+sim_dump.sc: VV_OPTS += --trace
```
dpetrisko, line 87: Is this actually needed? The verilation happens in dump.  Or is this for the case where you're sim_dump-ing without having built before?

muwyse: I see. I thought it wasn't working without the target, but maybe I missed something. I don't think the latter would work though, right? You need to build in the dumping to the verilation process, I don't think it can be flipped on after the fact at sim time.

dpetrisko: I think if you do
```
make clean
make sim_dump.sc
```

It will automatically build a simsc, and if you've set the VV_OPTS in the sim_dump.sc target, it will pass to the dump.sc to build it with dumping.  A cleaner solution may be

```
sim_dump.sc: build_dump.sc
```

muwyse: okay. should I remove this last commit? I think make build_dump.sc sim.sc works as expected, as you said, so sim_dump.sc isn't needed.

dpetrisko: Can you add this instead of your fix?

sim_dump.sc: build_dump.sc

That'll fix up the dependency at least.  Plus, you need to do build_dump sim_dump for VCS, so might as well have the target in verilator as well, for consistency

muwyse: New issue for this: https://github.com/black-parrot/black-parrot/issues/492

I think we need more than just:
sim_dump.sc: build_dump.sc
to make this work, no? This would build with dump enabled, but wouldn't actually run?

dpetrisko: oh, you're totally right.

The two options are
```
sim_dump.sc: build_dump.sc
sim_dump.sc:
  $(MAKE) sim.sc

and the way you had it:

sim_dump.sc: VV_OPTS += --trace
...
sim_dump.sc: build.sc
```

I think 2 is better, because it doesn't involve recursive make.

PR #495, opened 2020-05-12
ITLB / I$ power optimizations

tanglingshu:
# I-cache power optimization
Bypass I-TLB and tag_mem when possible. In addition, select the necessary bank of data_mem.

#### 1) General Case:
When incoming vaddr tag == previously fetched valid vtag and we’re not doing itlb_fence or itlb_fill, we bypass the i-tlb
If we bypass the i-tlb, then the physical tag is always valid (ptag_v_li) and the physical tag (ptag_li) is previously stored ptag. Note itlb_r_v_lo = 0 when bypass is enabled
#### 2) Special Case 1 (itlb_fence or itlb_fill)
While we’re performing a itlb_fence or itlb_fill, we do not bypass
Furthermore, we invalidate the current vtag that’s going to be stored in the prev_fetch_vtag_r
#### 3) Special Case 2 (itlb_miss_lo)
If a fetch results in a miss, we invalidate the vtag in prev_fetch_vtag_r. Note that itlb_miss_lo is determined on the same cycle as we’re storing prev_fetch_tag

#### 1) General Case:
When incoming vaddr index == previous valid index and we’re not writing to or poisoning tag_mem, we bypass the tag mem
If we’re reading and we bypass the tag mem. Latch_last_read_p=1 will hold onto the previous tag_mem output
#### 2) Special Case 1 (writing to tag mem)
Turn off bypass and we invalidate the previously stored tag_mem address until we read again
#### 3) Special Case 2 (poisoning)
Turn off bypass signal

#### 1) General Case:
If we decided to bypass tag_mem, we can use the hit_index_tl to help select the single data_mem bank that will be used. 
#### 2) Special Case 1 (ptag_i changes while tag_mem under bypass)
To solve this issue, vaddr_vtag and addr_vtag_tl were added. If the vtag changes during bypass, all banks must be enabled

### DIFF
--- a/bp_common/src/v/bp_tlb.v
+++ b/bp_common/src/v/bp_tlb.v
@@ -43,6 +43,35 @@ bsg_dff_reset #(.width_p(vtag_width_p))
    ,.data_i(vtag_i)
    ,.data_o(vtag_r)
   );
+
+// tlb bypass logic
+logic tlb_bypass;
+logic tlb_bypass_r;
+logic tlb_last_read_r;
+logic r_v_lo_bypass_r;
+bp_pte_entry_leaf_s r_entry_bypass_r;
+
+assign tlb_bypass = (vtag_i == vtag_r) & tlb_last_read_r;
+
+bsg_dff_reset #(.width_p(1))
+  tlb_bypass_reg
+  (.clk_i(clk_i)
+   ,.reset_i(reset_i)
+   ,.data_i(tlb_bypass)
+   ,.data_o(tlb_bypass_r)
+  );
+
+bsg_dff_reset_set_clear 
+  #(.width_p(1)
+    ,.clear_over_set_p(1)
+    )
+  tlb_last_read_reg
+  (.clk_i(clk_i)
+   ,.reset_i(reset_i)
+   ,.set_i((v_i & ~w_i))
+   ,.clear_i(~(v_i & ~w_i))
+   ,.data_o(tlb_last_read_r)
+  ); 
  
 bp_pte_entry_leaf_s r_entry, passthrough_entry;
 logic r_v_lo;
@@ -60,16 +89,34 @@ bsg_cam_1r1w_sync
    ,.w_tag_i(vtag_i)
    ,.w_data_i(entry_i)
 
-   ,.r_v_i(v_i & ~w_i)
+   ,.r_v_i(v_i & ~w_i & ~tlb_bypass)
    ,.r_tag_i(vtag_i)
 
    ,.r_data_o(r_entry)
    ,.r_v_o(r_v_lo)
    );
 
+bsg_dff_reset_en_bypass #(.width_p(entry_width_lp))
+  r_entry_bypass_reg
+  (.clk_i(clk_i)
+   ,.reset_i(reset_i)
+   ,.en_i(~tlb_bypass_r)
+   ,.data_i(r_entry)
+   ,.data_o(r_entry_bypass_r)
+  );
+
+bsg_dff_reset_en_bypass #(.width_p(1))
+  r_v_lo_bypass_reg
+  (.clk_i(clk_i)
+   ,.reset_i(reset_i)
+   ,.en_i(~tlb_bypass_r)
+   ,.data_i(r_v_lo)
+   ,.data_o(r_v_lo_bypass_r)
+  );  
+
 assign passthrough_entry = '{ptag: vtag_r, default: '0};
-assign entry_o    = translation_en_i ? r_entry : passthrough_entry;
-assign v_o        = translation_en_i ? r_v_r & r_v_lo : r_v_r;
+assign entry_o    = translation_en_i ? r_entry_bypass_r : passthrough_entry;
+assign v_o        = translation_en_i ? r_v_r & r_v_lo_bypass_r : r_v_r;
 assign miss_v_o   = r_v_r & ~v_o;
 
 endmodule
--- a/bp_common/src/v/bp_tlb.v
+++ b/bp_common/src/v/bp_tlb.v
@@ -48,7 +48,7 @@ bsg_dff_reset #(.width_p(vtag_width_p))
 logic tlb_bypass;
 logic tlb_bypass_r;
 logic tlb_last_read_r;
-logic r_v_lo_bypass_r;
+logic r_entry_bypass_v_r;
 bp_pte_entry_leaf_s r_entry_bypass_r;
 
 assign tlb_bypass = (vtag_i == vtag_r) & tlb_last_read_r;
@@ -106,17 +106,17 @@ bsg_dff_reset_en_bypass #(.width_p(entry_width_lp))
   );
 
 bsg_dff_reset_en_bypass #(.width_p(1))
-  r_v_lo_bypass_reg
+  r_entry_bypass_v_reg
   (.clk_i(clk_i)
    ,.reset_i(reset_i)
    ,.en_i(~tlb_bypass_r)
    ,.data_i(r_v_lo)
-   ,.data_o(r_v_lo_bypass_r)
+   ,.data_o(r_entry_bypass_v_r)
   );  
 
 assign passthrough_entry = '{ptag: vtag_r, default: '0};
 assign entry_o    = translation_en_i ? r_entry_bypass_r : passthrough_entry;
-assign v_o        = translation_en_i ? r_v_r & r_v_lo_bypass_r : r_v_r;
+assign v_o        = translation_en_i ? r_v_r & r_entry_bypass_v_r : r_v_r;
 assign miss_v_o   = r_v_r & ~v_o;
 
 endmodule
--- a/bp_common/src/v/bp_tlb.v
+++ b/bp_common/src/v/bp_tlb.v
@@ -51,7 +51,7 @@ logic tlb_last_read_r;
 logic r_entry_bypass_v_r;
 bp_pte_entry_leaf_s r_entry_bypass_r;
 
-assign tlb_bypass = (vtag_i == vtag_r) & tlb_last_read_r;
+assign tlb_bypass = (vtag_i == vtag_r) & tlb_last_read_r | ~translation_en_i;
 
 bsg_dff_reset #(.width_p(1))
   tlb_bypass_reg
--- a/bp_common/src/v/bp_tlb.v
+++ b/bp_common/src/v/bp_tlb.v
@@ -52,7 +52,7 @@ logic tlb_last_read_r;
 logic r_entry_bypass_v_r;
 bp_pte_entry_leaf_s r_entry_bypass_r;
 
-assign tlb_bypass = (vtag_i == vtag_r) & tlb_last_read_r | ~translation_en_i;
+assign tlb_bypass = ((vtag_i == vtag_r) & tlb_last_read_r) | ~translation_en_i;
 
 bsg_dff_reset #(.width_p(1))
   tlb_bypass_reg
--- a/bp_common/src/v/bp_tlb.v
+++ b/bp_common/src/v/bp_tlb.v
@@ -44,7 +44,8 @@ bsg_dff_reset #(.width_p(vtag_width_p))
    ,.data_o(vtag_r)
   );
 
-// tlb bypass logic
+// Tlb bypass logic bypasses tlb reads if translation is disabled 
+// or if the input vtags on consecutive reads are the same
 logic tlb_bypass;
 logic tlb_bypass_r;
 logic tlb_last_read_r;
@@ -61,18 +62,14 @@ bsg_dff_reset #(.width_p(1))
    ,.data_o(tlb_bypass_r)
   );
 
-bsg_dff_reset_set_clear 
-  #(.width_p(1)
-    ,.clear_over_set_p(1)
-    )
+bsg_dff_reset #(.width_p(1))
   tlb_last_read_reg
   (.clk_i(clk_i)
    ,.reset_i(reset_i)
-   ,.set_i((v_i & ~w_i))
-   ,.clear_i(~(v_i & ~w_i))
+   ,.data_i(v_i & ~w_i)
    ,.data_o(tlb_last_read_r)
-  ); 
- 
+  );
+
 bp_pte_entry_leaf_s r_entry, passthrough_entry;
 logic r_v_lo;
 bsg_cam_1r1w_sync
@@ -96,6 +93,8 @@ bsg_cam_1r1w_sync
    ,.r_v_o(r_v_lo)
    );
 
+// If the tlb is bypassed, r_entry_bypass_r is same as the previous
+// r_entry. Else, r_entry_bypass_r is from CAM's r_data_o
 bsg_dff_reset_en_bypass #(.width_p(entry_width_lp))
   r_entry_bypass_reg
   (.clk_i(clk_i)
--- a/bp_fe/src/v/bp_fe_icache.v
+++ b/bp_fe/src/v/bp_fe_icache.v
@@ -100,8 +100,8 @@ module bp_fe_icache
   assign cache_req_metadata_o = cache_req_metadata_cast_lo;
   
   logic [index_width_lp-1:0]            vaddr_index;
-
-  logic [word_offset_width_lp-1:0] vaddr_offset;
+  logic [vtag_width_p-1:0]              vaddr_vtag;
+  logic [word_offset_width_lp-1:0]      vaddr_offset;
 
   logic [icache_assoc_p-1:0]            way_v_tv_r; // valid bits of each way
   logic [way_id_width_lp-1:0]           way_invalid_index; // first invalid way
@@ -112,7 +112,8 @@ module bp_fe_icache
 
   assign vaddr_index      = vaddr_i[word_offset_width_lp+byte_offset_width_lp+:index_width_lp];
   assign vaddr_offset     = vaddr_i[byte_offset_width_lp+:word_offset_width_lp];
-   
+  assign vaddr_vtag       = vaddr_i[word_offset_width_lp+byte_offset_width_lp+index_width_lp+:vtag_width_p];
+
   // TL stage
   logic v_tl_r;
   logic tl_we;
@@ -132,15 +133,15 @@ module bp_fe_icache
       if (tl_we) begin
         page_offset_tl_r <= vaddr_i[bp_page_offset_width_gp-1:0];
         vaddr_tl_r       <= vaddr_i;
-        fencei_op_tl_r    <= fencei_v_i;
+        fencei_op_tl_r   <= fencei_v_i;
       end
     end
   end
 
   // tag memory
-  logic                                     tag_mem_v_li;
-  logic                                     tag_mem_w_li;
-  logic [index_width_lp-1:0]                tag_mem_addr_li;
+  logic                                                                tag_mem_v_li;
+  logic                                                                tag_mem_w_li;
+  logic [index_width_lp-1:0]                                           tag_mem_addr_li;
   logic [icache_assoc_p-1:0][$bits(bp_coh_states_e)+ptag_width_lp-1:0] tag_mem_data_li;
   logic [icache_assoc_p-1:0][$bits(bp_coh_states_e)+ptag_width_lp-1:0] tag_mem_w_mask_li;
   logic [icache_assoc_p-1:0][$bits(bp_coh_states_e)+ptag_width_lp-1:0] tag_mem_data_lo;
@@ -148,6 +149,7 @@ module bp_fe_icache
   bsg_mem_1rw_sync_mask_write_bit #(
     .width_p(icache_assoc_p*($bits(bp_coh_states_e)+ptag_width_lp))
     ,.els_p(icache_sets_p)
+    ,.latch_last_read_p(1)
   ) tag_mem (
     .clk_i(clk_i)
     ,.reset_i(reset_i)
@@ -199,11 +201,15 @@ module bp_fe_icache
   logic [icache_assoc_p-1:0]       hit_v_tl;
   logic [paddr_width_p-1:0]        addr_tl;
   logic [icache_assoc_p-1:0]       way_v_tl;
+  logic [index_width_lp-1:0]       vaddr_index_tl;
+  logic [vtag_width_p-1:0]         vaddr_vtag_tl;
    
   assign addr_tl = {ptag_i, vaddr_tl_r[0+:bp_page_offset_width_gp]};
-
   assign addr_tag_tl = addr_tl[block_offset_width_lp+index_width_lp+:ptag_width_lp];
   assign addr_word_offset_tl = addr_tl[byte_offset_width_lp+:word_offset_width_lp];
+
+  assign vaddr_index_tl = vaddr_tl_r[word_offset_width_lp+byte_offset_width_lp+:index_width_lp];
+  assign vaddr_vtag_tl = vaddr_tl_r[word_offset_width_lp+byte_offset_width_lp+index_width_lp+:vtag_width_p];
 
   for (genvar i = 0; i < icache_assoc_p; i++) begin: tag_comp_tl
     assign hit_v_tl[i]   = (tag_tl[i] == addr_tag_tl) && (state_tl[i] != e_COH_I);
@@ -218,9 +224,9 @@ module bp_fe_icache
      );
 
   // TV stage
-  logic v_tv_r;
-  logic tv_we;
-  logic uncached_tv_r;
+  logic                                                      v_tv_r;
+  logic                                                      tv_we;
+  logic                                                      uncached_tv_r;
   logic [paddr_width_p-1:0]                                  addr_tv_r;
   logic [vaddr_width_p-1:0]                                  vaddr_tv_r;
   logic [icache_assoc_p-1:0][ptag_width_lp-1:0]              tag_tv_r;
@@ -427,12 +433,39 @@ module bp_fe_icache
 
   // data mem
 
-  logic data_mem_v;
+  logic                      data_mem_v;
+  logic                      data_mem_last_read_r;
+  logic                      data_mem_bypass;
+  logic [icache_assoc_p-1:0] data_mem_bypass_select;
+
+  // data mem bypass bank select logic
+  bsg_dff_reset_set_clear #(
+    .width_p(1)
+    ,.clear_over_set_p(1)
+  ) data_mem_last_read_reg (
+    .clk_i(clk_i)
+    ,.reset_i(reset_i)
+    ,.set_i(tl_we)
+    ,.clear_i(~tl_we)
+    ,.data_o(data_mem_last_read_r)
+  );
+
+  bsg_decode #(
+    .num_out_p(icache_assoc_p)
+  ) data_mem_bank_v_decode (
+    .i(vaddr_offset ^ hit_index_tl)
+    ,.o(data_mem_bypass_select)
+  );
+
+  assign data_mem_bypass = (vaddr_vtag == vaddr_vtag_tl) & (vaddr_index == vaddr_index_tl) & data_mem_last_read_r;
+
   assign data_mem_v = (data_mem_pkt.opcode != e_cache_data_mem_uncached)
     & data_mem_pkt_yumi_o;
 
   assign data_mem_v_li = tl_we
-    ? {icache_assoc_p{1'b1}}
+    ? data_mem_bypass
+      ? data_mem_bypass_select 
+      : {icache_assoc_p{1'b1}}
     : {icache_assoc_p{data_mem_v}};
 
   assign data_mem_w_li = data_mem_pkt_yumi_o
@@ -458,7 +491,23 @@ module bp_fe_icache
   );
 
   // tag_mem
-  assign tag_mem_v_li = tl_we | tag_mem_pkt_yumi_o;
+  logic tag_mem_bypass;
+  logic tag_mem_last_read_r;
+
+  // tag_mem bypass logic
+  bsg_dff_reset_set_clear #(
+    .width_p(1)
+    ,.clear_over_set_p(1)
+  ) tag_mem_last_read_reg (
+    .clk_i(clk_i)
+    ,.reset_i(reset_i)
+    ,.set_i(tl_we)
+    ,.clear_i(tag_mem_w_li)
+    ,.data_o(tag_mem_last_read_r)
+  ); 
+
+  assign tag_mem_bypass = (vaddr_index == vaddr_index_tl) & tag_mem_last_read_r;
+  assign tag_mem_v_li = (tl_we & ~tag_mem_bypass) | tag_mem_pkt_yumi_o;
   assign tag_mem_w_li = ~tl_we & tag_mem_pkt_v_i;
   assign tag_mem_addr_li = tl_we
     ? vaddr_index
--- a/bp_fe/src/v/bp_fe_icache.v
+++ b/bp_fe/src/v/bp_fe_icache.v
@@ -437,25 +437,32 @@ module bp_fe_icache
   logic                      data_mem_last_read_r;
   logic                      data_mem_bypass;
   logic [icache_assoc_p-1:0] data_mem_bypass_select;
+  logic [icache_assoc_p-1:0] vaddr_offset_dec;
 
   // data mem bypass bank select logic
   bsg_dff_reset_set_clear #(
     .width_p(1)
-    ,.clear_over_set_p(1)
-  ) data_mem_last_read_reg (
-    .clk_i(clk_i)
-    ,.reset_i(reset_i)
-    ,.set_i(tl_we)
-    ,.clear_i(~tl_we)
-    ,.data_o(data_mem_last_read_r)
-  );
+    ,.clear_over_set_p(1)) 
+    data_mem_last_read_reg
+    (.clk_i(clk_i)
+     ,.reset_i(reset_i)
+     ,.set_i(tl_we)
+     ,.clear_i(~tl_we)
+     ,.data_o(data_mem_last_read_r)
+    );
 
-  bsg_decode #(
-    .num_out_p(icache_assoc_p)
-  ) data_mem_bank_v_decode (
-    .i(vaddr_offset ^ hit_index_tl)
-    ,.o(data_mem_bypass_select)
-  );
+  bsg_decode #(.num_out_p(icache_assoc_p)) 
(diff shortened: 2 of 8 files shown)


bp_fe/src/v/bp_fe_icache.v
```diff
@@ -145,15 +152,41 @@ module bp_fe_icache
   logic [icache_assoc_p-1:0][$bits(bp_coh_states_e)+ptag_width_lp-1:0] tag_mem_w_mask_li;
   logic [icache_assoc_p-1:0][$bits(bp_coh_states_e)+ptag_width_lp-1:0] tag_mem_data_lo;
 
+  // Pre-TL stage and TL stage tag mem bypass logic assignment
+  assign tag_mem_bypass = (prev_tag_mem_addr_li_r == tag_mem_addr_li) & prev_tag_mem_addr_li_v_r & ~tag_mem_w_li & ~poison_i;
+
+  always_ff @(posedge clk_i) begin
+    if (reset_i) begin
+      prev_tag_mem_addr_li_r    <= 1'b0;
```
dpetrisko, line 160: Use addr_tl_r instead

bp_fe/src/v/bp_fe_icache.v
```diff
@@ -145,15 +152,41 @@ module bp_fe_icache
   logic [icache_assoc_p-1:0][$bits(bp_coh_states_e)+ptag_width_lp-1:0] tag_mem_w_mask_li;
   logic [icache_assoc_p-1:0][$bits(bp_coh_states_e)+ptag_width_lp-1:0] tag_mem_data_lo;
 
+  // Pre-TL stage and TL stage tag mem bypass logic assignment
+  assign tag_mem_bypass = (prev_tag_mem_addr_li_r == tag_mem_addr_li) & prev_tag_mem_addr_li_v_r & ~tag_mem_w_li & ~poison_i;
+
+  always_ff @(posedge clk_i) begin
```
dpetrisko, line 158: Use a bsg_dff_reset_en

bp_fe/src/v/bp_fe_icache.v
```diff
@@ -145,15 +152,41 @@ module bp_fe_icache
   logic [icache_assoc_p-1:0][$bits(bp_coh_states_e)+ptag_width_lp-1:0] tag_mem_w_mask_li;
   logic [icache_assoc_p-1:0][$bits(bp_coh_states_e)+ptag_width_lp-1:0] tag_mem_data_lo;
 
+  // Pre-TL stage and TL stage tag mem bypass logic assignment
+  assign tag_mem_bypass = (prev_tag_mem_addr_li_r == tag_mem_addr_li) & prev_tag_mem_addr_li_v_r & ~tag_mem_w_li & ~poison_i;
```
dpetrisko, line 156: No reason to have ~tag_mem_w_li or ~poison_i, I don't think.  If these conditions are happening, then we're not reading from the mems anyway

bp_fe/src/v/bp_fe_icache.v
```diff
@@ -145,15 +152,41 @@ module bp_fe_icache
...
     .width_p(icache_assoc_p*($bits(bp_coh_states_e)+ptag_width_lp))
     ,.els_p(icache_sets_p)
+    ,.latch_last_read_p(1)
   ) tag_mem (
     .clk_i(clk_i)
     ,.reset_i(reset_i)
     ,.data_i(tag_mem_data_li)
     ,.addr_i(tag_mem_addr_li)
-    ,.v_i(~reset_i & tag_mem_v_li)
+    ,.v_i(~reset_i & tag_mem_v_li & ~tag_mem_bypass)
```
dpetrisko, line 189: combine tag_mem_bypass into tag_mem_v_li

bp_fe/src/v/bp_fe_icache.v
```diff
@@ -425,8 +452,19 @@ module bp_fe_icache
   assign data_mem_v = (data_mem_pkt.opcode != e_cache_data_mem_uncached)
     & data_mem_pkt_yumi_o;
 
+  // select the data_mem bank through tag bypass logic
+  logic [icache_assoc_p-1:0] data_mem_v_li_bank_select;
```
dpetrisko, line 456: data_mem_bypass_select

bp_fe/src/v/bp_fe_icache.v
```diff
@@ -425,8 +452,19 @@ module bp_fe_icache
...
+
+  bsg_decode #(
+    .num_out_p(icache_assoc_p)
+  ) data_mem_bank_v_decode (
+    .i(vaddr_offset ^ hit_index_tl)
+    ,.o(data_mem_v_li_bank_select)
+  );
+
   assign data_mem_v_li = tl_we
```
dpetrisko, line 465:
``` 
 assign data_mem_v_li = tl_we
      ? (tag_mem_bypass_tl_r & tag_mem_bypass)
         ? data_mem_v_li_bank_select
         : {icache_assoc_p{1'b1}}
      : {icache_assoc_p{data_mem_v}};
```

bp_fe/src/v/bp_fe_mem.v
```diff
@@ -83,11 +83,33 @@ assign mem_resp_o     = mem_resp_cast_o;
 logic instr_page_fault_lo, instr_access_fault_lo, icache_miss_lo, itlb_miss_lo;
 
 logic fetch_ready;
+logic prev_fetch_vtag_v_r;
```
dpetrisko, line 86: Move into TLB

bp_fe/src/v/bp_fe_mem.v
```diff
@@ -83,11 +83,33 @@ assign mem_resp_o     = mem_resp_cast_o;
...
 wire itlb_fill_v  = mem_cmd_v_i & (mem_cmd_cast_i.op == e_fe_op_tlb_fill);
 wire fetch_v      = fetch_ready & mem_cmd_v_i & (mem_cmd_cast_i.op == e_fe_op_fetch);
 wire fencei_v     = fetch_ready & mem_cmd_v_i & (mem_cmd_cast_i.op == e_fe_op_icache_fence);
 
+//itlb bypass pre-itlb logic
+wire itlb_vtag_bypass   = (prev_fetch_vtag_r == mem_cmd_cast_i.operands.fetch.vaddr.tag) & prev_fetch_vtag_v_r & ~(itlb_fence_v | itlb_fill_v);
+
+always_ff @(posedge clk_i) begin
+   if (reset_i) begin
+      prev_fetch_vtag_r    <= '0;
```
dpetrisko, line 99: Similar comment to I$ -- the last read is already latched in the tlb

bp_fe/src/v/bp_fe_mem.v
```diff
@@ -83,11 +83,33 @@ assign mem_resp_o     = mem_resp_cast_o;
...
+      prev_fetch_vtag_v_r  <= 1'b0;
+   end
+
+   else if (itlb_fence_v | itlb_fill_v | itlb_miss_lo) begin
```
dpetrisko, line 103: bsg_dff_reset_en

bp_fe/src/v/bp_fe_icache.v
tanglingshu: as in "prev_tag_mem_addr_tl_r"?

tanglingshu: or just "addr_tl_r"

bp_common/src/v/bp_tlb.v
```diff
@@ -32,6 +32,34 @@ bp_pte_entry_leaf_s r_entry, w_entry, passthrough_entry;
 logic [lg_els_lp-1:0] cam_w_addr, cam_r_addr, ram_addr;
 logic                 r_v, w_v, cam_r_v;
 
+// tlb bypass logic
+logic tlb_bypass;
+logic [vtag_width_p-1:0] prev_fetch_vtag_r;
+logic prev_fetch_vtag_v_r;
+
+bsg_dff_reset_en #(.width_p(vtag_width_p))
```
dpetrisko, line 40: reuse miss_addr, make it an enable register

bp_common/src/v/bp_tlb.v
```diff
@@ -32,6 +32,34 @@ bp_pte_entry_leaf_s r_entry, w_entry, passthrough_entry;
...
+  prev_fetch_vtag_reg
+  (.clk_i(clk_i)
+   ,.reset_i(reset_i)
+   ,.en_i(r_v)
+   ,.data_i(vtag_i)
+   ,.data_o(prev_fetch_vtag_r)
+   );
+
+bsg_dff_reset_set_clear #(
```
dpetrisko, line 49: just prev_vtag, not prev_fetch_vtag

bp_common/src/v/bp_tlb.v
```diff
@@ -42,6 +70,8 @@ assign ram_addr   = (w_i)? cam_w_addr : cam_r_addr;
 
 assign passthrough_entry.ptag = miss_vtag_o;
 
+
+
```
dpetrisko, line 74: get rid of random spaces

bp_fe/src/v/bp_fe_icache.v
```diff
@@ -453,7 +480,7 @@ module bp_fe_icache
   );
    
   // tag_mem
-  assign tag_mem_v_li = tl_we | tag_mem_pkt_yumi_o;
+  assign tag_mem_v_li = (tl_we | tag_mem_pkt_yumi_o) & ~tag_mem_bypass;
```
dpetrisko, line 483: Should be (tl_we & ~tag_mem_bypass) | tag_mem_pkt_yumi_o.  Otherwise, the bypass will suppress an unrelated fill

bp_fe/src/v/bp_fe_icache.v
```diff
@@ -425,8 +440,20 @@ module bp_fe_icache
...
+  bsg_decode #(
+    .num_out_p(icache_assoc_p)
+  ) data_mem_bank_v_decode (
+    .i(vaddr_offset ^ hit_index_tl)
+    ,.o(data_mem_bypass_select)
+  );
+
   assign data_mem_v_li = tl_we
-    ? {icache_assoc_p{1'b1}}
+    ? (tag_mem_bypass_tl_r & tag_mem_bypass & (vaddr_vtag == addr_vtag_tl))
```
dpetrisko, line 454: I don't understand this line.  If we bypassed and we're going to bypass?

bp_fe/src/v/bp_fe_icache.v
```diff
@@ -145,9 +152,23 @@ module bp_fe_icache
   logic [icache_assoc_p-1:0][$bits(bp_coh_states_e)+ptag_width_lp-1:0] tag_mem_w_mask_li;
   logic [icache_assoc_p-1:0][$bits(bp_coh_states_e)+ptag_width_lp-1:0] tag_mem_data_lo;
 
+  // Pre-TL stage and TL stage tag mem bypass logic assignment
+  assign prev_tag_mem_addr = vaddr_tl_r[word_offset_width_lp+byte_offset_width_lp+:index_width_lp];
+  assign tag_mem_bypass = (prev_tag_mem_addr == tag_mem_addr_li) & ~tag_mem_w_li & ~poison_i;
```
dpetrisko, line 157: I think only bypass if the prev read was valid? Otherwise if the previous read was poisoned or filled, we would not read again

dpetrisko: Basically, there are two separate conditions flopped here:
- Whether to bypass (used to determine read/not read in TL_n)
- Whether we did bypass (used to determine what to mux in TV_n)

bp_common/src/v/bp_tlb.v
```diff
@@ -106,13 +136,14 @@ bsg_cam_1r1w
...
     ,.els_p(dtlb_els_p)
+    ,.latch_last_read_p(1)
   )
   entry_ram
   (.clk_i(clk_i)
    ,.reset_i(reset_i)
    ,.data_i(w_entry)
    ,.addr_i(ram_addr)
-   ,.v_i(cam_r_v | w_v)
+   ,.v_i((cam_r_v | w_v) & ~tlb_bypass)
```
dpetrisko, line 146: w_v | (cam_r_v & ~tlb_bypass).  Otherwise, we will bypass a write which makes no sense.

bp_fe/src/v/bp_fe_icache.v
tanglingshu: You're right. I think because previously I set 

"tag_mem_v_li = (tl_we | tag_mem_pkt_yumi_o) & ~tag_mem_bypass;"

I suppressed an unrelated fill, thus I added ~tag_mem_w_li & ~poison_i to counter that. It should be solved now

tanglingshu: You're correct. Previously I did
assign tag_mem_v_li = (tl_we | tag_mem_pkt_yumi_o) & ~tag_mem_bypass; 
which suppressed some fill. This has been solved by creating a valid bit for prev_tag_mem_addr

bp_fe/src/v/bp_fe_icache.v
tanglingshu: I guess there are two ways to go about it. Method 1 is the current implementation

1) The "prev_tag_mem_addr_v_r" register is set on "tl_we (read)" and cleared on "tag_mem_w_li (write)". With this implementation, I bypass tag_mem more often. There are times where we don't read or write to the tag_mem, and this method allows me to bypass tag_mem during that period. 

For example:

a) tl_we=1, I'm reading vaddr_index "0xF"
b) prev_tag_mem_addr_v_r <= 1
c) I'm reading vaddr_index "0xF" again, tag_mem_bypass is set
d) goes on for a few cycle, I'm bypassing tag_mem
e) tl_we=0, but the vaddr_index remains "0xF". Here, prev_tag_mem_addr_v_r is still set because we haven't done a write yet (i,e,. tag_mem_w_li = 0)
f) for the subsequent cycles where tl_we=0, tag_mem_w_li=0, vaddr_index="0xF", I'm still bypassing the tag_mem because 1) vaddr_index is the same and 2) my prev_tag_mem_addr_v_r is set

Now suppose, tl_we=1 again, with vaddr_index = "0xF". I immediately perform tag_mem_bypass instead of bypassing a cycle later. However, the "hit_index_tl" doesn't come out until the tl stage, so I have to wait for a cycle to get the correct "hit_index_tl". 

here's the signal diagram: 
Method_1

That's why "tag_mem_bypass_tl_r" & "tag_mem_bypass" are together.

Method 2 would be instead of clearing "prev_tag_mem_addr_v_r" on "tag_mem_w_li (write)", I can clear it on "~tl_we (not reading)". This way, my "tag_mem_bypass" will always occur a cycle after "tl_we", guaranteeing that I have the correct "hit_index_tl" But with this implementation, I'm not going to bypass tag_mem as much as I would with method 1.

I know this explanation is a bit wordy, but let me know if this makes sense and what do you think is the best way to go about this.

bp_fe/src/v/bp_fe_icache.v
```diff
@@ -198,16 +231,16 @@ module bp_fe_icache
...
   logic                        hit_tl;
   logic [paddr_width_p-1:0]    addr_tl;
-  logic [icache_assoc_p-1:0]  way_v_tl;
-   
-  assign addr_tl = {ptag_i, vaddr_tl_r[0+:bp_page_offset_width_gp]};
+  logic [icache_assoc_p-1:0]   way_v_tl;
 
+  assign addr_tl = {ptag_i, vaddr_tl_r[0+:bp_page_offset_width_gp]};
+  assign addr_vtag_tl = vaddr_tl_r[word_offset_width_lp+byte_offset_width_lp+index_width_lp+:vtag_width_p];
   assign addr_tag_tl = addr_tl[block_offset_width_lp+index_width_lp+:ptag_width_lp];
```
dpetrisko, line 238: Change to addr_ptag, for clarity

bp_fe/src/v/bp_fe_icache.v
dpetrisko: >However, the "hit_index_tl" doesn't come out until the tl stage, so I have to wait for a cycle to get the correct "hit_index_tl".

Instead you can save prev_hit_index_tl in a bsg_dff_en_bypass, and use it directly.  This will avoid the bubble

bp_fe/src/v/bp_fe_icache.v
```diff
@@ -145,9 +153,34 @@ module bp_fe_icache
...
+    );
+
+  bsg_dff_reset_set_clear #(
+    .width_p(1)
+    ,.clear_over_set_p(1)
+    )
+    prev_tag_mem_addr_valid
+    (.clk_i(clk_i)
+     ,.reset_i(reset_i)
+     ,.set_i(tl_we)
```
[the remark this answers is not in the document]
dpetrisko: Yeah, this looks like a typo since above you say "The "prev_tag_mem_addr_v_r" register is set on "tl_we (read)" and cleared on "tag_mem_w_li (write)"

bp_fe/src/v/bp_fe_icache.v
```diff
@@ -145,9 +153,34 @@ module bp_fe_icache
...
+    ,.reset_i(reset_i)
+    ,.data_i(tag_mem_bypass)
+    ,.data_o(tag_mem_bypass_tl_r)
+    );
+
+  bsg_dff_reset_set_clear #(
+    .width_p(1)
+    ,.clear_over_set_p(1)
+    )
+    prev_tag_mem_addr_valid
```
dpetrisko, line 172: nit: prev_tag_mem_addr_v_reg

bp_fe/src/v/bp_fe_icache.v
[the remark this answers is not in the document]
tanglingshu: I think it boils down to I either do:
a)   clear on ~tl_we (don't read) and bank bypass select on tag_mem_bypass
```
bsg_dff_reset_set_clear #(
    .width_p(1)
    ,.clear_over_set_p(1)
    )
    prev_tag_mem_addr_valid
    (.clk_i(clk_i)
     ,.reset_i(reset_i)
     ,.set_i(tl_we)
     ,.clear_i(~tl_we)
     ,.data_o(prev_tag_mem_addr_v_r)
     );
...
  assign data_mem_v_li = tl_we
    ? (tag_mem_bypass & (vaddr_vtag == addr_ptag_tl))
      ? data_mem_bypass_select 
      : {icache_assoc_p{1'b1}}
    : {icache_assoc_p{data_mem_v}};

```

or 

b) clear on tag_mem_w_li (write) and bank bypass select on tag_mem_bypass and tag_mem_bypass_tl_r

```
bsg_dff_reset_set_clear #(
    .width_p(1)
    ,.clear_over_set_p(1)
    )
    prev_tag_mem_addr_valid
    (.clk_i(clk_i)
     ,.reset_i(reset_i)
     ,.set_i(tl_we)
     ,.clear_i(tag_mem_w_li)
     ,.data_o(prev_tag_mem_addr_v_r)
     );
...
  assign data_mem_v_li = tl_we
    ? (tag_mem_bypass & tag_mem_bypass_tl_r & (vaddr_vtag == addr_ptag_tl))
      ? data_mem_bypass_select 
      : {icache_assoc_p{1'b1}}
    : {icache_assoc_p{data_mem_v}};

```

But, let me try method b with prev_hit_index_tl

bp_fe/src/v/bp_fe_icache.v
tanglingshu: If this new "hit_index_tl" is different than "prev_hit_index_tl", then using "prev_hit_index_tl" would give me an undesired result. 

image

In this picture, at the marked magenta vertical line is where I would read the "prev_hit_index_tl" which gives me a 0. In reality, I should either read "hit_index_tl" at the line (which gives me a 1) or read "prev_hit_index_tl" a cycle after which also gives me a hit index of 1 (but this way I'd miss out one a cycle where I could bypass data_mem)

dpetrisko: bsg_dff_en_bypass will use the result the same cycle

bp_fe/src/v/bp_fe_icache.v
dpetrisko: reopening

bp_fe/src/v/bp_fe_icache.v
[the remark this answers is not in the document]
dpetrisko: I don't see any reason for having tag_mem_bypass_tl_r in method b.  (I assume addr_ptag_tl is supposed to be addr_vtag there)

bp_fe/src/v/bp_fe_icache.v
```diff
@@ -425,8 +452,20 @@ module bp_fe_icache
   assign data_mem_v = (data_mem_pkt.opcode != e_cache_data_mem_uncached)
     & data_mem_pkt_yumi_o;
 
+  // select the data_mem bank through tag bypass logic
+  logic [icache_assoc_p-1:0] data_mem_bypass_select;
+
+  bsg_decode #(
+    .num_out_p(icache_assoc_p)
+  ) data_mem_bank_v_decode (
+    .i(vaddr_offset ^ hit_index_tl)
```
dpetrisko, line 461: This has to be prev_hit_index_tl, otherwise the critical path here is going through the tag compare

dpetrisko: prev_hit_index_tl == hit_index_tv_r

bp_fe/src/v/bp_fe_icache.v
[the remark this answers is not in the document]
dpetrisko: I think we need to go over the schematic here, otherwise we're going to keep going in circles.  Can you update the diagrams on the google doc to the latest, and include the data mem bypass?

bp_common/src/v/bp_tlb.v
```diff
@@ -60,16 +89,34 @@ bsg_cam_1r1w_sync
...
+  r_entry_bypass_reg
+  (.clk_i(clk_i)
+   ,.reset_i(reset_i)
+   ,.en_i(~tlb_bypass_r)
+   ,.data_i(r_entry)
+   ,.data_o(r_entry_bypass_r)
+  );
+
+bsg_dff_reset_en_bypass #(.width_p(1))
+  r_v_lo_bypass_reg
```
dpetrisko, line 109: r_v_lo_bypass_reg -> r_entry_bypass_v_reg
r_v_lo_bypass_r -> r_entry_bypass_v_r

dpetrisko, line 109: r_v_lo_bypass_reg -> r_entry_bypass_v_reg
r_v_lo_bypass_r -> r_entry_bypass_v_r

bp_fe/src/v/bp_fe_icache.v
```diff
@@ -421,12 +426,39 @@ module bp_fe_icache
...
+  );
+
+  bsg_decode #(
+    .num_out_p(icache_assoc_p)
+  ) data_mem_bank_v_decode (
+    .i(vaddr_offset ^ hit_index_tl)
+    ,.o(data_mem_bypass_select)
+  );
+
+  assign data_mem_bypass = (vaddr_vtag == vaddr_vtag_tl) & (vaddr_index == vaddr_index_tl) & data_mem_last_read_r;
```
dpetrisko, line 453: This will alias if the TLB mappings change.  Either need to check ptag_i == ptag_tv_r, or clear the data_m em_last_read_reg on TLB modifications.

tanglingshu:
> Need to rebase on latest fe_dev to get the one-hot caches. Let me know if you need clarification on those changes.
> 
> Otherwise, LGTM, pending timing path analysis

May you clarify on what you mean by "need to rebase on latest fe_dev to get the one-hot caches"? What action is required on my end? Thanks

### **Synthesis Time Path Analysis:**
Technology Node: gf12
Operating Conditions: SSPG_0P72V_M40C
Freq/Per: 1GHz/1000ps
Library: IN14LPP_SC7P5T_84CPP_BASE_SSC16L_SSPG_0P72V_M40C
Wire Load Model Mode: top

**Path affected by tlb bypass changes in Icache:**
pc_gen_stage_reg -> fe/mem/itlb/vtag_i -> fe/mem/itlb/tlb_bypass_reg 
total arrival time = 956.26ps
total max timing slack=0.19ps
note: tlb_bypass_reg adds about 21.14ps delay

**Path affected by tlb bypass changes in dcache:**
dtlb/cam/r_tag_reg -> cam_tag_array -> one_hot_mem -> r_entry_bypass_reg(bsg_chip_bsg_dff_reset_en_bypass) -> dtlb/entry_o -> dcache_ptag -> dcache/wbuf 
total arrival time = 927.30ps
total max timing slack=1.05ps
note: r_entry_bypass_reg adds about 36.88ps delay

**Path affected by icache tag/data bypass logic:**
None shown in timing report with max path of 1000

Other notable timing paths unrelated to this change but might be helpful for future reference:
- icache: data_mem_pkt_way_r_reg -> read_mux_butterfly -> icache_UCE -> cmd_fifo (932.0ps/slack=0ps)
- hit_index_tv_r_reg -> data_set_select_mux -> final_data_mux -> pc_gen/instr_scan -> icache_vaddr_i -> icache/data_mem (767.24ps/slack=0.01ps, large setup time)
- fe_cmd_fifo (unhardened fifo) -> pc_gen -> bp_fe_bht (945.62ps/slack=0.1ps)

bp_fe/src/v/bp_fe_icache.v
tanglingshu: Just to clarify, are you looking for something like:
```
assign data_mem_bypass = (ptag_i == ptag_tv_r) & (vaddr_vtag == vaddr_vtag_tl) &(vaddr_index == vaddr_index_tl) & data_mem_last_read_r;
```
Out of curiosity, can the tlb mapping change during a bypass or a read? If I'm reading from the icache, doesn't the tlb mapping change stop me from reading from icache? (thus clearing data_mem_last_read_reg)

tanglingshu: If TLB mapping changes, is there a way to know there will be a mapping change before the TL stage? By the time that i know ptag_i is different (which is during the TL stage), it's already too late to enable the correct data_mem bank.

dpetrisko: Aha, good point

dpetrisko: Actually, it doesn’t matter if the virtual mappings have changed. As long as the physical tags match, it is still a hit

tanglingshu: I see your point and I thought about it initially as well. As long as ptag_i matches any of the tag_tl, then there is a hit. Then I can use the hit_index_tl and addr offset to decide which bank to enable. 

Whenever ptag_i changes, it changes in the TL stage so it's already too late for me to enable the correct bank. My understanding is that the vtag and ptag are 1-to-1 mapping. So, the only way ptag can change is when vtag changes. To go about ptag_i changing problem, I said that if there is a change vtag, then enable all of the data_mem bank. 

If however a vtag results in a tlb miss, I assume that the output of the data_mem is invalid anyways, so it doesn't matter how I select the data_mem banks.

Do you see another way to select the data_mem banks with ptag_i?

dpetrisko:
>May you clarify on what you mean by "need to rebase on latest fe_dev to get the one-hot caches"? What action is required on my end? Thanks

git rebase fe_dev

This will apply your changes on top of the latest fe_dev version

bp_fe/src/v/bp_fe_mem.v
```diff
@@ -108,8 +134,26 @@ bp_tlb
    ,.entry_o(itlb_r_entry)
    );
 
-wire [ptag_width_p-1:0] ptag_li     = itlb_r_entry.ptag;
-wire                    ptag_v_li   = itlb_r_v_lo;
+//itlb bypass post-itlb logic
```
dpetrisko, line 137: Isn't all of this logic in the tlb now?

bp_fe/src/v/bp_fe_icache.v
```diff
@@ -427,12 +433,46 @@ module bp_fe_icache
...
+    input_offset_decode 
+    (.i(vaddr_offset)
+     ,.o(vaddr_offset_dec)
+    );
+
+  bsg_adder_one_hot #(.width_p(icache_assoc_p))
+    data_mem_bank_select_adder
+    (.a_i(hit_v_tl)
+     ,.b_i(vaddr_offset_dec)
+     ,.o(data_mem_bypass_select)
```
dpetrisko, line 464: If we're computing data_mem_bypass_select here, seems like we should just save the result and use the selection for the data_mux in tv

dpetrisko: In which case, this is no longer a "bypass select", but bank_select_tl (and bank_select_tv_r)

bp_fe/src/v/bp_fe_mem.v
tanglingshu: Must've been a mistake during rebase, let me clean that up. Sorry for that

dpetrisko: Oh cool.  Thought I missed it during previous review!

bp_fe/src/v/bp_fe_icache.v
tanglingshu: Is the idea of using the bank selection in the tv stage to help the critical path?

dpetrisko: No, we still use it for the bypass.  My point is that we do this same decode and one-hot addition again in TV as input to the data mux.  We might as well remove the redundant logic

tanglingshu: yea yea, I get what you're saying. Sorry, had to re-read it to understand that. Let me try that :)

tanglingshu: The original _ld_data_way_select_ is done through a "bsg_adder_one_hot" with _hit_v_tv_r_ and _addr_word_offset_dec_tv_r_. However, I don't think _ld_data_way_select_ is equivalent to _bank_select_tv_r_. 

That logic would make sense if _bank_select_tl (aka. data_mem_bypass_select)_ is generated from _hit_v_tl_ and **_vaddr_offset_dec_tl_**. But the above logic is made from _hit_v_tl_ and _vaddr_offset_dec_. The _vaddr_offset_dec_ is from the input stage.

tanglingshu: Though I think you may still make it work. _data_mem_bypass_select_ basically becomes _bank_select_ and I'll have to put it through 2 stages of dffs (ie. _bank_select_tl_r_ and then _bank_select_tv_r_). I think that might work, but not 100% sure, gotta test that.

dpetrisko: No you're right.  The thought doesn't work and we need to do two additions.  It's unfortunate, but the fact they're one-hot makes it bearable

bp_common/src/v/bp_tlb.v
```diff
@@ -43,6 +43,35 @@ bsg_dff_reset #(.width_p(vtag_width_p))
...
   );
+
+// tlb bypass logic
+logic tlb_bypass;
+logic tlb_bypass_r;
+logic tlb_last_read_r;
+logic r_entry_bypass_v_r;
+bp_pte_entry_leaf_s r_entry_bypass_r;
+
+assign tlb_bypass = (vtag_i == vtag_r) & tlb_last_read_r | ~translation_en_i;
```
dpetrisko, line 54: Use parentheses to make OOO clear: ((vtag_i == vtag_r) & tlb_last_read_r) | ~translation_en_i;

bp_common/src/v/bp_tlb.v
```diff
@@ -43,6 +43,35 @@ bsg_dff_reset #(.width_p(vtag_width_p))
...
+
+bsg_dff_reset_set_clear 
+  #(.width_p(1)
+    ,.clear_over_set_p(1)
+    )
+  tlb_last_read_reg
+  (.clk_i(clk_i)
+   ,.reset_i(reset_i)
+   ,.set_i((v_i & ~w_i))
+   ,.clear_i(~(v_i & ~w_i))
```
dpetrisko, line 72: If we only are bypassing on the cycle after a read, we can just use a bsg_dff.

bp_common/src/v/bp_tlb.v
```diff
@@ -60,16 +89,34 @@ bsg_cam_1r1w_sync
...
 
    ,.r_data_o(r_entry)
    ,.r_v_o(r_v_lo)
    );
 
+bsg_dff_reset_en_bypass #(.width_p(entry_width_lp))
+  r_entry_bypass_reg
+  (.clk_i(clk_i)
+   ,.reset_i(reset_i)
+   ,.en_i(~tlb_bypass_r)
```
dpetrisko, line 103: Again, if we can only bypass on consecutive reads, might as well be a dff

bp_common/src/v/bp_tlb.v
```diff
@@ -60,16 +89,34 @@ bsg_cam_1r1w_sync
...
+   ,.data_i(r_entry)
+   ,.data_o(r_entry_bypass_r)
+  );
+
+bsg_dff_reset_en_bypass #(.width_p(1))
+  r_entry_bypass_v_reg
+  (.clk_i(clk_i)
```
dpetrisko, line 110: same here

bp_common/src/v/bp_tlb.v
tanglingshu: If I understood you correctly, if I use a regular bsg_dff_reset_en here, wouldn't I still need a mux to select between _r_entry_ (output of CAM's _r_data_o_) and the bypassed version _r_entry_bypass_r_? If _r_entry_bypass_r_ is only placed through bsg_dff_reset_en, then it's basically takes two cycles (1 cycle for _r_entry_ and 1 more _r_entry_r_ (aka _r_entry_bypass_r_))

Unless there is a latch_last_read_p with this CAM then I wouldn't need any of these DFFs lol

bp_common/src/v/bp_tlb.v
tanglingshu: same as above

bp_common/src/v/bp_tlb.v
dpetrisko: oh, I thought you were muxing below.  But it's just between passthrough no not passthrough Yep, agree this is fine

latch_last_read is _weird_ with 1r1w

dpetrisko:
Great work! Thanks for your patience on the reviews :)

PR #501, opened 2020-05-19
Updating TLB to use latest bsg_cam module

dpetrisko:
This module has better critical path through use of one-hot signals and a synchronous tag, async data lookup.

There's also a change to fix buffering in the bp_cfg module without relying on an always ready invariant in the mem adapters


farzamgl:
I have a few questions about the new replacement module:

- Why do we update the LRU bits on allocations? As far as I know in the TLB every allocation is followed by a read in the replayed instruction so that causes 2 LRU updated on every TLB miss.
- I noticed that there is a mux on the lru_way if we have empty entries in the CAM, but if I remember correctly, the pseudo-tree-LRU algorithm points to an empty entry if there is one so there might not be a need for the mux.

dpetrisko:
> Why do we update the LRU bits on allocations? As far as I know in the TLB every allocation is followed by a read in the replayed instruction so that causes 2 LRU updated on every TLB miss.

This is true for BlackParrot, but not in the general case. In the general case, it's dangerous not to update on allocate because multiple sequential allocates will overwrite each other. https://github.com/bespoke-silicon-group/basejump_stl/pull/289 

I believe the LRU updates are idempotent, though, so it shouldn't matter for BlackParrot.

> the pseudo-tree-LRU algorithm points to an empty entry

https://github.com/bespoke-silicon-group/basejump_stl/blob/master/bsg_misc/bsg_lru_pseudo_tree_decode.v
https://github.com/bespoke-silicon-group/basejump_stl/blob/master/bsg_misc/bsg_lru_pseudo_tree_encode.v

The pseudo tree modules don't have any knowledge of emptiness

farzamgl:
> The pseudo tree modules don't have any knowledge of emptiness

No, but I think, when starting from scratch, if there are ways that have not been accessed before, it'll not set the LRU way to a previously accessed way.

dpetrisko:
For an 8-way pseudo-LRU, If you fill all the ways except for 1 and then have an access pattern of 3 or more, you're not guaranteed to put in the empty way, because the pseudo-LRU has fewer state bits than needed.  For true LRU, you're right

farzamgl:
Oh I see. I was just looking at the access pattern of filling the TLB with 1 access to each entry.

PR #502, opened 2020-05-20
Added Linux, Buildroot, and Opensbi build flow

farzamgl:
This PR adds a Linux build flow with OpenSBI as the firmware and a Busybox based rootfs image generated by Buildroot.


bp_common/test/cfg/linux_defconfig
```diff
@@ -0,0 +1,3005 @@
+#
```
dpetrisko, line 1: can we not commit this? autogenerated?

external/Makefile.tools
```diff
@@ -48,6 +48,7 @@ gnu_build:
 	cd $(gnu_dir); \
 		./configure --prefix=$(BP_EXTERNAL_DIR) --with-arch=rv64ima --with-abi=lp64 --with-cmodel=medany 
 	$(MAKE) -C $(gnu_dir)
+	$(MAKE) -C $(gnu_dir) linux
```
dpetrisko, line 51: Does this definitely generate both toolchains?

external/Makefile.tools
```diff
@@ -48,6 +48,7 @@ gnu_build:
...
 	$(MAKE) -C $(gnu_dir) install
 	-cp $(BP_EXTERNAL_DIR)/riscv64-unknown-elf-dramfs/bin/dramfs_mklfs $(BP_BIN_DIR)
```
dpetrisko, line 53: Can you remove this the - on this cp while we're here?

Makefile.common
```diff
@@ -49,6 +49,5 @@ export PYTHON    ?= python
 #  SystemC paths are needed for verilator-driven builds
 export SYSTEMC_INCLUDE := $(BP_EXTERNAL_DIR)/include
```
dpetrisko, line 50: Does it work to move the SYSTEMC stuff to the Makefile.common also?

bp_common/test/Makefile.tests
```diff
@@ -45,6 +46,9 @@ spec_build:
 riscvdv_build:
 	$(MAKE) -C $(riscvdv_dir)
 
+linux_build: $(fw_payload)
```
dpetrisko, line 49: Need to add to a suite folder or the build system won't be able to find it. Just linux/linux.riscv I guess?

bp_common/test/cfg/linux_defconfig
farzamgl: Not fully autogenerated, it's generated by Linux's configuration menu.

external/Makefile.tools
farzamgl: Yeah I tested it from scratch.

bp_common/test/Makefile.tests
farzamgl: Which part of the build system?

Makefile.common
farzamgl: I gotta check. Is the related problem is building the Verilator itself or using Verilator?

bp_common/test/Makefile.tests
dpetrisko: make sim.v

Makefile.common
dpetrisko: Using verilator

bp_common/test/cfg/linux_defconfig
dpetrisko: Hm.  Do the other SDKs (Ariane, freedom) save this output, or do they generate it in the build somehow?

Makefile.common
farzamgl: Looks fine either way actually.

farzamgl:
> Awesome!! I assume you've tested on dromajo, have you tried it on fpga?

Hasn't been tested on BP yet (sim or FPGA). Only on Dromajo.

bp_common/test/Makefile.tests
farzamgl: Oh I see. We are categorizing the tests. Do I need to just need to change th cp destination or should I also specify somewhere that which suite linux belongs to?

bp_common/test/cfg/linux_defconfig
farzamgl: Most of them have a starting config file. I can clean it up later to see which flags are necessary to input to Linux generator and which ones are generated by default but it's out of the bounds of my knowledge right now.

bp_common/test/Makefile.tests
dpetrisko: You can just change the cp destination.  If you add to bp_common/test/Makefile.frag, it can autorebuild, but I think it's probably better to require explicit builds of linux since it's so intensive

PR #510, opened 2020-05-21
replace |& operator with 2>&1 | for greater compatibility

### LINKED ISSUE #504: Makefile.verilator depends on Bash 4
The `|&` operator, which is equivalent to `2>&1 |` is used in Makefile.verilator, but is only supported by Bash 4 and later. By default, many systems link the dash shell to /bin/sh, which apparently does not necessarily support the `|&` operator. This only manifests if /bin/sh points to something that doesn't support the operator, for example, dash, which appears to be the default on my system (Ubuntu on WSL / Windows 10).

code: https://github.com/black-parrot/black-parrot/blob/master/bp_common/syn/Makefile.verilator#L50

Question: should this operator be removed and use `2>&1 |` instead, or are there other features from Bash 4 that we use, and thus want to make Bash 4 (or equivalent) an explicit requirement?

Note: it is easy for me to reconfigure my system to point /bin/sh at bash, but this may not be true for all users, especially if they are running on managed machines they do not have full control over.

muwyse:
closes #504


dpetrisko:
Is this the only place it shows up?

muwyse:
That was only use I found. I did not find it in: ME, FE, BE, Top, or rest of bp_common directories. I did not check external.

PR #516, opened 2020-05-25
Adding citation information

dpetrisko:
IEEE micro paper is in early access so it can officially be cited as the canonical BlackParrot paper


taylor-bsg:
I would give the more conventional citation, authors, month, year. it looks like it will be July/August 2020.

https://www.sigarch.org/call-contributions/ieee-micro-special-issue-on-agile-and-open-source-hardware/

dpetrisko:
Hm, this is the citation that IEEE itself gives (https://ieeexplore.ieee.org/document/9097398) as a placeholder until it's actually published.  The real citation needs page numbers, months...

taylor-bsg:
Page numbers do not matter. IEEE micro publishes every two months, so each
issue is labeled with a month pair. So we have everything we need.

dpetrisko:
Fixed!

taylor-bsg:
Great!

PR #519, opened 2020-05-26
Adding CCE loopback module

dpetrisko:
Basically, the issue here is that local requests to in-core configuration registers will freeze the system if they don't correspond to a sink (cfg, clint, cache, etc.).

Instead, we can use this lightweight loopback module which just forwards back the request to the sender.  This makes a more robust system, since we don't have error responses in the bedrock network.

### DIFF
--- a/bp_me/src/v/cce/bp_cce_loopback.v
+++ b/bp_me/src/v/cce/bp_cce_loopback.v
@@ -1,0 +1,53 @@
+
+//
+// This module is an active tie-off. That is, requests to this module will return the header
+//   with a zero payload. This is useful to not stall the network in the case of an erroneous
+//   address, or prevent deadlock at network boundaries
+module bp_cce_loopback
+  import bp_common_pkg::*;
+  import bp_common_aviary_pkg::*;
+  import bp_cce_pkg::*;
+  import bp_common_cfg_link_pkg::*;
+  import bp_me_pkg::*;
+  #(parameter bp_params_e bp_params_p = e_bp_inv_cfg
+    `declare_bp_proc_params(bp_params_p)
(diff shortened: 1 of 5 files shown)


muwyse:
So this is only for softcore? By local in-core configuration registers, do you mean CSR's? Are these not stored in the core, but actually stored in memory? Or, are you referring to accesses that use a bad address in general, i.e., one that isn't located in any of the defined memory regions (clint, cfg, etc.), and so they just get lost in the system somewhere?

If the header is simply copied from cmd to response, that doesn't alter the destination or source fields, so won't a message in the loopback be routed to the incorrect place indefinitely? Do source and destination need to be swapped so the response message is properly routed back to the command source?

Otherwise, LGTM; makes sense.

dpetrisko:
> So this is only for softcore? 

Also for processor (perhaps more importantly). The consequence of a bad address in a single tile could be complete deadlock on the noc.

> By local in-core configuration registers, do you mean CSR's? Are these not stored in the core, but actually stored in memory? Or, are you referring to accesses that use a bad address in general, i.e., one that isn't located in any of the defined memory regions (clint, cfg, etc.), and so they just get lost in the system somewhere?

The latter.  Basically, addresses that are routed to a tile must be responded to.  But we decode the address and route to a specific device on the tile, so if no device matches, it will just hang.

> If the header is simply copied from cmd to response, that doesn't alter the destination or source fields, so won't a message in the loopback be routed to the incorrect place indefinitely? Do source and destination need to be swapped so the response message is properly routed back to the command source?

That is true if we don't deserialize, but these are full messages.  So the flow here is really 
1) LCE sends bad address in lce_req
2) CCE sends mem_cmd
3) CCE receives loopback mem_resp
4) CCE sends lce_cmd to LCE id in payload

PR #520, opened 2020-05-26
Disabling linux login prompt and adding test.sh

farzamgl:
This PR disables the Linux login prompt and automatically runs a script and powers off.  This allows us to run parallel simulations on the final binary since it's a finite standalone test.


farzamgl:
@dpetrisko this is a quick way I thought that we can fix the problem. Open to ideas to make it more polished though.

dpetrisko:
Oh, can you add parallel Linux regression CI script and a manual job in Gitlab for this? I can also do it tomorrow

farzamgl:
> Oh, can you add parallel Linux regression CI script and a manual job in Gitlab for this? I can also do it tomorrow

I think the Linux cosim doesn't work on sw_dev right now. We have to make a separate PR for that after updating it.

PR #522, opened 2020-05-27
UCE FSM modification for multicycle fill/evict

Songchun-Li:
UCE FSM modification for multicycle fill/evict
1. States for following func are changed:
    cache flush with dirty data (e_flush_write)
    cache miss load (e_read_req) 
    cache miss load with dirty data replacement (e_writeback_read_req, e_writeback_write_req)
2. 2 counters are added, fill_cnt and mem_cmd_cnt, to track the progress of multicycle fill/evict
3. A new register, mem_cmd_done_r is added. In multicycle fillm, this 1-bit reg indicates whether all the mem_cmd are sent. Together with this register, mem_cmds can be sent as long as L2 is ready, avoiding potential latency.

### DIFF
--- a/bp_me/src/v/cce/bp_uce.v
+++ b/bp_me/src/v/cce/bp_uce.v
@@ -217,40 +217,44 @@ module bp_uce
   end
 
   logic fill_up, fill_cnt_hit_target, mem_cmd_up, mem_cmd_cnt_hit_target;
-  logic one_mem_cmd;
-  assign one_mem_cmd = (block_width_p == fill_width_p);
+  if (fill_width_p == block_width_p) begin
+    assign fill_cnt = 1'b0;
+    assign mem_cmd_cnt = 1'b0;
+    assign fill_cnt_hit_target = 1'b1;
+    assign mem_cmd_cnt_hit_target = 1'b1;
+  end else begin
+    bsg_counter_clear_up #(
+      .max_val_p(block_size_in_fill_lp-1)
+      ,.init_val_p(0)
+      ,.disable_overflow_warning_p(1)
+      )
+    fill_counter
+      (.clk_i(clk_i)
+      ,.reset_i(reset_i)
 
-  bsg_counter_clear_up #(
-    .max_val_p(block_size_in_fill_lp-1)
-    ,.init_val_p(0)
-    ,.disable_overflow_warning_p(1)
-    )
-  fill_counter
-    (.clk_i(clk_i)
-    ,.reset_i(reset_i)
+      ,.clear_i('0)
+      ,.up_i(fill_up)
 
-    ,.clear_i('0)
-    ,.up_i(fill_up)
+      ,.count_o(fill_cnt)
+      );
+    assign fill_cnt_hit_target = (fill_cnt == block_size_in_fill_lp-1);
 
-    ,.count_o(fill_cnt)
-    );
-  assign fill_cnt_hit_target = (fill_cnt == block_size_in_fill_lp-1);
+    bsg_counter_clear_up #(
+      .max_val_p(block_size_in_fill_lp-1)
+      ,.init_val_p(0)
+      ,.disable_overflow_warning_p(1)
+      )
+    mem_cmd_counter
+      (.clk_i(clk_i)
+      ,.reset_i(reset_i)
 
-  bsg_counter_clear_up #(
-    .max_val_p(block_size_in_fill_lp-1)
-    ,.init_val_p(0)
-    ,.disable_overflow_warning_p(1)
-    )
-  mem_cmd_counter
-    (.clk_i(clk_i)
-    ,.reset_i(reset_i)
+      ,.clear_i('0)
+      ,.up_i(mem_cmd_up)
 
-    ,.clear_i('0)
-    ,.up_i(mem_cmd_up)
-
-    ,.count_o(mem_cmd_cnt)
-    );
-  assign mem_cmd_cnt_hit_target = (mem_cmd_cnt == block_size_in_fill_lp-1);
+      ,.count_o(mem_cmd_cnt)
+      );
+    assign mem_cmd_cnt_hit_target = (mem_cmd_cnt == block_size_in_fill_lp-1);
+  end
 
   logic [index_width_lp-1:0] index_cnt;
   logic index_up;
@@ -491,8 +495,8 @@ module bp_uce
               mem_cmd_cast_o.header.payload.way_id = lce_assoc_p'(cache_req_metadata_r.repl_way);
               mem_cmd_cast_o.header.payload.lce_id = lce_id_i;
               mem_cmd_v_o = mem_cmd_ready_i;
-              mem_cmd_up = one_mem_cmd ? 1'b0 : mem_cmd_v_o;
-              mem_cmd_done_n = one_mem_cmd;
+              mem_cmd_up = mem_cmd_v_o;
+              mem_cmd_done_n = mem_cmd_cnt_hit_target;
               state_n = mem_cmd_v_o
                         ? cache_req_metadata_r.dirty
                           ? e_writeback_evict
@@ -548,7 +552,7 @@ module bp_uce
             data_mem_pkt_v_o = load_resp_v_li;
 
             cache_req_critical_o = '0;
-            fill_up = one_mem_cmd ? 1'b0 : tag_mem_pkt_yumi_i & data_mem_pkt_yumi_i;
+            fill_up = tag_mem_pkt_yumi_i & data_mem_pkt_yumi_i;
             mem_resp_yumi_lo = tag_mem_pkt_yumi_i & data_mem_pkt_yumi_i;
             // request next sub-block
             mem_cmd_cast_o.header.msg_type       = e_cce_mem_rd;
@@ -556,7 +560,7 @@ module bp_uce
             mem_cmd_cast_o.header.size           = e_mem_msg_size;
             mem_cmd_cast_o.header.payload.way_id = lce_assoc_p'(cache_req_metadata_r.repl_way);
             mem_cmd_cast_o.header.payload.lce_id = lce_id_i;
-            mem_cmd_v_o = one_mem_cmd ? 1'b0 : (mem_cmd_ready_i & ~mem_cmd_done_r);
+            mem_cmd_v_o = mem_cmd_ready_i & ~mem_cmd_done_r;
             mem_cmd_up = mem_cmd_v_o;
             mem_cmd_done_n = mem_cmd_cnt_hit_target & mem_cmd_v_o;
 
@@ -594,7 +598,7 @@ module bp_uce
             data_mem_pkt_v_o = load_resp_v_li;
 
             cache_req_critical_o = '0;
-            fill_up = one_mem_cmd ? 1'b0 : tag_mem_pkt_yumi_i & data_mem_pkt_yumi_i;
+            fill_up = tag_mem_pkt_yumi_i & data_mem_pkt_yumi_i;
             mem_resp_yumi_lo = tag_mem_pkt_yumi_i & data_mem_pkt_yumi_i;
             // request next sub-block
             mem_cmd_cast_o.header.msg_type       = e_cce_mem_rd;
@@ -602,7 +606,7 @@ module bp_uce
             mem_cmd_cast_o.header.size           = e_mem_msg_size;
             mem_cmd_cast_o.header.payload.way_id = lce_assoc_p'(cache_req_metadata_r.repl_way);
             mem_cmd_cast_o.header.payload.lce_id = lce_id_i;
-            mem_cmd_v_o = one_mem_cmd ? 1'b0 : (mem_cmd_ready_i & ~mem_cmd_done_r);
+            mem_cmd_v_o = mem_cmd_ready_i & ~mem_cmd_done_r;
             mem_cmd_up = mem_cmd_v_o;
             mem_cmd_done_n = mem_cmd_cnt_hit_target & mem_cmd_v_o;
 
--- a/bp_me/src/v/cce/bp_uce.v
+++ b/bp_me/src/v/cce/bp_uce.v
@@ -20,10 +20,14 @@ module bp_uce
     , localparam num_dwords_per_bank_lp = bank_width_lp / dword_width_p
     , localparam byte_offset_width_lp  = `BSG_SAFE_CLOG2(bank_width_lp>>3)
     // Words per line == associativity
-    , localparam word_offset_width_lp  = `BSG_SAFE_CLOG2(assoc_p)
-    , localparam block_offset_width_lp = (word_offset_width_lp + byte_offset_width_lp)
+    , localparam bank_offset_width_lp  = `BSG_SAFE_CLOG2(assoc_p)
+    , localparam block_offset_width_lp = (bank_offset_width_lp + byte_offset_width_lp)
     , localparam index_width_lp = `BSG_SAFE_CLOG2(sets_p)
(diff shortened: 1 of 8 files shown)


dpetrisko:
why is the entire uce a diff? It's very difficult to review like this

Songchun-Li:
> why is the entire uce a diff? It's very difficult to review like this

UCE modification has been redone.

bp_me/src/v/cce/bp_uce.v
```diff
@@ -32,9 +36,10 @@ module bp_uce
     , localparam cache_stat_mem_pkt_width_lp = `bp_cache_stat_mem_pkt_width(sets_p, assoc_p)
 
     // Block size parameterisations -
-    , localparam is_blockwidth_512 = (block_width_p == 512)
-    , localparam is_blockwidth_256 = (block_width_p == 256)
-    , localparam is_blockwidth_128 = (block_width_p == 128)
+    , localparam is_fillwidth_512 = (fill_width_p == 512)
```
dpetrisko, line 39: rather than use this variable, directly set the e_mem_size here as a parameter

e.g. localparam e_mem_msg_size block_size = xxx

bp_me/src/v/cce/bp_uce.v
```diff
@@ -194,6 +200,53 @@ module bp_uce
...
+    assign bank_index =  mem_cmd_cnt;
+    assign fill_index_shift = mem_resp_cast_i.header.addr[byte_offset_width_lp+:bank_offset_width_lp];
+  end else begin
+    assign bank_index = mem_cmd_cnt << bank_sub_offset_width_lp;
+    assign fill_index_shift = mem_resp_cast_i.header.addr[byte_offset_width_lp+:bank_offset_width_lp] >> bank_sub_offset_width_lp;
+  end
+
+  logic fill_up, fill_cnt_hit_target, mem_cmd_up, mem_cmd_cnt_hit_target;
+  logic one_mem_cmd;
+  assign one_mem_cmd = (block_width_p == fill_width_p);
```
dpetrisko, line 216: I think this is unnecessary? If block_width == fill_width, then the counter degenerates into a flop, but all of the equations should still work.

dpetrisko: Well, you may need to use clear_i

bp_me/src/v/cce/bp_uce.v
```diff
@@ -456,27 +540,73 @@ module bp_uce
...
+            tag_mem_pkt_v_o = load_resp_v_li;
+
+            data_mem_pkt_cast_o.opcode = e_cache_data_mem_write;
+            data_mem_pkt_cast_o.index  = mem_resp_cast_i.header.addr[block_offset_width_lp+:index_width_lp];
+            data_mem_pkt_cast_o.way_id = mem_resp_cast_i.header.payload.way_id[0+:`BSG_SAFE_CLOG2(assoc_p)];
+            data_mem_pkt_cast_o.data   = mem_resp_cast_i.data;
+            data_mem_pkt_cast_o.fill_index = 1'b1 << fill_index_shift;
+            data_mem_pkt_v_o = load_resp_v_li;
+
+            cache_req_critical_o = tag_mem_pkt_yumi_i & data_mem_pkt_yumi_i;
```
dpetrisko, line 563: Just tie to zero for now, this is incorrect

bp_me/src/v/cce/bp_uce.v
```diff
@@ -194,6 +200,53 @@ module bp_uce
...
+    ,.reset_i(reset_i)
+
+    ,.clear_i('0)
+    ,.up_i(fill_up)
+
+    ,.count_o(fill_cnt)
+    );
+  assign fill_cnt_hit_target = (fill_cnt == block_size_in_fill_lp-1);
+
+  bsg_counter_clear_up #(
```
dpetrisko, line 234: Rather than have a separate done register, you can have max_val_p == block_size_in_fill_lp.  Then done is when the counter reaches max value

bp_me/src/v/cce/bp_uce.v
Songchun-Li: The reason I added one_mem_cmd is that the counters are actually flops when block_width == fill_width and block_width == 2*fill_width. 
If block_width == fill_width, the fill_cnt_hit_target and mem_cmd_cnt_hit_target is high when mem_cmd_cnt and  fill_cnt is 0, which means that we don't even need a flop, so one_mem_cmd disables the flop.
If block_width == 2 * fill_width, the fill_cnt_hit_target and mem_cmd_cnt_hit_target is high when mem_cmd_cnt and  fill_cnt is 1, then the flop is used to count.

bp_me/src/v/cce/bp_uce.v
Songchun-Li: The mem_cmd_done_r reg here is to indicate that all the mem_cmd is sent.
For example, with the mem_cmd_cnt counts from 0 to 7, UCE sends out 8 mem_cmd to L2, using the mem_cmd_cnt to generate the addr. When all the mem_cmd_cnt is sent, **mem_cmd_cnt will be cleared back to 0**. Therefore, mem_cmd_cnt_hit_target will be low again and can not save the info that all mem_cmd is sent. Reg mem_cmd_done_r save this info until the current state ends.

bp_me/src/v/cce/bp_uce.v
Songchun-Li: Or I can rewrite this with conditional generation, so the counter will not be generated when block_width == fill_width, and set the two signal fill_cnt_hit_target, mem_cmd_cnt_hit_target to be higher, so that we can get rid of the one_mem_cmd

bp_me/src/v/cce/bp_uce.v
Songchun-Li: Using max_val_p == block_size_in_fill_lp will work well when UCE writes back dirty data, because UCE transists to next state right after it sends out all mem_cmd. However, the senario I mentioned above is at the e_read_req, where UCE will still need to wait for all the mem_resps coming back, and therefore the mem_cmd_cnt will go back to 0 before it transists to next state. Reg mem_cmd_done_r save this info until the current state ends.

Songchun-Li:
All the comments are resolved, except the mem_cmd_done_r reg which needs further discussion. These 2 new commits pass sims with the combination cfgs of block_width=512, fill_width=block_width or  bank_width, assoc = 8, 4, 2,

bp_me/src/v/cce/bp_uce.v
dpetrisko: I think max_val_p == block_size_in_fill is still the way to go.  It can be cleared on transition so as to not lose any cycles

bp_me/src/v/cce/bp_uce.v
```diff
@@ -20,21 +20,31 @@ module bp_uce
...
     , localparam cache_tag_mem_pkt_width_lp = `bp_cache_tag_mem_pkt_width(sets_p, assoc_p, ptag_width_p)
     , localparam cache_data_mem_pkt_width_lp = `bp_cache_data_mem_pkt_width(sets_p, assoc_p, block_width_p, fill_width_p)
     , localparam cache_stat_mem_pkt_width_lp = `bp_cache_stat_mem_pkt_width(sets_p, assoc_p)
 
-    // Block size parameterisations -
-    , localparam is_blockwidth_512 = (block_width_p == 512)
-    , localparam is_blockwidth_256 = (block_width_p == 256)
-    , localparam is_blockwidth_128 = (block_width_p == 128)
+    // Fill size parameterisations -
+    , localparam bp_mem_msg_size_e e_mem_msg_size = (fill_width_p == 512)
```
dpetrisko, line 39: block_msg_size_lp

bp_me/src/v/cce/bp_uce.v
```diff
@@ -194,6 +205,57 @@ module bp_uce
   wire miss_v_li       = miss_load_v_li | miss_store_v_li;
   wire uc_load_v_li    = cache_req_v_r & cache_req_r.msg_type inside {e_uc_load};
 
+  logic [fill_cnt_width_lp-1:0] fill_cnt, mem_cmd_cnt;
```
dpetrisko, line 208: Comment here to explain what it's doing

bp_me/src/v/cce/bp_uce.v
```diff
@@ -194,6 +205,57 @@ module bp_uce
...
+    fill_counter
+      (.clk_i(clk_i)
+      ,.reset_i(reset_i)
+
+      ,.clear_i('0)
+      ,.up_i(fill_up)
+
+      ,.count_o(fill_cnt)
+      );
+    assign fill_cnt_hit_target = (fill_cnt == block_size_in_fill_lp-1);
```
dpetrisko, line 240: fill_done

bp_me/src/v/cce/bp_uce.v
```diff
@@ -194,6 +205,57 @@ module bp_uce
...
+    mem_cmd_counter
+      (.clk_i(clk_i)
+      ,.reset_i(reset_i)
+
+      ,.clear_i('0)
+      ,.up_i(mem_cmd_up)
+
+      ,.count_o(mem_cmd_cnt)
+      );
+    assign mem_cmd_cnt_hit_target = (mem_cmd_cnt == block_size_in_fill_lp-1);
```
dpetrisko, line 256: mem_cmd_done

bp_me/src/v/cce/bp_uce.v
```diff
@@ -230,6 +292,14 @@ module bp_uce
      );
   wire way_done = (way_cnt == assoc_p-1);
 
+  logic mem_cmd_done_n, mem_cmd_done_r;
```
dpetrisko, line 295: use bsg_dff_reset_set_clear

dpetrisko: Should be set when 
mem_cmd_done & mem_cmd_v_o
and cleared when 
cache_req_complete_o | way_up

bp_me/src/v/cce/bp_uce.v
dpetrisko: I think you are right that there should be a flop here.  Reason being that when handling critical word first, we will start the counter in the middle of the count and loop around, so the done condition will be dependent on critical word instead of on max count.  And we will need the overflow behavior to be consistent

bp_me/src/v/cce/bp_uce.v
```diff
@@ -194,6 +205,57 @@ module bp_uce
...
+  end
+
+  logic fill_up, fill_cnt_hit_target, mem_cmd_up, mem_cmd_cnt_hit_target;
+  if (fill_width_p == block_width_p) begin
+    assign fill_cnt = 1'b0;
+    assign mem_cmd_cnt = 1'b0;
+    assign fill_cnt_hit_target = 1'b1;
+    assign mem_cmd_cnt_hit_target = 1'b1;
+  end else begin
+    bsg_counter_clear_up #(
```
dpetrisko, line 226: move fill_cnt and mem_cmd_cnt declaratoin inside this else block, so that the if statement just has fill_done and mem_cmd_done

bp_me/src/v/cce/bp_uce.v
```diff
@@ -165,19 +175,20 @@ module bp_uce
...
   wire is_ready         = (state_r == e_ready);
-  wire is_send_req      = (state_r == e_send_req);
-  wire is_writeback_read = (state_r == e_writeback_read);
-  wire is_writeback_req = (state_r == e_writeback_req);
-  wire is_write_request = (state_r == e_write_wait);
-  wire is_read_request  = (state_r == e_read_wait);
+  wire is_send_critical = (state_r == e_send_critical);
+  wire is_writeback_evict = (state_r == e_writeback_evict); // read dirty data from cache to UCE
+  wire is_writeback_read = (state_r == e_writeback_read_req); // read data from L2 to cache
+  wire is_writeback_wb   = (state_r == e_writeback_write_req); // read dirty data in UCE
```
dpetrisko, line 189: // send dirty data from UCE to L2?

bp_me/src/v/cce/bp_uce.v
```diff
@@ -230,6 +292,14 @@ module bp_uce
      );
   wire way_done = (way_cnt == assoc_p-1);
 
+  logic mem_cmd_done_n, mem_cmd_done_r;
+  always_ff @(posedge clk_i) begin
```
dpetrisko, line 296: Style nit: begin and end on their own line, 2-space indent
```
always_ff @(posedge clk_i)
  begin
    ...
  end
```

dpetrisko: throughout

bp_me/src/v/cce/bp_uce.v
```diff
@@ -254,6 +324,16 @@ module bp_uce
   assign credits_full_o = (credit_count_lo == coh_noc_max_credits_p);
   assign credits_empty_o = (credit_count_lo == 0);
 
+  logic [fill_width_p-1:0] writeback_data;
+  bsg_mux #(
```
dpetrisko, line 328: Similar style nit: Declare modules in this style. We want to maintain style within files when making changes.  (Throughout)

```
  bsg_dff_en_bypass
   #(.width_p(1))
   metadata_v_reg
    (.clk_i(clk_i)

     ,.en_i(cache_req_v_i | cache_req_metadata_v_i)
     ,.data_i(cache_req_metadata_v_i)
     ,.data_o(cache_req_metadata_v_r)
     );
```

bp_me/src/v/cce/bp_uce.v
Songchun-Li: Moving mem_cmd_cnt declaration inside the else block may not be a good idea, because mem_cmd_cnt is not just for tracking, but also used to generate the addr in mem_cmd. And mem_cmd_cnt also has a conditional expression that is somehow overlap with the conditional generation of the mem cmd counter

Songchun-Li: Moving the mem_cmd_cnt declaration inside the else block may not be a good idea, because mem_cmd_cnt is also used to generate the addr of mem_cmd with conditional expression from line 216-225.
Moving the declaration will need to integrate this two if statement in one. In this case, we need to further divede the else statement to two and need to duplicate the instatiation of the two counter like this. And the code will be much longer.
```
if (fill_width_p == block_width_p)
  begin
      assign fill_done = 1'b1;
      assign mem_cmd_done = 1'b1;
      assign bank_index =  '0;
      assign fill_index_shift = '0;
  end
else if (fill_size_in_bank_lp == 1)
  begin
      logic [fill_cnt_width_lp-1:0] fill_cnt, mem_cmd_cnt;
      assign bank_index =  mem_cmd_cnt;
      assign fill_index_shift = mem_resp_cast_i.header.addr[byte_offset_width_lp+:bank_offset_width_lp];
      // two counters here
  end
else
  begin
      logic [fill_cnt_width_lp-1:0] fill_cnt, mem_cmd_cnt;
      assign bank_index = mem_cmd_cnt << bank_sub_offset_width_lp;
      assign fill_index_shift = mem_resp_cast_i.header.addr[byte_offset_width_lp+:bank_offset_width_lp] >> bank_sub_offset_width_lp;
       // two counters here
  end
  
```

Songchun-Li:
The issues mentioned are fixed, except that the declararion of mem_cmd_cnt is not yet moved because of the reason mentioned in the above comment.

bp_me/src/v/cce/bp_uce.v
dpetrisko: Makes sense! I didn't think about the fact it would be used as the offset

Issue #483, opened 2020-05-07
ME regression passes in VCS, but not Verilator


dpetrisko:
2eaabec0 in sw_dev.

```make -C bp_me build.v sim.v``` passes

```make -C bp_me build.sc sim.sc``` fails with trace mismatch

I think this may have snuck in with the new Verilator version.  If that's the case, we can talk about reverting it, but it would be nice to include the latest build system changes they've committed.  I've anecdotally found that when VCS works, but Verilator doesn't, it tends to be non-compliant SV code, which would be nice to root out

(I've just disabled the Verilator run in CI for now)

muwyse:
Very puzzling. I ran both and diffed the CCE trace's. The Verilator one simply has fewer lines, but all the lines it contains match exactly with the VCS trace.

The LCE traces show that in the Verilator simulation a store appears to be dropped by the LCE. Still investigating, but it is almost certainly a bug in the mock LCE. My guess is there is a logic bug with committing the store after getting coherence permissions for the block, or the nonsynth nature of the mock LCE is causing problems.

The upside here is it appears to be a bug with the nonsynth mock LCE, not the actual processor or CCE.

muwyse:
Root cause appears to be a regression in Verilator tool itself related to bit selection from a 2D array/vector. Mock LCE has been updated to use different syntax to avoid, with comment about problem.

Fixed by #523

PR #523, opened 2020-05-28
Enable flexible LCE Request data width; fix mock LCE Verilator issue

muwyse:
This PR adds support for larger data widths in the LCE Request message/network (Issue #404). The ME and top were updated to support these changes. The LCE Request network coming from the D$/I$ has 64-bits of data, but the CCE now supports up to cce_block_width_p of data. The I/O tiles are configured to use cce_block_width_p requests. Further changes are needed for the accelerator configurations.

This PR also provides a workaround fix for a Verilator specific issue in the mock LCE (Issue #483). It appears something in the Verilator tool itself changed regarding bit selection from a 2D vector. The mock LCE was also rewritten to use bsg_mem instances for the tag, data, and dirty bit memories, and for some general cleanup.

### DIFF
--- a/bp_common/src/include/bp_common_me_if.vh
+++ b/bp_common/src/include/bp_common_me_if.vh
@@ -46,7 +46,7 @@
  *
  */
 
-`define declare_bp_lce_cce_if(cce_id_width_mp, lce_id_width_mp, paddr_width_mp, lce_assoc_mp, data_width_mp, cce_block_width_mp) \
+`define declare_bp_lce_cce_if(cce_id_width_mp, lce_id_width_mp, paddr_width_mp, lce_assoc_mp, cce_block_width_mp) \
                                                                                                          \
 /*                                                                                                       \
  * bp_lce_cce_req_s defines an LCE request sent by an LCE to a CCE on a cache miss. An LCE enters        \
@@ -59,6 +59,7 @@
  * non_exclusive indicates if the requesting cache prefers non-exclusive read-access                     \
  * addr is the cache missing address                                                                     \
  * lru_way_id indicates the way within the target set that will be used to fill the miss in to           \
+ *                                                                                                       \
  */                                                                                                      \
   typedef struct packed                                                                                  \
   {                                                                                                      \
@@ -73,7 +74,7 @@
                                                                                                          \
   typedef struct packed                                                                                  \
   {                                                                                                      \
-    logic [data_width_mp-1:0]                    data;                                                   \
+    logic [cce_block_width_mp-1:0]               data;                                                   \
     bp_lce_cce_req_header_s                      header;                                                 \
   } bp_lce_cce_req_s;                                                                                    \
                                                                                                          \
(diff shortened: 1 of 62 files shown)


bp_me/test/common/bp_me_nonsynth_mock_lce.v
```diff
@@ -284,24 +221,22 @@ module bp_me_nonsynth_mock_lce
...
   assign dword_offset = cmd.paddr[5:3];
   assign byte_offset = cmd.paddr[2:0];
 
   // Data word (64-bit) targeted by current trace replay command
-  logic [dword_width_p-1:0] load_data;
-  assign load_data = data_cur[dword_width_p*dword_offset +: dword_width_p];
+  logic [dword_width_p-1:0] load_dword;
+  assign load_dword = dword_width_p'(data_lo[mshr_r.lru_way] >> (dword_width_p*dword_offset));
+  // This doesn't work in Verilator...why?
+  //assign load_dword = data_lo[mshr_r.lru_way][dword_width_p*dword_offset +: dword_width_p];
```
dpetrisko, line 239: Put the failing version number

bp_common/src/include/bp_common_me_if.vh
```diff
@@ -73,10 +78,16 @@
...
-    logic [data_width_mp-1:0]                    data;                                                   \
+    logic [lce_req_max_data_width_mp-1:0]        data;                                                   \
     bp_lce_cce_req_header_s                      header;                                                 \
   } bp_lce_cce_req_s;                                                                                    \
                                                                                                          \
+  typedef struct packed                                                                                  \
+  {                                                                                                      \
+    logic [cce_block_width_mp-1:0]               data;                                                   \
+    bp_lce_cce_req_header_s                      header;                                                 \
+  } bp_lce_cce_block_req_s;                                                                              \
```
dpetrisko, line 89: There's an interesting question here.  These interface structures are really "protocols".  And the protocol data width will be maximum 512 bits, based on the size field.

When we implement streaming mem_cmd/mem_resp interfaces, rather than block_width or req_width, the data parameter will refer to how wide the data bus width is (i.e. noc width, cache width).  In that case, having multiple structs isn't really necessary.  We're really just saying -- for this instantiation of the interface, we only expect data less than or equal to this size.

When you change the interface of the A better approach to removing extra buffering is to move that parameter to the wormhole adapters themselves.  For instance, by changing bp_me/src/v/wormhole/bp_me_wormhole_packet_encode_lce_req.v, this actually doesn't change the wormhole adapter buffer size at all, because that's based on len * protocol data width.

So I would say just combine bp_lce_cce_req_s and bp_lce_cce_block_req_s; we can look at the overhead in the backend, but most likely the vast majority of that will be better solved when we tackle serialization / deserialization

muwyse: Let me take another pass at it. I added the second struct because bp_tile instantiates both the LCEs and CCEs, and until we have the serdes changes, I figured we didn't want the D$ and I$ LCEs to have block_width data with the LCE Req network. I have a new idea for this though, which is to use the header struct already defined and concat it with a dword_width data signal. That should take care of routing a narrow signal from LCEs to adapters for now.

Agreed, these issues disappear with removing serialization/deserialization. That should be next step (or maybe just rolled in with this one)

dpetrisko: I like the header struct + data solution for the CCE, since that's (similar to) what the serialization approach will look like.  Maybe we can also use that in the LCE?

I'd suggest  getting rid of the lce_req_max_width_p parameter in the interface definition completely, (always CCE block width).  The extra wires anywhere the full packet is instantiated will get synthesized out, and there are no actual excess registers here (which would also get synthesized out)

We should definitely merge this before approaching the serialization project, which will take a few weeks at least.

muwyse:
Updated to make LCE Req network carry up to cce_block_width_p of data, same as cmd and resp networks. All modules connect with full width at the moment, but the interface can be broken down to header struct + data where desired to reduce width. Further changes to the encoder (lce_req wormhole encoder) may be useful to reduce width of signals through the block and along the path from LCE to wormhole adapter. LCEs may also benefit from switch to header+data on the interface.

dpetrisko:
LGTM!

PR #526, opened 2020-05-30
One-hotifying caches

dpetrisko:
This PR converts the caches to one-hot encoding for their muxes.  In order to support this changes, we need to modify our banking scheme from bank_id = way_id ^ word_offset to bank_id = way_id [mod +] word_offset.  For this we use the one-hot adder module recently added to basejump_stl.

This should give a fairly large critical path improvement with little area overhead.

@Shashank-Vijay @sripathi-muralitharan Take a look to see if the changes make sense?

@taylor-bsg do you have an idea if there are more efficient ways to synthesize circular shift? We should add a barrel shifter to basejump_stl, but I'm guessing this will synthesize to the same.

```
  // Circular left shift
  assign bar = (foo << x) | (foo >> WIDTH-x)

  // Circular right shift
  assign bar = {foo, foo} >> x;
```

### DIFF
--- a/bp_be/src/v/bp_be_mem/bp_be_dcache/bp_be_dcache.v
+++ b/bp_be/src/v/bp_be_mem/bp_be_dcache/bp_be_dcache.v
@@ -1005,8 +1005,14 @@ module bp_be_dcache
       : {data_mem_mask_width_lp{1'b1}};
   end
 
-  // Circular left shift
-  assign lce_data_mem_write_data = (data_mem_pkt.data << (data_mem_pkt.way_id*bank_width_lp)) | (data_mem_pkt.data >> (dcache_block_width_p-data_mem_pkt.way_id*bank_width_lp));
+  wire [`BSG_SAFE_CLOG2(dcache_block_width_p)-1:0] write_data_rot_li = data_mem_pkt.way_id*bank_width_lp;
+  bsg_rotate_left #(
+    .width_p(dcache_block_width_p)
+  ) write_data_rotate (
+    .data_i(data_mem_pkt.data)
+    ,.rot_i(write_data_rot_li)
+    ,.o(lce_data_mem_write_data)
+  );
 
   // tag_mem
   //
@@ -1142,8 +1148,14 @@ module bp_be_dcache
     end
   end
 
-  // Circular right shift
-  assign data_mem_o = {data_mem_data_lo, data_mem_data_lo} >> (data_mem_pkt_way_r*bank_width_lp);
+  wire [`BSG_SAFE_CLOG2(dcache_block_width_p)-1:0] read_data_rot_li = data_mem_pkt_way_r*bank_width_lp;
+  bsg_rotate_right #(
+    .width_p(dcache_block_width_p)
+  ) read_data_rotate (
+    .data_i(data_mem_data_lo)
+    ,.rot_i(read_data_rot_li)
+    ,.o(data_mem_o)
+  );
 
   // As an optimization, we snoop the data_mem_pkts to see if there
   // are any matching entries in the write buffer and disallow the
--- a/bp_be/src/v/bp_be_mem/bp_be_dcache/bp_be_dcache.v
+++ b/bp_be/src/v/bp_be_mem/bp_be_dcache/bp_be_dcache.v
@@ -387,7 +387,7 @@ module bp_be_dcache
   end
 
   bsg_decode
-   #(.num_out_p(icache_assoc_p))
+   #(.num_out_p(dcache_assoc_p))
    offset_decode
     (.i(addr_word_offset_tl)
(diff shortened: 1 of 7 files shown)


bp_be/src/v/bp_be_mem/bp_be_dcache/bp_be_dcache.v
```diff
@@ -387,25 +386,12 @@ module bp_be_dcache
...
-    #(.width_p(dcache_assoc_p)
-      ,.lo_to_hi_p(1)
-      )
-    pe_store_hit_tl
-    (.i(store_hit_tl)
-      ,.v_o(store_hit_v_tl)
-      ,.addr_o(store_hit_way_tv_r_tl)
-      );
+  bsg_decode
+   #(.num_out_p(icache_assoc_p))
```
sripathi-muralitharan, line 390: Shouldn't this be dcache_assoc_p?

bp_be/src/v/bp_be_mem/bp_be_dcache/bp_be_dcache.v
```diff
@@ -844,13 +845,22 @@ module bp_be_dcache
   logic [bank_width_lp-1:0] ld_data_way_picked;
   logic [dword_width_p-1:0] ld_data_dword_picked;
   logic [dword_width_p-1:0] bypass_data_masked;
+  logic [dcache_assoc_p-1:0] ld_data_way_select;
+
+  bsg_adder_one_hot
+   #(.width_p(icache_assoc_p))
```
sripathi-muralitharan, line 851: Shouldn't this be dcache_assoc_p?

bp_be/src/v/bp_be_mem/bp_be_dcache/bp_be_dcache.v
```diff
@@ -844,13 +845,22 @@ module bp_be_dcache
...
+   select_adder
+    (.a_i(load_hit_tv_r)
```
sripathi-muralitharan, line 853: I think for the way computation, you would have to use load_hit_way_tv

bp_be/src/v/bp_be_mem/bp_be_dcache/bp_be_dcache.v
dpetrisko: :)

bp_be/src/v/bp_be_mem/bp_be_dcache/bp_be_dcache.v
dpetrisko: load_hit_way_tv is the encoded version. We want the one-hot version here

bp_be/src/v/bp_be_mem/bp_be_dcache/bp_be_dcache.v
```diff
@@ -991,15 +1005,9 @@ module bp_be_dcache
...
-    .width_p(bank_width_lp)
-    ,.els_p(dcache_assoc_p)
-  ) write_mux_butterfly (
-    .data_i(data_mem_pkt.data)
-    ,.sel_i(data_mem_pkt.way_id)
-    ,.data_o(lce_data_mem_write_data)
-  );
- 
+  // Circular left shift
+  assign lce_data_mem_write_data = (data_mem_pkt.data << (data_mem_pkt.way_id*bank_width_lp)) | (data_mem_pkt.data >> (icache_block_width_p-data_mem_pkt.way_id*bank_width_lp));
```
sripathi-muralitharan: I guess you should use dcache_block_width_p here.

bp_be/src/v/bp_be_mem/bp_be_dcache/bp_be_dcache.v
sripathi-muralitharan: Oh right. Gotcha. For a moment there, I thought that was the one-hot version.

taylor-bsg:
Dan,

bsg_rotate_right.v and bsg_rotate_left.v

M

taylor-bsg:
@tommydcjung Adding because Tommy might find this interesting.

dpetrisko:
> Dan, bsg_rotate_right.v and bsg_rotate_left.v

Nice :)


I believe these will synthesize sub-optimally in a non-flattened flow. This will always generate a barrel shifter of the granularity of a bit i.e. a full log512*512 bit mux tree. However, this case calls for a shifter of granularity bank_width i.e. a log8*(512). 

I’ll verify the synthesis behavior and add a granularity parameter to the basejump rotators.

taylor-bsg:
Usually the tools will push constants through even when not flattened, so
in the case you mentioned, the low bits of the select should be zero, which
should hopefully partially evaluate away the lower layers of the mux tree.
But it is definitely useful to check this.

M

PR #527, opened 2020-05-30
Remove defunct pmp csr support

dpetrisko:
We don't actually support PMP, and supporting the CSR without trapping on violations is a violation of the spec.


dpetrisko:
Will do

dpetrisko:
passes

PR #531, opened 2020-06-01
Core Modifications to support Fetch + Op Atomics

sripathi-muralitharan:
The changes in this PR are mainly to the instruction decoder to identify the fetch + op atomics as valid instructions and not trap.

These were tested with the OpenPiton setup and all RISC-V amo-p tests pass.


bp_be/src/v/bp_be_calculator/bp_be_instr_decoder.v
```diff
@@ -282,10 +286,136 @@ always_comb
           // Note: could do a more efficent decoding here by having atomic be a flag
           //   And having the op simply taken from funct3
           unique casez (instr)
-            `RV64_LRW: decode.fu_op = e_lrw;
-            `RV64_SCW: decode.fu_op = e_scw;
-            `RV64_LRD: decode.fu_op = e_lrd;
-            `RV64_SCD: decode.fu_op = e_scd;
+            `RV64_LRW:      decode.fu_op = e_lrw;
+            `RV64_SCW:      decode.fu_op = e_scw;
+            `RV64_AMOSWAPW: begin
```
dpetrisko, line 291: nit; begin on next line like the others

dpetrisko:
Looks fine to me for the purpose of piton integration.  I think the dcache has reached the point where we need a separate decoder module, and possibly pipeline structs to reduce duplication and verbosity.

Possibly it also makes sense to reuse op_w from the decoder rather than expand to 6 bit fu_ops.  The dcache packet could then become addr + op + size.  Would reduce cache decoder size and some flops.  Just thinking out loud, though

bp_be/src/v/bp_be_calculator/bp_be_instr_decoder.v
sripathi-muralitharan: :)

sripathi-muralitharan:
Yeah, we could re-use the ops and refactor the dcache_pkt to have all the information we need. Since, we are going to refactor the caches a whole bunch for the late writeback feature that we are going to add in later, we could put in a new D$ decoder and the pipeline structs. Certainly would enhance management and readability of signals.

PR #533, opened 2020-06-02
[WIP] Better riscv-dv integration

dpetrisko:
Added a ton more riscv-dv tests and updated them to use M-mode.  This uncovered several problems in dromajo https://github.com/bsg-external/dromajo/pull/3 and several potential bugs in BP. There are ~40/150 failing randomly generated tests.  Most likely the same bug appears multiple times, but nothing like a good 🐛 🔍 !

- [ ] Find bugs
- [ ] Add targeted tests for bugs
- [ ] Add riscv-dv run (with seed) to CI
- [ ] Add long-running risvc-dv (without seed) to manual CI, to find bugs in the background

Upstream riscv-dv is working on a python-based generator, which we should move to instead of the UVM, so that open-source users can use this as well.

### DIFF
--- a/bp_be/src/v/bp_be_mem/bp_be_csr.v
+++ b/bp_be/src/v/bp_be_mem/bp_be_csr.v
@@ -367,7 +367,7 @@ always_comb
         end
       else if (csr_cmd.csr_op == e_sfence_vma)
         begin
-          if (is_s_mode & mstatus_lo.tvm)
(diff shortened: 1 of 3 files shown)


dpetrisko:
Some of the errors are coming from: https://github.com/google/riscv-dv/issues/588

taylor-bsg:
Lol .. We should propose BlackParrot as a verification infrastructure for
RISCV-DV

Issue #494, opened 2020-05-12
Unify LCEs


dpetrisko:
Currently, we have an icache_lce and dcache_lce which perform largely the same functions. This is a code maintenance burden, since bug fixes / enhancements in one need to happen in another.  There's relatively small overhead in having unused FSM states, and almost no extra data storage between them. If we find benefits to specializing, we can do so with parameterization and decomposition rather than maintaining two sets of components.

This also clears up the story of the BedRock network: LCE, CCE, UCE (and eventually PCE). The accelerator story is also made more clear, as they just use the standard LCE rather than the dcache

Action items
- [x] Combine bp_be_dcache_lce and bp_fe_icache_lce into a single lce
- [x] Move to memory end (update flists, etc)

PR #535, opened 2020-06-04
[WiP]: ME LCE - Unified LCE to replace FE and BE LCE's

muwyse:
This PR addresses issue #494 and creates a unified LCE to replace the FE and BE LCE's.

The goal is to reduce code maintenance as there is a minimal difference between the existing BE and FE LCE's. The FE LCE is effectively a subset of the BE LCE.


muwyse:
6/4/20 AM status: some RISCV tests failing, debugging root cause.

dpetrisko:
(Once you're done debugging) delete the lces in the FE/BE

bp_me/src/v/lce/bp_lce_cmd.v
```diff
@@ -0,0 +1,677 @@
...
+
+            end
+
+            // Invalidate Tag - write tag mem and send Invalidate Ack
+            e_lce_cmd_inv: begin
+              tag_mem_pkt.index = lce_cmd_addr_index;
+              tag_mem_pkt.way_id = lce_cmd_way_id;
+              tag_mem_pkt.state = e_COH_I;
+              tag_mem_pkt.opcode = e_cache_tag_mem_invalidate;
+              tag_mem_pkt_v_o = ~tag_mem_cmd_sent_r;
```
dpetrisko, line 415: How about tag_mem_pkt_v_o = lce_resp_ready_i;  and get rid of the sent register? (same with other sent registers)

muwyse: That is fine with me. I maintained the behavior of the old LCEs here. The tradeoff is sending tag/data/stat potentially multiple times in the case when more than one packet type needs to be sent, or to send a packet multiple times if the LCE-CCE message handshake does not occur. (i.e., send a tag packet while trying to send an LCE response but the response message can't send in the first cycle and the tag packet can.)

I think this is fine because the CCE "manages" the LCE and the LCE should be able to issue the same read/write command on any of the packet ports multiple times without issue (i.e., multiple tag mem packets to invalidate a tag). Basically, I think the operations can be replayed even if successful without harm.

dpetrisko: Gotcha.  Yeah, I'm not too familiar with the old lce code.  The UCE works this way, with the tag_mem_packet being sent multiple times, and I agree it should be fine.  The way I look at it is once the initial invalidate comes in, the xCE is free to do whatever it wants to the mems until cache_req_complete_o comes back

muwyse:
status update: debugging multicore; best guess is bug with transfers or writebacks, which have historically been the issue when updating LCEs

muwyse:
ready for review; tests look good

bp_me/src/v/lce/bp_lce_cmd.v
```diff
@@ -0,0 +1,612 @@
...
+
+    , output logic                                   stat_mem_pkt_v_o
+    , output logic [cache_stat_mem_pkt_width_lp-1:0] stat_mem_pkt_o
+    , input                                          stat_mem_pkt_yumi_i
+    , input [stat_info_width_lp-1:0]                 stat_mem_i
+
+    // request complete signals
+    // cached requests and uncached loads block in the caches, but uncached stores do not
+    // cache_req_complete_o is routed to the cache to indicate a blocking request is complete
+    , output logic                                   cache_req_complete_o
```
dpetrisko, line 78: Interesting.  I know this is how the UCE works as well, but it's kind of weird to have a single type of request not send cache_req_complete.   Do you see any problems with changing so that cache_req_complete is also sent on uncached stores?

bp_me/src/v/lce/bp_lce_cmd.v
```diff
@@ -0,0 +1,612 @@
...
+     (.clk_i(clk_i)
+      ,.reset_i(reset_i)
+      ,.clear_i(cnt_clear)
+      ,.up_i(cnt_inc)
+      ,.count_o(cnt_r)
+      );
+
+  // coherence request size for cached requests
+  // block size smaller than 8-bytes not supported
+  bp_mem_msg_size_e cmd_block_size =
```
dpetrisko, line 200: make a localparam or this won't synthesize

bp_me/src/v/lce/bp_lce_cmd.v
```diff
@@ -0,0 +1,612 @@
...
+    #(.width_p(1))
+    sync_done_reg
+     (.clk_i(clk_i)
+      ,.en_i(sync_done_en)
+      ,.data_i(sync_done_li)
+      ,.data_o(sync_done_o)
+      );
+
+  // data buffer and data buffered registers
+  logic data_buffered_r, data_buffered_en, data_buffered_li;
```
dpetrisko, line 151: I'm kind of partial to the way we do this in the UCE:
https://github.com/black-parrot/black-parrot/blob/master/bp_me/src/v/cce/bp_uce.v#L88

Basically, you always latch the dirty data after a read command.  This brings the logic out of the FSM, which is usually a good idea to reduce complexity

bp_me/src/v/lce/bp_lce_req.v
```diff
@@ -0,0 +1,285 @@
...
+  bp_me_addr_to_cce_id
+   #(.bp_params_p(bp_params_p))
+   req_map
+    (.paddr_i(lce_req.header.addr)
+     ,.cce_id_o(req_cce_id_lo)
+     );
+
+  // coherence request size for cached requests
+  // block size smaller than 8-bytes not supported
+  bp_mem_msg_size_e req_block_size =
```
dpetrisko, line 157: localparam

bp_me/src/v/lce/bp_lce_cmd.v
muwyse: I think we address this after this PR. There might be some subtle things we need to rework or consider. Basically, the dcache doesn't consider an uncached store a miss, so if we raise cache_req_complete for uncached store there will be a mismatch between misses and req_complete signals.

In the LCE, every request is a request that counts against its credit counter so the LCE doesn't need to differentiate. (unless it becomes responsible for tracking multiple outstanding requests and ordering responses to the cache).

dpetrisko: works for me!

bp_me/src/v/lce/bp_lce_cmd.v
muwyse: I see your point; I'll change this. It should also eliminate the data_buffered register.

PR #536, opened 2020-06-07
Adding Multicore Cosimulation and Checkpointing

farzamgl:
This PR adds the capability to run multicore tests with the Dromajo cosimulation and also lets us create checkpoints for these tests.
- Bumped Dromajo with multicore capability
- Initialize the Dromajo cosimulation instance using RTL parameters as input flags
- Minor testbench and Makefile changes (Added `SAMPLE_NCPUS` flag for checkpoint generation)

### DIFF
--- a/bp_top/test/common/bp_nonsynth_cosim.v
+++ b/bp_top/test/common/bp_nonsynth_cosim.v
@@ -15,6 +15,7 @@ module bp_nonsynth_cosim
     , input                                   freeze_i
     , input                                   en_i
 
+    , input [31:0]                            num_core_i
     , input [`BSG_SAFE_CLOG2(num_core_p)-1:0] mhartid_i
     , input [63:0]                            config_file_i
     , input [31:0]                            cosim_instr_i
@@ -32,23 +33,20 @@ module bp_nonsynth_cosim
     , input                                   interrupt_v_i
     , input [dword_width_p-1:0]               cause_i
 
-    , output logic [num_core_p-1:0]           finish_o
+    , output logic                            finish_o
     );
 
-import "DPI-C" context function void init_dromajo(string cfg_f_name);
+import "DPI-C" context function void dromajo_init(string cfg_f_name, int hartid, int ncpus);
 import "DPI-C" context function bit  dromajo_step(int      hart_id,
                                                   longint pc,
                                                   int insn,
                                                   longint wdata);
 import "DPI-C" context function void dromajo_trap(int hart_id, longint cause);
 
-logic finish;
-
 always_ff @(negedge reset_i)
   if (en_i)
     begin
-      $display("Running with Dromajo cosimulation");
-      init_dromajo(config_file_i);
+      dromajo_init(config_file_i, mhartid_i, num_core_i);
     end
 
   bp_be_decode_s decode_r;
--- a/bp_top/test/common/dromajo_cosim.cpp
+++ b/bp_top/test/common/dromajo_cosim.cpp
@@ -4,14 +4,20 @@
 #include "stdlib.h"
 #include <string>
 
+using namespace std;
+
 dromajo_cosim_state_t* dromajo_pointer;
 uint64_t d_address = 0;
 uint64_t d_count = 0;
 
-extern "C" void init_dromajo(char* cfg_f_name) {
-  char *argv[] = {(char*)"Variane", cfg_f_name};
+extern "C" void dromajo_init(char* cfg_f_name, int hartid, int ncpus) {
 
-  dromajo_pointer = dromajo_cosim_init(2, argv);
+  if(hartid == 0) {
+    cout << "Running with Dromajo cosimulation" << endl;
+    string ncpus_str = "--ncpus=" + to_string(ncpus);
+    char *argv[] = {(char*)"", cfg_f_name, (char*)(&ncpus_str[0])};
+    dromajo_pointer = dromajo_cosim_init(3, argv);
+  }
 }
 
 extern "C" bool dromajo_step(int      hart_id,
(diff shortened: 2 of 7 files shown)


bp_top/test/common/bp_nonsynth_cosim.v
```diff
@@ -32,23 +33,20 @@ module bp_nonsynth_cosim
...
                                                   int insn,
                                                   longint wdata);
 import "DPI-C" context function void dromajo_trap(int hart_id, longint cause);
 
-logic finish;
-
 always_ff @(negedge reset_i)
   if (en_i)
     begin
-      $display("Running with Dromajo cosimulation");
```
dpetrisko, line 50: Keep with mhartid_i == 0?

dpetrisko: nvm, I see the C++ entry

bp_common/test/cfg/sample.cfg
```diff
@@ -4,7 +4,7 @@
   "bios": "prog.mainram",
   "load": "prog",
   "memory_base_addr": 0x80000000,
-  "memory_size": $SAMPLE_MEMSIZE,
```
dpetrisko, line 7: Can we keep this parameter? 256 is a large startup overhead for small samples

bp_top/test/common/dromajo_cosim.cpp
```diff
@@ -4,14 +4,20 @@
...
 uint64_t d_count = 0;
 
-extern "C" void init_dromajo(char* cfg_f_name) {
-  char *argv[] = {(char*)"Variane", cfg_f_name};
+extern "C" void dromajo_init(char* cfg_f_name, int hartid, int ncpus) {
 
-  dromajo_pointer = dromajo_cosim_init(2, argv);
+  if(hartid == 0) {
+    cout << "Running with Dromajo cosimulation" << endl;
+    string ncpus_str = "--ncpus=" + to_string(ncpus);
```
dpetrisko, line 17: This is really clever, I like it

bp_top/test/tb/bp_tethered/testbench.v
```diff
@@ -292,43 +286,39 @@ bind bp_be_top
...
+     ,.commit_instr_i(be_calculator.commit_pkt.instr)
+
+     ,.rd_w_v_i(be_checker.scheduler.wb_pkt.rd_w_v)
+     ,.rd_addr_i(be_checker.scheduler.wb_pkt.rd_addr)
+     ,.rd_data_i(be_checker.scheduler.wb_pkt.rd_data)
+
+     ,.interrupt_v_i(be_mem.csr.trap_pkt_cast_o._interrupt)
+     ,.cause_i(be_mem.csr.trap_pkt_cast_o.cause)
+     // TODO: Find a way to access the finish_o signals for each core
+     ,.finish_o()
```
dpetrisko, line 318: Not sure what the best thing is here, either

bp_common/test/cfg/sample.cfg
farzamgl: I'll add the SAMPLE_MEMSIZE flag back in the Makefiles(for smaller checkpoint files). But just to clarify, the setting here is for initializing the Dromajo instance in cosimulation. Which has no overhead and there is no point in it being smaller than the memory size.

bp_top/test/common/dromajo_cosim.cpp
farzamgl: I think ultimately we can use this kind of parameter passing to not use an input cfg file for cosim.

bp_top/test/tb/bp_tethered/testbench.v
farzamgl: Yeah, I'm still stuck on what to do with this problem. Right now it only affects the profiler information dumping. But it's really annoying that I can't output to testbench signals. Only if there was a way to tag a constant number to a bind instance.

farzamgl:
> Looks good! Minor comments. What programs have you tested this with?

atomic_queue_demo(2, 4, 8)

bp_common/test/cfg/sample.cfg
dpetrisko: The dromajo checkpoint memory size == the mem checkpoint size, isn’t it? So the nbf loader takes longer?

farzamgl: Yes, if we're loading in the memory with nbf loader.

bp_top/test/tb/bp_tethered/testbench.v
taylor-bsg: Not totally sure what you are trying to do here; but here's some code. ;-)

// Code your testbench here
// or browse Examples

`define QUALIFIER wor
//`define QUALIFIER wire


module myctr #(parameter start_p=0) (input clk_i);

  logic [9:0] ctr;
  
  initial 
    begin
      ctr = start_p;
    end
  
  always @(posedge clk_i)
	begin
      $display("%m: %b\n",ctr);
      ctr <= ctr + 1;
    end
      
endmodule

module bindme(input [9:0] ctr_i, output `QUALIFIER   finished_wor_o);
  
  assign finished_wor_o = (ctr_i[9] === 1);
  
endmodule

module main();
  
  logic clk_i;
  
  initial
    begin
      clk_i = 0;      
    end
  
  always
    begin
      while (1)
        begin
	        #20;
    	    clk_i = ~clk_i;
        end
    end
  
  `QUALIFIER finished_wor_lo;
  
  myctr #(20) ctr0(clk_i) ;
  myctr #(10) ctr1(clk_i) ;
  myctr #(0)  ctr2(clk_i) ;
  
  bind myctr bindme zoop(.ctr_i(ctr), .finished_wor_o(main.finished_wor_lo));
  
  always @(negedge clk_i)
    if (finished_wor_lo)
      $finish();
    
endmodule

PR #538, opened 2020-06-10
Unlocking different domains using config loader

Shashank-Vijay:
Added support for:

- Enabling different domains using mmio_cfg_loader
- Using enabled domain data to identify address access faults
- Checking access to domains other than current in host


bp_common/src/v/bp_pma.v
```diff
@@ -6,15 +6,59 @@ module bp_pma
...
+     ,.reset_val_p(1)
+     )
+     domain_reg
+     (.clk_i(clk_i)
+     ,.reset_i(reset_i)
+     ,.en_i(domain_w_v_i)
+     // We want the 0th domain to be enabled always
+     ,.data_i((domain_i | 8'h1))
+     ,.data_o(domain_data_r)
+     );
```
dpetrisko, line 44: domain_data_o = domain_data_r | 8'h1;

This way, we don't need to write it to take effect

bp_be/src/v/bp_be_mem/bp_be_mem_top.v
```diff
@@ -317,13 +319,28 @@ bp_tlb
...
+assign cfg_domain_data_o = domain_data;
+assign cfg_sac_data_o = sac_data;
+
 bp_pma
  #(.bp_params_p(bp_params_p))
  pma
-  (.ptag_v_i(dtlb_r_v_lo)
+  (.clk_i(clk_i)
+   ,.reset_i(reset_i)
+
```
dpetrisko, line 333: add cfg_bus port to this module.  We prefer the reads/writes to CSRs to happen directly off the cfg bus

bp_top/test/common/bp_nonsynth_host.v
```diff
@@ -124,6 +126,9 @@ always_ff @(negedge clk_i)
     end
     if (getchar_data_cmd_v & io_cmd_v_i)
       pop();
+
+    if (io_cmd_v_i & (domain_id != '0))
+      $display("Warning: Accesing domain: %0h", domain_id);
```
dpetrisko, line 131: Accesing illegal domain: %0h.  Sending loopback message!

bp_top/test/common/bp_nonsynth_host.v
```diff
@@ -159,14 +165,29 @@ bsg_two_fifo
...
 
+assign domain_io_resp_lo =
+  '{header: '{msg_type              : io_cmd_cast_i.header.msg_type
+              ,addr                 : io_cmd_cast_i.header.addr
+              ,payload: '{state        : io_cmd_cast_i.header.payload.state
+                         ,way_id       : io_cmd_cast_i.header.payload.way_id
+                         ,prefetch     : io_cmd_cast_i.header.payload.prefetch
+                         ,uncached     : io_cmd_cast_i.header.payload.uncached
+                         ,speculative  : io_cmd_cast_i.header.payload.speculative
+                         ,lce_id       : lce_id_width_p'(3)
```
dpetrisko, line 185: ?

sripathi-muralitharan: Trying to create some message with a non existent LCE ID so that it doesn't get inferred incorrectly as a command to some other place.

dpetrisko: I see.  Should just forward back the header.  In fact core_io_resp should as well (since that's what it's doing anyway

dpetrisko: The LCE id is used to identify the requester.  It probably isn't checked in this instance, but a CCE might be confused if it receives an LCE without an outstanding request

sripathi-muralitharan: We were afraid that by sending back the actual header from the message, the response arbitration can return this data back to the module from where it originated which might cause issues in blocking because these messages might not be dequeued. If the response arbiter doesn't find the LCE ID to be valid, then the message is just dropped on the response path. But then, we did this with the softcore so we didn't really worry about the CCE much. I guess you are right about the CCE part.

dpetrisko: The response absolutely should go back to the sender, that's the whole point of the loopback :)

dpetrisko: Otherwise, a cache sends a request to a bad domain, it doesn't return.  Now we just have an outstanding credit forever, and eventual deadlock

dpetrisko: This problem should also show up in the UCE

sripathi-muralitharan: I guess since we didn't stress this too much that the credit problem didn't show up. Makes sense to me now.

dpetrisko: BTW, were you writing a C test for this?  That would be a good case to add, store then fence

sripathi-muralitharan: Yeah, wrote a preliminary test in this which stores to addresses belonging to these domains. Need to make a better version of that test though.

dpetrisko:
LGTM for openpiton stuff.  I think there are some simplifications we'll need before merging into mainline

PR #539, opened 2020-06-12
Partial Fill / Eviction Cache Service Interface Modification

Songchun-Li:
Rebase the work of IF modification for partial fill and UCE FSM modification for multi-cycle fill/eviction to me_dev. Tested on unicore with multicycle-fill, and on single core with full cache block fill.

### DIFF
--- a/bp_be/src/v/bp_be_mem/bp_be_dcache/bp_be_dcache.v
+++ b/bp_be/src/v/bp_be_mem/bp_be_dcache/bp_be_dcache.v
@@ -80,23 +80,25 @@ module bp_be_dcache
   import bp_common_aviary_pkg::*;
  #(parameter bp_params_e bp_params_p = e_bp_inv_cfg
    `declare_bp_proc_params(bp_params_p)
-   `declare_bp_cache_service_if_widths(paddr_width_p, ptag_width_p, dcache_sets_p, dcache_assoc_p, dword_width_p, dcache_block_width_p, dcache)
+   `declare_bp_cache_service_if_widths(paddr_width_p, ptag_width_p, dcache_sets_p, dcache_assoc_p, dword_width_p, dcache_block_width_p, dcache_fill_width_p, dcache)
 
     , parameter debug_p=0
     , parameter lock_max_limit_p=8
 
+    , localparam way_id_width_lp=`BSG_SAFE_CLOG2(dcache_assoc_p)
     , localparam cfg_bus_width_lp= `bp_cfg_bus_width(vaddr_width_p, core_id_width_p, cce_id_width_p, lce_id_width_p, cce_pc_width_p, cce_instr_width_p)
-    , localparam block_size_in_words_lp = dcache_assoc_p
+    , localparam block_size_in_banks_lp = dcache_assoc_p
     , localparam bank_width_lp = dcache_block_width_p / dcache_assoc_p
     , localparam num_dwords_per_bank_lp = bank_width_lp / dword_width_p
     , localparam bypass_data_mask_width_lp = (dword_width_p >> 3)
     , localparam data_mem_mask_width_lp = (bank_width_lp >> 3)
     , localparam byte_offset_width_lp = `BSG_SAFE_CLOG2(bank_width_lp>>3)
-    , localparam word_offset_width_lp = `BSG_SAFE_CLOG2(block_size_in_words_lp)
-    , localparam block_offset_width_lp=(word_offset_width_lp+byte_offset_width_lp)
+    , localparam bank_offset_width_lp = `BSG_SAFE_CLOG2(block_size_in_banks_lp)
+    , localparam block_offset_width_lp=(bank_offset_width_lp+byte_offset_width_lp)
     , localparam index_width_lp=`BSG_SAFE_CLOG2(dcache_sets_p)
     , localparam ptag_width_lp=(paddr_width_p-bp_page_offset_width_gp)
-    , localparam way_id_width_lp=`BSG_SAFE_CLOG2(dcache_assoc_p)
+    , localparam block_size_in_fill_lp = dcache_block_width_p / dcache_fill_width_p
+    , localparam fill_size_in_bank_lp = dcache_fill_width_p / bank_width_lp
 
     , localparam dcache_pkt_width_lp=`bp_be_dcache_pkt_width(bp_page_offset_width_gp,dword_width_p)
     , localparam tag_info_width_lp=`bp_be_dcache_tag_info_width(ptag_width_lp)
@@ -137,6 +139,7 @@ module bp_be_dcache
     , output cache_req_metadata_v_o
 
     , input cache_req_complete_i
+    , input cache_req_critical_i
 
     // data_mem
     , input data_mem_pkt_v_i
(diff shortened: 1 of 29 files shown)


dpetrisko:
@muwyse , I'll review this thoroughly, but take a look. Especially think about how it will relate to serialization / deserialization in the LCEs

bp_me/src/v/lce/bp_lce.v
```diff
@@ -112,6 +114,8 @@ module bp_lce
       $error("LCE block width must be a whole number of bytes and power of two");
     assert(block_width_p <= 1024) else
       $error("LCE block width must be no greater than 128 bytes");
+    assert(fill_width_p != block_width_p) else
+      $error("LCE block width must be equal to fill width. Partial fill is not supported");
```
muwyse, line 118: should this assertion be "==" ?

bp_be/src/v/bp_be_mem/bp_be_dcache/bp_be_dcache.v
```diff
@@ -989,36 +1002,51 @@ module bp_be_dcache
...
+
     assign data_mem_data_li[i] = wbuf_yumi_li
       ? {num_dwords_per_bank_lp{wbuf_entry_out.data}}
       : lce_data_mem_write_data[i];
-  
+
     assign data_mem_mask_li[i] = wbuf_yumi_li
       ? wbuf_mask
-      : {data_mem_mask_width_lp{1'b1}};
+      : {data_mem_mask_width_lp{data_mem_write_bank_mask[i]}}; // use fill_index to generate write_mask
```
muwyse, line 1019: put comments on line above, not inline

bp_be/src/v/bp_be_mem/bp_be_dcache/bp_be_dcache.v
```diff
@@ -80,32 +80,36 @@ module bp_be_dcache
...
+
+    , parameter writethrough_p=0
+    , parameter debug_p=0
+
     , parameter lock_max_limit_p=8
 
+    , localparam way_id_width_lp=`BSG_SAFE_CLOG2(dcache_assoc_p)
     , localparam cfg_bus_width_lp= `bp_cfg_bus_width(vaddr_width_p, core_id_width_p, cce_id_width_p, lce_id_width_p, cce_pc_width_p, cce_instr_width_p)
-    , localparam block_size_in_words_lp = dcache_assoc_p
+    , localparam block_size_in_banks_lp = dcache_assoc_p
```
muwyse, line 92: if this is being set to dcache_assoc_p, perhaps get rid of block_size_in_banks_lp and use dcache_assoc_p directly.

bp_be/src/v/bp_be_mem/bp_be_dcache/bp_be_dcache.v
```diff
@@ -80,32 +80,36 @@ module bp_be_dcache
...
-   
-    , parameter debug_p=0 
+   `declare_bp_cache_service_if_widths(paddr_width_p, ptag_width_p, dcache_sets_p, dcache_assoc_p, dword_width_p, dcache_block_width_p, dcache_fill_width_p, dcache)
+
+    , parameter writethrough_p=0
+    , parameter debug_p=0
+
     , parameter lock_max_limit_p=8
 
+    , localparam way_id_width_lp=`BSG_SAFE_CLOG2(dcache_assoc_p)
```
muwyse, line 90: a more canonical name would be lg_dcache_assoc_lp

bp_common/src/include/bp_common_cache_service_if.vh
```diff
@@ -100,18 +100,19 @@ typedef enum logic [1:0] {
...
+    logic [block_data_width_mp/fill_width_mp-1:0] fill_index;                                \
+    bp_cache_data_mem_opcode_e                    opcode;                                   \
   }  bp_``cache_name_mp``_data_mem_pkt_s
 
-`define bp_cache_data_mem_pkt_width(sets_mp, ways_mp, data_width_mp) \
-  (`BSG_SAFE_CLOG2(sets_mp)+`BSG_SAFE_CLOG2(ways_mp)+data_width_mp \
-   +`bp_cache_data_mem_opcode_width)
+`define bp_cache_data_mem_pkt_width(sets_mp, ways_mp, block_data_width_mp, fill_width_mp) \
+  (`BSG_SAFE_CLOG2(sets_mp)+`BSG_SAFE_CLOG2(ways_mp)+fill_width_mp \
+   +block_data_width_mp/fill_width_mp+`bp_cache_data_mem_opcode_width)
```
muwyse, line 115: add parentheses around division for clarity on order of operations

bp_common/src/include/bp_common_cache_service_if.vh
```diff
@@ -100,18 +100,19 @@ typedef enum logic [1:0] {
...
-    logic [`BSG_SAFE_CLOG2(ways_mp)-1:0]      way_id;                                   \
-    logic [data_width_mp-1:0]                 data;                                     \
-    bp_cache_data_mem_opcode_e                opcode;                                   \
+`define declare_bp_cache_data_mem_pkt_s(sets_mp, ways_mp, block_data_width_mp, fill_width_mp, cache_name_mp) \
+  typedef struct packed                                                                     \
+  {                                                                                         \
+    logic [`BSG_SAFE_CLOG2(sets_mp)-1:0]          index;                                    \
+    logic [`BSG_SAFE_CLOG2(ways_mp)-1:0]          way_id;                                   \
+    logic [fill_width_mp-1:0]                     data;                                     \
+    logic [block_data_width_mp/fill_width_mp-1:0] fill_index;                                \
```
muwyse, line 109: Add parentheses around division for clarity on order of operations.

muwyse: Add additional check in nonsynth_if_verif that block_data_width_mp divides evenly by fill_width_mp. I'm assuming we won't want to support a fill width of, e.g., 40-bits when the block width is 512-bits, right? In other words, we will support fill width of like 16, 32, or 64-bits and block widths of 128, 256, 512, which gives us a nice integer ratio of block width : fill width.

dpetrisko:
>I think the idea is we want to support something like critical word first fill. But, what are the constraints? As the xCE what rules must I follow in terms of ordering fills for a block? What constraints are there in the relationship between dword_width, fill_width, and block_width? How do these relate to the allowed bank_widths and banking schemes in the cache? I think it could be beneficial to limit the allowable configurations to those that make sense. For example, bank_width must be an integer multiple of dword_width, or, fill_width == bank_width.

Excellent questions.  We have the descriptions for this in the Cache Service Interface Doc, but we should put them in the interface specification with this PR

bp_me/src/v/lce/bp_lce.v
Songchun-Li: Currently, multi-cycle fill is only implemented in the UCE. And I made the interface modification on the LCE only, so it still works with fill_width_p == block_width but not with partial fill. That's why an error is asserted when fill_width_p != block_width_p.

dpetrisko: Yes, that's what mark is saying.  This will raise the assertion if fill_width == block_width...

Songchun-Li: Got it. I just missed the "else" and have the contrary understanding

Songchun-Li:
- Request changed are maded. 
- Brief description of multi-cycle fill/eviction are appended in the comment at the top of both d-cache and i-cache.
- Additional checks on the condition, fill_width = N*bank_width <= block_width is added in nonsynth_if_verif

bp_fe/syn/flist.vcs
```diff
@@ -241,7 +241,7 @@ $BP_ME_DIR/src/v/wormhole/bp_me_wormhole_packet_encode_mem_cmd.v
 $BP_ME_DIR/src/v/wormhole/bp_me_wormhole_packet_encode_mem_resp.v
 ## TOP
 $BP_TOP_DIR/src/v/bp_nd_socket.v
-$BP_TOP_DIR/src/v/bp_cacc_vdp.v
+# $BP_TOP_DIR/src/v/bp_cacc_vdp.v
```
dpetrisko, line 244: Need to fix the accelerator before merging

Songchun-Li:
Finish rebasing to the latest me_dev. And the accelerator and some other module in flist are fixed

PR #540, opened 2020-06-12
Cleanup BE dev in prep for future changes

dpetrisko:
This PR does not introduce functional changes.  It is in place in preparation for 2 cycle loads and compressed instructions, which require a different method for exceptions and interrupts travelling down the pipe.

Main changes:
- Formalize PTW interface with ptw_miss_pkt and ptw_fill_pkt
- Interrupts now happen directly in mem3 if there are no currently committing instructions
- Exceptions now aggregate in the exception pipe, which will make it easier to properly handle instruction misaligned exceptions generated by the control pipe
- This PR also removes the unmaintained unified CLINT, CLINT nodes, and deprecated MC link


dpetrisko:
It also boots linux.

Note, I'll hold off merging to dev until we have the two cycle loads, since there are some hacky round trip paths and arbitration needed to support the current 3 cycle load operation.  But I'm trying to break up the PR to make it easier to review (it's still substantial I realize)

bp_be/src/v/bp_be_calculator/bp_be_calculator_top.v
```diff
@@ -251,40 +256,73 @@ bp_be_pipe_mul
...
+     ,.kill_ex3_i(exc_stage_r[2].poison_v) 
+  
+     ,.decode_i(reservation_r.decode)
+     ,.pc_i(reservation_r.pc)
+     ,.instr_i(reservation_r.instr)
+     ,.rs1_i(reservation_r.rs1)
+     ,.rs2_i(reservation_r.rs2)
+     ,.imm_i(reservation_r.imm)
+  
+     ,.mmu_cmd_o(mmu_cmd_o)
```
farzamgl, line 277: Can we change these to mem_* to match the other ports?

bp_be/src/v/bp_be_mem/bp_be_mem_top.v
```diff
@@ -458,23 +424,26 @@ assign dcache_cmd_v    = mmu_cmd_v_i;
...
-assign dcache_pkt_v    = (ptw_busy)? ptw_dcache_v : dcache_cmd_v;
-
 always_comb
   begin
     if(ptw_busy) begin
-      dcache_pkt = ptw_dcache_pkt;
+      dcache_pkt_v    = ptw_dcache_v;
+      dcache_pkt      = ptw_dcache_pkt;
+      dcache_ptag     = ptw_dcache_ptag;
+      dcache_tlb_miss = ~ptw_dcache_ptag_v;
```
farzamgl, line 433: Why? PTW accesses are physical.

dpetrisko: Yeah, it's not strictly necessary.  However, the PTW could fail PMA/PMP in which case I don't know what to do. AFAIK, RISC-V doesn't have the concept of a double fault to handle this case

bp_be/src/v/bp_be_calculator/bp_be_calculator_top.v
dpetrisko: Yeah good call

PR #541, opened 2020-06-16
Unified LCE and write buffer hotfix

muwyse:
This PR introduces a unified LCE to replace the FE and BE specific LCEs.

This change also provides a fix for the dcache write buffer arbitration to disallow data_mem packets from the LCE when until the buffer is empty. This is conservative, but prevents the buffer from overflowing, which can happen if an LCE presents a data_mem packet to the dcache while it is processing a series of store hits.


dpetrisko:
Can we write a test for the write buffer bug? Seemed very difficult to track down

muwyse:
> Can we write a test for the write buffer bug? Seemed very difficult to track down

It may be possible, although might need to think about how. The tricky part is that it would require precisely controlling the timing of stores and data_mem packets to the cache. So, load a cache block, then do a sequence of store hits and injecting a data_mem packet to compete with the write buffer. And then somehow measuring the result, which I suppose is just loading back the stored values?

dpetrisko:
>It may be possible, although might need to think about how. The tricky part is that it would require precisely controlling the timing of stores and data_mem packets to the cache. So, load a cache block, then do a sequence of store hits and injecting a data_mem packet to compete with the write buffer. And then somehow measuring the result, which I suppose is just loading back the stored values?

I think the following 2 core test should work.  In a loop:
- core X reads line B
- core X reads line A word 0-7
- core X increments words 0-7
- core X stores words 0-7
- core Y reads line B

Possibly need to add more core Ys to trigger it, though.

muwyse:
> I think the following 2 core test should work. In a loop:
> 
> * core X reads line B
> * core X reads line A word 0-7
> * core X increments words 0-7
> * core X stores words 0-7
> * core Y reads line B
> 
> Possibly need to add more core Ys to trigger it, though.

I think an assertion on the overflow condition will be far more effective. A test like this is too highly dependent on timing and likely needs to be mostly hand-written assembly, which is burdensome to write and maintain. There is also no guarantee it remains effective as timings of the components change (dcache should be stable, but all the ME components and latencies through networks might change depending on microarch, network org, etc.)

dpetrisko:
> I think an assertion on the overflow condition will be far more effective. A test like this is too highly dependent on timing and likely needs to be mostly hand-written assembly, which is burdensome to write and maintain. There is also no guarantee it remains effective as timings of the components change (dcache should be stable, but all the ME components and latencies through networks might change depending on microarch, network org, etc.)

That's a good point. Really, this type of backpressure should be an addition to the D$ testbench, where we can precisely control the timings.  Won't hold up this PR for that that

Issue #393, opened 2020-03-16
Change mmu_cmd.vaddr to split into page_offset and vtag


taylor-bsg:
https://github.com/black-parrot/black-parrot/blob/9b3cea5c453d307f893044be6cd5c09f907cab9e/bp_be/src/v/bp_be_mem/bp_be_mem_top.v#L476

Should just take low 12 bits of vaddr, rather than using these fields, since we are just using this for translation.

dpetrisko:
Depends on what parameters we want to fix.  I think it's reasonable to have a dcache which only accepts a 12 bit vindex and reconfigures sets, based on cacheline size internally to support this, so this change is probably fine.

dpetrisko:
There's also the question of allowing synonyms in the I$...

taylor-bsg:
My point is, this code doesn't need to consider the index or the offset, it is only used for VM, which has no relationship to the cache parameters. There is no benefit to writing the code this way. It can just blindly take the low 12 bits. In all other places except this one, it just uses a constant which evaluates to 12.  So defining it this way is just creating an unnecessary mismatch in width for no benefit.

dpetrisko:
Oh, I misinterpreted and thought we were talking about the D$ packet, not the mmu_cmd.  Agree

dpetrisko:
So based off a discussion with Andrew W, this is actually incorrect behavior. We need to pass the entire 64-bit effective address in the mem_cmd, so that 
1) we can check that bits 39-64 match bit 38 in the case of translation enabled 
2) passthrough bit 39 in the case of translation disabled 

https://groups.google.com/a/groups.riscv.org/d/msgid/isa-dev/CABXpatqk%2BGinxy%3DV_u7K5iYB6N_wYuLeNA5YZSEhPy0BEqU6dQ%40mail.gmail.com?utm_medium=email&utm_source=footer

taylor-bsg:
I think you have attached this comment to the wrong issue... this comment is saying that the page offset field for VM should be populated just by taking the low 12 bits instead of this expression.

M

dpetrisko:
No, this is the right issue. The mmu_cmd struct needs to change to be 64 bit address, not split into page offset and vtag as the issue title says

dpetrisko:
fixed by #543

Issue #394, opened 2020-03-16
Don't recompute page_offset_width_p


taylor-bsg:
https://github.com/black-parrot/black-parrot/blob/9b3cea5c453d307f893044be6cd5c09f907cab9e/bp_be/src/v/bp_be_mem/bp_be_mem_top.v#L30

Should use bp_page_offset_width_gp instead of computing this.

taylor-bsg:
This and the other bug are the probably the only two places that are holding back VM from working as you change the cache block size.

dpetrisko:
fixed by #658

PR #543, opened 2020-06-18
Using effective addresses instead of virtual addresses in the MMU

sripathi-muralitharan:
This PR introduces the fix for a bug that was found in bare metal execution where the specification is to use 64-bit effective addresses instead of 39-bit virtual addresses.

This PR also closes issues #393 and #394.

### DIFF
--- a/bp_be/src/v/bp_be_calculator/bp_be_pipe_mem.v
+++ b/bp_be/src/v/bp_be_calculator/bp_be_pipe_mem.v
@@ -20,7 +20,7 @@ module bp_be_pipe_mem
    // Generated parameters
    , localparam decode_width_lp        = `bp_be_decode_width
    , localparam exception_width_lp     = `bp_be_exception_width
-   , localparam mmu_cmd_width_lp       = `bp_be_mmu_cmd_width(vaddr_width_p)
+   , localparam mmu_cmd_width_lp       = `bp_be_mmu_cmd_width
    , localparam csr_cmd_width_lp       = `bp_be_csr_cmd_width
    , localparam mem_resp_width_lp      = `bp_be_mem_resp_width(vaddr_width_p)
 
@@ -59,7 +59,7 @@ module bp_be_pipe_mem
    );
 
 // Declare parameterizable structs
-`declare_bp_be_mmu_structs(vaddr_width_p, ppn_width_p, lce_sets_p, cce_block_width_p/8)
+`declare_bp_be_mmu_structs
 
 // Cast input and output ports 
 bp_be_decode_s    decode;
@@ -101,7 +101,7 @@ bsg_shift_reg
 
 logic [reg_data_width_lp-1:0] offset;
 
-assign offset = decode.offset_sel ? '0 : imm_i[0+:vaddr_width_p];
+assign offset = decode.offset_sel ? '0 : imm_i;
 
 assign mem1_cmd_v = (decode.dcache_r_v | decode.dcache_w_v) & ~kill_ex1_i;
 
@@ -109,11 +109,13 @@ wire fe_exc_v = (decode.fu_op == e_op_instr_misaligned)
                 | (decode.fu_op == e_op_instr_access_fault)
                 | (decode.fu_op == e_op_instr_page_fault)
                 | (decode.fu_op == e_itlb_fill);
+
+wire pc_sigext = pc_i[38];
 always_comb 
   begin
     mem1_cmd.mem_op   = decode.fu_op;
     mem1_cmd.data     = rs2_i;
-    mem1_cmd.vaddr    = fe_exc_v ? pc_i : (rs1_i + offset);
+    mem1_cmd.eaddr    = fe_exc_v ? {{25{pc_sigext}}, pc_i} : (rs1_i + offset);
   end
 
 assign csr_cmd_v_o = csr_cmd_v_lo & ~kill_ex3_i;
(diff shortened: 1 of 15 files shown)


bp_be/src/v/bp_be_calculator/bp_be_pipe_mem.v
```diff
@@ -101,19 +101,21 @@ bsg_shift_reg
...
+assign offset = decode.offset_sel ? '0 : imm_i;
 
 assign mem1_cmd_v = (decode.dcache_r_v | decode.dcache_w_v) & ~kill_ex1_i;
 
 wire fe_exc_v = (decode.fu_op == e_op_instr_misaligned)
                 | (decode.fu_op == e_op_instr_access_fault)
                 | (decode.fu_op == e_op_instr_page_fault)
                 | (decode.fu_op == e_itlb_fill);
+
+wire pc_sigext = pc_i[38];
```
dpetrisko, line 113: vaddr_width_p-1

bp_be/src/v/bp_be_calculator/bp_be_pipe_mem.v
```diff
@@ -101,19 +101,21 @@ bsg_shift_reg
...
 always_comb 
   begin
     mem1_cmd.mem_op   = decode.fu_op;
     mem1_cmd.data     = rs2_i;
-    mem1_cmd.vaddr    = fe_exc_v ? pc_i : (rs1_i + offset);
+    mem1_cmd.eaddr    = fe_exc_v ? {{25{pc_sigext}}, pc_i} : (rs1_i + offset);
```
dpetrisko, line 118: eaddr_width-vaddr_width

bp_be/src/v/bp_be_mem/bp_be_mem_top.v
```diff
@@ -488,7 +504,7 @@ always_comb
     else begin
       // We assume that mem op == dcache op
       dcache_pkt.opcode      = bp_be_dcache_opcode_e'(mmu_cmd.mem_op);
-      dcache_pkt.page_offset = {mmu_cmd.vaddr.index, mmu_cmd.vaddr.offset};
+      dcache_pkt.page_offset = mmu_cmd.eaddr[bp_page_offset_width_gp-1:0];
```
dpetrisko, line 507: slice up e.g. 0+:page_offset_width_p

bp_be/src/v/bp_be_mem/bp_be_mem_top.v
```diff
@@ -311,7 +311,7 @@ bp_tlb
 
    ,.v_i(dtlb_r_v | dtlb_w_v)
    ,.w_i(dtlb_w_v)
-   ,.vtag_i((dtlb_w_v)? dtlb_w_vtag : dtlb_r_vtag)
```
dpetrisko, line 314: Looking at this, I think a cleaner solution is to only have the TLB deal with vtags.  

This means taking the translation_en bypass out of the TLB and bypassing around it.  That way, the ports stay the same width.  That'll also prevent needing that eaddr_page_fault flop below (because the TLB won't miss if it's not read at all)

bp_be/src/v/bp_be_mem/bp_be_mem_top.v
```diff
@@ -464,9 +464,25 @@ end
 wire data_priv_page_fault = ((priv_mode_lo == `PRIV_MODE_S) & ~mstatus_sum_lo & dtlb_r_entry.u)
                               | ((priv_mode_lo == `PRIV_MODE_U) & ~dtlb_r_entry.u);
 wire data_write_page_fault = is_store_r & (~dtlb_r_entry.w | ~dtlb_r_entry.d);
-
-assign load_page_fault_v  = mmu_cmd_v_r & dtlb_r_v_lo & translation_en_lo & ~is_store_r & data_priv_page_fault;
-assign store_page_fault_v = mmu_cmd_v_r & dtlb_r_v_lo & translation_en_lo & is_store_r & (data_priv_page_fault | data_write_page_fault);
+wire eaddr_page_fault = (mmu_cmd_v_r & (mmu_cmd_r.eaddr[rv64_eaddr_width_gp-1:vaddr_width_p] != {25{mmu_cmd_r.eaddr[vaddr_width_p-1]}}));
```
dpetrisko, line 467: eaddr_width_p-vaddr_width_p for bit replication

bp_be/src/v/bp_be_mem/bp_be_mem_top.v
```diff
@@ -504,15 +520,17 @@ assign load_access_fault_v  = load_op_tl_lo & (mode_fault_v | did_fault_v | sac_
...
 
 // D-TLB connections
+// We use the tag generated from the effective address for reads.
+// For writes, we zero extend the vtag generated from the PTW
 assign dtlb_r_v     = dcache_cmd_v & ~fencei_cmd_v;
-assign dtlb_r_vtag  = mmu_cmd.vaddr.tag;
+assign dtlb_r_etag  = mmu_cmd.eaddr[rv64_eaddr_width_gp-1:bp_page_offset_width_gp];
 assign dtlb_w_v     = ptw_tlb_w_v & ~itlb_not_dtlb_resp;
-assign dtlb_w_vtag  = ptw_tlb_w_vtag;
+assign dtlb_w_etag  = {25'h0, ptw_tlb_w_vtag};
```
dpetrisko, line 528: Now's a good to make the TLB 1R1W instead of 1RW, since the backing memories are 1r1w... We notice it here because the read port needs to be etag whereas the write port needs to be vtag

bp_be/src/v/bp_be_mem/bp_be_mem_top.v
dpetrisko: It'll also get rid of the concept of "etag"

sripathi-muralitharan: Do you mean the following has to be done:
1. Remove the output mux out of the TLB so that the TLB still uses the vtags only. 
2. Control the read valid on the TLB based on eaddr_page_fault (which will now be calculated on an incoming mmu_cmd instead of the one cycle delayed version)?

taylor-bsg:
Dan, regarding TLB being 1R1W, we should definitely support the TLB being
implemented as 1RW ram..

dpetrisko:
Yeah that’s reasonable

PR #545, opened 2020-06-18
Redirect basejump to upstream master

dpetrisko:
To solve integration issues with other projects using basejump stl, our new policy is to track master as closely as possible.  Leaving the BlackParrot basejump fork up for legacy builds

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -24,8 +24,8 @@
 
 [submodule "external/basejump_stl"]
   path = external/basejump_stl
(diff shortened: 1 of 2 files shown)


taylor-bsg:
Sounds good! Note there is an errant unrelated file change though.

dpetrisko:
Heh, it actually is related.  This is fixing the code to be compliant without having: https://github.com/bespoke-silicon-group/basejump_stl/pull/249

Otherwise, there's a syntax error

black-parrot:
got it!

PR #546, opened 2020-06-23
Cache engine PPA fixes

dpetrisko:
- Uncached stores now have 1/cycle throughput in LCE
- Removing bypass paths from cache engines (these paths were critical in both unicore and multicore)
- Clarified the v_r semantics which will help with the negedge 2-cycle loads

These will cause a minor IPC regression on writebacks and flushes, but we can recover it with bypass flops using the same v_r semantics if we find the performance impact to be too great

### DIFF
--- a/bp_me/src/v/lce/bp_lce_cmd.v
+++ b/bp_me/src/v/lce/bp_lce_cmd.v
@@ -131,7 +131,6 @@ module bp_lce_cmd
     ,e_wb
     ,e_wb_dirty_rd
     ,e_wb_dirty_send
-    ,e_null_wb
     ,e_coh_ack
   } lce_cmd_state_e;
   lce_cmd_state_e state_r, state_n;
@@ -147,12 +146,10 @@ module bp_lce_cmd
   bp_cache_data_mem_pkt_s data_mem_pkt;
   bp_cache_tag_mem_pkt_s tag_mem_pkt;
   bp_cache_stat_mem_pkt_s stat_mem_pkt;
-  bp_cache_stat_info_s stat_mem_cast_i;
 
   assign data_mem_pkt_o = data_mem_pkt;
   assign tag_mem_pkt_o = tag_mem_pkt;
   assign stat_mem_pkt_o = stat_mem_pkt;
-  assign stat_mem_cast_i = stat_mem_i;
 
   // sync done register - goes high when all sync command/acks complete
   logic sync_done_en, sync_done_li;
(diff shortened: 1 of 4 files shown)


bp_me/src/v/lce/bp_lce_cmd.v
```diff
@@ -501,27 +545,27 @@ module bp_lce_cmd
...
+                  : stat_buf_v_r & lce_resp_v_o
+                    ? e_ready
+                    : e_wb;
+          
       end
 
       // Writeback dirty block - read from data memory, write to stat memory to clear dirty bit
       e_wb_dirty_rd: begin
 
+        /* TODO: NEED TO NOT USE STAT_PKT CAST DIRECTLY */
```
muwyse, line 568: What is meant by this comment? Why can't stat_pkt cast be used directly?

bp_me/src/v/lce/bp_lce_req.v
```diff
@@ -172,38 +176,38 @@ module bp_lce_req
...
-            , e_uc_load: begin
+            e_uc_store: begin
+              lce_req_v_o = lce_req_ready_i;
+
+              lce_req.data[0+:dword_width_p] = cache_req.data[0+:dword_width_p];
+              lce_req.header.size = bp_mem_msg_size_e'(cache_req.size);
+              lce_req.header.addr = cache_req.addr;
+              lce_req.header.msg_type = e_lce_req_type_uc_wr;
+            end
+            e_uc_load: begin
```
muwyse, line 210: I think it is possible for lce_req_ready_i to be low on the cycle that the cache_req_i arrives. In this situation, for an uncached store, what happens? The uc store will not send because lce_req_ready_i == 0; will cache_req (cache_req_i) be valid the next cycle? If it is not valid the next cycle, won't the uc store potentially be dropped on the floor and forgotten? 

My understanding is the input is valid for a single cycle, hence why it was sent to a bypass DFF previously. If that poses a timing issue (i.e., too long a path), we probably need a cache_req_r_v signal/register and/or a way to retry the LCE request send using the buffered request info in the following cycle(s).

bp_me/src/v/lce/bp_lce_cmd.v
dpetrisko: oh sorry, vestigial comment :P

bp_me/src/v/lce/bp_lce_req.v
dpetrisko: https://github.com/black-parrot/black-parrot/blob/me_dev_feature_ppafixes/bp_me/src/v/lce/bp_lce.v#L245

It's illegal for cache_req_v_i to be high if lce_req is not ready

dpetrisko: I'll make it more explicit, though

muwyse: okay, is there an assertion that can check this? The LCE Req Module currently reports ready_o when it has credits available, but that isn't linked to lce_req_ready_i and its status. I don't think it is likely, in fact I think it might not be possible by construction in our current design, that cache_req_v_i will be high and lce_req_ready_i will be low. That either needs to be commented and checked as an illegal condition, or the module should be able to handle that case.

dpetrisko: Oh, you're totally right. I missed this.  In the UCE, what we do is condition ready on the mem port being ready.  I'll do that here for the LCE as well

PR #547, opened 2020-06-24
Enabling option to boot from synthesizable bootrom 

dpetrisko:
This is the first step in enabling execution-driven debug mode.  This PR enables program execution from uncached regions and adds a boot rom flow, which can be used to add arbitrary code execution to the boot flow, before loading the actual program in DRAM

### DIFF
--- a/bp_common/test/src/bootrom/bootrom.S
+++ b/bp_common/test/src/bootrom/bootrom.S
@@ -1,0 +1,41 @@
+
+.section .text.start
+.globl _start
+_start:
+    li x0, 0
+    li x1, 0
+    li x2, 0
+    li x3, 0
+    li x4, 0
+    li x5, 0
+    li x6, 0
+    li x7, 0
+    li x8, 0
+    li x9, 0
+    li x10, 0
+    li x11, 0
+    li x12, 0
+    li x13, 0
+    li x14, 0
+    li x15, 0
+    li x16, 0
+    li x17, 0
+    li x18, 0
+    li x19, 0
+    li x20, 0
+    li x21, 0
+    li x22, 0
+    li x23, 0
+    li x24, 0
+    li x25, 0
+    li x26, 0
+    li x27, 0
+    li x28, 0
+    li x29, 0
+    li x30, 0
+    li x31, 0
+    
+    li x31, 0x80000000
+    jr x31
+halt:
+    j halt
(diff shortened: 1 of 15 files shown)


bp_fe/src/v/bp_fe_pc_gen.v
```diff
@@ -432,7 +432,7 @@ always_comb
       end
   end
 
-assign mem_poison_o         = ~pc_gen_stage_n[1].v;
```
farzamgl, line 435: Why?

bp_fe/src/v/bp_fe_mem.v
```diff
@@ -205,10 +205,8 @@ wire is_uncached_mode = (cfg_bus_cast_i.icache_mode == e_lce_mode_uncached);
 wire mode_fault_v = (is_uncached_mode & ~uncached_li);
 // TODO: Enable other domains by setting enabled dids with cfg_bus
 wire did_fault_v = (ptag_li[ptag_width_p-1-:io_noc_did_width_p] != '0);
-// Don't allow speculative access to local tile memory
-wire local_fault_v = (ptag_li < (dram_base_addr_gp >> page_offset_width_p));
```
farzamgl, line 209: Maybe only allow access to the bootroom region instead of removing the check?
Also, should we give load access to this region in D$?

bp_common/test/src/bootrom/bootrom.S
```diff
@@ -0,0 +1,41 @@
...
+    li x24, 0
+    li x25, 0
+    li x26, 0
+    li x27, 0
+    li x28, 0
+    li x29, 0
+    li x30, 0
+    li x31, 0
+    
+    li x31, 0x80000000
```
farzamgl, line 38: Can we test this with the dpc and dret right now?

bp_fe/src/v/bp_fe_pc_gen.v
dpetrisko: This is actually not quite right, it should be flush or override.

But ~pc_gen_stage_n[1].v breaks executing from the uncached region, because what happens is:
1) We send uncached request, flush
2) execute rolled back instruction
3) pc_gen_stage_n[1].v is low, poison.  This kills the uncached load data
4) Repeat infinitely

bp_fe/src/v/bp_fe_mem.v
dpetrisko: Check is outdated now.  There's a loopback controller in the tile so that spurious requests will just return 0

bp_common/test/src/bootrom/bootrom.S
dpetrisko: Debug mode is still broken I think, will look into it in separate PR

Issue #399, opened 2020-03-19
Add Linux cosim regression test


dpetrisko:
Linux boot is roughly 265M instructions to the login prompt.  This means we could split it into roughly 30 10m instruction checkpoints and run in parallel to have a ~6 hour linux regression.  We can add this to our dev CI branch to run nightly (or have people run locally).

TODO:
- [x] Add run N instruction capability to testbench (most likely in RTL)
- [x] Add generate K checkpoints of size N capability to makefiles
- [x] Write regression script which generates the checkpoints and runs them in parallel

Issue #346, opened 2020-02-26
Unified NBF Loader


dpetrisko:
Currently, on boot we have three loaders which perform roughly the same function.

1) CFG loader, which loads configuration registers
2) NBF loader, which copies a program into DRAM
3) NBF loader (with checkpoint), which copies a checkpoint info into BP architectural state.

These are multiplexed onto a MMIO master to DMA into BlackParrot.  The MMIO config loader is just a less flexible version of the NBF loader, so it should go away entirely.  Secondly, copying the program into DRAM should be done as part of the checkpoint (Preloading can still be an option, but having all in one is useful for e.g. running traces on simulated ASIC).

PR #548, opened 2020-06-24
Misc Software Changes

farzamgl:
- Updated nbf.py so we write the memory, core data, and checkpoint state through the nbf loader(#346)
- Linux parallel cosim runs without errors(#399)
- Fixed the address CSRs' sign extension
- Fixed a bug in FE/BE handshake when loading a non M mode checkpoint
- Added command line Dromajo configuration and removed the .cfg files
- Updated perch's util functions


farzamgl:
@dpetrisko can you add the Linux CI target to the gitlab script?

bp_be/src/v/bp_be_checker/bp_be_director.v
```diff
@@ -216,8 +216,13 @@ always_comb
     fe_cmd_v = 1'b0;
     flush_o = 1'b0;
 
+    // Do not send anything on reset
```
dpetrisko, line 219: What bug is this fixing? Before boot, there won't be any instructions in the system so no messages should be sent

bp_fe/src/v/bp_fe_pc_gen.v
```diff
@@ -94,7 +94,6 @@ wire attaboy_v        = fe_cmd_v_i & (fe_cmd_cast_i.opcode == e_op_attaboy);
 wire cmd_nonattaboy_v = fe_cmd_v_i & (fe_cmd_cast_i.opcode != e_op_attaboy);
 
 wire trap_v = pc_redirect_v & (fe_cmd_cast_i.operands.pc_redirect_operands.subopcode == e_subop_trap);
-wire translation_v = pc_redirect_v & (fe_cmd_cast_i.operands.pc_redirect_operands.subopcode == e_subop_translation_switch);
```
dpetrisko, line 97: Why get rid of this?

bp_be/src/v/bp_be_checker/bp_be_director.v
farzamgl: It could send some cmds during writing the satp CSR with the cfg link.

bp_fe/src/v/bp_fe_pc_gen.v
farzamgl: This subopcode is not used in either FE or BE

bp_be/src/v/bp_be_checker/bp_be_director.v
dpetrisko: Ah, I see.  Semi-related, is privilege mode also loaded as part of the checkpoint?

bp_fe/src/v/bp_fe_pc_gen.v
dpetrisko: That's true, but we should actually use it in the BE

bp_be/src/v/bp_be_checker/bp_be_director.v
farzamgl: Yes. Into the BE.

bp_fe/src/v/bp_fe_pc_gen.v
farzamgl: In what situations?

dpetrisko:
> can you add the Linux CI target to the gitlab script?

What's the bash command to run it?

farzamgl:
> What's the bash command to run it?

`make run_psample.v SUITE=linux PROG=linux SAMPLE_INSTR_P=10000000 SAMPLE_MEMSIZE=256`

bp_fe/src/v/bp_fe_pc_gen.v
farzamgl: Alright so I'll add it back but I think it should change the shadow translation_en not the shadow priv(hence the name) right?

dpetrisko: yeah probably.  The way it's set up, it doesn't really matter since the priv mode is always just priv_mode_n

dpetrisko:
Pushed the linux regression.  The issue that'll come up is that our CI server is throttled to a certain number of cores (I think 10 per job?).  We should bring up a dedicated machine with no usage limit to really rip through the regression

PR #549, opened 2020-06-25
2-cycle load-use for D$

dpetrisko:
This PR provides the data cache with a 2-cycle load-use latency. It does so by shifting the data and tag memories to a negative edge read, splitting the extra cycle into 1/2 cycle for address calculation and 1/2 cycle for data muxing. This requires some complexity in the cache engines to synchronize between the clock edges and provide a full cycle read.  Another change is to make the data muxing in the D$ more efficient to better use that 1/2 cycle.

It's been pushed through the backend flow without additional critical paths and pushed through the FPGA flow to boot linux

### DIFF
--- a/bp_be/src/v/bp_be_mem/bp_be_dcache/bp_be_dcache.v
+++ b/bp_be/src/v/bp_be_mem/bp_be_dcache/bp_be_dcache.v
@@ -286,9 +286,9 @@ module bp_be_dcache
   logic [dword_width_p-1:0] data_tl_r;
   logic gdirty_r;
 
-  assign tl_we = v_i & cache_req_ready_i;
+  assign tl_we = v_i & ~poison_i;
   
-  always_ff @ (posedge clk_i) begin
+  always_ff @ (negedge clk_i) begin
     if (reset_i) begin
       v_tl_r <= 1'b0;
     end
@@ -330,7 +330,7 @@ module bp_be_dcache
       ,.els_p(dcache_sets_p)
     )
     tag_mem
-      (.clk_i(clk_i)
+      (.clk_i(~clk_i)
       ,.reset_i(reset_i)
       ,.v_i(tag_mem_v_li)
       ,.w_i(tag_mem_w_li)
@@ -355,7 +355,7 @@ module bp_be_dcache
         ,.els_p(dcache_sets_p*dcache_assoc_p)
         )
       data_mem
-        (.clk_i(clk_i)
+        (.clk_i(~clk_i)
         ,.reset_i(reset_i)
         ,.v_i(data_mem_v_li[i])
         ,.w_i(data_mem_w_li)
@@ -431,7 +431,7 @@ module bp_be_dcache
   // fencei does not require a ptag
   assign tv_we = v_tl_r & ~poison_i & (ptag_v_i | fencei_op_tl_r);
 
-  always_ff @ (posedge clk_i) begin
+  always_ff @(negedge clk_i) begin
     if (reset_i) begin
       v_tv_r <= 1'b0;
 
@@ -720,23 +720,23 @@ module bp_be_dcache
 
     if(load_miss_tv) begin
       cache_req_cast_o.msg_type = e_miss_load;
-      cache_req_v_o = cache_req_ready_i;
+      cache_req_v_o = cache_req_ready_i & ~poison_i;
     end
     else if(store_miss_tv | lr_miss_tv) begin
       cache_req_cast_o.msg_type = e_miss_store;
-      cache_req_v_o = cache_req_ready_i;
+      cache_req_v_o = cache_req_ready_i & ~poison_i;
     end
     else if(wt_req) begin
       cache_req_cast_o.msg_type = e_wt_store;
-      cache_req_v_o = cache_req_ready_i;
+      cache_req_v_o = cache_req_ready_i & ~poison_i;
     end
     else if(uncached_load_req) begin
       cache_req_cast_o.msg_type = e_uc_load;
-      cache_req_v_o = cache_req_ready_i;
+      cache_req_v_o = cache_req_ready_i & ~poison_i;
     end
     else if(uncached_store_req) begin
       cache_req_cast_o.msg_type = e_uc_store;
-      cache_req_v_o = cache_req_ready_i;
+      cache_req_v_o = cache_req_ready_i & ~poison_i;
     end
     else if(fencei_req) begin
       // Don't flush on fencei when coherent
@@ -765,15 +765,15 @@ module bp_be_dcache
   // output stage
   // Cache Miss Tracking logic
   logic cache_miss_r;
-  wire miss_tracker_en_li = cache_req_v_o & ~uncached_store_req & ~fencei_req & ~wt_req;
-  bsg_dff_reset_en
-   #(.width_p(1))
+  bsg_dff_reset_set_clear
+   #(.width_p(1), .clear_over_set_p(1))
    cache_miss_tracker
     (.clk_i(clk_i)
      ,.reset_i(reset_i)
 
-     ,.en_i(miss_tracker_en_li | cache_req_complete_i)
-     ,.data_i(cache_req_v_o)
+     // We don't wait for ack on uncached stores or writethrough requests
+     ,.set_i(cache_req_v_o & ~uncached_store_req & ~wt_req)
+     ,.clear_i(cache_req_complete_i)
      ,.data_o(cache_miss_r)
      );
   assign ready_o = cache_req_ready_i & ~cache_miss_r;
@@ -833,7 +833,7 @@ module bp_be_dcache
      ,.disable_overflow_warning_p(1)
      )
    lock_counter
-    (.clk_i(clk_i)
+    (.clk_i(~clk_i)
      ,.reset_i(reset_i)
 
      ,.clear_i(lock_clr)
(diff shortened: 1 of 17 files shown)


taylor-bsg:
Super cool!

bp_be/src/v/bp_be_mem/bp_be_dcache/bp_be_dcache.v
```diff
@@ -455,7 +455,7 @@ module bp_be_dcache
       addr_bank_offset_dec_tv_r <= '0;
     end
     else begin
-      v_tv_r <= tv_we;
+      v_tv_r <= tv_we & ~poison_i;
 
       if (tv_we) begin
```
muwyse, line 460: should this be (tv_we & ~poison_i), since the v_tv_r also uses that condition?

bp_be/test/tb/bp_be_dcache/wrapper.v
```diff
@@ -238,6 +238,8 @@ module wrapper
         ,.block_width_p(dcache_block_width_p)
         ,.timeout_max_limit_p(4)
         ,.credits_p(coh_noc_max_credits_p)
+        ,.data_mem_negedge_p(1)
+        ,.tag_mem_negedge_p(1)
```
muwyse, line 242: what about stat_mem_negedge_p? I see it is used below for UCE/LCE, but not passed here.

bp_top/src/v/bp_core.v
```diff
@@ -257,6 +257,8 @@ module bp_core
       ,.fill_width_p(dcache_fill_width_p)
       ,.timeout_max_limit_p(4)
       ,.credits_p(coh_noc_max_credits_p)
+      ,.data_mem_negedge_p(1)
+      ,.tag_mem_negedge_p(1)
```
muwyse, line 261: stat_mem_negedge_p?

bp_top/src/v/bp_unicore.v
```diff
@@ -198,7 +198,10 @@ module bp_unicore
      ,.assoc_p(dcache_assoc_p)
      ,.sets_p(dcache_sets_p)
      ,.block_width_p(dcache_block_width_p)
-     ,.fill_width_p(dcache_fill_width_p))
+     ,.fill_width_p(dcache_fill_width_p)
+     ,.data_mem_negedge_p(1)
+     ,.tag_mem_negedge_p(1)
```
muwyse, line 203: stat_mem_negedge_p?

bp_be/test/tb/bp_be_dcache/wrapper.v
```diff
@@ -351,6 +353,8 @@ module wrapper
      ,.assoc_p(dcache_assoc_p)
      ,.sets_p(dcache_sets_p)
      ,.block_width_p(dcache_block_width_p)
+     ,.data_mem_negedge_p(1)
+     ,.tag_mem_negedge_p(1)
```
muwyse, line 357: stat_mem_negedge_p?

bp_be/src/v/bp_be_mem/bp_be_dcache/bp_be_dcache.v
```diff
@@ -286,14 +286,14 @@ module bp_be_dcache
   logic [dword_width_p-1:0] data_tl_r;
   logic gdirty_r;
 
-  assign tl_we = v_i & cache_req_ready_i;
+  assign tl_we = v_i;
```
muwyse, line 289: I assume this means that when a cache operation misses and needs the LCE, but the LCE is not ready for the miss, the cache will either replay the instruction (missing again, but hoping LCE is ready on the replay) or the cache will hold the missing operation and cache_req_pkt until the LCE is ready, while reporting a cache miss to the core?

bp_be/src/v/bp_be_mem/bp_be_dcache/bp_be_dcache.v
```diff
@@ -286,14 +286,14 @@ module bp_be_dcache
...
   
-  always_ff @ (posedge clk_i) begin
+  always_ff @ (negedge clk_i) begin
     if (reset_i) begin
       v_tl_r <= 1'b0;
     end
     else begin
-      v_tl_r <= tl_we;
+      v_tl_r <= tl_we & ~poison_i;
       if (tl_we) begin
```
Shashank-Vijay, line 297: Same comment as Mark, should this be tl_we and ~poison_i?

bp_be/src/v/bp_be_mem/bp_be_dcache/bp_be_dcache.v
```diff
@@ -765,15 +765,15 @@ module bp_be_dcache
...
+  bsg_dff_reset_set_clear
+   #(.width_p(1), .clear_over_set_p(1))
    cache_miss_tracker
     (.clk_i(clk_i)
      ,.reset_i(reset_i)
 
-     ,.en_i(miss_tracker_en_li | cache_req_complete_i)
-     ,.data_i(cache_req_v_o)
+     // We don't wait for ack on uncached stores or writethrough requests
+     ,.set_i(cache_req_v_o & ~uncached_store_req & ~wt_req)
```
Shashank-Vijay, line 775: What about ~fencei_req that was previously there in miss_tracker_en_li?

bp_be/src/v/bp_be_mem/bp_be_dcache/bp_be_dcache.v
Shashank-Vijay: Or is it like, since we have the poison_i protection on v_tl_r and v_tv_r and these are used along with any other signal in that pipeline stage, we don't need the check again for those signals?

bp_be/src/v/bp_be_mem/bp_be_dcache/bp_be_dcache.v
dpetrisko: So the reason for the change is to ease critical paths on flushes.  Gating the flops is a power optimization to reduce toggling, not functional.  The tradeoff is 1 more toggle on a pipeline flush vs tightening the path to the flops

dpetrisko: I'll add a comment

bp_be/test/tb/bp_be_dcache/wrapper.v
dpetrisko: The D$ only has data mem and tag mem on negedge, stat mem is on posedge, along with the writebuffer.  I'll add this to the comment in the D$

bp_be/src/v/bp_be_mem/bp_be_dcache/bp_be_dcache.v
dpetrisko: Yeah, basically we will only dispatch to the cache at cycle 0 if dcache_ready.  In cycle 1 (TL), the cache_req_ready may go low, but it doesn't really matter since we're not sending to cache_req in that cycle.  In cycle 2 (TV), if we miss, it will be replayed in the pipeline, no matter what. The two cases:
1) cache_req_ready, the miss goes out and we stall the pipe on cache_req_ready
2) ~cache_req_ready, the miss doesn't go out, and we stall the pipe on cache_req_ready

I'll put this in the d$ block comment

bp_be/src/v/bp_be_mem/bp_be_dcache/bp_be_dcache.v
dpetrisko: From above:
>So the reason for the change is to ease critical paths on flushes. Gating the flops is a power optimization to reduce toggling, not functional. The tradeoff is 1 more toggle on a pipeline flush vs tightening the path to the flops

bp_be/src/v/bp_be_mem/bp_be_dcache/bp_be_dcache.v
dpetrisko: fencei_req sends a cache_req_complete signal now. It's a little less confusing because it doesn't really need to be a special case.  uc_store and wt are special cases because we want to avoid that combinational round trip

bp_be/src/v/bp_be_mem/bp_be_dcache/bp_be_dcache.v
muwyse: Okay, makes sense. My concern was that ~cache_req_ready in TV would cause the request to not go to the LCE, but the cache miss tracking logic to record the operation as a miss and get stuck waiting for the LCE to complete the miss (which the LCE never actually got because the cache never sent the packet)

bp_be/test/tb/bp_be_dcache/wrapper.v
muwyse: okay, thanks

dpetrisko:
> At a high level, will this restrict the performance/frequency of the design in the future in any way since we are moving the mems to the negative edges?

There are two places where it's possible this impacts frequency.  The 12-bit mem address calculation and the bypass path.
- The address calculation seems to be fine as long as we use high performance cells
- The bypass path may end up being critical in which case we can take an extra cycle for byte muxing.  We should have no problems with full word bypasses

> Also, I think a simplified timing diagram(s) showing how a hit and miss work would be helpful for the docs. But, that won't hold up this code change.

I agree with that, it's a little mind bending.  I have some sketches I'll add along with the other docs I'm working on

PR #552, opened 2020-06-26
ME documentation update

muwyse:


dpetrisko:
Awesome!

dpetrisko:
Where do you think discussion of the I/O CCE and UCE should live? Probably in the microarchitecture guide, with a reference to them here?

muwyse:
Yes, sounds good.

PR #563, opened 2020-07-02
Fixed Multicore Cosim Stat Dumping

### LINKED ISSUE #561: Fix multicore checkpoint stat dumping
https://github.com/black-parrot/black-parrot/blob/master/bp_top/test/tb/bp_tethered/testbench.v#L236

There is an issue with doing a module-level bind driving a single bit of a multi-bit signal.  Basically, the signal being driven is dependent on core_id, but the dot reference doesn't like this late resolution.  Solutions include a manual (iterative) bind, a high impedance signal, ...

farzamgl:
Fixes #561


dpetrisko:
Awesome!

PR #565, opened 2020-07-03
merge me_dev to dev - full MOESIF support in LCE and ucode CCE

muwyse:
This change adds full support for the MOESIF protocol (and any valid subset) in the LCEs and CCEs, as well as updates the LCEs for a more optimized MESI protocol. Microcode for MOESIF is provided. The ucode CCE default protocol remains MESI, but comparison of MESI and MOESIF can be performed in the future to determine if its default protocol should change.

The FSM CCE is updated to execute the optimized MESI protocol.


bp_common/src/include/bp_common_cache_service_if.vh
```diff
@@ -81,6 +81,8 @@ typedef enum logic [2:0] {
  e_cache_tag_mem_invalidate,
  // set tag and coherence state for given index and way_id
  e_cache_tag_mem_set_tag,
+ // set coherence state for given index and way_id
+ e_cache_tag_mem_set_state,
```
dpetrisko, line 85: Maybe we should get rid of invalidate, since it's strictly less useful than set_state?

bp_common/src/include/bp_common_me_if.vh
```diff
@@ -220,7 +220,7 @@ typedef enum logic [3:0]
   ,e_lce_cmd_tr              = 4'b1000 // transfer block
   ,e_lce_cmd_st_tr           = 4'b1001 // set state and transfer block
   ,e_lce_cmd_st_tr_wb        = 4'b1010 // set state, transfer, and writeback block
-  ,e_lce_cmd_uc_data         = 4'b1011 // unached data to LCE, i.e, up to 64-bits data
+  ,e_lce_cmd_uc_data         = 4'b1011 // unached data to LCE
```
dpetrisko, line 223: unached -> uncached

bp_me/src/v/cce/bp_cce_msg.v
```diff
@@ -771,14 +773,32 @@ module bp_cce_msg
...
               end else begin
+                // all commands set the way_id field
                 lce_cmd.header.way_id = way_i;
+
+                // commands including a set state operation set the state field
                 if ((decoded_inst_i.lce_cmd == e_lce_cmd_st)
-                    | (decoded_inst_i.lce_cmd == e_lce_cmd_st_wakeup)) begin
+                    | (decoded_inst_i.lce_cmd == e_lce_cmd_st_wakeup)
+                    | (decoded_inst_i.lce_cmd == e_lce_cmd_st_wb)) begin
+                  // TODO: decoder sets coh_state_i to mshr.next_coh_state
```
dpetrisko, line 783: Is this TODO important?

muwyse:
> A high level question: since CCEs are the coordinators of this protocol, and we already support $-$ transfers on shared (right?), what's the concrete difference of an F state in the LCE? For O, I assume it's preventing a writeback, so that makes sense to me

MESI does not do cache to cache transfers if a block is in the S state, because a block in S has no defined LCE owner (the directory is considered the owner). The F state adds ownership to S, allowing the cache to cache transfer of a clean read-only block. It would be possible for the directory to select some cache with the block in S and tell it to transfer the block to another cache but that behavior could be interpreted as violating the coherence protocol because S is not an ownership state. The challenge is deciding which cache to use as the source for the transfer so one cache doesn't get overloaded with transfer requests for blocks. That said, adding the F state forces all transfers for the block to come from one cache, so it is unclear to me if that is better/worse than random selection of some cache to source the transfer. I think both are better than fetching from memory, though, since the block is already in an LCE.

Yes, the O state prevents a writeback of a dirty block (in M) that is requested for reading by another cache, allowing a dirty block to be shared by declaring an owner for it. When the owner (in O) replaces the block it will be forced to write it back or if another cache does a write request then ownership will transfer to the requester (granting block in M state) and it can stay dirty (i.e., no writeback).

bp_me/src/v/cce/bp_cce_msg.v
muwyse: I don't think it is critical right now. This is more of a question about the encoding of the ucode instructions and if there is room in queue operations to specify the mux select signal.

I'll adjust the comment.

bp_common/src/include/bp_common_cache_service_if.vh
muwyse: Makes sense. I think that should be a small change and I can roll into this.

PR #566, opened 2020-07-03
Adding ability to boot from bootrom

dpetrisko:
- Adds a separately addressed bootrom to host
- Restores bootrom checking in dromajo
- Removes config bus read ports from module to reduce fanout and critical path

Depends on [BaseJump STL PR](https://github.com/bespoke-silicon-group/basejump_stl/pull/317)

### DIFF
--- a/bp_me/src/v/cce/bp_cce_fsm.v
+++ b/bp_me/src/v/cce/bp_cce_fsm.v
@@ -74,9 +74,6 @@ module bp_cce_fsm
    , input                                             mem_cmd_ready_i
   );
 
-  // stub cfg ucode output, since FSM CCE has no ucode
-  assign cfg_cce_ucode_data_o = '0;
-
   //synopsys translate_off
   initial begin
     assert (lce_sets_p > 1) else $error("Number of LCE sets must be greater than 1");
--- a/bp_me/src/v/cce/bp_cce_fsm.v
+++ b/bp_me/src/v/cce/bp_cce_fsm.v
@@ -48,7 +48,6 @@ module bp_cce_fsm
 
    // Config channel
    , input [cfg_bus_width_lp-1:0]                      cfg_bus_i
-   , output [cce_instr_width_p-1:0]                    cfg_cce_ucode_data_o
 
    // LCE-CCE Interface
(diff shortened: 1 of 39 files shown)


dpetrisko:
Works with/without cosim

bp_me/src/v/cce/bp_cce_fsm.v
```diff
@@ -48,7 +48,6 @@ module bp_cce_fsm
 
    // Config channel
    , input [cfg_bus_width_lp-1:0]                      cfg_bus_i
-   , output [cce_instr_width_p-1:0]                    cfg_cce_ucode_data_o
```
muwyse, line 51: remove the assignment/stub of this port below: https://github.com/black-parrot/black-parrot/pull/566/files#diff-7641973b6d9ca28e7ffc1a8afa78ed8cR77-R78

bp_me/src/v/cce/bp_cce_wrapper.v
```diff
@@ -29,7 +29,13 @@ module bp_cce_wrapper
...
    // Configuration Interface
    , input [cfg_bus_width_lp-1:0]                      cfg_bus_i
-   , output [cce_instr_width_p-1:0]                    cfg_cce_ucode_data_o
+
+   // ucode programming interface, synchronous read, direct connection to RAM
+   , input                                             ucode_v_i
+   , input                                             ucode_w_i
+   , input [cce_pc_width_p-1:0]                        ucode_addr_i
+   , input [cce_instr_width_p-1:0]                     ucode_data_i
+   , output [cce_instr_width_p-1:0]                    ucode_data_o
```
muwyse, line 38: Not sure how SV works here, but do these ports need to be added to bp_cce_fsm and then stubbed? Or does the (.*) connections for ports ignore ones that don't exist?

dpetrisko: the latter

bp_me/src/v/cce/bp_cce_fsm.v
dpetrisko: fixed!

muwyse:
LGTM

PR #571, opened 2020-07-07
Streaming design between UCE and L2

Songchun-Li:
1. Added a new parameter _cce_mem_if_data_width_ to define the data width in cmd/resp packets between UCE and L2.  
2. Added _name_mp_ on _bp_me_if_ and _bp_me_if_width_ definition, so that L2 can have different data width on its upstream and down stream sides.
3. Added logic to convert the data width in io_cmd/resp. In this way, the IO port of unicore remains the same.
4. Added a streaming version of L2 cache adapter (_bp_me_cce_to_cache_stream_) which is used only when cache _fill_width_ == _cce_mem_if_data_width_ == 64 bits. This new deisgn reduces the area occupation, whiling reduce the latency with streaming design in mem_resp generation.


dpetrisko:
Lots of good enhancements here, will need to think about how best to integrate. 

Can you separate out the mem_if changes into a separate PR? (Don’t close this one)  That’ll make it a little more manageable

PR #572, opened 2020-07-09
Mem IF changes for streaming design

Songchun-Li:
This PR is separate out from the  streaming design PR for the convenience to integrate the several changes.
This PR includes:
1. Definition changes on bp_me_if and bp_me_if_width 
2. Data width transformation for IO ports and the mem ports to main memory without L2 cache

### DIFF
--- a/bp_common/src/include/bp_common_aviary_defines.vh
+++ b/bp_common/src/include/bp_common_aviary_defines.vh
@@ -117,7 +117,6 @@ typedef struct packed
   integer acache_assoc;
   integer acache_block_width;
   integer acache_fill_width;
-  integer cce_mem_if_data_width;
   integer cce_pc_width;
   integer ucode_cce;
 
@@ -231,7 +230,6 @@ typedef struct packed
   , localparam acache_assoc_p             = proc_param_lp.acache_assoc                             \
   , localparam acache_block_width_p       = proc_param_lp.acache_block_width                       \
   , localparam acache_fill_width_p        = proc_param_lp.acache_fill_width                        \
-  , localparam cce_mem_if_data_width_p    = proc_param_lp.cce_mem_if_data_width                    \
   , localparam lce_assoc_p                = `BSG_MAX(dcache_assoc_p,                               \
                                                      `BSG_MAX(icache_assoc_p, acache_assoc_p))     \
   , localparam lce_assoc_width_p          = `BSG_SAFE_CLOG2(lce_assoc_p)                           \
--- a/bp_common/src/include/bp_common_aviary_defines.vh
+++ b/bp_common/src/include/bp_common_aviary_defines.vh
@@ -117,6 +117,7 @@ typedef struct packed
   integer acache_assoc;
   integer acache_block_width;
   integer acache_fill_width;
+  integer cce_mem_if_data_width;
   integer cce_pc_width;
   integer ucode_cce;
 
@@ -230,6 +231,7 @@ typedef struct packed
   , localparam acache_assoc_p             = proc_param_lp.acache_assoc                             \
   , localparam acache_block_width_p       = proc_param_lp.acache_block_width                       \
   , localparam acache_fill_width_p        = proc_param_lp.acache_fill_width                        \
+  , localparam cce_mem_if_data_width_p    = proc_param_lp.cce_mem_if_data_width                    \
   , localparam lce_assoc_p                = `BSG_MAX(dcache_assoc_p,                               \
                                                      `BSG_MAX(icache_assoc_p, acache_assoc_p))     \
   , localparam lce_assoc_width_p          = `BSG_SAFE_CLOG2(lce_assoc_p)                           \
--- a/bp_me/src/asm/include/bp_common_me_if.h
+++ b/bp_me/src/asm/include/bp_common_me_if.h
@@ -95,12 +95,12 @@ typedef enum {
 
 // Mem Commands
 typedef enum {
-  e_bp_mem_rd               = 0
-  ,e_bp_mem_wr              = 1
-  ,e_bp_mem_uc_rd           = 2
-  ,e_bp_mem_uc_wr           = 3
-  ,e_bp_mem_pre             = 4
-} bp_mem_cmd_type_e;
+  e_mem_msg_rd               = 0
+  ,e_mem_msg_wr              = 1
+  ,e_mem_msg_uc_rd           = 2
+  ,e_mem_msg_uc_wr           = 3
+  ,e_mem_msg_pre             = 4
+} bp_mem_msg_e;
 
 #define bp_cce_mem_cmd_type_width 4
 
--- a/bp_me/src/asm/include/bp_common_me_if.h
+++ b/bp_me/src/asm/include/bp_common_me_if.h
@@ -95,12 +95,12 @@ typedef enum {
 
 // Mem Commands
 typedef enum {
-  e_cce_mem_rd               = 0
-  ,e_cce_mem_wr              = 1
-  ,e_cce_mem_uc_rd           = 2
-  ,e_cce_mem_uc_wr           = 3
-  ,e_cce_mem_pre             = 4
-} bp_cce_mem_cmd_type_e;
+  e_bp_mem_rd               = 0
+  ,e_bp_mem_wr              = 1
+  ,e_bp_mem_uc_rd           = 2
+  ,e_bp_mem_uc_wr           = 3
+  ,e_bp_mem_pre             = 4
+} bp_mem_cmd_type_e;
 
 #define bp_cce_mem_cmd_type_width 4
 
--- a/bp_me/src/include/v/bp_me_cce_mem_if.vh
+++ b/bp_me/src/include/v/bp_me_cce_mem_if.vh
@@ -1,7 +1,7 @@
 /**
  * bp_me_cce_mem_if.vh
  *
- * This file defines the interface between the CCE and memory.
+ * This file defines the interface between the UCE/CCE and memory.
  *
  */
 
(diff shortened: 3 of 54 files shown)


bp_common/src/include/bp_common_aviary_defines.vh
```diff
@@ -117,6 +117,7 @@ typedef struct packed
   integer acache_assoc;
   integer acache_block_width;
   integer acache_fill_width;
+  integer cce_mem_if_data_width;
```
dpetrisko, line 120: I don't think we need this as a parameter at the top level.  It's really a derivative system property depending on adapter configuration, etc

bp_me/src/include/v/bp_me_cce_mem_if.vh
```diff
@@ -61,7 +61,7 @@ typedef enum logic [3:0]
  * speculative is set if the request was issued speculatively by the CCE
  */
 
-`define declare_bp_cce_mem_msg_payload_s(lce_id_width_mp, lce_assoc_mp) \
+`define declare_bp_cce_mem_msg_payload_s(lce_id_width_mp, lce_assoc_mp, name_mp) \
```
dpetrisko, line 64: Probably rename cce_mem -> mem throughout, since this memory format is not specific to CCE.  Thoughts @muwyse ?

bp_me/src/v/cache/bp_me_cache_slice.v
```diff
@@ -12,18 +12,19 @@ module bp_me_cache_slice
...
+   `declare_bp_me_if_widths(paddr_width_p, cce_mem_if_data_width_p, lce_id_width_p, lce_assoc_p, uce_l2)
+   `declare_bp_me_if_widths(paddr_width_p, cce_block_width_p, lce_id_width_p, lce_assoc_p, cce_mem)
 
    , localparam cfg_bus_width_lp = `bp_cfg_bus_width(vaddr_width_p, core_id_width_p, cce_id_width_p, lce_id_width_p, cce_pc_width_p, cce_instr_width_p)
    )
   (input                                clk_i
    , input                              reset_i
 
-   , input [cce_mem_msg_width_lp-1:0]   mem_cmd_i
+   , input [uce_l2_msg_width_lp-1:0]    mem_cmd_i
```
dpetrisko, line 23: uce_mem, not uce_l2

bp_me/src/v/cache/bp_me_cce_to_cache.v
```diff
@@ -1,6 +1,7 @@
 /*
  * bp_me_cce_to_cache.v
- *
+ * 
+ * Shrinking the latency for
```
dpetrisko, line 4: Remove comment

bp_me/src/v/cache/bp_me_cce_to_cache.v
```diff
@@ -15,12 +16,12 @@ module bp_me_cce_to_cache
 
   #(parameter bp_params_e bp_params_p = e_bp_inv_cfg
     `declare_bp_proc_params(bp_params_p)
```
dpetrisko, line 18: Add ports widths as parameters here instead

bp_top/src/v/bp_cfg.v
```diff
@@ -9,7 +9,7 @@ module bp_cfg
  import bp_common_cfg_link_pkg::*;
  #(parameter bp_params_e bp_params_p = e_bp_inv_cfg
    `declare_bp_proc_params(bp_params_p)
-   `declare_bp_me_if_widths(paddr_width_p, cce_block_width_p, lce_id_width_p, lce_assoc_p)
+   `declare_bp_me_if_widths(paddr_width_p, cce_mem_if_data_width_p, lce_id_width_p, lce_assoc_p, cce_mem)
```
dpetrisko, line 12: can be dword_width

bp_me/src/v/cce/bp_cce_loopback.v
```diff
@@ -11,7 +11,7 @@ module bp_cce_loopback
   import bp_me_pkg::*;
   #(parameter bp_params_e bp_params_p = e_bp_inv_cfg
     `declare_bp_proc_params(bp_params_p)
-    `declare_bp_me_if_widths(paddr_width_p, cce_block_width_p, lce_id_width_p, lce_assoc_p)
```
dpetrisko, line 14: Can be dword width

bp_top/src/v/bp_clint_slice.v
```diff
@@ -10,7 +10,7 @@ module bp_clint_slice
  import bp_me_pkg::*;
  #(parameter bp_params_e bp_params_p = e_bp_inv_cfg
    `declare_bp_proc_params(bp_params_p)
-   `declare_bp_me_if_widths(paddr_width_p, cce_block_width_p, lce_id_width_p, lce_assoc_p)
```
dpetrisko, line 13: Can be dword width

bp_top/src/v/bp_unicore.v
```diff
@@ -1,4 +1,7 @@
-
+// TODO: the mem_port data width between UCE and L2 is defined by cce_mem_if_data_width_p,
```
dpetrisko, line 1: Remove comment and leave parameter as cce_block_width_p

bp_top/src/v/bp_unicore.v
```diff
@@ -91,29 +98,29 @@ module bp_unicore
   logic [1:0] credits_full_li, credits_empty_li;
   logic timer_irq_li, software_irq_li, external_irq_li;
 
-  bp_cce_mem_msg_s [2:0] proc_cmd_lo;
+  bp_uce_l2_msg_s [2:0] proc_cmd_lo;
```
dpetrisko, line 101: uce_mem

bp_top/src/v/bp_unicore.v
```diff
@@ -334,25 +341,51 @@ module bp_unicore
      );
 
   // Assign incoming I/O as basically another UCE interface
-  assign proc_cmd_lo[2] = io_cmd_i;
```
dpetrisko, line 337: Is this just trying to solve width mismatch issues?  You can just do 

larger = larger_width'(smaller)
smaller = larger[0+:smaller_width].  Because the data payload is at the end, this is safe.

bp_top/src/v/bp_cfg.v
Songchun-Li: I understand that the data size for cfg, clint and etc. is 64 bit, but I am not sure if it is a good idea to set it to dword_width. The reason is I am not sure if we are going to support varying data size between UCE and L2 like 128, 256 bits in the future.

bp_me/src/include/v/bp_me_cce_mem_if.vh
muwyse: Agreed. bp_mem_if perhaps? Or do we want bp_me_mem_if? I vote for the first option and that this header move to bp_common/src/include/

dpetrisko: >I vote for the first option and that this header move to bp_common/src/include/

Works for me!

bp_top/src/v/bp_cfg.v
dpetrisko: Not sure I understand.  What does the UCE/L2 port width have to do with the CFG/CLINT port width?

Songchun-Li: What I mean is setting data width in cfg/clint/loopback to be 64 bit may lead to some port mismatch issues. For example, if the UCE issues a mem_cmd with 128 bit data packet and the mem_cmd_i port in cfg/clint is 64 bit, there will be a port mismatch issues. I know it does not affect the functional correctness as the missing 64 bits are useless, just want to make sure.

Songchun-Li:
Required change are made except the bp_me_if in bp_me_cce_mem_if.vh and related renaming. (Comes in next commit)
The parameter cce_mem_if_data_width is removed from top level definition and now generated in the unicore based on the cache fill_width, so it can passed to UCE and L2.

Songchun-Li:
Renaming on bp_mem_if is done, but the comment about CCE in it has not been modified yet.

bp_me/src/asm/include/bp_common_me_if.h
```diff
@@ -95,12 +95,12 @@ typedef enum {
 
 // Mem Commands
 typedef enum {
-  e_cce_mem_rd               = 0
-  ,e_cce_mem_wr              = 1
-  ,e_cce_mem_uc_rd           = 2
-  ,e_cce_mem_uc_wr           = 3
-  ,e_cce_mem_pre             = 4
-} bp_cce_mem_cmd_type_e;
+  e_bp_mem_rd               = 0
```
dpetrisko, line 98: @muwyse what do you think of the names here? e_mem_rd is probably too likely to cause name conflicts.  But we usually don't use _bp_ in enums

muwyse: One option that would align with the bp_mem_msg_size_e enum would be to call the enum type bp_mem_msg_e and each name would be e_mem_msg_XXX (e_mem_msg_rd, e_mem_msg_wr, etc.)

A slightly more verbose option is bp_mem_msg_type_e and e_mem_msg_type_XXX

muwyse: agreed that e_mem_rd is a likely name conflict waiting to happen

Songchun-Li: I vote for the bp_mem_msg_type_e and e_mem_msg_type_XXX, because bp_mem_msg_e is similar to bp_``name_mp``_msg_s, the top level message struct.

dpetrisko: bp_mem_msg_type_e and e_mem_msg_XXX sounds good!

PR #577, opened 2020-07-17
Delaying byte muxing by a cycle

dpetrisko:
The load-fwd path is the current critical path in the BE.  Subword loads are not usually in critical loops.  Therefore, we allow early 2-cycle double word loads and delay subword loads by a cycle.  Float loads will also be on the 3-cycle path.

Changes:
- Update D$ description block
- Added early/final data ports to the D$
- Updated hazard detection logic
- Update D$ testbench to not rely on zero store data and use final data
- Removed misleading cache miss signal, since it doesn't indicate whether a miss is due to poison or tag miss

### DIFF
--- a/bp_be/src/v/bp_be_calculator/bp_be_instr_decoder.v
+++ b/bp_be/src/v/bp_be_calculator/bp_be_instr_decoder.v
@@ -55,7 +55,8 @@ module bp_be_instr_decoder
       // Destination pipe
       decode.pipe_ctl_v    = '0;
       decode.pipe_int_v    = '0;
-      decode.pipe_mem_v    = '0;
+      decode.pipe_mem_early_v = '0;
+      decode.pipe_mem_final_v = '0;
       decode.pipe_mul_v    = '0;
       decode.pipe_fma_v    = '0;
       decode.pipe_long_v   = '0;
@@ -199,7 +200,8 @@ module bp_be_instr_decoder
           end
         `RV64_LOAD_OP :
           begin
-            decode.pipe_mem_v = 1'b1;
+            decode.pipe_mem_early_v =  (instr inside {`RV64_LD});
+            decode.pipe_mem_final_v = ~(instr inside {`RV64_LD});
             decode.irf_w_v    = 1'b1;
             decode.dcache_r_v = 1'b1;
             decode.mem_v      = 1'b1;
@@ -216,7 +218,7 @@ module bp_be_instr_decoder
           end
         `RV64_STORE_OP :
           begin
-            decode.pipe_mem_v = 1'b1;
+            decode.pipe_mem_early_v = 1'b1;
             decode.dcache_w_v = 1'b1;
             decode.mem_v      = 1'b1;
             unique casez (instr)
@@ -233,7 +235,7 @@ module bp_be_instr_decoder
               `RV64_FENCE   : begin end
               `RV64_FENCE_I :
                 begin
-                  decode.pipe_mem_v  = 1'b1;
+                  decode.pipe_mem_early_v = 1'b1;
                   decode.dcache_w_v  = 1'b1;
                   decode.serial_v    = 1'b1;
                   decode.fu_op       = e_dcache_op_fencei;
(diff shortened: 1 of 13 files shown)


taylor-bsg:
I think it's worth looking at the original critical path to see if there is any fat to trim..

dpetrisko:
It's a half cycle path, all of which is taken up by the 512->64-bit mux + bypass

taylor-bsg:
FYI This kind of path, you often handle by having a hierarchical bypass
mux, for example muxing the early arriving signals first, and then having a
2-input mux for the late arriving signal.  What is being skipped?

1. the storebuffer
2. shifting for bytes
3. Sign extension

M

dpetrisko:
2/3

dpetrisko:
Note this is how Rocket/SiFive handle this case, and they have a full cycle...

bp_be/src/v/bp_be_calculator/bp_be_instr_decoder.v
```diff
@@ -274,7 +276,7 @@ module bp_be_instr_decoder
           end
         `RV64_AMO_OP:
           begin
-            decode.pipe_mem_v = 1'b1;
+            decode.pipe_mem_early_v = 1'b1;
```
Shashank-Vijay, line 279: Shouldn't the amo word instructions go through the final_data path? Also, same thing for when we have L1 atomics.

bp_be/src/v/bp_be_calculator/bp_be_pipe_mem.v
```diff
@@ -381,8 +390,8 @@ module bp_be_pipe_mem
   assign store_access_fault_v = dtlb_r_v_lo & is_store_mem1 & (mode_fault_v | did_fault_v | sac_fault_v);
 
   assign tlb_miss_v_o           = dtlb_miss_v;
-  assign cache_miss_v_o         = is_req_mem2 & ~dcache_v;
-  assign fencei_v_o             = is_fencei_mem2 & dcache_v;
+  assign cache_miss_v_o         = is_req_mem2 & ~dcache_early_v;
```
Shashank-Vijay, line 393: So for non-LD load operations, both early_v_o and final_v_o would be 1? Could we instead use the information in the decode struct to choose one of the two conditions for early and final and change the logic in the dcache to send one or the other but not both? That would be another mux here though, and I'm not sure if that would affect the critical path.

bp_be/src/v/bp_be_calculator/bp_be_instr_decoder.v
dpetrisko: L1 64-bit atomics would be early.  32-bit could go either way.  You're right that L2 atomics will be on the slow path out of the D$, but these datapath signals are explicitly for the fixed-latency operations.  The L2 atomics will require calculator datapath support for a late writeback independent of these changes

bp_be/src/v/bp_be_calculator/bp_be_pipe_mem.v
dpetrisko: There's really no reason to do that.  The early_v_o signal serves as the hit signal for any sized operation

Shashank-Vijay: Yeah, realized that!

bp_be/src/v/bp_be_calculator/bp_be_instr_decoder.v
Shashank-Vijay: Yup, got it. I'll incorporate this in my late writeback design slides

dpetrisko: Thinking about this more.  L2 atomics requests will be sized, but not sign extended.  We'll want to reuse the data mux stage sign extension for them, so they will inherently be late.  This brings up some interesting requirements with stall behavior (need to preserve info in DM stage as a pseudo-MSHR), food for thought

PR #579, opened 2020-07-20
Move memory response processing out of FSM for CCE's uncached only mode

muwyse:


bp_me/src/v/cce/bp_cce_msg.v
```diff
@@ -172,11 +175,10 @@ module bp_cce_msg
   // sent and responses received. If the config bus changes the CCE mode to
   // normal, but there are still outstanding accesses, the message module will
   // wait to transition modes until
-  localparam max_counter_val_lp = 2**10;
   logic uc_cnt_inc, uc_cnt_dec;
-  logic [`BSG_WIDTH(max_counter_val_lp)-1:0] uc_cnt;
+  logic [`BSG_WIDTH(max_mem_requests_p)-1:0] uc_cnt;
   bsg_counter_up_down
-    #(.max_val_p(max_counter_val_lp)
+    #(.max_val_p(max_mem_requests_p)
```
dpetrisko, line 181: We should probably track this for cached requests too?

Maybe use https://github.com/bespoke-silicon-group/basejump_stl/blob/master/bsg_dataflow/bsg_flow_counter.v, inc on mem_cmd, dec on mem_resp?

dpetrisko:
Probably we want the whole uncached logic in 1 state, to 1) avoid concurrency bugs like this 2) get better performance on uncached stores (1 cycle min vs 2 cycle min).  I'd advise putting all of the logic in uc_ready

```
if (resp)
  // ack mem_resp
  // send lce_cmd
else
  // ack lce_req
  // send mem_cmd
```

bp_me/src/v/cce/bp_cce_msg.v
muwyse: Yes, we could track for all mem_cmd/mem_resp if we want. This will be more effort to implement because it needs to be taken into account for ucode instructions when sending mem_cmd, possibly stalling the ucode engine if there are no credits left. I think it should be fairly straightforward though.

muwyse:
> Probably we want the whole uncached logic in 1 state, to 1) avoid concurrency bugs like this 2) get better performance on uncached stores (1 cycle min vs 2 cycle min). I'd advise putting all of the logic in uc_ready

Good call. I don't think the uncached only mode FSM will have any complications with this approach.

bp_me/src/v/cce/bp_cce_msg.v
dpetrisko: 1) Do all mem_cmd / mem_resp come from this module?
2) Is there always exactly 1 mem_resp per mem_cmd?

If so, it makes it easy because you don't have to track per message, just use the v_o and yumi_o signals directly.  If not, we can do it in a separate PR because it's a bit more involved

muwyse: Yes and Yes.

The only consideration is that when the ucode engine is executing and sending a memory command, it will need to stall if there are no credits left, even if mem_cmd_ready_i is high.

I suggest a separate PR for it so this one gets merged in sooner.

dpetrisko: works for me!

PR #580, opened 2020-07-21
Added DRAMSim3

farzamgl:
Added DRAMSim3 and its adapter to bp_mem.


external/Makefile.tools
```diff
@@ -11,7 +12,6 @@ spike_dir       := $(BP_EXTERNAL_DIR)/riscv-isa-sim
 dromajo_dir     := $(BP_EXTERNAL_DIR)/dromajo
 axe_dir         := $(BP_EXTERNAL_DIR)/axe
 dramsim2_dir    := $(BP_EXTERNAL_DIR)/DRAMSim2
```
dpetrisko, line 14: probably want to get rid of dramsim2 here too?

bp_top/test/tb/bp_tethered/Makefile.params
```diff
@@ -27,8 +24,6 @@ export USE_DRAMSIM2_LATENCY_P ?= 0
 export USE_MAX_LATENCY_P      ?= 1
 export USE_RANDOM_LATENCY_P   ?= 0
 
-export BP_SIM_CLK_PERIOD ?= 10
```
dpetrisko, line 30: How is the clock period now defined?

bp_me/test/common/bp_mem_to_dram.v
```diff
@@ -0,0 +1,355 @@
...
+  ,.write_data_i(reorder_w_data_li)
+
+  ,.fifo_deq_v_o(reorder_deq_v_lo)
+  ,.fifo_deq_data_o(reorder_deq_data_lo)
+  ,.fifo_deq_yumi_i(reorder_deq_yumi_li)
+
+  ,.empty_o()
+  );
+
+  bsg_cam_1r1w_tag_array
```
dpetrisko, line 260: Do you have a diagram for this module? I'm confused as to why we need both the reorder fifo and the cam

bp_me/test/common/bp_mem_to_dram.v
```diff
@@ -0,0 +1,355 @@
...
+  ,output       [write_mask_width_lp-1:0]          dma_mask_o
+  ,output                                          dma_data_v_o
+  ,input                                           dma_data_yumi_i
+  // Receiving cache block                                        
+  ,input        [data_width_p-1:0]                 dma_data_i
+  ,input        [paddr_width_p-1:0]                dma_addr_i
+  ,input                                           dma_data_v_i
+  ,output                                          dma_data_ready_o
+  // Cmd input
+  ,input        [cce_mem_msg_width_lp-1:0]         mem_cmd_i
```
dpetrisko, line 46: Could we just connect this to an SRAM and get rid of bp_mem?

bp_me/test/common/bp_mem.v
```diff
@@ -3,11 +3,14 @@
  * bp_mem.v
  */
 
+`define dram_pkg bsg_dramsim3_hbm2_8gb_x128_pkg
```
dpetrisko, line 6: Do we only have HBM configs validated? We probably want a LPDDR model. If it's not available, that can be a followup PR though

bp_me/test/common/bp_mem_to_dram.v
farzamgl: The reorder FIFO generates an id for each cmd that we issue. We store the address->id translation in the CAM. When the read is done DRAM provides the read address. We lookup the corresponding id from the CAM and write the {id, data} to the reorder FIFO.

bp_me/test/common/bp_mem.v
farzamgl: The Verilog code in BP should be the same but I've to test it and possibly add another addressing mode in Basejump. I'll make another PR for that.

external/Makefile.tools
farzamgl: Yeah. Also, we can remove both DRAMSim2 and DRAMSim3 submodules.

bp_top/test/tb/bp_tethered/Makefile.params
farzamgl: Good point actually. Apparently the module needs to be clocked at the same period that's defined in the config file. So I think I've to add another clock region.

bp_me/test/common/bp_mem_to_dram.v
farzamgl: What do you mean?

bp_me/test/common/bp_mem_to_dram.v
dpetrisko: Makes sense!

bp_me/test/common/bp_mem.v
dpetrisko: kk!

bp_me/test/common/bp_mem_to_dram.v
dpetrisko: Oh, I misunderstood what this module was doing

bp_me/test/common/bp_mem_to_dram.v
```diff
@@ -0,0 +1,355 @@
...
+  , localparam lg_block_size_in_words_lp = `BSG_SAFE_CLOG2(block_size_in_words_lp)
+  , localparam bsg_cache_dma_pkt_width_lp = `bsg_cache_dma_pkt_width(paddr_width_p)
+  , localparam reorder_fifo_els_lp = fifo_els_p * block_size_in_words_lp
+  , localparam lg_reorder_fifo_els_lp = `BSG_SAFE_CLOG2(reorder_fifo_els_lp)
+  )
+  
+  (// Cache DMA side
+   input                                           clk_i
+  ,input                                           reset_i
+  // Sending address and write_en
```
dpetrisko: Nit, but can you list the mem_cmd/mem_resp ports first, since they're the "primary" ports here

bp_top/test/tb/bp_tethered/Makefile.params
dpetrisko: Yeah, we'll need a CDC in the bp to dram adapter, and I think we'll also need this parameter (or define) for the core clock

bp_me/test/common/bp_mem.v
```diff
@@ -45,88 +45,108 @@ module bp_mem
...
+assign dma_addr_li = paddr_width_p'(dram_read_done_ch_addr_lo[0]);
+
+genvar i;
+for (i=1; i<`dram_pkg::num_channels_p; i=i+1) begin
+  assign dram_v_li[i] = '0;
+  assign dram_w_li[i] = '0;
+  assign dram_data_v_li[i] = '0;
+end
+
+bp_mem_to_dram
```
dpetrisko, line 51: We probably want to support some fixed latency model as well.  So not necessarily the transducer approach we had before, but some parameterization in here which selects between the dramsim3 (delay_model+ storage) and some fixed latency + ideal SRAM. Maybe look at bsg_manycore for how they handle this? I know they have a "fast and fake" memory

bp_me/test/common/bp_mem.v
```diff
@@ -45,88 +45,108 @@ module bp_mem
...
+assign dma_data_yumi_li = dram_data_yumi_lo[0];
+
+assign dram_ch_addr_li[0] = dma_pkt_lo.addr;
+assign dram_data_li[0] = dma_data_lo;
+assign dram_mask_li[0] = dma_mask_lo;
+assign dma_data_li = dram_data_lo[0];
+assign dma_addr_li = paddr_width_p'(dram_read_done_ch_addr_lo[0]);
+
+genvar i;
+for (i=1; i<`dram_pkg::num_channels_p; i=i+1) begin
```
dpetrisko, line 45: So we only support 1 channel? How could we scale this to multiple channels in the future?

taylor-bsg:
Yes, fast n fake sets the latency to the minimum so that we can do
functional tests in the minimum possible time.

bp_me/test/common/bp_mem_to_dram.v
farzamgl: I think you're looking at an outdated version.

bp_me/test/common/bp_mem.v
farzamgl: Yeah. Good idea.

bp_me/test/common/bp_mem.v
farzamgl: Yes. I guess we can use an arbiter between the adapter and the different DRAM channels but I've to take a closer look at how the different channel timings work to make sure that it'll actually make a difference.

bp_top/test/tb/bp_tethered/Makefile.params
```diff
@@ -40,18 +33,15 @@ export TB_PARAMS  = \
...
                     -pvalue+cosim_memsize_p=$(SAMPLE_MEMSIZE) \
                     -pvalue+cosim_instr_p=$(SAMPLE_INSTR_P) \
                     -pvalue+warmup_instr_p=$(SAMPLE_WARMUP_P) \
-                    -pvalue+use_dramsim2_latency_p=$(USE_DRAMSIM2_LATENCY_P) \
-                    -pvalue+use_max_latency_p=$(USE_MAX_LATENCY_P) \
-                    -pvalue+use_random_latency_p=$(USE_RANDOM_LATENCY_P)
+                    -pvalue+dram_fixed_latency_p=$(DRAM_FIXED_LATENCY)
 
 export DUT_DEFINES = 
-export TB_DEFINES  = +define+BP_SIM_CLK_PERIOD=$(BP_SIM_CLK_PERIOD)
```
dpetrisko, line 54: Think we still need this

bp_me/test/common/bp_mem.v
```diff
@@ -45,88 +33,222 @@ module bp_mem
...
-  (.clk_i(clk_i)
-   ,.reset_i(reset_i)
-
-   ,.v_i(dram_v_li)
-   ,.w_i(dram_w_li)
-   ,.addr_i(dram_addr_li)
-   ,.ready_o(dram_ready_lo)
-
-   ,.v_o(delayed_dram_v_lo)
-   ,.yumi_i(delayed_dram_yumi_li)
```
dpetrisko, line 105: The extra features of bp_mem_storage_sync are 
1) preloading mem
2) rebasing hex files

I don't think it's really necessary anymore, since DRAMSIM3 won't support preloading / rebasing etc.

I'd suggest refactoring bp_mem to be:

```
                               latency_counter (1 per channel) - bsg_mem_multiport
bp_mem_to_dram / 
                              \
                                bsg_nonsynth_dramsim3
```

and delete bp_mem_transducer, bp_mem_delay_model, bp_mem_storage_sync, bp_mem_utils.cpp

bp_common/software/py/nbf.py
```diff
@@ -135,7 +135,8 @@ def init_dram(self):
     for k in sorted(self.dram_data.keys()):
       addr = k
       opcode = self.get_opcode(addr)
-      self.print_nbf(opcode, addr, self.dram_data[k])
+      if self.dram_data[k] != 0:
```
dpetrisko, line 138: Can we make this a parameter in the nbf loader?  For example, on FPGA, zero-ing out is necessary

bp_me/test/common/bp_mem.v
farzamgl: The `bsg_mem_multiport` doesn't support zeroing out the memory on reset though. We need that to reduce the checkpoint loading time to keep the simulation times feasible for benchmarks with bigger memory space. I can remove the two functionalities though and delete the other files.

dpetrisko: Gotcha, we can add the zero-ing functionality to some basejump module in a separate PR

bp_common/software/py/nbf.py
```diff
@@ -191,10 +193,11 @@ def dump(self):
   parser = argparse.ArgumentParser()
   parser.add_argument('ncpus', type=int, help='number of BlackParrot cores')
   parser.add_argument('ucode_file', metavar='ucode.mem', help='CCE ucode file')
```
dpetrisko, line 195: Unrelated, but should be little effort.  Can you make the ucode_file optional here as well?

farzamgl: But why? Is there a way to load the ucode right now except for the nbf loader?

farzamgl: Well actually now that's also true for the mem argument haha.

PR #586, opened 2020-07-24
Concentrator-based Design between UCE/CCE and Mem

Songchun-Li:
Progress code review on concentrator-based design between UCE and mem in unicore cfg

An interim version is implemented. (Schematics: Slide 5 @ https://docs.google.com/presentation/d/1zYQiMsIGADDS19U77o_-9BlmiorkVAnXTqtTkVT1oAg/edit?usp=sharing). This interim version works with multiple headers + data packets.

Note. It seems there is a significant mIPC drop compared to the baseline, perhaps because filt_width is set to be 64 and no extra FIFOs is added so that it can not do 1 uncached store / cycle.

### DIFF
--- a/bp_me/src/v/wormhole/bp_me_wormhole_packet_encode_mem_cmd.v
+++ b/bp_me/src/v/wormhole/bp_me_wormhole_packet_encode_mem_cmd.v
@@ -18,15 +18,16 @@ module bp_me_wormhole_packet_encode_mem_cmd
   import bp_me_pkg::*;
   #(parameter bp_params_e bp_params_p = e_bp_inv_cfg
     `declare_bp_proc_params(bp_params_p)
-    `declare_bp_mem_if_widths(paddr_width_p, cce_block_width_p, lce_id_width_p, lce_assoc_p, cce_mem)
 
     , parameter flit_width_p = "inv"
     , parameter cord_width_p = "inv"
     , parameter cid_width_p = "inv"
     , parameter len_width_p = "inv"
+    , parameter data_width_p = "inv"
 
+    `declare_bp_mem_if_widths(paddr_width_p, data_width_p, lce_id_width_p, lce_assoc_p, cce_mem) 
     , localparam mem_cmd_packet_width_lp = 
-        `bp_mem_wormhole_packet_width(flit_width_p, cord_width_p, len_width_p, cid_width_p, cce_mem_msg_width_lp-cce_block_width_p, cce_block_width_p)
+        `bp_mem_wormhole_packet_width(flit_width_p, cord_width_p, len_width_p, cid_width_p, cce_mem_msg_width_lp-data_width_p, data_width_p)
     )
    (input [cce_mem_msg_width_lp-1:0]       mem_cmd_i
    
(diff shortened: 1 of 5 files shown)


dpetrisko:
>Note. It seems there is a significant mIPC drop compared to the baseline, perhaps because filt_width is set to be 64 and no extra FIFOs is added so that it can not do 1 uncached store / cycle.

What test? Most of them don't use uncached stores.

Instead, I would believe a drop caused by serialization / deserialization on each side of the concentrator.  We probably want flit size to increase to the size of an uncached store (header + 64 bit == 128 bit?). Then the PISOs in CLINT/CFG etc are simply fifos.  

The other possibility is that there are bubbles sneaking in other places. The concentrator should not, but I'm not 100% sure on the wormhole adapters

bp_me/src/v/wormhole/bp_me_wormhole_packet_encode_mem_cmd.v
```diff
@@ -18,15 +18,16 @@ module bp_me_wormhole_packet_encode_mem_cmd
   import bp_me_pkg::*;
   #(parameter bp_params_e bp_params_p = e_bp_inv_cfg
     `declare_bp_proc_params(bp_params_p)
-    `declare_bp_mem_if_widths(paddr_width_p, cce_block_width_p, lce_id_width_p, lce_assoc_p, cce_mem)
 
     , parameter flit_width_p = "inv"
     , parameter cord_width_p = "inv"
     , parameter cid_width_p = "inv"
     , parameter len_width_p = "inv"
+    , parameter data_width_p = "inv"
```
dpetrisko, line 26: Rather than add a data parameter here, let's convert these and other wormhole modules to work with headers only.  That will be more flexible supporting other handshakes in the future

bp_top/src/v/bp_unicore.v
```diff
@@ -345,113 +355,216 @@ module bp_unicore
...
+      wire is_cfg_cmd          = local_cmd_li & (device_cmd_li == cfg_dev_gp);
+      wire is_clint_cmd        = local_cmd_li & (device_cmd_li == clint_dev_gp);
+      wire is_io_cmd           = (local_cmd_li & (device_cmd_li inside {boot_dev_gp, host_dev_gp})) | is_other_domain;
+      wire is_cache_cmd        = ~local_cmd_li || (local_cmd_li & (device_cmd_li == cache_dev_gp));
+      wire is_loopback_cmd     = local_cmd_li & ~is_cfg_cmd & ~is_clint_cmd & ~is_io_cmd & ~is_cache_cmd;
+      
+      
+      always_comb
+        begin
+          if (is_cfg_cmd) cmd_dst_cid[i] = cid_width_lp'(4);
```
dpetrisko, line 386: This CID should be device_cmd_li from above

bp_top/src/v/bp_unicore.v
```diff
@@ -15,6 +15,11 @@ module bp_unicore
    , localparam uce_mem_data_width_lp = `BSG_MAX(icache_fill_width_p, dcache_fill_width_p) 
    `declare_bp_mem_if_widths(paddr_width_p, uce_mem_data_width_lp, lce_id_width_p, lce_assoc_p, uce_mem)
    `declare_bp_mem_if_widths(paddr_width_p, cce_block_width_p, lce_id_width_p, lce_assoc_p, cce_mem)
+   , localparam flit_width_lp = 64
```
dpetrisko, line 18: Should use mem network parameters in aviary

bp_top/src/v/bp_unicore.v
```diff
@@ -345,113 +355,216 @@ module bp_unicore
...
+
+        ,.src_cord_i('0)
+        ,.src_cid_i(cid_width_lp'(i))
+        ,.dst_cord_i('0)
+        ,.dst_cid_i(cmd_dst_cid[i])
+
+        ,.packet_o(proc_cmd_packet_lo[i])
+        );
+
+      bsg_wormhole_router_adapter
```
dpetrisko, line 413: This might be where cycles are going. I believe this module can only handle 1 request at a time, so it's possible I$/D$ contention is slower in this system than the previous

dpetrisko:
Closing since we're moving to the stream-based approach instead of concentrator

PR #593, opened 2020-07-28
Wrap around logic for ordering addresses sent to the memory

Shashank-Vijay:
This PR adds support for sending the critical fill address request first over the mem_cmd signal in the UCE. The higher fill addresses follow and a wrap around to the lower fill addresses occurs once the highest fill address of the current block is sent.

### DIFF
--- a/bp_me/src/v/cce/bp_uce.v
+++ b/bp_me/src/v/cce/bp_uce.v
@@ -581,7 +581,7 @@ module bp_uce
         e_send_critical:
           if (miss_v_li)
             begin
-              mem_cmd_cast_o.header.msg_type       = e_cce_mem_rd;
+              mem_cmd_cast_o.header.msg_type       = e_mem_msg_rd;
               mem_cmd_cast_o.header.addr           = critical_addr;
               mem_cmd_cast_o.header.size           = block_msg_size_lp;
               mem_cmd_cast_o.header.payload.way_id = lce_assoc_p'(cache_req_metadata_r.repl_way);
--- a/bp_me/src/v/cce/bp_uce.v
+++ b/bp_me/src/v/cce/bp_uce.v
@@ -283,6 +283,9 @@ module bp_uce
   logic [fill_cnt_width_lp-1:0] mem_cmd_cnt;
   logic [block_size_in_fill_lp-1:0] fill_index_shift;
   logic [bank_offset_width_lp-1:0] bank_index;
+  logic [fill_cnt_width_lp-1:0] mem_cmd_init_slice, mem_cmd_init_slice_r;
+  logic [paddr_width_p-1:0] critical_addr;
+
   if (fill_size_in_bank_lp == 1)
     begin
       assign bank_index =  mem_cmd_cnt;
@@ -295,14 +298,15 @@ module bp_uce
     end
 
   logic fill_up, fill_done, mem_cmd_up, mem_cmd_done;
-  if (fill_width_p == block_width_p)
-    begin
+  if (fill_width_p == block_width_p) 
+    begin : fill_equals_block
       assign mem_cmd_cnt = 1'b0;
       assign fill_done = 1'b1;
       assign mem_cmd_done = 1'b1;
+      assign critical_addr = {cache_req_r.addr[paddr_width_p-1:block_offset_width_lp], block_offset_width_lp'(0)};
     end
-  else
-    begin
+  else 
+    begin : fill_less_than_block
       logic [fill_cnt_width_lp-1:0] fill_cnt;
       bsg_counter_clear_up
        #(.max_val_p(block_size_in_fill_lp-1)
@@ -319,20 +323,25 @@ module bp_uce
         );
       assign fill_done = (fill_cnt == block_size_in_fill_lp-1);
 
-      bsg_counter_clear_up
+      bsg_counter_set_en
        #(.max_val_p(block_size_in_fill_lp-1)
-        ,.init_val_p(0)
-        ,.disable_overflow_warning_p(1))
+        ,.reset_val_p(0))
        mem_cmd_counter
         (.clk_i(clk_i)
         ,.reset_i(reset_i)
 
-        ,.clear_i('0)
-        ,.up_i(mem_cmd_up)
-
+        ,.set_i(cache_req_v_i)
+        ,.en_i(mem_cmd_up)
+        ,.val_i(mem_cmd_init_slice)
         ,.count_o(mem_cmd_cnt)
         );
-      assign mem_cmd_done = (mem_cmd_cnt == block_size_in_fill_lp-1);
+      
+      assign mem_cmd_init_slice = cache_req_cast_i.addr[block_offset_width_lp-1-:fill_cnt_width_lp]; 
+      assign mem_cmd_init_slice_r = cache_req_r.addr[block_offset_width_lp-1-:fill_cnt_width_lp]; 
+      assign mem_cmd_done = (mem_cmd_init_slice_r == fill_cnt_width_lp'(0))
+                            ? cache_req_v_r & (mem_cmd_cnt == ((1'b1 << fill_cnt_width_lp)-1))
+                            : cache_req_v_r & (mem_cmd_cnt == (mem_cmd_init_slice_r-1));
+      assign critical_addr = {cache_req_r.addr[paddr_width_p-1:(block_offset_width_lp-fill_cnt_width_lp)], (block_offset_width_lp-fill_cnt_width_lp)'(0)};
     end
 
   logic [index_width_lp-1:0] index_cnt;
@@ -572,8 +581,8 @@ module bp_uce
         e_send_critical:
           if (miss_v_li)
             begin
-              mem_cmd_cast_o.header.msg_type       = e_mem_msg_rd;
-              mem_cmd_cast_o.header.addr           = {cache_req_r.addr[paddr_width_p-1:block_offset_width_lp], block_offset_width_lp'(0)};
+              mem_cmd_cast_o.header.msg_type       = e_cce_mem_rd;
+              mem_cmd_cast_o.header.addr           = critical_addr;
               mem_cmd_cast_o.header.size           = block_msg_size_lp;
               mem_cmd_cast_o.header.payload.way_id = lce_assoc_p'(cache_req_metadata_r.repl_way);
               mem_cmd_cast_o.header.payload.lce_id = lce_id_i;
--- a/bp_me/src/v/cce/bp_uce.v
+++ b/bp_me/src/v/cce/bp_uce.v
@@ -307,7 +307,7 @@ module bp_uce
     end
   else 
     begin : fill_less_than_block
-      logic [fill_cnt_width_lp-1:0] first_addr, last_addr;
+      logic [fill_cnt_width_lp-1:0] first_cmd_cnt, last_cmd_cnt;
       logic [fill_cnt_width_lp-1:0] fill_cnt;
       bsg_counter_clear_up
        #(.max_val_p(block_size_in_fill_lp-1)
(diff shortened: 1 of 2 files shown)


bp_me/src/v/cce/bp_uce.v
```diff
@@ -319,20 +323,25 @@ module bp_uce
...
+        ,.set_i(cache_req_v_i)
+        ,.en_i(mem_cmd_up)
+        ,.val_i(mem_cmd_init_slice)
         ,.count_o(mem_cmd_cnt)
         );
-      assign mem_cmd_done = (mem_cmd_cnt == block_size_in_fill_lp-1);
+      
+      assign mem_cmd_init_slice = cache_req_cast_i.addr[block_offset_width_lp-1-:fill_cnt_width_lp]; 
+      assign mem_cmd_init_slice_r = cache_req_r.addr[block_offset_width_lp-1-:fill_cnt_width_lp]; 
+      assign mem_cmd_done = (mem_cmd_init_slice_r == fill_cnt_width_lp'(0))
```
dpetrisko, line 341: why are there two conditions for this?

bp_me/src/v/cce/bp_uce.v
```diff
@@ -319,20 +323,25 @@ module bp_uce
...
-        ,.up_i(mem_cmd_up)
-
+        ,.set_i(cache_req_v_i)
+        ,.en_i(mem_cmd_up)
+        ,.val_i(mem_cmd_init_slice)
         ,.count_o(mem_cmd_cnt)
         );
-      assign mem_cmd_done = (mem_cmd_cnt == block_size_in_fill_lp-1);
+      
+      assign mem_cmd_init_slice = cache_req_cast_i.addr[block_offset_width_lp-1-:fill_cnt_width_lp];
```
dpetrisko, line 339: mem_cmd_init_slice -> first_addr
mem_cmd_init_slice_r -> last_addr

bp_me/src/v/cce/bp_uce.v
dpetrisko: If you're worried about underflow, it should work fine.  -1 == max value, unsigned

docs/interface_specification.md
```diff
@@ -152,6 +152,7 @@ A UCE implementation must support the following operations:
 - Handle both uncached and cached requests
 - Support both write-through and write-back protocols
 - Support credit-based flow control, to support fencing in the core
+- Support ordering addresses starting from the requested address (aligned based on the fill width), continuing with the higher addresses and wrapping around to the lower addresses
```
dpetrisko, line 155: This should be in the mem if section, not the UCE section.

"Misaligned addresses return data wrapped around the request size using the following scheme:

Request 0x0: [d c b a]
Request 0x2: [b a d c]"

In this cache service section, we should also clarify that subword requests replicate data to up to dword size

bp_me/src/v/cce/bp_uce.v
Shashank-Vijay: Yeah the condition was added for the edge case, because a single (x - 1) condition wasn't working. I checked with the single condition now and it was still not computing the final address slice correctly so I just used two's complement addition in the condition and removed the mux.

docs/interface_specification.md
Shashank-Vijay: Will change that

bp_me/src/v/cce/bp_uce.v
```diff
@@ -319,20 +323,23 @@ module bp_uce
...
-
+        ,.set_i(cache_req_v_i)
+        ,.en_i(mem_cmd_up)
+        ,.val_i(mem_cmd_init_slice)
         ,.count_o(mem_cmd_cnt)
         );
-      assign mem_cmd_done = (mem_cmd_cnt == block_size_in_fill_lp-1);
+      
+      assign mem_cmd_init_slice = cache_req_cast_i.addr[block_offset_width_lp-1-:fill_cnt_width_lp]; 
+      assign mem_cmd_final_slice = (cache_req_r.addr[block_offset_width_lp-1-:fill_cnt_width_lp] + {fill_cnt_width_lp{1'b1}});
```
dpetrisko, line 340: This doesn't look right to me.  Wouldn't this just be the slice after the first one?

bp_me/src/v/cce/bp_uce.v
dpetrisko: This is a rename suggestion

bp_me/src/v/cce/bp_uce.v
```diff
@@ -295,14 +298,15 @@ module bp_uce
...
 
   logic fill_up, fill_done, mem_cmd_up, mem_cmd_done;
-  if (fill_width_p == block_width_p)
-    begin
+  if (fill_width_p == block_width_p) 
+    begin : fill_equals_block
       assign mem_cmd_cnt = 1'b0;
       assign fill_done = 1'b1;
       assign mem_cmd_done = 1'b1;
+      assign critical_addr = {cache_req_r.addr[paddr_width_p-1:block_offset_width_lp], block_offset_width_lp'(0)};
```
dpetrisko, line 306: Think we can use critical_addr directly (not that it really matters, but just for simplicity)

bp_me/src/v/cce/bp_uce.v
```diff
@@ -283,6 +283,9 @@ module bp_uce
   logic [fill_cnt_width_lp-1:0] mem_cmd_cnt;
   logic [block_size_in_fill_lp-1:0] fill_index_shift;
   logic [bank_offset_width_lp-1:0] bank_index;
+  logic [fill_cnt_width_lp-1:0] mem_cmd_init_slice, mem_cmd_final_slice;
```
dpetrisko, line 286: these variables can be defined in the else clause of the below block

bp_me/src/v/cce/bp_uce.v
```diff
@@ -319,20 +323,23 @@ module bp_uce
...
+        ,.en_i(mem_cmd_up)
+        ,.val_i(mem_cmd_init_slice)
         ,.count_o(mem_cmd_cnt)
         );
-      assign mem_cmd_done = (mem_cmd_cnt == block_size_in_fill_lp-1);
+      
+      assign mem_cmd_init_slice = cache_req_cast_i.addr[block_offset_width_lp-1-:fill_cnt_width_lp]; 
+      assign mem_cmd_final_slice = (cache_req_r.addr[block_offset_width_lp-1-:fill_cnt_width_lp] + {fill_cnt_width_lp{1'b1}});
+      assign mem_cmd_done = cache_req_v_r & (mem_cmd_cnt == mem_cmd_final_slice);
+      assign critical_addr = {cache_req_r.addr[paddr_width_p-1:(block_offset_width_lp-fill_cnt_width_lp)], (block_offset_width_lp-fill_cnt_width_lp)'(0)};
```
dpetrisko, line 342: Meta-comments. 
1) We prefer slice width notation [foo+:bar_width] 
2) what does the quantity (block_offset_width_lp-fill_cnt_width_lp) represent? Would fill_offset_width_lp be more descriptive?

bp_me/src/v/cce/bp_uce.v
Shashank-Vijay: If fill_cnt_width is 3 bits, the second operand is replicating 1 3 times, so 3'b111. Instead of doing x - 3'b001, it's doing x + 3'b111 (2's complement). I only did this because the minus operation was not giving the correct result, not sure why. I'll check again

bp_me/src/v/cce/bp_uce.v
Shashank-Vijay: The quantity (block_offset_width_lp-fill_cnt_width_lp) represents the number of bits (based on the fill_width) needed to construct the address sent over the mem_cmd. 

Examples:
1. block_width = 512, fill_width = 64: The quantity is (6 - 3) = 3, meaning only the last 3 bits are made zero and the rest of the address is sent as it is, thereby sending the critical address correctly.
2. block_width = 256, fill_width = 128: The quantity is (5 - 1) = 4, meaning only the last 4 bits are made zero. The remaining 1 bit in the block offset is sent as it is and that identifies the critical address (whether it is 0 or 1). 

Yeah, it would be better to name it as fill_offset_width_lp. I'll use the [foo+:bar_width] notation with <bar width> as (paddr_width_p - fill_offset_width_lp).

bp_me/src/v/cce/bp_uce.v
Shashank-Vijay: Do you mean use just one critical addr for both the if generate blocks? The reason why I split is because fill_cnt_width_lp is `BSG_SAFE_CLOG2(1) when block_width == fill_width, which gives a result of 1 whereas we need a 0.

bp_me/src/v/cce/bp_uce.v
Shashank-Vijay: Changed it to (x-1)

bp_me/src/v/cce/bp_uce.v
dpetrisko: oh sorry typo, I meant cache_req_r.addr. I don't think the block alignment is necessary

bp_me/src/v/cce/bp_uce.v
```diff
@@ -319,20 +324,23 @@ module bp_uce
...
+        ,.en_i(mem_cmd_up)
+        ,.val_i(first_addr)
         ,.count_o(mem_cmd_cnt)
         );
-      assign mem_cmd_done = (mem_cmd_cnt == block_size_in_fill_lp-1);
+      
+      assign first_addr = cache_req_cast_i.addr[block_offset_width_lp-1-:fill_cnt_width_lp];
+      assign last_addr = (cache_req_r.addr[block_offset_width_lp-1-:fill_cnt_width_lp] - fill_cnt_width_lp'(1));
+      assign mem_cmd_done = cache_req_v_r & (mem_cmd_cnt == last_addr);
+      assign critical_addr = {cache_req_r.addr[paddr_width_p-1-:(paddr_width_p-fill_offset_width_lp)], (fill_offset_width_lp)'(0)};
```
dpetrisko, line 343: cache_req_r.addr[paddr_width_p-1-:(paddr_width_p-fill_offset_width_lp)]
->
cache_req_r.addr[paddr_width_p-1-:fill_offset_width_lp]
?

bp_me/src/v/cce/bp_uce.v
```diff
@@ -319,20 +324,23 @@ module bp_uce
...
-
+        ,.set_i(cache_req_v_i)
+        ,.en_i(mem_cmd_up)
+        ,.val_i(first_addr)
         ,.count_o(mem_cmd_cnt)
         );
-      assign mem_cmd_done = (mem_cmd_cnt == block_size_in_fill_lp-1);
+      
+      assign first_addr = cache_req_cast_i.addr[block_offset_width_lp-1-:fill_cnt_width_lp];
+      assign last_addr = (cache_req_r.addr[block_offset_width_lp-1-:fill_cnt_width_lp] - fill_cnt_width_lp'(1));
```
dpetrisko, line 341: apologies, bad rename suggestion on my part. These aren't actually addresses.  How about first_cmd_cnt, last_cmd_cnt?

Additionally, [block_offset_width_lp-1-:fill_cnt_width_lp] -> [fill_offset_width_lp+:fill_cnt_width_lp].  The up-slice is a little clearer as to what's happening

bp_me/src/v/cce/bp_uce.v
dpetrisko: I could be wrong though. Is it necessary?

bp_me/src/v/cce/bp_uce.v
dpetrisko, line 343: Hm, I actually think paddr_width_p-1:fill_offset_width_lp was clearer here, since we're also doing the concatenation.  Soft suggestion, up to you

bp_me/src/v/cce/bp_uce.v
Shashank-Vijay: It was done in the previous version as well. Doing the block alignment here will ensure that for fill_width == block_width, the address sent over the mem_cmd is the start of that block. If we send cache_req_r..addr directly, we might send a mem_cmd like (addr: 80000018, size: 64 bytes for block_width 512). I'm not sure what would happen then (if there would be an alignment somewhere further down?)

bp_me/src/v/cce/bp_uce.v
Shashank-Vijay: That missed my eye too! I'll change it

bp_me/src/v/cce/bp_uce.v
Shashank-Vijay: Sure, I can change that back

PR #594, opened 2020-07-30
Update dev to sw_dev

dpetrisko:
- Added support for DRAMSim3 memory system
- Removing dramsim2/3 submodules
- Moving VCS timeouts to testbench makefiles
- Skip zero DRAM entries in NBF (quicker restore time for checkpoints)


dpetrisko:
Holding off until #596

PR #596, opened 2020-08-01
Adding backdoor load to bp_mem, both fixed latency and dramsim

dpetrisko:
Title.  Basically, allows preloading from readmemh even while using bsg_mem SRAMs.

A future enhancement may do this all in C host code, such that we can deallocate an array of the readmemh data, even though it's only read/written once.


dpetrisko:
@farzamgl This works for both dramsim3 and fixed latency, preload_p and not preload_p.  It's failing for all of the sample targets though.  And trying with sw_dev, I think the same problem exists there. Can you try and see if you run into issues?

PR #599, opened 2020-08-06
Supporting multiple caches in the D$ testbench and changes in I$ testbench

Shashank-Vijay:
This PR adds the following:

- Multiple caches in the dcache testbench
- LCE tracer in the dcache testbench
- Multicycle fill test in the trace script


bp_be/test/tb/bp_be_dcache/wrapper.v
```diff
@@ -60,342 +65,726 @@ module wrapper
...
+           ,.lce_cmd_o(lce_cmd_lo[i])
+           ,.lce_cmd_v_o(lce_cmd_v_lo[i])
+           ,.lce_cmd_ready_i(lce_cmd_ready_li[i])
+
+           ,.credits_full_o(credits_full_lo[i])
+           ,.credits_empty_o(credits_empty_lo[i])
+           );
+
+           // Supporting logic for multicore case
+           if (num_caches_p > 1)
```
dpetrisko, line 319: No need to special case this, the multicore case can handle the single core case

bp_be/test/tb/bp_be_dcache/wrapper.v
```diff
@@ -60,342 +65,726 @@ module wrapper
...
+              ,.timeout_max_limit_p(4)
+              ,.credits_p(coh_noc_max_credits_p)
+              ,.data_mem_invert_clk_p(1)
+              ,.tag_mem_invert_clk_p(1)
+             )
+           dcache_lce
+           (.clk_i(clk_i)
+           ,.reset_i(reset_i)
+
+           ,.lce_id_i(/*cfg_bus_cast_i.dcache_id*/lce_id_width_p'(i))
```
dpetrisko, line 271: Should punch these cfg_bus signals out of the wrapper and drive them from the testbench

bp_be/test/tb/bp_be_dcache/wrapper.v
```diff
@@ -60,342 +65,726 @@ module wrapper
...
+           ,.lce_cmd_ready_i(lce_cmd_ready_li[i])
+
+           ,.credits_full_o(credits_full_lo[i])
+           ,.credits_empty_o(credits_empty_lo[i])
+           );
+
+           // Supporting logic for multicore case
+           if (num_caches_p > 1) 
+             begin : multiple
+               bp_me_cce_id_to_cord
```
dpetrisko, line 321: These are always 0 since there's one CCE, right?

bp_be/test/tb/bp_be_dcache/wrapper.v
```diff
@@ -60,342 +65,726 @@ module wrapper
...
+                   );
+               
+               assign lce_req_packet_lo[i].payload = lce_req_lo[i];
+               assign lce_req_packet_lo[i].cid = req_cce_cid_li[i];
+               assign lce_req_packet_lo[i].cord = req_cce_cord_li[i];
+               assign lce_req_packet_lo[i].len = coh_noc_len_width_p'(0);
+
+               // Request adapter to convert the request packet to the link
+               // format
+               bsg_wormhole_router_adapter_in
```
dpetrisko, line 336: If it's a single flit, can directly assign to the fields in the packet

bp_be/test/tb/bp_be_dcache/wrapper.v
```diff
@@ -60,342 +65,726 @@ module wrapper
...
+                    ,.packet_i(lce_req_packet_lo[i])
+                    ,.v_i(lce_req_v_lo[i])
+                    ,.ready_o(lce_req_ready_li[i])
+
+                    ,.link_i(lce_req_link_li[i])
+                    ,.link_o(lce_req_link_lo[i])
+                    );
+
+               
+               bp_me_lce_id_to_cord
```
dpetrisko, line 355: Same thing here, can directly assign cord to 0

bp_be/test/tb/bp_be_dcache/wrapper.v
```diff
@@ -60,342 +65,726 @@ module wrapper
...
+                  );
+
+               assign lce_cmd_packet_lo[i].payload = lce_cmd_lo[i];
+               assign lce_cmd_packet_lo[i].cid = lce_cmd_lo[i].header.dst_id;
+               assign lce_cmd_packet_lo[i].cord = lce_cord_li[i];
+               assign lce_cmd_packet_lo[i].len = coh_noc_len_width_p'(0);
+
+               // Command out adapter to convert the command packet to the
+               // link format
+               bsg_wormhole_router_adapter
```
dpetrisko, line 370: Same comment as above adapter

bp_be/test/tb/bp_be_dcache/wrapper.v
```diff
@@ -60,342 +65,726 @@ module wrapper
...
+                    ,.link_i(lce_cmd_link_li[i])
+                    ,.link_o(lce_cmd_link_lo[i])
+
+                    ,.packet_o(lce_cmd_packet_li[i])
+                    ,.v_o(lce_cmd_v_li[i])
+                    ,.yumi_i(lce_cmd_yumi_lo[i])
+                    );
+               assign lce_cmd_li[i] = lce_cmd_packet_li[i].payload;
+
+               bp_me_cce_id_to_cord
```
dpetrisko, line 393: Same comment, just 0

bp_be/test/tb/bp_be_dcache/wrapper.v
```diff
@@ -60,342 +65,726 @@ module wrapper
...
+                    ,.packet_i(lce_resp_packet_lo[i])
+                    ,.v_i(lce_resp_v_lo[i])
+                    ,.ready_o(lce_resp_ready_li[i])
+
+                    ,.link_i(lce_resp_link_li[i])
+                    ,.link_o(lce_resp_link_lo[i])
+                    );
+             end
+           else
+             // Single core case: Connecting the signals directly without
```
dpetrisko, line 427: Remove, single core can be handled the same as the rest of the LCE configs

bp_be/test/tb/bp_be_dcache/wrapper.v
```diff
@@ -60,342 +65,726 @@ module wrapper
...
+             ,.ready_o(fifo_mem_cmd_ready_li)
+
+             ,.v_o(mem_cmd_v_o)
+             ,.data_o(mem_cmd_o)
+             ,.yumi_i(mem_cmd_v_o & mem_cmd_ready_i)
+             );
+         end
+     end
+
+   if ((uce_p == 0) & (num_caches_p > 1))
```
dpetrisko, line 563: Put this in the lce case above for clarity

bp_be/test/tb/bp_be_dcache/wrapper.v
```diff
@@ -60,342 +65,726 @@ module wrapper
...
+            ,.links_o(lce_resp_link_li)
+
+            ,.concentrated_link_i(resp_concentrated_link_li)
+            ,.concentrated_link_o(resp_concentrated_link_lo)
+            );
+     end   
+
+   if (uce_p == 0)
+     begin : cce
+       logic mem_resp_v_to_cce, mem_resp_yumi_from_cce, mem_resp_ready_lo;
```
dpetrisko, line 722: Same thing, put in the lce case

bp_be/test/tb/bp_be_dcache/wrapper.v
Shashank-Vijay: Oh yeah, I made that distinction so that we wouldn't have the concentrators and adapters in the single core case, but it doesn't really matter in a testbench I guess. I'll change it

bp_be/test/tb/bp_be_dcache/wrapper.v
Shashank-Vijay: They are driven from the testbench but the dcache_id was assigned as '0 in the always comb case. That's why i just just used the genvar. I'll remove the commented part

bp_be/test/tb/bp_be_dcache/wrapper.v
Shashank-Vijay: That's true. The module's output could also potentially be a problem because the dst_id fed into them are not always 0 as we discussed yesterday. I'll remove the modules and do a direct assignment

bp_be/test/tb/bp_be_dcache/wrapper.v
Shashank-Vijay: I kept this outside that case because we want only one of this for a wrapper. The above lce case is inside a for generate loop. I can merge the cce part below and this though

bp_be/test/tb/bp_be_dcache/wrapper.v
```diff
@@ -60,342 +65,584 @@ module wrapper
...
+       ,.mem_resp_i(mem_resp_to_cce)
+       ,.mem_resp_v_i(mem_resp_v_to_cce)
+       ,.mem_resp_yumi_o(mem_resp_yumi_from_cce)
+
+       ,.mem_cmd_o(mem_cmd_from_cce)
+       ,.mem_cmd_v_o(mem_cmd_v_from_cce)
+       ,.mem_cmd_ready_i(mem_cmd_ready_to_cce)
+       );
+
+       // Inbound Mem to CCE
```
dpetrisko, line 613: Do we need fifos here? The testbench is helpful

bp_be/test/tb/bp_be_dcache/wrapper.v
```diff
@@ -60,342 +65,584 @@ module wrapper
...
+             ,.ready_o(fifo_mem_cmd_ready_li)
+
+             ,.v_o(mem_cmd_v_o)
+             ,.data_o(mem_cmd_o)
+             ,.yumi_i(mem_cmd_v_o & mem_cmd_ready_i)
+             );
+         end
+     end
+
+   if (uce_p == 0)
```
dpetrisko, line 449: Combine this with the previous uce_p == 0 case

bp_be/test/tb/bp_be_dcache/wrapper.v
```diff
@@ -60,342 +65,584 @@ module wrapper
...
+            );
+       assign cce_lce_resp_li = cce_lce_resp_packet_li.payload;
+
+       assign req_concentrated_link_li = '{data          : req_concentrated_link_lo.data
+                                           ,v            : req_concentrated_link_lo.v
+                                           ,ready_and_rev: cce_lce_req_link_lo.ready_and_rev
+                                           };
+
+       // Request concentrator
+       bsg_wormhole_concentrator_in
```
dpetrisko, line 518: // We use concentrators just for the 1->N arbitration capabilities, rather than serialization

bp_be/test/tb/bp_be_dcache/wrapper.v
Shashank-Vijay: We need the mem_resp fifo. I checked without both and a few tests timed out. When I added the resp fifo back, everything works. Probably because of multiple responses in quick succession during writeback? I'll remove the cmd fifo though, doesn't seem to be required.

dpetrisko: Probably worth seeing why that is.  Generally that means there's a handshake somewhere not being obeyed

dpetrisko:
@Shashank-Vijay Let's see if we can figure out if we actually need that mem_resp fifo.  Other than that, looks good to go

dpetrisko (approved):

PR #600, opened 2020-08-06
Merging the store data into the store miss response

Shashank-Vijay:
This PR adds the following:

- A store_merge module that merges the store data at the appropriate place in the store miss fill response
- Connections to this module in both UCE and LCE
- Connections for the cache_req_safe signal

### DIFF
--- a/bp_be/src/v/bp_be_mem/bp_be_dcache/bp_be_dcache.v
+++ b/bp_be/src/v/bp_be_mem/bp_be_dcache/bp_be_dcache.v
@@ -647,8 +647,8 @@ module bp_be_dcache
     cache_req_cast_o = '0;
 
     // Assigning sizes to cache miss packet
-    if (uncached_tv_r | wt_req | l2_amo_req) begin
-      if (decode_tv_r.double_op)
+    if (uncached_tv_r | wt_req | l2_amo_req | store_miss_tv) begin
+      unique if (decode_tv_r.double_op)
         cache_req_cast_o.size = e_size_8B;
       else if (decode_tv_r.word_op)
         cache_req_cast_o.size = e_size_4B;
@@ -656,6 +656,8 @@ module bp_be_dcache
         cache_req_cast_o.size = e_size_2B;
       else if (decode_tv_r.byte_op)
         cache_req_cast_o.size = e_size_1B;
+      else
+        cache_req_cast_o.size = e_size_8B;
     end
     else
       cache_req_cast_o.size = e_size_64B;
@@ -712,7 +714,7 @@ module bp_be_dcache
     end
 
     cache_req_cast_o.addr = paddr_tv_r;
-    cache_req_cast_o.data = data_tv_r;
+    cache_req_cast_o.data = store_miss_tv ? wbuf_entry_in.data : data_tv_r;
   end
 
   // Cache metadata is valid after the request goes out
--- a/bp_be/src/v/bp_be_mem/bp_be_dcache/bp_be_dcache.v
+++ b/bp_be/src/v/bp_be_mem/bp_be_dcache/bp_be_dcache.v
@@ -145,6 +145,7 @@ module bp_be_dcache
 
     , input cache_req_complete_i
     , input cache_req_critical_i
+    , input cache_req_safe_i
 
     // data_mem
     , input data_mem_pkt_v_i
@@ -745,7 +746,7 @@ module bp_be_dcache
      ,.clear_i(cache_req_complete_i)
      ,.data_o(cache_miss_r)
      );
-  assign ready_o = cache_req_ready_i & ~cache_miss_r;
+  assign ready_o = (cache_req_ready_i & ~cache_miss_r) | cache_req_safe_i;
 
   assign early_v_o = v_tv_r & ((uncached_tv_r & (decode_tv_r.load_op & uncached_load_data_v_r))
                               | (uncached_tv_r & (decode_tv_r.store_op & cache_req_ready_i))
--- a/bp_be/src/v/bp_be_mem/bp_be_dcache/bp_be_dcache.v
+++ b/bp_be/src/v/bp_be_mem/bp_be_dcache/bp_be_dcache.v
@@ -746,7 +746,7 @@ module bp_be_dcache
      ,.clear_i(cache_req_complete_i)
      ,.data_o(cache_miss_r)
      );
-  assign ready_o = (cache_req_ready_i & ~cache_miss_r) | cache_req_safe_i;
+  assign ready_o = cache_req_ready_i & ~cache_miss_r;
 
   assign early_v_o = v_tv_r & ((uncached_tv_r & (decode_tv_r.load_op & uncached_load_data_v_r))
                               | (uncached_tv_r & (decode_tv_r.store_op & cache_req_ready_i))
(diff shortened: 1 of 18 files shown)


bp_be/src/v/bp_be_mem/bp_be_dcache/bp_be_dcache.v
```diff
@@ -743,7 +746,7 @@ module bp_be_dcache
      ,.clear_i(cache_req_complete_i)
      ,.data_o(cache_miss_r)
      );
-  assign ready_o = cache_req_ready_i & ~cache_miss_r;
+  assign ready_o = (cache_req_ready_i & ~cache_miss_r) | cache_req_safe_i;
```
dpetrisko, line 749: Take this out, we're not handling cache_req_safe in the dcache yet

bp_be/src/v/bp_be_mem/bp_be_dcache/bp_be_dcache.v
```diff
@@ -712,7 +715,7 @@ module bp_be_dcache
     end
 
     cache_req_cast_o.addr = paddr_tv_r;
-    cache_req_cast_o.data = data_tv_r;
+    cache_req_cast_o.data = store_miss_tv ? wbuf_entry_in.data : data_tv_r;
```
dpetrisko, line 718: Before, we only used the raw data from TV, without store byte muxing.  I can see why we may use the replicated data, but why not use it all the time, to avoid the mux?

bp_fe/src/v/bp_fe_icache.v
```diff
@@ -385,7 +386,7 @@ module bp_fe_icache
      ,.data_o(cache_miss)
      );
 
-  assign vaddr_ready_o = cache_req_ready_i & ~cache_miss & ~cache_req_v_o;
+  assign vaddr_ready_o = (cache_req_ready_i & ~cache_miss & ~cache_req_v_o) | cache_req_safe_i;
```
dpetrisko, line 389: Revert

bp_me/src/v/lce/bp_lce.v
```diff
@@ -215,6 +224,38 @@ module bp_lce
       ,.lce_cmd_ready_i(lce_cmd_ready_i)
       );
 
+  // Write merge logic
```
dpetrisko, line 227: Can't we tell if it's a store miss based on FSM state? Regardless, this should all be in the lce_req block

bp_top/src/v/bp_core.v
```diff
@@ -138,6 +139,7 @@ module bp_core
 
      ,.icache_req_complete_i(icache_req_complete_lo)
      ,.icache_req_critical_i(icache_req_critical_lo)
+     ,.icache_req_safe_i('0)
```
dpetrisko, line 142: Should come from LCE?

bp_me/src/v/lce/bp_store_data_merge.v
```diff
@@ -0,0 +1,102 @@
...
+    begin : fill_dword
+      assign addr_fill_slice = '0;
+    end
+  else if (fill_width_p == block_width_p)
+    begin : fill_block
+      assign addr_fill_slice = (cache_req_cast_i.addr[block_offset_width_lp-1:3] << 6);
+    end
+  else 
+    begin : fill_chunks
+      assign addr_fill_slice = (cache_req_cast_i.addr[block_offset_width_lp-fill_cnt_width_lp-1:3] << 6);
```
dpetrisko, line 59: 6?

bp_me/src/v/lce/bp_store_data_merge.v
```diff
@@ -0,0 +1,102 @@
...
+   , localparam bank_offset_width_lp   = `BSG_SAFE_CLOG2(assoc_p)
+   , localparam block_offset_width_lp  = (bank_offset_width_lp + byte_offset_width_lp)
+   , localparam block_size_in_fill_lp  = block_width_p / fill_width_p
+   , localparam fill_cnt_width_lp      = `BSG_SAFE_CLOG2(block_size_in_fill_lp)
+
+   `declare_bp_cache_service_if_widths(paddr_width_p, ptag_width_p, sets_p, assoc_p, dword_width_p, block_width_p, fill_width_p, cache)
+  )
+  (
+    input [fill_width_p-1:0] data0_i
+    , input [cache_req_width_lp-1:0] cache_req_i
```
dpetrisko, line 35: This module shouldn't speak cache_req, just data1_i, size and address

bp_me/src/v/lce/bp_store_data_merge.v
```diff
@@ -0,0 +1,102 @@
...
+      assign addr_fill_slice = (cache_req_cast_i.addr[block_offset_width_lp-1:3] << 6);
+    end
+  else 
+    begin : fill_chunks
+      assign addr_fill_slice = (cache_req_cast_i.addr[block_offset_width_lp-fill_cnt_width_lp-1:3] << 6);
+    end
+
+  assign byte_select = (cache_req_cast_i.addr[2:0] << 3);
+
+  bsg_mux_bitwise
```
dpetrisko, line 64: should use bsg_mux_segmented with segment 8 for byte granularity

bp_me/src/v/lce/bp_lce.v
Shashank-Vijay: No, both load miss and store miss are handled in the same state e_send_cached_req in lce_req. We would need the check for msg_type. The reason why this logic is currently in the top LCE module is because the data_mem_pkt comes from the lce_cmd module. It didn't seem clean to send that packet into the lce_req module and send it to the output from there, when the other packets go to the output from the command module. We wouldn't even need the cache_req_r signal pulled from the lce_req module since we have the cache_req_i signal entering the top module. I just did that to reuse the register instead of creating one more at the top level.

bp_top/src/v/bp_core.v
Shashank-Vijay: Yeah, forgot to add the signals for the LCE case. Will do now

bp_me/src/v/lce/bp_store_data_merge.v
Shashank-Vijay: Multiply by 64 to get the correct dword offset. I've written a comment there now

Shashank-Vijay: Now that we have byte granularity, it is a shift by 3

bp_be/src/v/bp_be_mem/bp_be_dcache/bp_be_dcache.v
Shashank-Vijay: Yeah we could use just the replicated data

dpetrisko:
I'm not convinced this is the cleanest way to do this.  Since we need the capability to "re-read" the memories when recovering from cache misses, we can also have a "re-write".  

Basically, when a read miss is resolved, we reread the memories at the address stored in TV. Then the hit proceeds as normal.  When a write miss is resolved, we don't need to reread the memories, but we can just store the data in TV to the write buffer afterwards

dpetrisko:
Closing since we're moving towards the snoop+recover approach

dpetrisko:
Feel free to open a new PR with just the cache_req_safe changes, since those are orthogonal

PR #601, opened 2020-08-06
Fixing up flow for tirekick instructions

dpetrisko:
- Adding bootrom to repo
- Adding nbf generation to tire kick flow
- Adding CI test for tire kick


bp_common/test/mem/bootrom.dump
```diff
@@ -0,0 +1,56 @@
+
+/mnt/users/ssd0/no_backup/petrisko/pre-alpha-release/bp_top/syn/results/vcs/bp_tethered.e_bp_unicore_cfg.none.sim.bp_tests.hello_world/bootrom.riscv:     file format elf64-littleriscv
```
muwyse, line 2: Is it possible / do you want to remove this full path since it is specific to you?

bp_common/test/mem/hello_world.dump
```diff
@@ -1,5 +1,54 @@
 
-/mnt/users/ssd0/no_backup/petrisko/pre-alpha-release/bp_top/syn/results/vcs/bp_tethered.e_bp_softcore_cfg.none.sim.bp_tests.hello_world/prog.riscv:     file format elf64-littleriscv
+/mnt/users/ssd0/no_backup/petrisko/pre-alpha-release/bp_top/syn/results/vcs/bp_tethered.e_bp_unicore_cfg.none.sim.bp_tests.hello_world/prog.riscv:     file format elf64-littleriscv
```
muwyse, line 2: same as above

bp_common/test/mem/bootrom.dump
dpetrisko: Good catch, will remove

PR #602, opened 2020-08-06
Separate out header / data from wormhole definitions

### LINKED ISSUE #464: Mem-Wormhole Packet incorrect padding
Found these in simulation, correct padding should be 
flit_width_mp-((2*cord_width_mp+2*cid_width_mp+len_width_mp+msg_width_mp)%flit_width_mp)

https://github.com/black-parrot/black-parrot/blob/cd949112f849edc1a102fdde71375d156fa55798/bp_me/src/include/v/bp_mem_wormhole.vh#L31
https://github.com/black-parrot/black-parrot/blob/cd949112f849edc1a102fdde71375d156fa55798/bp_me/src/include/v/bp_mem_wormhole.vh#L64

dpetrisko:
This PR is mostly in preparation for wormhole streamer changes, which require separation of header/data ports for various modules.  No functional change / no PPA change.

- Separating out header / data in wormhole definitions
- Fixing wormhole padding definitions in bp_mem_wormhole.vh (Fixes #464)
- Refactored wormhole encoders to only translate header->header
- Minor bug fix: Line 512 of bp_tile _lo -> _v_lo

### DIFF
--- a/bp_me/src/include/v/bp_mem_wormhole.vh
+++ b/bp_me/src/include/v/bp_mem_wormhole.vh
@@ -23,7 +23,7 @@
                                         \
   typedef struct packed                 \
   {                                     \
-    logic [(flit_width_mp-(2*cord_width_mp+2*cid_width_mp+len_width_mp+$bits(msg_hdr_name_mp))%flit_width_mp)-1:0] \
+    logic [`bp_mem_wormhole_packet_pad_width(flit_width_mp, cord_width_mp, len_width_mp, cid_width_mp, $bits(msg_hdr_name_mp))-1:0] \
                               pad;      \
     msg_hdr_name_mp           msg_hdr;  \
     bp_mem_router_header_s    wh_hdr;   \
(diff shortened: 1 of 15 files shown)


bp_me/src/include/v/bp_mem_wormhole.vh
```diff
@@ -11,58 +11,96 @@
...
-    logic [`bp_mem_wormhole_packet_pad_width(flit_width_mp, cord_width_mp, len_width_mp, cid_width_mp, msg_width_mp)-1:0] \
-                              pad;      \
-    logic [msg_width_mp-1:0]  msg;      \
     logic [cid_width_mp-1:0]  src_cid;  \
     logic [cord_width_mp-1:0] src_cord; \
     logic [cid_width_mp-1:0]  cid;      \
     logic [len_width_mp-1:0]  len;      \
     logic [cord_width_mp-1:0] cord;     \
-  }  struct_name_mp
+  }  bp_mem_router_header_s;            \
```
muwyse, line 22: If I understand correctly, the bp_mem_router_header_s is basically a superset of bsg_wormhole_router_header_s as defined in basejump_stl/bsg_noc/bsg_wormhole_router.vh, which has only cord and len fields. We add the others for our specific needs?

dpetrisko: Yup exactly. The wormhole router itself only considers cord/len.  On top of it, there are other fields like CID that form the "higher level protocol", usually related either to more detailed routing information, or return path formation.

For the mem network, this includes the source cord/cid so that we can return the message properly. In the coh network, the source LCE id is already embedded in the msg_hdr, which is enough to reconstruct the return path. So we just need the cid

PR #606, opened 2020-08-10
add memory credit counter to ucode CCE

### LINKED ISSUE #581: Track memory network credits in ucode CCE
https://github.com/black-parrot/black-parrot/blob/8a1e61660a0ea2b137b8d750349b7242e5b25ef1/bp_me/src/v/cce/bp_cce_msg.v#L179

Currently, we track only outstanding uncached requests. We should instead increment on mem_cmd_v_o and decrement on mem_resp_yumi_o, using bsg_flow_counter

muwyse:
This PR closes #581 by adding a memory credit counter to the ucode CCE. It also fixes the transition from uncached only to normal operation mode.

### DIFF
--- a/bp_me/src/v/cce/bp_cce_msg.v
+++ b/bp_me/src/v/cce/bp_cce_msg.v
@@ -34,9 +34,6 @@ module bp_cce_msg
     // counter width (used for e.g., stall and performance counters)
     , localparam counter_width_lp          = 64
 
-    // maximum number of memory requests in uncached only mode
-    , parameter max_mem_requests_p = mem_noc_max_credits_p
-
     // Interface Widths
     , localparam mshr_width_lp             = `bp_cce_mshr_width(lce_id_width_p, lce_assoc_p, paddr_width_p)
     , localparam cfg_bus_width_lp          = `bp_cfg_bus_width(vaddr_width_p, core_id_width_p, cce_id_width_p, lce_id_width_p, cce_pc_width_p, cce_instr_width_p)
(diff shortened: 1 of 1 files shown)


bp_me/src/v/cce/bp_cce_msg.v
```diff
@@ -385,15 +391,16 @@ module bp_cce_msg
 
         // request logic
         // cached requests will stall on the input port until normal mode is entered
-        if (lce_req_v_i) begin
+        // The uncached mode FSM stops issuing requests as soon as the config bus sets the mode
+        // to normal mode to ensure that the mode transition happens.
+        if (lce_req_v_i && ~(cfg_bus_cast.cce_mode == e_cce_mode_normal)) begin
```
dpetrisko, line 396: Why is this necessary?

bp_me/src/v/cce/bp_cce_msg.v
```diff
@@ -170,29 +167,44 @@ module bp_cce_msg
...
+      ,.reset_i(reset_i)
+      // memory commands consume credits
+      ,.v_i(mem_cmd_v_o)
+      ,.ready_i(mem_cmd_ready_i)
+      // memory responses return credits
+      ,.yumi_i(mem_resp_yumi_o)
+      ,.count_o(mem_credit_count_lo)
+      );
+
+  wire mem_credit_free = (mem_credit_count_lo < mem_noc_max_credits_p);
```
dpetrisko, line 188: nit: credits_full and credits_empty are more typical names

bp_me/src/v/cce/bp_cce_msg.v
dpetrisko: Is there a case where an incoming lce_req causes deadlock, or is just a "switch into cached mode ASAP"?

bp_me/src/v/cce/bp_cce_msg.v
muwyse: credits_empty = (credit_count  == 0), correct?

bp_me/src/v/cce/bp_cce_msg.v
muwyse: The uncached mode FSM needs to stop issuing requests when the config bus tells it to change modes, but the decision to run the uncached mode FSM or normal mode code is made based on the registered value of the mode. The registered mode only transitions once all memory responses are drained. This check ensures that the uncached mode stops issuing new memory commands while it drains memory responses.

Without the check it would be possible for the uncached FSM to ignore the mode transition and continue to process uncached requests, so long as there was always an outstanding memory response.

bp_me/src/v/cce/bp_cce_msg.v
dpetrisko: yep

bp_me/src/v/cce/bp_cce_msg.v
dpetrisko: Makes sense! My take is that there's no real "correct" timing for switching modes, since the config bus can be driven remotely and is asynchronous to the rest of the system. Software which cares about which cmds/responses are cached/uncached need to employ a global fence operation regardless.  So I think this check can be removed without affecting correctness

muwyse: okay; code has been adjusted. Mode register is removed and the CCE transitions with the signal from the config bus. It is up to software to employ a fence if a stricter separation between modes is required. Normal mode will process and drain any outstanding memory responses just as uncached mode would have.

PR #612, opened 2020-08-23
BP Stream Design Between UCE and L2 adapter in unicore

Songchun-Li:
This pull request include multiple changes to implement BP Stream Protocol between UCE and L2, which are
1. Separated out the header and data port for UCE.
2. Added lock signals for mem_cmd/resp.
3. Updated the logic with the new locking arbiter and the new cache adapter.
4. Add bp_me_cache_slice_new for the new cache adapter.

Note. The current design sends multiple cmds for cache read when the fill width is less than a cache block.


bp_me/src/v/cce/bp_uce_stream.v
```diff
@@ -0,0 +1,729 @@
+// This uce_stream module sends only one request for cache block load
```
dpetrisko, line 1: name bp_uce.  Otherwise, we don't get a diff and it's very hard to review

bp_top/src/v/bp_unicore.v
```diff
@@ -435,9 +476,11 @@ module bp_unicore
 
       assign resp_fifo_v_li[i] = resp_selected_v_li & (resp_fifo_selected_li.header.payload.lce_id == i);
       assign resp_fifo_li[i] = resp_fifo_selected_li;
+      assign resp_fifo_lock_li[i] = resp_fifo_v_li[i] & resp_selected_lock_li;
     end
-
+  
   /* TODO: Extract local memory map to module */
+  // when the counter is on , all these staff are also locked
```
dpetrisko, line 483: what does this comment mean?

bp_top/src/v/bp_unicore.v
```diff
@@ -447,43 +490,49 @@ module bp_unicore
...
+  assign cache_cmd_lock_li      = is_cache_cmd & |cmd_fifo_yumi_li & cmd_selected_lock_li; // Only L2 will used the locking signal
 
   bp_cce_loopback
    #(.bp_params_p(bp_params_p))
    loopback
     (.clk_i(clk_i)
      ,.reset_i(reset_i)
 
-     ,.mem_cmd_i(loopback_cmd_li)
+     ,.mem_cmd_i(loopback_cmd_li[0+:uce_mem_msg_width_lp])
```
dpetrisko, line 506: why is this width mismatch occurring?

bp_top/src/v/bp_unicore.v
```diff
@@ -447,43 +490,49 @@ module bp_unicore
...
      ,.mem_resp_v_o(loopback_resp_v_lo)
      ,.mem_resp_yumi_i(loopback_resp_yumi_li)
      );
 
   if (l2_en_p)
     begin : l2
       logic mem_resp_ready_lo;
-      bp_me_cache_slice
-       #(.bp_params_p(bp_params_p))
+      bp_me_cache_slice_stream
```
dpetrisko, line 518: I don't see this module in the PR?

bp_top/src/v/bp_unicore.v
```diff
@@ -507,3 +556,72 @@ module bp_unicore
     end
 
 endmodule
+
+
+module bsg_locking_arb_fixed_new #( parameter inputs_p="inv"
```
dpetrisko, line 561: Can just point the submodule to your basejump_stl branch with this module

bp_top/src/v/bp_unicore.v
```diff
@@ -189,8 +190,10 @@ module bp_unicore
      );
 
   wire [1:0][lce_id_width_p-1:0] lce_id_li = {cfg_bus_lo.dcache_id, cfg_bus_lo.icache_id};
+  wire dcache_uce_resp_lock_li = proc_resp_lock_li[1] & proc_resp_v_li[1];
+  wire icache_uce_resp_lock_li = proc_resp_lock_li[0] & proc_resp_v_li[0];
 
-  bp_uce
+  bp_uce_stream
```
dpetrisko, line 196: I think you need to update the no_l2 case as well, with the stream2burst module

bp_top/src/v/bp_unicore.v
```diff
@@ -189,8 +190,10 @@ module bp_unicore
      );
 
   wire [1:0][lce_id_width_p-1:0] lce_id_li = {cfg_bus_lo.dcache_id, cfg_bus_lo.icache_id};
+  wire dcache_uce_resp_lock_li = proc_resp_lock_li[1] & proc_resp_v_li[1];
```
dpetrisko, line 193: Shouldn't have to gate with resp_v_li, right? Lock should only be high when the stream is valid

Songchun-Li:
I was working on the other branch to implement the stream design with the pumps, so the latest work is not in this PR. I will rebase this branch to the last change I made soon.

Songchun-Li:
I've pushed my work of implementing bp_stream with pumps at two sides here, where I got the last benchmark, except the locking arbiter is removed from the unicore file. (I ran the tests again to make sure it is good for review) 

However, as the work is based on another PR of pump modules, the work here is also based on the me_dev_feature_dmanoserial branch. Should I change the merge target to me_dev_feature_dmanoserial rather than me_dev?

Songchun-Li:
This PR is rebased to the latest pump modules update. It is tested and is ready for review.

Songchun-Li:
Change to used the original locking arbiter with the change of last signal in the stream pumps. Passed cache hammer and riscv tests, and is ready for review

bp_me/src/v/cache/bp_me_cce_to_cache.sv
```diff
@@ -21,16 +21,16 @@ module bp_me_cce_to_cache
...
     , parameter block_offset_width_lp=(word_offset_width_lp+byte_offset_width_lp)
     
-    , parameter bsg_cache_pkt_width_lp=`bsg_cache_pkt_width(paddr_width_p,dword_width_p)
-    , parameter counter_width_lp=`BSG_SAFE_CLOG2(cce_block_width_p/dword_width_p)
+    , parameter bsg_cache_pkt_width_lp=`bsg_cache_pkt_width(paddr_width_p,cache_data_width_p)
 
-    , localparam stream_words_lp = cce_block_width_p/cache_data_width_p
-    , localparam data_len_width_lp = `BSG_SAFE_CLOG2(stream_words_lp)
-    , localparam stream_offset_width_lp = `BSG_SAFE_CLOG2(cache_data_width_p / 8)
+    , localparam block_width_lp = `BSG_MAX(icache_block_width_p, dcache_block_width_p)
```
dpetrisko, line 31: This is == cce_block_width_p, just use that

bp_me/src/v/wormhole/bp_burst_to_stream.sv
```diff
@@ -0,0 +1,195 @@
+
+module bp_burst_to_stream
+ import bp_common_pkg::*;
+ import bp_common_aviary_pkg::*;
+ import bp_me_pkg::*;
+ #(parameter bp_params_e bp_params_p = e_bp_default_cfg
+   `declare_bp_proc_params(bp_params_p)
+   // Assuming in_data_width_p(burst) <= out_data_width_p(stream)
+   , parameter in_data_width_p  = "inv"
```
dpetrisko, line 9: Assume in == out and remove all buffering from this module

bp_me/src/v/wormhole/bp_stream_to_burst.sv
```diff
@@ -0,0 +1,125 @@
+
+module bp_stream_to_burst
+ import bp_common_pkg::*;
+ import bp_common_aviary_pkg::*;
+ import bp_me_pkg::*;
+ #(parameter bp_params_e bp_params_p = e_bp_default_cfg
+   `declare_bp_proc_params(bp_params_p)
+   // Assuming in_data_width_p(stream) >= out_data_width_p(burst)
+   , parameter in_data_width_p  = "inv"
```
dpetrisko, line 9: Assume in == out and remove all buffering from this module

bp_me/src/v/cache/bp_me_cce_to_cache.sv
```diff
@@ -154,15 +154,67 @@ module bp_me_cce_to_cache
...
+    end
+  bsg_rotate_left
+   #(.width_p(data_mask_width_lp))
+   mask_rotator
+    (.data_i(set_mask)
+    ,.rot_i(mask_shift_amt_li)
+    ,.o(shifted_mask_lo)
+    );
   
+  if (icache_fill_width_p != dcache_fill_width_p)
```
dpetrisko, line 186: This seems strange. Why would we need to rotate the return data in this case?

bp_me/src/v/cache/bp_me_cce_to_cache.sv
```diff
@@ -154,15 +154,67 @@ module bp_me_cce_to_cache
...
     ,.stream_cnt_o(mem_resp_cnt_lo)
     ,.stream_done_o(mem_resp_done_lo)
     );
   assign cache_yumi_o = mem_resp_yumi_lo | (is_clear & cache_v_i);
+
+  logic [data_mask_width_lp-1:0] set_mask, shifted_mask_lo;
+  logic [lg_data_mask_width_lp-1:0] mask_shift_amt_li;
+  always_comb
+    begin
+      mask_shift_amt_li = mem_cmd_stream_addr_lo[byte_offset_width_lp+:lg_cache_data_size_in_bytes_lp] << 3;
```
dpetrisko, line 170: We only allow misaligned requests.  I think this block is handling the more general case of misaligned requests

bp_me/src/v/cache/bp_me_cce_to_cache.sv
dpetrisko: Possibly due to misaligned request handling as commented above?

bp_top/src/v/bp_unicore.sv
```diff
@@ -377,62 +387,68 @@ module bp_unicore
...
   bp_bedrock_uce_mem_msg_s [2:0] cmd_fifo_lo;
-  logic [2:0] cmd_fifo_v_lo, cmd_fifo_yumi_li;
-
-  bp_bedrock_uce_mem_msg_s [2:0] resp_fifo_li;
-  logic [2:0] resp_fifo_v_li, resp_fifo_ready_lo;
-
+  logic [2:0] cmd_fifo_last_lo, cmd_fifo_v_lo, cmd_fifo_yumi_li;
   for (genvar i = 0; i < 3; i++)
     begin : fifo
       bsg_two_fifo
```
dpetrisko, line 397: I think we wanted to get rid of these fifos? We'd need the UCE mem interface to be ready_and instead of ready_then, which should be fine.

Probably another PR, but I'm leaving this comment to remember

bp_top/src/v/bp_unicore.sv
```diff
@@ -377,62 +387,68 @@ module bp_unicore
...
-         ,.ready_o(resp_fifo_ready_lo[i])
-
-         ,.data_o(proc_resp_li[i])
-         ,.v_o(proc_resp_v_li[i])
-         ,.yumi_i(proc_resp_yumi_lo[i])
-         );
     end
 
+    // resp fifo for I/O only
+    bp_bedrock_uce_mem_msg_s io_resp_lo;
```
dpetrisko, line 414: Why do we need this fifo, if the output interface is helpful?

bp_top/src/v/bp_unicore.sv
```diff
@@ -377,62 +387,68 @@ module bp_unicore
...
-   cmd_arbiter
-    (.ready_i(cmd_arb_ready_li)
-     ,.reqs_i(cmd_fifo_v_lo)
-     ,.grants_o(cmd_fifo_yumi_li)
-     );
+  wire cmd_arb_unlock = |(cmd_fifo_yumi_li & cmd_fifo_last_lo);
+  bsg_locking_arb_fixed
+   #(.inputs_p(3)
+   ,.lo_to_hi_p(0))
+   cmd_locking_arb_fixed_new
```
dpetrisko, line 442: just cmd_arb

dpetrisko (changes requested):
Looks really cool. Major comment is to rework burst2stream to remove buffering, some clarification and simplification requested in cce_to_cache

bp_me/src/v/wormhole/bp_burst_to_stream.sv
Songchun-Li: The sipo/piso are removed from these modules. Should I remove the one_fifos as well?

bp_me/src/v/cache/bp_me_cce_to_cache.sv
Songchun-Li: image
I found a issue in l1_hetero where icache_block_width == icache_fill_width == 128 and dcache_block_width == dcache_fill_width == 256. In this case, cache_data_width == 256, and the 128-bit icache_block can be either halves of the 256 bits data, which is identified by the mask (0000_ffff or ffff_0000)in cache_pkt.
And  the returned data should be shifted right when the requested data is at the upper half (ffff_0000)of the 256 bits. (Highlited with the red line in the screenshot)

bp_me/src/v/cache/bp_me_cce_to_cache.sv
Songchun-Li: The reason I added this logic is to handle the more general case of misaligned requests is that there are many misaligned write requests into memory with the size of 64 bits or 32 bits (less than the bus size when fill_width > 64), when the instruction is loaded at initialization.
The following screenshot shows an example whose cache_data_width == 256 bits and 64 bits instructions are loaded. The 1 bits (set_mask) are left shifted to the corresponding position to store instructions.  
image

bp_me/src/v/cache/bp_me_cce_to_cache.sv
dpetrisko: use bsg_bus_pack

bp_me/src/v/wormhole/bp_burst_to_stream.sv
dpetrisko: yep

bp_me/src/v/wormhole/bp_stream_to_burst.sv
```diff
@@ -0,0 +1,102 @@
...
+
+  // has_data: send the first stream header, with N stream data pkt
+  // ~has_data: send the only stream header only
+  wire has_data = payload_mask_p[in_msg_header_cast_i.msg_type];
+  
+  bp_bedrock_out_msg_header_s header_lo;
+  logic header_ready_lo, data_ready_lo;
+  assign in_msg_ready_and_o = streaming_r ? data_ready_lo : header_ready_lo;
+
+  bsg_one_fifo
```
dpetrisko, line 70: these need to be two fifos in order to prevent bubbles

bp_me/src/v/cache/bp_me_cce_to_cache.sv
```diff
@@ -211,9 +247,22 @@ module bp_me_cce_to_cache
             endcase
 
             cache_pkt.addr = is_tagfl ? tagfl_addr : mem_cmd_header_lo.addr;
-            cache_pkt.data = mem_cmd_data_lo;
+            // replicate slice data to work together with write mask
+            case(cmd_data_slice_cnt)
+              lg_cache_data_size_in_bytes_lp'(2): cache_pkt.data = {2{mem_cmd_data_lo[0+:(cache_data_width_p/2)]}};
```
dpetrisko, line 252: The data should already be replicated on the mem_cmd bus.  If not, use bsg_bus_pack in the UCE

bp_me/src/v/wormhole/bp_stream_to_burst.sv
```diff
@@ -0,0 +1,102 @@
...
+    ,.data_o(streaming_r)
+    );
+
+  // has_data: send the first stream header, with N stream data pkt
+  // ~has_data: send the only stream header only
+  wire has_data = payload_mask_p[in_msg_header_cast_i.msg_type];
+  
+  bp_bedrock_out_msg_header_s header_lo;
+  logic header_ready_lo, data_ready_lo;
+  assign in_msg_ready_and_o = streaming_r ? data_ready_lo : header_ready_lo;
```
dpetrisko, line 68: You can't accept a new header if data isn't ready. Need to &

bp_me/src/v/wormhole/bp_burst_to_stream.sv
```diff
@@ -0,0 +1,160 @@
...
+
+  assign out_msg_data_o = has_data ? data_lo : '0; // prevent meanless data from fifo.
+  assign out_msg_v_o = header_v_lo & (data_v_lo | ~has_data);
+
+  bp_bedrock_out_msg_header_s out_msg_header_cast_o;
+  // increment the address for bp_stream
+  logic is_last_cnt;
+  if (stream_words_lp == 1)
+   begin: full_block_stream
+      assign is_last_cnt = out_msg_v_o;
```
dpetrisko, line 102: Is this clause needed for functionality, or are you optimizing out the counters?

bp_me/src/v/cache/bp_me_cce_to_cache.sv
```diff
@@ -211,9 +247,22 @@ module bp_me_cce_to_cache
...
+              lg_cache_data_size_in_bytes_lp'(4): cache_pkt.data = {4{mem_cmd_data_lo[0+:(cache_data_width_p/4)]}};
+              lg_cache_data_size_in_bytes_lp'(8): cache_pkt.data = {8{mem_cmd_data_lo[0+:(cache_data_width_p/8)]}}; 
+              default:  cache_pkt.data = mem_cmd_data_lo;
+            endcase
             // This mask is only used for the LM/SM operations for >64 bit mask operations
-            cache_pkt.mask = '1;
+            case(mem_cmd_header_lo.size)
```
dpetrisko, line 258: You can just assign it, cache_pkt.mask is not checked for explicitly sized operations

bp_me/src/v/cce/bp_uce.sv
```diff
@@ -543,19 +537,18 @@ module bp_uce
...
+            cmd_header_lo.size     = block_msg_size_lp;
+            cmd_payload.lce_id     = lce_id_i;
+            cmd_header_lo.payload  = cmd_payload;
+            cmd_data_lo            = writeback_data;
+            cmd_v_lo = mem_cmd_ready_i & dirty_data_v_r & dirty_tag_v_r;
+
+            way_up = mem_cmd_done;
+            index_up = way_done & mem_cmd_done;
+
+            state_n = (mem_cmd_done & index_done & way_done)
```
dpetrisko, line 551: Is done and-ed with mem_cmd_v_o in the stream_pump module?

Songchun-Li: Yep

bp_me/src/v/wormhole/bp_burst_to_stream.sv
Songchun-Li: The counters are needed only when block_width > bus_width (stream_words_lp>1), so we need counters to generate  sub-block address and to track the progress to generate last_o signal. I think this if clause would optimize the area when the counters are not necessary

bp_me/src/v/cache/bp_me_cce_to_cache.sv
Songchun-Li: The data replication is actually for instruction data loading from the I/O port and have nothing to do with the UCE, so I used bsg_bus_pack for mem_cmd_data in the cce_to_cache

bp_me/src/v/cache/bp_me_cce_to_cache.sv
Songchun-Li: I understand that cache_pkt.mask is always 8'hff when cache_data_width == 64, so we can always assign it to '1. However, the question I have is, when cache_data_width > 64, for example, 128, the mask should be 16'hxxxx, right? For cache block load/store cmd from UCE, it is 16'hffff because it makes full use of the data width. However, when the instruction data is stored into L2 from the I/O initially, it got stored in size of 64 bits or 32 bits sometimes. In this case, the mask should be 16'h00ff or 16'hff00, and this is the reason I make it checking the size

bp_top/src/v/bp_unicore.sv
```diff
@@ -139,17 +139,9 @@ module bp_unicore
                           };
 
   bp_bedrock_uce_mem_msg_s cache_cmd_li;
-  bp_bedrock_cce_mem_msg_s cache_cmd;
   logic cache_cmd_v_li, cache_cmd_ready_lo, cache_cmd_last_li;
   bp_bedrock_uce_mem_msg_s cache_resp_lo;
-  bp_bedrock_cce_mem_msg_s cache_resp;
   logic cache_resp_v_lo, cache_resp_yumi_li, cache_resp_last_lo;
-  assign cache_cmd = '{header: cache_cmd_li.header
```
Songchun-Li, line 145: @dpetrisko This conversion leads to width mis-matches because the  cache_cmd/resp_data_width of l2_cache_slice and stream2burst are both uce_mem_data_width_lp, rather than cce_block_width_p. I think the conversion logic here should be removed, right?

dpetrisko: yep, can be removed

bp_me/src/v/wormhole/bp_burst_to_stream.sv
```diff
@@ -0,0 +1,160 @@
...
+    ,.data_i(in_msg_data_i)
+    ,.ready_o(in_msg_data_ready_and_o)
+    ,.v_i(in_msg_data_v_i & in_msg_data_ready_and_o)
+
+    ,.data_o(data_lo)
+    ,.v_o(data_v_lo)
+    ,.yumi_i(out_msg_ready_and_i & out_msg_v_o & has_data)
+    );
+
+  assign out_msg_data_o = has_data ? data_lo : '0; // prevent meanless data from fifo.
```
dpetrisko, line 94: don't gate data that already has a valid signal, it's just extra hardware

bp_me/src/v/wormhole/bp_stream_to_burst.sv
```diff
@@ -0,0 +1,102 @@
...
+
+  // has_data: send the first stream header, with N stream data pkt
+  // ~has_data: send the only stream header only
+  wire has_data = payload_mask_p[in_msg_header_cast_i.msg_type];
+  
+  bp_bedrock_out_msg_header_s header_lo;
+  logic header_ready_lo, data_ready_lo;
+  assign in_msg_ready_and_o = streaming_r ? data_ready_lo : (header_ready_lo & data_ready_lo);
+
+  bsg_two_fifo
```
dpetrisko, line 70: Similar to above, just header fifo

bp_me/src/v/wormhole/bp_burst_to_stream.sv
```diff
@@ -0,0 +1,160 @@
...
+  localparam in_data_bytes_lp = in_data_width_p/8;
+  localparam out_data_bytes_lp = out_data_width_p/8;
+  localparam stream_words_lp = block_width_p/out_data_width_p;
+  localparam data_len_width_lp = (stream_words_lp>0) ? `BSG_SAFE_CLOG2(stream_words_lp) : 0;
+  localparam stream_offset_width_lp = `BSG_SAFE_CLOG2(out_data_width_p / 8);
+
+  bp_bedrock_in_msg_header_s in_msg_header_cast_i;
+  assign in_msg_header_cast_i = in_msg_header_i;
+
+  bp_bedrock_out_msg_header_s out_header_lo;
```
dpetrisko, line 57: This can be reduced to the single header buffer and passing through the data signals

Songchun-Li:
The data fifo in burst_to_stream has been removed, but I think it may not be reasonable to remove the data fifo in stream_to_burst.

`in_msg_ready_and_o` of stream_to_burst is connected to the cmd arbiter, and it is demanded by the arbiter. Therefore, the `in_msg_header_i` and `in_msg_data_i` only holds for 1 cycle when they are granted. If there is a fifo for header and no fifo for data, when the header is valid on the burst side, the data signal is already changed. Therefore, I think a buffer for data is necessary to hold the data for extra cycles. Or should I add a parameter, `ready_then_valid_p`, and generate the dafa_fifo when `ready_then_valid_p == 1`?

And I added a alternative solution in [bp_stream_to_burst_nofifo](https://github.com/black-parrot/black-parrot/pull/612/commits/2e9a30e24dd20c4629e4aed56933d51fff25901e#diff-f9622fbb9c13b85d03554a6e38bc2de068fa6a799cf94c366f90fc8deae19956), where both the header fifo and data fifo are removed. The cons. is that both `out_msg_header_ready_and_i` and `out_msg_data_ready_and_i` become demanding.

bp_me/src/v/cache/bp_me_cce_to_cache.sv
```diff
@@ -154,15 +152,32 @@ module bp_me_cce_to_cache
...
     ,.fsm_ready_and_o(mem_resp_yumi_lo)
 
     ,.stream_cnt_o(mem_resp_cnt_lo)
     ,.stream_done_o(mem_resp_done_lo)
     );
   assign cache_yumi_o = mem_resp_yumi_lo | (is_clear & cache_v_i);
-  
+
+  if (icache_fill_width_p != dcache_fill_width_p)
+    begin: hetero_l1_fill_width
```
dpetrisko, line 165: Just always use bus_pack. I know it works in the case of homo l1, but as long as mem_resp_data_width > cache_data, we should use it.

(as an aside, I think bsg_bus_pack supports data_i > data_o width, but I'm not sure. If it doesn't it should be extremely quick to fix)

bp_me/src/v/cce/bp_uce.sv
```diff
@@ -75,13 +75,17 @@ module bp_uce
...
     , output logic                                   mem_cmd_v_o
     , input                                          mem_cmd_ready_i
 
-    , input [uce_mem_msg_width_lp-1:0]               mem_resp_i
+    , input [uce_mem_msg_header_width_lp-1:0]        mem_resp_header_i
+    , input [fill_width_p-1:0]                       mem_resp_data_i
+    , input                                          mem_resp_last_i
     , input                                          mem_resp_v_i
-    , output logic                                   mem_resp_yumi_o
+    , output logic                                   mem_resp_ready_o
```
dpetrisko, line 88: Make as many of these ready_and as possible, most of the yumis can be converted. I see there are some intermediate yumi signal names connected to ready_and_ports, which can be cleaned

dpetrisko (changes requested):
Minor changes requested, but mostly I just need to look closer at the arbitration logic in the unicore. I think it's right, but it's a little complex and I need wrap my head around it.

Removing buffering in the conversion modules is the real blocker here

bp_me/src/v/cache/bp_me_cce_to_cache.sv
Songchun-Li: The conditional generate statement here is removed.
 And I add some changes on the size_i signal. The reason is when `fill_width < block_width`, `size` in the resp header which is equal to block size, is larger than cache_data_width_p, then crash the mux output in the bsg_bus_pack.

bp_me/src/v/cce/bp_uce.sv
Songchun-Li: Shall I also replce the yumis with ready_and at the L2 sides? Or just the UCE side?

dpetrisko: L2 too

bp_top/src/v/bp_unicore.sv
```diff
@@ -449,31 +443,40 @@ module bp_unicore
...
-  bsg_arb_fixed
-   #(.inputs_p(5), .lo_to_hi_p(0))
-   resp_arbiter
-    (.ready_i(resp_arb_ready_li)
-     ,.reqs_i({loopback_resp_v_lo, cache_resp_v_lo, io_resp_v_i, clint_resp_v_lo, cfg_resp_v_lo})
-     ,.grants_o({loopback_resp_yumi_li, cache_resp_yumi_li, io_resp_yumi_o, clint_resp_yumi_li, cfg_resp_yumi_li})
-     );
+  logic resp_selected_last_lo;
+  wire resp_arb_ready_li = &proc_resp_ready_lo;
+  wire resp_arb_unlock = |({loopback_resp_yumi_li, cache_resp_ready_and_li, io_resp_yumi_o, clint_resp_yumi_li, cfg_resp_yumi_li} & {loopback_resp_v_lo, cache_resp_last_lo, io_resp_v_i, clint_resp_v_lo, cfg_resp_v_lo});
```
Songchun-Li, line 448: Shall I take care of the other yumi besides cache cmd/resp ?

dpetrisko: No, it's fine. Out of scope for this PR

bp_me/src/v/cache/bp_me_cache_slice_stream.sv
```diff
@@ -0,0 +1,158 @@
+
```
dpetrisko, line 1: Replace bp_me_cache_slice.sv with bp_me_cache_slice_stream.sv, no need to keep the obsolete version around

bp_me/src/v/cache/bp_me_cce_to_cache.sv
```diff
@@ -215,23 +223,23 @@ module bp_me_cce_to_cache
             // This mask is only used for the LM/SM operations for >64 bit mask operations
             cache_pkt.mask = '1;
             cache_pkt_v_o  = cache_pkt_ready_i & mem_cmd_v_lo;
-            // send yumi signal back to pump_out
-            mem_cmd_yumi_li = cache_pkt_v_o;
+            // send ready_and signal back to pump_out
+            mem_cmd_ready_and_li = cache_pkt_v_o;
```
dpetrisko, line 227: ready_and can't depend on a v_o signal, should be equal to cache_pkt_ready_i I think

bp_me/src/v/cache/bp_me_cce_to_cache.sv
```diff
@@ -215,23 +223,23 @@ module bp_me_cce_to_cache
...
 
             mem_resp_header_li = mem_cmd_header_lo; // return the same critical addr back to stream_fifo
             mem_resp_v_li      = cache_pkt_v_o;
 
-            cmd_state_n = (mem_cmd_stream_new_lo & mem_cmd_yumi_li) ? e_cmd_stream : e_cmd_ready;
+            cmd_state_n = (mem_cmd_stream_new_lo & mem_cmd_ready_and_li) ? e_cmd_stream : e_cmd_ready;
```
dpetrisko, line 232: ready_and_li & v_lo?

bp_me/src/v/wormhole/bp_stream_to_burst.sv
```diff
@@ -0,0 +1,84 @@
...
+    (.clk_i(clk_i)
+    ,.reset_i(reset_i)
+    ,.set_i(in_msg_v_i & in_msg_ready_and_o)
+    ,.clear_i(in_msg_last_i & in_msg_v_i & in_msg_ready_and_o)
+    ,.data_o(streaming_r)
+    );
+
+  assign in_msg_ready_and_o = streaming_r ? out_msg_data_ready_and_i : (out_msg_header_ready_and_i & out_msg_data_ready_and_i);
+
+  bsg_dff_en_bypass
```
dpetrisko, line 64: I think we only need to save the address here, not the entire header. Same with burst_to_stream

bp_me/src/v/cache/bp_me_cce_to_cache.sv
Songchun-Li: mem_cmd_stream_new_lo is ANDed from v_lo within the stream_pump_in. Shall I remove the v_lo from the new_lo signal in the stream_pump, or shall I add a duplicated v_lo here to make it 
`mem_cmd_stream_new_lo &  & mem_cmd_v_lo & mem_cmd_ready_and_l`

dpetrisko: oh let's remove it from stream pump, it's more standard to have them be independent

bp_me/src/v/wormhole/bp_stream_to_burst.sv
Songchun-Li: I made some changes in the conversion mdules with what I learnt in the other PRs. 

-  In the stream_to_burst, the header buffer is removed and both header and data are passed through, as only the msg_type is used and it remains the same in all headera within a stream transcation.
- In the burst_to_stream, the header 2 fifo is replaced with a bypass register, as we need to hold the header from burst (last for a cycle only), to keeping sending header with different address to the stream side. And this change saves a cycle for each transcation.

bp_me/src/v/cache/bp_me_cache_slice_stream.sv
Songchun-Li: The obsolete version is removed, and the stream version is adopted with lite2stream and stream2lite in bp_tile and bp_l2e_tile.

bp_me/src/v/cache/bp_me_cce_to_cache.sv
```diff
@@ -117,29 +113,85 @@ module bp_me_cce_to_cache
...
+        .data_i(cache_pkts_mask_mux_li)
+        ,.sel_i(cache_pkts_mask_sel_li)
+        ,.data_o(cache_pkt_mask_lo)
+      );
+    end
+  else 
+    begin:full_SM
+      assign cache_pkt_data_lo = mem_cmd_data_lo; 
+      assign cache_pkt_mask_lo = '1;    
+    end
```
Songchun-Li, line 171: This piece of code is added to handle data and mask of cache_pkt, when writeback data size is less than cache_data_width_p.

For example, if icache_fill_width == 256 and dcache_fill_width == 128, the cache_data_width == 256. When dcache writes back a dirty data block to L2, the 128 bits data can be either lower half of upper half of the 256 bits data in the cache_pkt. In current design, mask is set to be all 1s for SM op, so that the other half data will be overwritten mistakenly. Therefore, I added this logic to generate proper mask and data for this partial SM operation.

In fact, there is no cfg (icache_fill_width < dcache_fill_width ) now using this logic, and it is just an initiative for potential issue. I am not sure if it's worthwhile to support it.

dpetrisko:
```
  localparam bp_proc_param_s bp_unicore_l1_hetero_override_p =
    '{icache_sets         : 256
      ,icache_assoc       : 2
      ,icache_block_width : 128
      ,icache_fill_width  : 128
      ,dcache_sets        : 128
      ,dcache_assoc       : 4
      ,dcache_block_width : 256
      ,dcache_fill_width  : 256
      ,default : "inv"
      };
```

Does this config trigger these conditions?

Songchun-Li: Nope, because icache does not writeback. In this cfg, it reads all 256 bits data and `resp_data_bus_pack` is in charge of copying the upper half 128 bits data to the lower 128 bits half based on the address.

dpetrisko: We certainly should support it though

bp_me/src/v/cache/bp_me_cce_to_cache.sv
```diff
@@ -117,29 +113,85 @@ module bp_me_cce_to_cache
...
+       #(.width_p(cache_data_width_p)
+       ,.unit_width_p(dword_width_p))
+       cmd_data_bus_pack
+        (.data_i(mem_cmd_data_lo)
+        ,.sel_i('0)
+        ,.size_i(mem_cmd_size_li)
+        ,.data_o(cache_pkt_data_lo)
+        ); 
+
+      localparam lg_dword_mask_width_lp=`BSG_SAFE_CLOG2(cache_data_width_p/dword_width_p);
```
dpetrisko, line 130: I feel like a dword is a red herring, surely block_size_in_fill is more relevant?

bp_me/src/v/cache/bp_me_cce_to_cache.sv
```diff
@@ -206,32 +269,33 @@ module bp_me_cce_to_cache
...
-            cache_pkt.opcode = is_read ? LM : SM; // or LD : SD?
+            cache_pkt.opcode = is_read ? LM : SM;
             cache_pkt.addr = mem_cmd_stream_addr_lo;
             cache_pkt.mask = '1;
             cache_pkt.data = mem_cmd_data_lo;
             cache_pkt_v_o = cache_pkt_ready_i;
-            // send yumi signal back to pump_out
-            mem_cmd_yumi_li = cache_pkt_v_o;
+            // send ready_and signal back to pump_out
+            mem_cmd_ready_and_li = cache_pkt_v_o;
```
dpetrisko, line 298: This is weird. Shouldn't have a ready driven by a v_o

dpetrisko: Haven't looked too closely, but possibly:
- mem_cmd_ready_and_li = cache_pkt_ready_li
- cache_pkt_v_o = mem_cmd_ready_and_li & mem_cmd_v_lo

bp_me/src/v/wormhole/bp_stream_pump_in.sv
```diff
@@ -122,18 +122,34 @@ module bp_stream_pump_in
...
+      always_comb 
+        begin
           casez(num_block_in_msg_size)
-            data_len_width_lp'(1):  sub_block_adddr_tuned = sub_block_adddr;
-            data_len_width_lp'(2):  sub_block_adddr_tuned = { mem_header_lo.addr[(stream_offset_width_lp+data_len_width_lp-1)+:1], sub_block_adddr[0+:(stream_offset_width_lp+data_len_width_lp-1)]};
-            data_len_width_lp'(4):  sub_block_adddr_tuned = { mem_header_lo.addr[(stream_offset_width_lp+data_len_width_lp-2)+:2], sub_block_adddr[0+:(stream_offset_width_lp+data_len_width_lp-2)]};
-            default:                sub_block_adddr_tuned = mem_header_lo.addr[0+:(stream_offset_width_lp+data_len_width_lp)];
+            mask_len_width_lp'(1):  sel_mask = {data_len_width_lp{1'b1}};
+            mask_len_width_lp'(2):  sel_mask = {data_len_width_lp{1'b1}} >> 1;
+            mask_len_width_lp'(4):  sel_mask = {data_len_width_lp{1'b1}} >> 2;
```
dpetrisko, line 136: Should use a for loop so this is generalizable

bp_me/src/v/wormhole/bp_stream_pump_out.sv
```diff
@@ -86,13 +89,39 @@ module bp_stream_pump_out
...
       
-      assign is_stream = stream_mask_p[fsm_base_header_cast_i.msg_type] & ~(first_cnt == last_cnt);
-      assign stream_cnt_o = set_cnt ? first_cnt : current_cnt;
-      assign is_last_cnt = (stream_cnt_o == last_cnt) | ~is_stream;
+          is_stream = stream_mask_p[fsm_base_header_cast_i.msg_type] & ~(first_cnt == last_cnt);
+          stream_cnt_o = set_cnt ? first_cnt : current_cnt;
+          is_last_cnt = (stream_cnt_o == last_cnt) | ~is_stream;
+        end
+
+      // Generate proper wrap-around address for differenct incoming msg size dynamically,
```
dpetrisko, line 103: differenct -> different

bp_me/src/v/wormhole/bp_stream_pump_out.sv
```diff
@@ -86,13 +89,39 @@ module bp_stream_pump_out
...
-      assign stream_cnt_o = set_cnt ? first_cnt : current_cnt;
-      assign is_last_cnt = (stream_cnt_o == last_cnt) | ~is_stream;
+          is_stream = stream_mask_p[fsm_base_header_cast_i.msg_type] & ~(first_cnt == last_cnt);
+          stream_cnt_o = set_cnt ? first_cnt : current_cnt;
+          is_last_cnt = (stream_cnt_o == last_cnt) | ~is_stream;
+        end
+
+      // Generate proper wrap-around address for differenct incoming msg size dynamically, 
+      // if stream_data_width_p < incoming msg size < block_width_p, the higher bits of 
+      // stream_cnt are overriden depending on the size.
```
dpetrisko, line 105: I'm still very confused by this logic. I'm sure it's right, but can you add ASCII art of how this works in practice?

bp_me/src/v/wormhole/bp_stream_pump_out.sv
```diff
@@ -110,16 +139,10 @@ module bp_stream_pump_out
...
-          // Generate proper wrap-around address for differenct incoming msg size dynamically, 
-          // if stream_data_width_p < incoming msg size < block_width_p, the width of stream_cnt < data_len_width_lp
-          casez(num_block_in_msg_size)
-            data_len_width_lp'(1): sub_block_adddr_tuned = sub_block_adddr;
-            data_len_width_lp'(2): sub_block_adddr_tuned = { fsm_base_header_cast_i.addr[(stream_offset_width_lp+data_len_width_lp-1)+:1], sub_block_adddr[0+:(stream_offset_width_lp+data_len_width_lp-1)]};
-            data_len_width_lp'(4): sub_block_adddr_tuned = { fsm_base_header_cast_i.addr[(stream_offset_width_lp+data_len_width_lp-2)+:2], sub_block_adddr[0+:(stream_offset_width_lp+data_len_width_lp-2)]};
-            default:               sub_block_adddr_tuned = fsm_base_header_cast_i.addr[0+:(stream_offset_width_lp+data_len_width_lp)];
-          endcase
-          mem_header_cast_o.addr = { fsm_base_header_cast_i.addr[paddr_width_p-1:stream_offset_width_lp+data_len_width_lp], sub_block_adddr_tuned };
+          mem_header_cast_o.addr = { fsm_base_header_cast_i.addr[paddr_width_p-1:stream_offset_width_lp+data_len_width_lp]
```
dpetrisko, line 142: Nice! I like how much simpler this got

bp_me/src/v/wormhole/bp_stream_to_burst.sv
```diff
@@ -0,0 +1,77 @@
+
+module bp_stream_to_burst
+ import bp_common_pkg::*;
+ import bp_common_aviary_pkg::*;
+ import bp_me_pkg::*;
+ #(parameter bp_params_e bp_params_p = e_bp_default_cfg
+   `declare_bp_proc_params(bp_params_p)
+   // Assuming in_data_width_p(stream) == out_data_width_p(burst)
+   , parameter in_data_width_p  = "inv"
+   , parameter out_data_width_p = in_data_width_p
```
dpetrisko, line 10: Add an assertion that in_data_width_p == out_data_width_p

dpetrisko: same for burst_to_stream

bp_top/src/v/bp_unicore.sv
```diff
@@ -449,31 +443,40 @@ module bp_unicore
...
+  bsg_locking_arb_fixed
+   #(.inputs_p(5)
+   ,.lo_to_hi_p(0))
+   resp_locking_arb_fixed_new
```
dpetrisko, line 452: resp_locking_arb_fixed_new -> resp_arb

dpetrisko (changes requested):
Thanks! Looking really good. Clarifying questions and minor changes requested

bp_me/src/v/cache/bp_me_cache_slice.sv
```diff
@@ -13,115 +13,71 @@ module bp_me_cache_slice
  #(parameter bp_params_e bp_params_p = e_bp_default_cfg
```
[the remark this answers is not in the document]
Songchun-Li: The cache slice is removed, and I did not make much changes to the dma ports, not sure if I should simplify those, like the "unused" dma_mem_resp_header and its handshake

bp_me/src/v/cache/bp_me_cce_to_cache.sv
```diff
@@ -117,29 +113,85 @@ module bp_me_cce_to_cache
...
+          bsg_expand_bitmask 
+          #(.in_width_p(cache_data_width_p/slice_width_lp)
+          ,.expand_p(2**(i+byte_offset_width_lp))) 
+          exp 
+            (.i(decode_lo)
+            ,.o(cache_pkts_mask_mux_li[i])
+          );
+        end
+
+      wire [`BSG_SAFE_CLOG2(mask_sel_mux_els_lp)-1:0] cache_pkts_mask_sel_li = mem_cmd_header_lo.size - 2'h3;
```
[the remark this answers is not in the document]
Songchun-Li: This piece of logic is an variation of the logic in [bsg_cache](https://github.com/bespoke-silicon-group/basejump_stl/blob/3d85eefce4bfa815ab796f29b986e3c327e9b3dd/bsg_cache/bsg_cache.v#L684). The difference is that, in bsg_cache, the minimum size is byte and the maximum is dword, and here, the minimum here is dword.

Therefore, mem_cmd_header_lo.size gives size by 2^x bytes, however, the minimum unit here is dwords (8 bytes), therefore we have get the sizes by 2^(x-3) dwords  here to select corresponding mask.

In fact, I was thinking about adding this larger size opcode support to the bsg_cache, like LQ/SQ(quad_word) for 128 bits, but not sure if it is a good idea.

bp_me/src/v/cache/bp_me_cce_to_cache.sv
```diff
@@ -117,29 +113,85 @@ module bp_me_cce_to_cache
   wire is_csr   = (mem_cmd_header_lo.addr < dram_base_addr_gp);
   wire is_tagfl = is_csr & ((mem_cmd_header_lo.addr[0+:20] == cache_tagfl_base_addr_gp));
   wire [paddr_width_p-1:0] tagfl_addr = {mem_cmd_data_lo[0+:lg_sets_lp+lg_ways_lp], block_offset_width_lp'(0)};
+  
+  if ((cache_data_width_p > dword_width_p) && (dcache_fill_width_p < icache_fill_width_p))
+    begin:partial_SM
+      wire [`BSG_WIDTH(`BSG_SAFE_CLOG2(cache_data_width_p/dword_width_p))-1:0] mem_cmd_size_li = `BSG_MIN(mem_cmd_header_lo.size - 3'h3, `BSG_SAFE_CLOG2(cache_data_width_p/dword_width_p));
```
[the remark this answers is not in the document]
Songchun-Li: Yes, the data is always in lower bits, but it is replicated here because sometimes the data should be at higher bits for memory store. For example, when dcache_fill_witdh == 128 and cache_data_width == 256, the store data can be
```
data: 128 bits data, 128'h0
mask: 16'hffff, 16'h0
```
or
```
data: 128'h0, 128 bits data
mask: 16'h0, 16'hffff
```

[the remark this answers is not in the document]
Songchun-Li: Similarly, by adding ops LQ/SQ(quad_word) into bsg_cache makes the codes here look much cleaner.

bp_me/src/v/cache/bp_me_cce_to_cache.sv
Songchun-Li: Indeed, using fill_width is more relevant, because using the minimum fill_width as the base unit in this loop-generated logic also save the area and power when the minimum fill width > dword, but it may be more difficult for coding.

bp_me/src/v/wormhole/bp_stream_pump_out.sv
Songchun-Li: Here is simplified a lot with comments added!

bp_top/src/v/bp_l2e_tile.sv
```diff
@@ -328,36 +334,85 @@ module bp_l2e_tile
...
+        ,.out_msg_o(cache_mem_resp_lo)
+        ,.out_msg_v_o(cache_mem_resp_v_lo)
+        ,.out_msg_ready_and_i(cache_mem_resp_yumi_li)
+        );
 
-         ,.mem_resp_data_i(dma_mem_resp_data_li)
-         ,.mem_resp_data_v_i(dma_mem_resp_data_v_li)
-         ,.mem_resp_data_ready_o(dma_mem_resp_data_ready_lo)
-         );
+      bp_me_cache_slice
```
dpetrisko: Remove cache slice

bp_me/src/v/cache/bp_me_cache_slice.sv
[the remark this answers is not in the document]
dpetrisko: delete the file so it's not accidentally in use (like in l2e tile)

bp_me/src/v/cache/bp_me_cce_to_cache.sv
dpetrisko: I think using SM for all stores will simplify the logic with very little / no overhead

bp_top/src/v/bp_l2e_tile.sv
Songchun-Li: It's been removed already, you should be able to see it with all the commits

bp_me/src/v/cache/bp_me_cache_slice.sv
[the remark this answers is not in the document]
Songchun-Li: It's been removed already, you should be able to see it with all the commits. And I also tested the multicore_1/2_l2e_cfg

Songchun-Li:
The mentioned issues are all resolved and the partial sm logic is also simplifed. 

And to test this logic, I wrote a new test, cache_dword_hammer [here](https://github.com/Songchun-Li/bp_tests/blob/dev_cache_dword_hammer/src/cache_dword_hammer.c), which load/store on the lowest dword of every cache block in the D-cache.  

I also test it with a new l1_hetero cfg as following, so that the dcache fill_width is less than the l2 data width and the condition statement is triggered.
```
icache_sets         : 128
,icache_assoc       : 4
,icache_block_width : 256
,icache_fill_width  : 256
,dcache_sets        : 256
,dcache_assoc       : 2
,dcache_block_width : 128
,dcache_fill_width  : 128
```

Above 2 features are just used in testing and not added to this PR yet. Let me know if I should add any of them.

bp_me/src/v/cache/bp_me_cce_to_cache.sv
```diff
@@ -118,28 +115,100 @@ module bp_me_cce_to_cache
   wire is_tagfl = is_csr & ((mem_cmd_header_lo.addr[0+:20] == cache_tagfl_base_addr_gp));
   wire [paddr_width_p-1:0] tagfl_addr = {mem_cmd_data_lo[0+:lg_sets_lp+lg_ways_lp], block_offset_width_lp'(0)};
 
+  if ((min_fill_width_lp > dword_width_p) && (min_fill_width_lp < cache_data_width_p))
```
dpetrisko, line 118: Just use SM for all store operations

bp_me/src/v/cache/bp_me_cce_to_cache.sv
```diff
@@ -118,28 +115,100 @@ module bp_me_cce_to_cache
...
+        begin:cache_pkt_sel
+          localparam slice_width_lp = (min_fill_width_lp*(2**i));
+
+          // Data
+          logic [slice_width_lp-1:0] slice_data;
+          assign slice_data = mem_cmd_data_lo[0+:slice_width_lp];
+          assign cache_pkt_data_mux_li[i] = {(cache_data_width_p/slice_width_lp){slice_data}};
+
+          // Mask
+          if (i == data_sel_mux_els_lp-1)
```
dpetrisko, line 139: The else here should degenerate to '1, no?

bp_me/src/v/cache/bp_me_cce_to_cache.sv
```diff
@@ -151,18 +220,29 @@ module bp_me_cce_to_cache
...
+    ,.stream_cnt_o(/* unused */)
     ,.stream_done_o(mem_resp_done_lo)
     );
-  assign cache_yumi_o = mem_resp_yumi_lo | (is_clear & cache_v_i);
-  
+  assign cache_yumi_o = mem_resp_ready_and_lo | (is_clear & cache_v_i);
+
+  wire [`BSG_WIDTH(`BSG_SAFE_CLOG2(cache_data_width_p>>3))-1:0] mem_resp_size_li = `BSG_MIN(mem_resp_header_lo.size, `BSG_SAFE_CLOG2(cache_data_width_p/8));
+  bsg_bus_pack
+   #(.width_p(cache_data_width_p)
```
dpetrisko, line 237: This should always be the mem_resp data width to comply with the interface. You could either pad the cache data with 0 or modify bus_pack to have input_width_p and output_width_p parameters

bp_me/src/v/cache/bp_me_cce_to_cache.sv
```diff
@@ -151,18 +220,29 @@ module bp_me_cce_to_cache
...
+   ,.unit_width_p(8))
```
dpetrisko, line 238: omit default setting

dpetrisko:
@Songchun-Li Minor fixes.  Need to rebase on me_dev for cache_dma changes.  Otherwise I think we can start merging this in. 

There are going to be some conflicts with this other arbitration-related PR: https://github.com/black-parrot/black-parrot/pull/805/files, so we'll have to think about how this affects the stream pump in particular

bp_me/src/v/cache/bp_me_cce_to_cache.sv
Songchun-Li:
```
localparam lg_cache_size_in_fill_lp =`BSG_SAFE_CLOG2(cache_size_in_fill_lp);
localparam data_sel_mux_els_lp = lg_cache_size_in_fill_lp+1;
``` 
The main issue is when cache_size_in_fill_lp == 1, data_sel_mux_els_lp == 2 because of th bsg safe clog2. By replacing it with clog2, the special case here is reomved.

Songchun-Li:
For new, it passes weird config regression with multi-core cfg with the lite2stream conversion. Working on some port-width mismatches issus. 

Besides,  config like l1_wide and l1_tiny has minimum fill_width (bank width) larger than 64 bits, but l2_data_width is restricted to 64 bits in the if_verif, how should I handle this issue? @dpetrisko

Songchun-Li:
Hi @dpetrisko! The commits are rebased to the latest me_dev. All the issues mentioned or found are resolved. It passes regress tests including icache, dcache, me, weird cfg, and all different unicore cfgs running riscv_testslist with different fill width varying from minimum fill width to block width. It's now ready for final review. Lmk if you have any suggestion on bp_common_aviary_pkgdef.svh.

bp_top/syn/Makefile
```diff
@@ -12,7 +12,7 @@ export REPORT_PATH  := $(BP_TOP_DIR)/syn/reports
 export TOUCH_PATH   := $(BP_TOP_DIR)/syn/touchfiles
 
 export TB    ?= bp_tethered
-export CFG   ?= e_bp_default_cfg
+export CFG   ?= e_bp_unicore_cfg
```
dpetrisko, line 15: keep as default

bp_top/test/common/bp_nonsynth_if_verif.sv
```diff
@@ -74,11 +74,13 @@ module bp_nonsynth_if_verif
     $fatal("Error: Cache fill width should be less or equal to L1 cache block width");
   if ((icache_fill_width_p % (icache_block_width_p/icache_assoc_p) != 0) || (dcache_fill_width_p % (dcache_block_width_p / dcache_assoc_p) != 0))
     $fatal("Error: Cache fill width should be a multiple of cache bank width");
+  // TODO: Combine icache & dcache data width into l1 fill width
+  // The cce_to_cache supports hetero fill width only when fill width == block width 
+  // if (icache_fill_width_p != dcache_fill_width_p)
```
dpetrisko, line 79: Keep assertion and remove comment

dpetrisko:
Thanks! Okay I'm going to merge this into the stream pump branch and start bringing things over to mainline. I'll let you know if there are any changes needed once I'm looking at the true diff

PR #615, opened 2020-09-17
Disable co-simulation checks in debug mode

sripathi-muralitharan:
This PR adds code to disable co-simulation checking during debug mode. This will be helpful for the next version of the bootrom which will set up the configuration as well copy the program to the memory. 

In the future, an optional checking of the boot flow will also be provided. 

Note: A specific commit in Dromajo still needs to be reverted (https://github.com/bsg-external/dromajo/commit/4485477adcfb7f16a62dfb66b95ec1d62d8c1a32) but since I do not have access to the repository, I am unable to make this change.

### DIFF
--- a/bp_top/test/common/dromajo_cosim.cpp
+++ b/bp_top/test/common/dromajo_cosim.cpp
@@ -21,7 +21,6 @@ extern "C" void dromajo_init(char* cfg_f_name, int hartid, int ncpus, int memory
     string memsize_str = "--memory_size=" + to_string(memory_size);
     string mmio_str = "--mmio_range=0x20000:0x80000000";
     char* load_str = "--load=prog";
-    char* bootrom_str = "--bootrom=bootrom.bin";
 
     if(checkpoint) {
       char* argv[] = {"dromajo", (char*)(&ncpus_str[0]), (char*)(&memsize_str[0]), (char*)(&mmio_str[0]), load_str, "prog.elf"};
--- a/bp_top/test/common/dromajo_cosim.cpp
+++ b/bp_top/test/common/dromajo_cosim.cpp
@@ -28,8 +28,8 @@ extern "C" void dromajo_init(char* cfg_f_name, int hartid, int ncpus, int memory
       dromajo_pointer = dromajo_cosim_init(6, argv);
     }
     else {
-      char* argv[] = {"dromajo", (char*)(&ncpus_str[0]), (char*)(&memsize_str[0]), (char*)(&mmio_str[0]), bootrom_str, "prog.elf"};
-      dromajo_pointer = dromajo_cosim_init(6, argv);
+      char* argv[] = {"dromajo", (char*)(&ncpus_str[0]), (char*)(&memsize_str[0]), (char*)(&mmio_str[0]), "prog.elf"};
+      dromajo_pointer = dromajo_cosim_init(5, argv);
     }
   }
 }
@@ -45,7 +45,7 @@ extern "C" bool dromajo_step(int      hartid,
                                      wdata,
                                      0,
                                      true,
-                                     false);
+                                     true);
   if(exit_code != 0)
     return true;
   else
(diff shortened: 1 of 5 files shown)


farzamgl:
Looks good. I think you also have to disable the debug checks in the Dromajo submodule.

dpetrisko:
> I am unable to make this change.

You can make a PR from your own fork

bp_top/test/common/dromajo_cosim.cpp
```diff
@@ -28,8 +28,8 @@ extern "C" void dromajo_init(char* cfg_f_name, int hartid, int ncpus, int memory
       dromajo_pointer = dromajo_cosim_init(6, argv);
     }
     else {
-      char* argv[] = {"dromajo", (char*)(&ncpus_str[0]), (char*)(&memsize_str[0]), (char*)(&mmio_str[0]), bootrom_str, "prog.elf"};
```
dpetrisko, line 31: I think this breaks checkpointing? @farzamgl ?

dpetrisko (changes requested):

sripathi-muralitharan:
> Looks good. I think you also have to disable the debug checks in the Dromajo submodule.

Submitted a PR for it in the Dromajo repository.

bp_top/test/common/dromajo_cosim.cpp
farzamgl: No, the checkpoint option is in the if clause.

dpetrisko: Ah I see.  So dromajo is just using the default bootrom in all cases then?

@sripathi-muralitharan delete the bootrom_str variable in this file and the bootrom.bin copying in the makefile flow, if it's unused.

sripathi-muralitharan: Well shouldn't we keep it because we are going to add a BOOTCHECK flag later on right?

farzamgl: It's using the *.boorom generated file in the checkpointing case. So yes we don't need the --bootrom option.

dpetrisko: > It's using the *.bootrom generated file in the checkpointing case. So yes we don't need the --bootrom option.

Got it.

BOOTCHECK is now (this PR) or never.  I lean towards never, since any errors in bootrom will be pretty easily detectable.  We could also add a test where we generate a bootrom with arbitrary data, and then the program is literally "add x1, x0; add x2, x0... csrr mhartid, csrr mcause; ...."?  The weirdness would be setting a non-default bootrom in the flow, not sure if it's worth the headache

farzamgl: Yes, let's just focus on the new boot flow. It's probably difficult maintaining both of them.

sripathi-muralitharan: >@sripathi-muralitharan delete the bootrom_str variable in this file and the bootrom.bin copying in the makefile flow, if it's unused.

Did not remove the bootrom.bin since its used to create the bootrom.mem required to initialize the nonsynth_rom (bootrom) in the host.

dpetrisko: You can go directly from riscv to mem as seen here: https://github.com/black-parrot/black-parrot/blob/master/bp_common/test/Makefile.tests#L82.  We only went from the binary because it was already available

sripathi-muralitharan: Oh alright. Added another commit with this change. Now bootrom.bin files are not generated as part of the flow.

dpetrisko (approved):
LGTM!

bp_top/test/tb/bp_tethered/testbench.v
```diff
@@ -245,6 +245,7 @@ bind bp_be_top
                ? calculator.pipe_sys.csr.scause_li
                : calculator.pipe_sys.csr.mcause_li
                )
+     ,.is_debug_mode_i(calculator.pipe_sys.csr.is_debug_mode)
```
dpetrisko, line 248: minor thing, can you fix the indentation here?

dpetrisko:
I'm going to merge this to sw_dev rather than top_dev, since the rest of the changes will be more related to software

PR #616, opened 2020-09-18
Removing instr_v field from BE

dpetrisko:
Minor fix to remove the "instr_v" field from various structures.  There were some weird edge cases differentiating between v and instr_v, now they are all combined into v.  (tested on linux boot)

### DIFF
--- a/bp_be/src/include/bp_be_ctl_defines.vh
+++ b/bp_be/src/include/bp_be_ctl_defines.vh
@@ -193,7 +193,6 @@ typedef enum logic
 typedef struct packed
 {
   logic                             v;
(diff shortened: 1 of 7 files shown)


farzamgl:
The edge cases were FE exceptions, right? How do we identify them now?

dpetrisko:
Yeah, exactly.  The trick we used was not incrementing NPC on itlb fills, so that when we did the fill we would still be at the same address.  Now, we increment NPC and just reset it to the fill address on fe_cmd, just like a normal trap

farzamgl (approved):

PR #619, opened 2020-09-20
Adding burst interface to unicore memory

dpetrisko:
Adds a burst interface for the L2 cache in unicore.  This has two main advantages:
- Reuse same L2 slice with unicore and multicore
- Pave the way for additional improvements e.g. stream on the input side of the cache slice

Depends on #618


bp_top/src/v/bp_unicore.v
```diff
@@ -485,25 +492,34 @@ module bp_unicore
...
+      assign mem_resp_data_yumi_o = mem_resp_data_ready_lo & mem_resp_data_v_i;
     end
   else
     begin : no_l2
-      assign mem_cmd_cast_o = cce_mem_msg_width_lp'(cache_cmd_li);
-      assign mem_cmd_v_o = cache_cmd_v_li;
-      assign cache_cmd_ready_lo = mem_cmd_ready_i;
+      //assign mem_cmd_cast_o = cce_mem_msg_width_lp'(cache_cmd_li);
+      //assign mem_cmd_v_o = cache_cmd_v_li;
+      //assign cache_cmd_ready_lo = mem_cmd_ready_i;
```
muwyse, line 518: remove comment?

bp_top/src/v/bp_unicore.v
```diff
@@ -485,25 +492,34 @@ module bp_unicore
...
 
-      assign cache_resp_lo = mem_resp_cast_i[0+:uce_mem_msg_width_lp];
-      assign cache_resp_v_lo = mem_resp_v_i;
-      assign mem_resp_yumi_o = cache_resp_yumi_li;
+      //assign cache_resp_lo = mem_resp_cast_i[0+:uce_mem_msg_width_lp];
+      //assign cache_resp_v_lo = mem_resp_v_i;
+      //assign mem_resp_yumi_o = cache_resp_yumi_li;
```
muwyse, line 522: remove comment?

bp_me/src/v/wormhole/bp_lite_to_burst.v
```diff
@@ -0,0 +1,107 @@
...
+  `declare_bp_mem_if(paddr_width_p, cce_block_width_p, lce_id_width_p, lce_assoc_p, out_mem);
+  bp_in_mem_msg_s mem_cast_i;
+  assign mem_cast_i = mem_i;
+
+  localparam in_data_bytes_lp = in_data_width_p/8;
+  localparam out_data_bytes_lp = out_data_width_p/8;
+  localparam burst_words_lp = in_data_width_p/out_data_width_p;
+  localparam burst_offset_width_lp = `BSG_SAFE_CLOG2(out_data_bytes_lp);
+
+  // We could make this a two fifo to get more throughput
```
muwyse, line 45: Any idea what throughput improvement would be for a two-fifo and the associated cost in power or area? Should we make this tradeoff? Should this tradeoff be parameterized?

bp_me/src/v/wormhole/bp_lite_to_burst.v
```diff
@@ -0,0 +1,107 @@
...
+ import bp_common_pkg::*;
+ import bp_common_aviary_pkg::*;
+ import bp_me_pkg::*;
+ #(parameter bp_params_e bp_params_p = e_bp_default_cfg
+   `declare_bp_proc_params(bp_params_p)
+
+   , parameter in_data_width_p  = "inv"
+   , parameter out_data_width_p = "inv"
+
+   , parameter logic master_p = 0
```
muwyse, line 12: What is the behavior difference for this module when master versus client?

bp_me/src/v/wormhole/bp_lite_to_burst.v
```diff
@@ -0,0 +1,107 @@
...
+     ,.data_o(mem_header_o)
+     ,.v_o(mem_header_v_o)
+     ,.yumi_i(mem_header_yumi_i)
+     );
+
+  wire is_wr = mem_cast_i.header.msg_type inside {e_mem_msg_uc_wr, e_mem_msg_wr};
+  localparam data_len_width_lp = `BSG_SAFE_CLOG2(burst_words_lp);
+  wire [data_len_width_lp-1:0] num_burst_cmds = (master_p ^ is_wr)
+    ? 1'b1
+    : `BSG_MAX(((1'b1 << mem_cast_i.header.size) / out_data_bytes_lp), 1'b1);
```
muwyse, line 66: Same as question above about master versus client behavior. Maybe add a comment explaining this. I think this has to do with whether the mem_i is mem_cmd or mem_resp?

muwyse (approved):
Code LGTM. A couple questions for clarification.

Also, this PR is effectively a superset of #618, correct?

bp_me/src/v/wormhole/bp_lite_to_burst.v
dpetrisko: So it's actually a functionality thing right now. The way this module works is by setting the burst counters when something goes _into_ the fifo, then bursting out of the fifo.  The downside is that there is a bubble between consecutive requests.  I think v2 could solve this problem (though I'm not sure if it's worth it; consecutive bursts are probably? rare)

bp_me/src/v/wormhole/bp_lite_to_burst.v
dpetrisko: master = cmd_o, resp_i
client = cmd_i, resp_o

If a master operation is a read, then the command will be 1 beat, response will be N beats.  If a client operation is a read, then the response will be N beats.

I think rather than master, "forward_p" might be clearer? I'll write this in a comment to explain

bp_me/src/v/wormhole/bp_lite_to_burst.v
dpetrisko: yep, exactly.  Explained above, will add comment

bp_top/src/v/bp_unicore.v
dpetrisko: whoops! Forgot to add in the no_l2 configuration

bp_me/src/v/wormhole/bp_burst_to_lite.v
```diff
@@ -0,0 +1,115 @@
...
+   , output logic                            mem_header_ready_o
+
+   , input [in_data_width_p-1:0]             mem_data_i
+   , input                                   mem_data_v_i
+   , output logic                            mem_data_ready_o
+
+   // Client BP Lite
+   , output logic [out_mem_msg_width_lp-1:0] mem_o
+   , output logic                            mem_v_o
+   , input                                   mem_yumi_i
```
muwyse, line 32: Shouldn't this be ready_i since handshake for BP Lite is valid&ready?

bp_me/src/v/wormhole/bp_lite_to_burst.v
```diff
@@ -0,0 +1,107 @@
...
+
+   // Master BP Lite
+   , input [in_mem_msg_width_lp-1:0]                mem_i
+   , input                                          mem_v_i
+   , output logic                                   mem_ready_o
+
+   // Client BP Burst
+   , output logic [out_mem_msg_header_width_lp-1:0] mem_header_o
+   , output logic                                   mem_header_v_o
+   , input logic                                    mem_header_yumi_i
```
muwyse, line 28: same as above. Protocols are all ready&valid, but this module introduces yumi signals

bp_me/src/v/wormhole/bp_lite_to_burst.v
```diff
@@ -0,0 +1,107 @@
...
+
+   , output logic [out_data_width_p-1:0]            mem_data_o
+   , output logic                                   mem_data_v_o
+   , input                                          mem_data_yumi_i
```
muwyse, line 32: same as above

bp_me/src/v/wormhole/bp_lite_to_stream.v
```diff
@@ -0,0 +1,138 @@
...
+   // Master BP Lite
+   , input [in_mem_msg_width_lp-1:0]                mem_i
+   , input                                          mem_v_i
+   , output logic                                   mem_ready_o
+
+   // Client BP Stream
+   , output logic [out_mem_msg_header_width_lp-1:0] mem_header_o
+   , output logic [out_data_width_p-1:0]            mem_data_o
+   , output logic                                   mem_v_o
+   , input                                          mem_yumi_i
```
muwyse, line 29: Same as above. Why yumi and not ready?

bp_me/src/v/wormhole/bp_stream_to_lite.v
```diff
@@ -0,0 +1,114 @@
...
+   , input [in_mem_msg_header_width_lp-1:0]  mem_header_i
+   , input [in_data_width_p-1:0]             mem_data_i
+   , input                                   mem_v_i
+   , output logic                            mem_ready_o
+   , input                                   mem_lock_i
+
+   // Client BP Lite
+   , output logic [out_mem_msg_width_lp-1:0] mem_o
+   , output logic                            mem_v_o
+   , input                                   mem_yumi_i
```
muwyse, line 30: same as above

bp_me/src/v/wormhole/bp_burst_to_lite.v
dpetrisko: Yep, you're totally right! I'll change to ready/valid & since it's an interconnect component

dpetrisko: Fixed

PR #621, opened 2020-09-22
Bootrom as Config Loader

sripathi-muralitharan:
This PR adds enhancements to the bootrom. 

The bootrom now loads the configuration (I$, D$ and CCE modes, domains and sac) and the CCE ucode in-place of the NBF script which was the earlier methodology. 

The NBF script still loads the program and sets the boot PC.


dpetrisko:
>sets the boot PC.

You can get rid of this.  The config bus no longer sets the boot pc

bp_common/software/py/ucode2boot.py
```diff
@@ -0,0 +1,21 @@
+import argparse
```
dpetrisko, line 1: what's this file doing?

bp_common/test/src/bootrom/bootrom.S
```diff
@@ -1,3 +1,21 @@
...
+#define CCE_INSTR_END 0xFFFFFFFFFFFFFFFF
+
+// Configuration parameters
+#define CFG_CCE_UCODE_BASE_ADDR 0x00208000
+
+#define CFG_CCE_MODE_ADDR 0x00200081
+#define CFG_DCACHE_MODE_ADDR 0x00200043
+#define CFG_ICACHE_MODE_ADDR 0x00200022
+
+#define CFG_DOMAIN_MASK_ADDR 0x00200009
```
dpetrisko, line 17: Don't bother setting domain or sac masks here.  These are extremely rare conditions to actually change

Makefile
```diff
@@ -72,7 +72,7 @@ tools: | $(TARGET_DIRS)
 
 progs: tools
 	git submodule update --init --recursive $(BP_COMMON_DIR)/test
-	$(MAKE) -C $(BP_COMMON_DIR)/test perch bp_tests riscv_tests beebs coremark bootrom
```
dpetrisko, line 75: Hm, we'll still want to provide a default bootrom to users.  Would it be possible to have a default bootrom that we relink with different ucode as part of the simulation flow?

As an aside, where does the bootrom get built in this alternate flow?

bp_top/test/tb/bp_tethered/Makefile.vcs
```diff
@@ -47,6 +47,8 @@ $(SIM_DIR)/bootrom.mem: $(SIM_DIR)/bootrom.riscv
 	$(RISCV_OBJCOPY) -O verilog --reverse-bytes=8 --verilog-data-width=8 $< $@
 
 NBF_INPUTS ?= --ncpus=$(NCPUS) --ucode=cce_ucode.mem
+NBF_INPUTS += --domains=$(DOMAINS_P)
```
dpetrisko, line 50: same comment as above, probably unnecessary

bp_top/test/tb/bp_tethered/Makefile.verilator
```diff
@@ -41,6 +41,8 @@ $(SIM_DIR)/cce_ucode.mem: $(CCE_MEM_PATH)/$(CCE_MEM)
 	@cp $^ $@
 
 NBF_INPUTS = --ncpus=$(NCPUS) --ucode=cce_ucode.mem
+NBF_INPUTS += --domains=$(DOMAINS_P)
```
dpetrisko, line 44: same comment as above, probably unnecessary

bp_common/software/py/ucode2boot.py
sripathi-muralitharan: Currently, the CCE ucode is 34 bits. Adding the .bin version of the ucode file gets incorrectly interpreted by the compiler. So, this script expands the ucode instructions to 64 bits and then converts them to hex.

bp_common/test/src/bootrom/bootrom.S
sripathi-muralitharan: So, whats the alternative here? Leave them in the NBF script? Put them as part of the default_cfg?

Makefile
sripathi-muralitharan: In this flow, the bootrom gets generated whenever the user executes a program in simulation. 

We could have a default bootrom but then the way things are set up right now, a .riscv file is not re-built if it already exists in the mem folder. 

As an aside, based on earlier comments, if we are finally deciding to get rid of the domain and sac parameters from the bootrom, then maybe the bootrom can be put back into this command. The only reason it was removed was because of the -D define options being passed in during compile time (for domain and sac) to the bootrom and the fact that the CCE ucode binary was also being linked in.

bp_top/test/tb/bp_tethered/Makefile.vcs
sripathi-muralitharan: Requested for clarification above.

bp_top/test/tb/bp_tethered/Makefile.verilator
sripathi-muralitharan: Requested for clarification above.

bp_common/software/py/ucode2boot.py
dpetrisko: Ah gotcha, maybe add a comment for that

bp_common/test/src/bootrom/bootrom.S
dpetrisko: Don't touch them at all.  User software can deal with them if they need to

Makefile
dpetrisko: We need a default bootrom so that when people clone the repo, they can just run the examples without recompiling.  Recompiling the bootrom requires the full riscv toolchain.

>As an aside, based on earlier comments, if we are finally deciding to get rid of the domain and sac parameters from the bootrom, then maybe the bootrom can be put back into this command. The only reason it was removed was because of the -D define options being passed in during compile time (for domain and sac) to the bootrom and the fact that the CCE ucode binary was also being linked in.

Cool, that works

sripathi-muralitharan:
I am working on a bug that was manifested for the multi-core version of BP with the CCE using the ucode. I am fixing it now. Please do not merge until the bug is fixed.

dpetrisko:
Looks awesome! Some minor things to touch up once the bug is fixed, but nothing big

Makefile
```diff
@@ -18,6 +18,7 @@ prep_lite: | $(TARGET_DIRS)
 	$(MAKE) libs
```
dpetrisko, line 18: I think you can revert all these changes?  bootrom gets built as part of programs, no reason to add a new target for it

bp_common/software/py/nbf.py
```diff
@@ -162,16 +162,16 @@ def dump(self):
...
     # Write I$, D$, and CCE modes
-    self.print_nbf_allcores(3, cfg_base_addr + cfg_reg_icache_mode, 1)
-    self.print_nbf_allcores(3, cfg_base_addr + cfg_reg_dcache_mode, 1)
-    self.print_nbf_allcores(3, cfg_base_addr + cfg_reg_cce_mode, 1)
+    # self.print_nbf_allcores(3, cfg_base_addr + cfg_reg_icache_mode, 1)
+    # self.print_nbf_allcores(3, cfg_base_addr + cfg_reg_dcache_mode, 1)
+    # self.print_nbf_allcores(3, cfg_base_addr + cfg_reg_cce_mode, 1)
 
     # Write PC to the DRAM base
     self.print_nbf_allcores(3, cfg_base_addr + cfg_reg_npc, 0x103000)
```
dpetrisko, line 177: You can comment this as well

bp_common/software/py/nbf.py
```diff
@@ -162,16 +162,16 @@ def dump(self):
     self.print_nbf_allcores(3, cfg_base_addr + cfg_reg_reset, 0)
 
     # Write CCE ucode
```
dpetrisko, line 164: Add a comment saying that this is now handled by bootrom

bp_common/test/mem/bootrom.dump
```diff
@@ -1,56 +0,0 @@
-
```
dpetrisko, line 1: We want to re-add a default bootrom

bp_common/test/mem/bootrom.mem
```diff
@@ -1,11 +0,0 @@
-@00000000
```
dpetrisko, line 1: Same here

bp_common/test/src/bootrom/Makefile
```diff
@@ -1,8 +1,26 @@
 
-RISCV_GCC = $(CROSS_COMPILE)gcc -fPIC -march=rv64im -mabi=lp64 -mcmodel=medany -static -nostdlib -nostartfiles
```
dpetrisko, line 2: was fPIC causing problems?

bp_common/test/src/bootrom/Makefile
```diff
@@ -1,8 +1,26 @@
...
+RISCV_GCC = $(CROSS_COMPILE)gcc -march=rv64im -mabi=lp64 -mcmodel=medany -static -nostdlib -nostartfiles
 
-build:
-	$(RISCV_GCC) bootrom.S -o bootrom.riscv -Tlink.ld -static -Wl,--no-gc-sections
+UCODE2BOOT = $(BP_COMMON_DIR)/software/py/ucode2boot.py
+
+ifdef CCE_MEM_PATH
+BOOT_INPUTS = --ucode=$(CCE_MEM_PATH)/$(CCE_MEM)
```
dpetrisko, line 7: What's this used for?

bp_common/test/src/bootrom/bootrom.S
```diff
@@ -35,13 +50,43 @@ _start:
...
+load_config:
+    li t0, 1
+    li t1, CFG_DCACHE_MODE_ADDR
+    sd t0, 0(t1)
+    
+    li t1, CFG_ICACHE_MODE_ADDR
+    sd t0, 0(t1)
+
+    li t1, CFG_CCE_MODE_ADDR
+    sd t0, 0(t1)
```
dpetrisko, line 76: Should have a fence here, I think

dpetrisko (changes requested):

Makefile
sripathi-muralitharan: Well, I put it after the ucode because the default bootrom should have a ucode linked in right? So once the ucode is generated, I added in the bootrom generation.

bp_common/test/src/bootrom/Makefile
sripathi-muralitharan: Yup, it was creating position independent code and because of that the GOT was being generated. I got rid of this option and put everything in one section for a flat compilation.

bp_common/test/src/bootrom/Makefile
sripathi-muralitharan: This is variable is useful when the default bootrom is being generated as part of `make prep_lite` or `make prep`. During the actual simulation, the user might want to switch the protocol, so the target changes the input to the script when doing this as part of the simulation flow. Now, I am not sure if we still completely support the other protocols or we have moved away from it, but I was just "thinking ahead"

bp_common/test/src/bootrom/bootrom.S
sripathi-muralitharan: Well, these are stores to uncached addresses. I don't think a fence will have any effect here. Is that right?

Makefile
dpetrisko: make sense!

bp_common/test/src/bootrom/Makefile
dpetrisko: Makes sense!

Makefile
dpetrisko: The default bootrom is actually committed to the repo, so we don't have to worry about it being generated.  The non-default bootrom should actually be generated as part of the simulation collateral, so part of the bp_tethered/Makefile.vcs flow.  That way, the user could specify by setting variables

bp_common/test/src/bootrom/bootrom.S
dpetrisko: That is not correct.  A fence ensures that the store completes before moving to the next instruction

bp_common/test/src/bootrom/Makefile
dpetrisko: Ah, gotcha.  We don't actually want to generate the bootrom at make prep time, since the default should be committed to the repo.  So probably don't need this

dpetrisko:
@sripathi-muralitharan Every LGTM. We just need to add the bootom generation step to the SIM_COLLATERAL Makefile.vcs/verilator and it'll be good to go

sripathi-muralitharan:
Well, what do you mean? The SIM_COLLATERAL in both files have that target (`bootrom.riscv`, `bootrom.mem` and `bootrom.dump`) already. Could you please clarify what you mean exactly?

dpetrisko:
>Well, what do you mean? The SIM_COLLATERAL in both files have that target (bootrom.riscv, bootrom.mem and bootrom.dump) already. Could you please clarify what you mean exactly?

That just copies the default bootrom, which up until now was the only bootrom.  Since we need to regenerate the bootrom upon switching CCE ucode, we need to actually compile it as part of that step.

bp_common/test/src/bootrom/bootrom.S
```diff
@@ -35,13 +52,54 @@ _start:
...
-    csrw dpc, x31
+    li t1, CFG_ICACHE_MODE_ADDR
+    or t1, t1, t4
+    sd t0, 0(t1)
+
+    li t1, CFG_CCE_MODE_ADDR
+    or t1, t1, t4
+    sd t0, 0(t1)
+
+    fence.i
```
dpetrisko, line 87: This should be a fence, not a fencei

dpetrisko (changes requested):

dpetrisko:
Cool, thanks! This will be super useful for the future

PR #622, opened 2020-09-23
BP Stream pump modules update

Songchun-Li:
This pull request contains 3 major updates:
1. _bp_stream_pump_in_ and _bp_stream_pump_out_ are attached to _bp_cce_to_cache_new_. And the FSM and relevant logic in the _bp_cce_to_cache_new_ is updated to work with pumps.
2. _bp_lite_to_stream_ is updated according to the update of BP Stream protocol, including no lock for signal beat cmd/resp and no lock for the last packet in a sequence of cmds/resps.
3. _bp_cache_slice_new_ replaces the old one in _bp_unicore_, and the memory interface of _bp_unicore_ is updated. Also, _bp_burst_to_lite_ and _bp_lite_to_stream_ are added and used to connected _bp_unicore_ in the wrapper.

The PR passed riscv_testlist for fill_width=512, and is currently running beebs_teslist.

Note. _e_bp_unicore_no_l2_cfg_ is not yet updated with the new _bp_unicore_ memory interface.


Songchun-Li:
It also pass beebs_testlist

dpetrisko:
Please do the following to clean up the diff:
- rename _new modules to replace old modules directly
- rebase on me_dev
- checkout the new burst converters from me_dev and use those

Songchun-Li:
It has been rebased on me_dev and updated with the burst to lite logic on me_dev. And the _new modules replaces the old ones as well. It passed both riscv and beebs testlist

bp_top/syn/flist.vcs
```diff
@@ -226,6 +226,9 @@ $BP_FE_DIR/src/v/bp_fe_top.v
 $BP_ME_DIR/src/v/lce/bp_lce.v
 $BP_ME_DIR/src/v/lce/bp_lce_req.v
 $BP_ME_DIR/src/v/lce/bp_lce_cmd.v
+# STREAM PUMP
+$BP_ME_DIR/src/v/stream_pump/bp_stream_pump_in.v
```
dpetrisko, line 230: Move these to /wormhole

bp_top/src/v/bp_tile.v
```diff
@@ -554,7 +554,7 @@ for (genvar i = 0; i < 2; i++)
       logic cache_resp_header_v_li, cache_resp_header_ready_lo;
       logic [dword_width_p-1:0] cache_resp_data_li;
       logic cache_resp_data_v_li, cache_resp_data_ready_lo;
-      bp_me_cache_slice
+      bp_me_cache_slice_new
```
dpetrisko, line 557: This should just be cache slice, right?

bp_me/src/v/wormhole/bp_stream_to_lite.v
```diff
@@ -111,7 +111,7 @@ module bp_stream_to_lite
     //  if (mem_v_i)
     //    $display("[%t] Stream received: %p %x", $time, mem_header_cast_i, mem_data_i);
 
-    //  if (mem_ready_i & mem_v_o)
+    //  if (mem_yumi_i)
```
dpetrisko, line 114: undo

bp_me/src/v/wormhole/bp_lite_to_stream.v
```diff
@@ -55,19 +55,23 @@ module bp_lite_to_stream
      ,.reset_i(reset_i)
 
      ,.data_i(mem_cast_i.header)
-     ,.ready_o(mem_ready_o)
      ,.v_i(mem_v_i)
 
      ,.data_o(header_lo)
      ,.v_o(mem_v_lo)
      ,.yumi_i(mem_yumi_li)
+
```
dpetrisko, line 63: undo

bp_me/src/v/wormhole/bp_lite_to_stream.v
```diff
@@ -55,19 +55,23 @@ module bp_lite_to_stream
...
-  wire is_wr = mem_cast_i.header.msg_type inside {e_mem_msg_uc_wr, e_mem_msg_wr};
   localparam data_len_width_lp = `BSG_SAFE_CLOG2(stream_words_lp);
-  wire [data_len_width_lp-1:0] num_stream_cmds = (forward_p ^ is_wr)
+  localparam data_ptr_width_lp = `BSG_WIDTH(stream_words_lp);
+
+  wire is_wr = mem_cast_i.header.msg_type inside {e_mem_msg_uc_wr, e_mem_msg_wr};
+  wire [data_ptr_width_lp-1:0] num_stream_cmds_1base = (forward_p ^ is_wr)
     ? 1'b1
     : `BSG_MAX(((1'b1 << mem_cast_i.header.size) / out_data_bytes_lp), 1'b1);
+  wire [data_len_width_lp-1:0] num_stream_cmds = num_stream_cmds_1base - 1'b1;
```
dpetrisko, line 74: undo, the -1 is more clear in len

bp_me/src/v/wormhole/bp_lite_to_stream.v
```diff
@@ -76,49 +80,51 @@ module bp_lite_to_stream
...
      ,.data_i(mem_cast_i.data)
-     ,.len_i(num_stream_cmds - 1'b1)
+     ,.len_i(num_stream_cmds)
      ,.v_i(mem_v_i)
 
      ,.data_o(mem_data_o)
      ,.v_o(mem_v_o)
-     ,.yumi_i(mem_ready_i & mem_v_o)
-
+     ,.yumi_i(mem_v_o & mem_ready_i)
```
dpetrisko, line 88: undo, ready & valid is more standard

bp_me/src/v/wormhole/bp_lite_to_stream.v
```diff
@@ -76,49 +80,51 @@ module bp_lite_to_stream
...
+    ,.en_i(mem_v_i)
+    ,.data_o(num_stream_cmds_r)
+    );
+    
   // We wouldn't need this counter if we could peek into the PISO...
-  localparam data_ptr_width_lp = `BSG_WIDTH(stream_words_lp);
-  logic [data_ptr_width_lp-1:0] first_cnt, last_cnt, current_cnt;
-  bsg_counter_set_en
-   #(.max_val_p(stream_words_lp), .reset_val_p(0))
+  logic [data_len_width_lp-1:0] data_cnt;
```
dpetrisko, line 105: I think you just want to take my changes here as well.  Doing a count + add does not handle the wraparound case properly

bp_me/src/v/cache/bp_me_cce_to_cache.v
```diff
@@ -4,347 +4,266 @@
...
+   #(.width_p(block_offset_width_lp))
+   addr_wraparound_adder
+    (.a_i(critical_addr)
+    ,.b_i(address_offset)
+    ,.s_o(sub_block_addr)
+    ,.c_o(carry_unused)
+    );
+  assign stream_addr = {cmd_addr[paddr_width_p-1:block_offset_width_lp], sub_block_addr};
+
+  // A new version of the stream fifo
```
dpetrisko, line 125: delete comment, this won't make sense after being merged

bp_me/src/v/cache/bp_me_cache_slice.v
```diff
@@ -51,6 +51,56 @@ module bp_me_cache_slice
    );
 
   `declare_bp_mem_if(paddr_width_p, cce_block_width_p, lce_id_width_p, lce_assoc_p, cce_mem);
+  bp_cce_mem_msg_s mem_cmd, mem_resp;
+  assign mem_cmd = mem_cmd_i;
+  assign mem_resp_o = mem_resp;
+
```
dpetrisko, line 57: Add 

TODO: convert ports to stream interface to avoid ser/des

bp_me/src/v/cache/bp_me_cce_to_cache.v
```diff
@@ -4,347 +4,266 @@
...
 
-        if ((mem_cmd_lo.header.addr < dram_base_addr_gp) && (mem_cmd_lo.header.addr[0+:20] == cache_tagfl_base_addr_gp))
-          begin
-            cache_pkt.opcode = TAGFL;
-            cache_pkt.addr = {cmd_data[0][0+:lg_sets_lp+lg_ways_lp], block_offset_width_lp'(0)};
+            cache_pkt.addr = is_tagfl ? tagfl_addr : cmd_addr;
+            cache_pkt.data = mem_cmd_data_lo;
+            // This mask is only used for the LM/SM operations for >64 bit mask operations
+            cache_pkt.mask = '1;
+            cache_pkt_v_o  = cache_pkt_ready_i & mem_resp_ready_lo & mem_cmd_v_lo;
```
dpetrisko, line 225: don't need to wait for mem_resp_ready here.  The cache will stall if it's not ready at the output

bp_me/src/v/cache/bp_me_cce_to_cache.v
```diff
@@ -4,347 +4,266 @@
...
+  wire is_csr   = (mem_cmd_header_lo.addr < dram_base_addr_gp);
+  wire is_tagfl = is_csr & ((mem_cmd_header_lo.addr[0+:20] == cache_tagfl_base_addr_gp));
+  wire [paddr_width_p-1:0] tagfl_addr = {mem_cmd_data_lo[0+:lg_sets_lp+lg_ways_lp], block_offset_width_lp'(0)};
+  wire [paddr_width_p-1:0] cmd_addr   = mem_cmd_header_lo.addr;
+
+  logic [paddr_width_p-1:0] stream_addr;
+  logic [block_offset_width_lp-1:0] critical_addr, address_offset, sub_block_addr;
+  logic carry_unused;
+  assign critical_addr = mem_cmd_header_lo.addr[0+:block_offset_width_lp];
+  assign address_offset = mem_cmd_cnt_lo << fill_offset_width_lp;
```
dpetrisko, line 114: Should be able to just inject the stream cnt with the correct offset, rather than doing an addition.  See how it's done in the lite2stream module

bp_me/src/v/cache/bp_me_cce_to_cache.v
```diff
@@ -4,347 +4,266 @@
...
-          begin
-            cmd_counter_n = cmd_counter_r + 1;
-            if (cmd_counter_r == cmd_max_count_r)
-              begin
-                cmd_counter_n = '0;
-                cmd_state_n = READY;
-                mem_cmd_yumi_li = 1'b1;
-              end
+            // Transition back to ready once we've completely sent out the stream
+            cmd_state_n = mem_cmd_done_lo ? e_cmd_ready : e_cmd_stream;
```
dpetrisko, line 246: does done raise high on the last entry or after the last entry has been sent?

bp_me/src/v/stream_pump/bp_stream_pump_out.v
```diff
@@ -0,0 +1,167 @@
...
+  , input                                      mem_yumi_i
+  
+  // FSM side
+  , input        [mem_msg_header_width_lp-1:0] header_i
+  , input        [fill_width_p-1:0]            data_i
+  , input                                      v_i
+  , output logic                               yumi_o
+
+  // control signals
+  , input                                      new_i
```
dpetrisko, line 35: I don't think the new_i signal is useful here, since we'll know if it's new from the fsm

bp_me/src/v/stream_pump/bp_stream_pump_out.v
```diff
@@ -0,0 +1,167 @@
...
+  , output logic                               done_o
+  );
+
+  `declare_bp_mem_if(paddr_width_p, fill_width_p, lce_id_width_p, lce_assoc_p, mem);
+  
+  `bp_cast_i(bp_mem_msg_header_s, header);
+  `bp_cast_o(bp_mem_msg_header_s, mem_header);
+
+  enum logic [1:0] {e_reset, e_single, e_pass_stream, e_combine_stream} state_n, state_r;
+  wire is_block_read  = header_cast_i.msg_type inside {e_mem_msg_rd};
```
dpetrisko, line 46: uncached ops and regular ops can both be 1-512 bytes.  To see if it's a stream operation or not, you have to look at data size

bp_me/src/v/stream_pump/bp_stream_pump_out.v
```diff
@@ -0,0 +1,167 @@
...
+  assign critical_addr = header_cast_i.addr[0+:block_offset_width_lp];
+  assign address_offset = cnt_o << fill_offset_width_lp;
+  bsg_adder_ripple_carry 
+   #(.width_p(block_offset_width_lp))
+   wraparound_adder
+    (.a_i(critical_addr)
+    ,.b_i(address_offset)
+    ,.s_o(sub_block_addr)
+    ,.c_o(carry_unused)
+    );
```
dpetrisko, line 79: as above, concatenate rather than add

bp_me/src/v/stream_pump/bp_stream_pump_in.v
```diff
@@ -0,0 +1,181 @@
...
+
+  // bus side
+  , input        [mem_msg_header_width_lp-1:0] mem_header_i
+  , input        [fill_width_p-1:0]            mem_data_i
+  , input                                      mem_v_i
+  , input                                      mem_lock_i
+  , output logic                               mem_ready_o
+  
+  // FSM side
+  , output logic [mem_msg_header_width_lp-1:0] header_o
```
dpetrisko, line 28: probably makes sense to split out into 'base_header' and address_o.  That way, don't have to do the address recalculation repeatedly.  Same comment for pump out

bp_me/src/v/cache/bp_me_cce_to_cache.v
Songchun-Li: Done signal is sent from the pump out when the the last entry has been sent, so the FSM here can simply transit back to ready when it gets a done pulse.

bp_me/src/v/wormhole/bp_lite_to_stream.v
Songchun-Li: But it seems there is some issues with the logic 
`last_cnt  = first_cnt - 1'b1;`
It stalls when it works with uncache commands with only 64 bit data

bp_me/src/v/stream_pump/bp_stream_pump_out.v
Songchun-Li: The idea of new here is that it inducates the upcoming header will do stream. And those headers with data less than cache_data_width do not have high new_i signal. Indeed, we can work with the size info in header to identify if it will do stream, but the new signal is simply passed from the pump_in module and does not add much hardware overhead.

bp_me/src/v/stream_pump/bp_stream_pump_in.v
Songchun-Li: Probably a good idea for pump_in module, then the cnt_o is not necessary. However,it may not for the pump_out module.  At UCE side, we will need the cnt_o to fetch data from dirty_data_r and calculate the address for each entry for the very first time. And at L2 side, the FSM sends only the data from bsg_cache to pump_out and not does not provide any address info, so the pump_out have to recalculate the address.

bp_me/src/v/wormhole/bp_lite_to_stream.v
Songchun-Li: For example, when the uncached write 64 bit data into with address 0x8000_0000, we get the first-cnt of 0, then with -1 we have last_cnt of 0xf where the cnts are 4-bit-wide, but it should be 1 in fact.

Songchun-Li: The latest commit basically picks up your changes and add something to solve some issues I saw in simulation. The remainning issue is to generate the different last_cnt based on size info.

bp_me/src/v/cache/bp_me_cce_to_cache.v
Songchun-Li: Solved in the commit, "Change the address add to concatenate"

bp_me/src/v/stream_pump/bp_stream_pump_out.v
Songchun-Li: Solved in "Change the addr calculateion to concatenate in pump_out“

bp_me/src/v/stream_pump/bp_stream_pump_out.v
Songchun-Li: Solved in the commit, "Make the pump module transists to stream states based on size info"

bp_me/src/v/stream_pump/bp_stream_pump_out.v
Songchun-Li: Update: Removing the new_i should be helpful for decoupling the fsm and pump_out, and it is removed in the commit "Remove the new_i for pump_out"

Songchun-Li:
The mentioned issues are mostly fixed with a few commits, except the "address_o" one.

bp_me/src/v/wormhole/bp_stream_pump_in.v
```diff
@@ -0,0 +1,190 @@
...
+ import bp_cce_pkg::*;
+ import bp_common_pkg::*;
+ import bp_common_aviary_pkg::*;
+ import bp_me_pkg::*;
+ import bsg_cache_pkg::*;
+ #(parameter bp_params_e bp_params_p = e_bp_default_cfg
+   `declare_bp_proc_params(bp_params_p)
+   
+   , parameter cache_data_width_p = dword_width_p
+   // 1 : master: read is N to N, write is 1 to 1
```
dpetrisko, line 11: Not sure I understand this designation, what does N to N mean here?

bp_me/src/v/wormhole/bp_stream_pump_in.v
```diff
@@ -0,0 +1,190 @@
...
+    (.clk_i(clk_i)
+     ,.reset_i(reset_i)
+
+     ,.set_i(new_o) 
+     ,.en_i(cnt_up)
+     ,.val_i(first_cnt + 1'b1)
+     ,.count_o(current_cnt)
+     );
+  assign first_cnt = header_cast_o.addr[stream_offset_width_lp+:data_len_width_lp];
+  assign last_cnt  = first_cnt - 1'b1;
```
dpetrisko, line 88: again, I think last_cnt = first_cnt + the stream length

bp_me/src/v/wormhole/bp_stream_pump_out.v
```diff
@@ -0,0 +1,166 @@
...
+    (.clk_i(clk_i)
+     ,.reset_i(reset_i)
+
+     ,.set_i(set_cnt) 
+     ,.en_i(cnt_up)
+     ,.val_i(first_cnt + 1'b1)
+     ,.count_o(current_cnt)
+     );
+  assign first_cnt = header_cast_i.addr[stream_offset_width_lp+:data_len_width_lp];
+  assign last_cnt  = first_cnt - 1'b1;
```
dpetrisko, line 70: same here

bp_me/src/v/cache/bp_me_cce_to_cache.v
```diff
@@ -4,347 +4,246 @@
  */
  
 module bp_me_cce_to_cache
```
dpetrisko, line 6: This module looks great!

dpetrisko (changes requested):

bp_me/src/v/wormhole/bp_stream_pump_in.v
Songchun-Li: For example, at master(UCE) sidem, for read operation, the pump_in receives N beats of headers/data and sends N beats of data to the UCE FSM. As for write operation, the pump_in receives 1 beat of write response and send 1 beat of write response to the FSM. This is how N to N and 1 to 1 come.
The interest part is at the client side, that we have 1 to N for read operation. That is because we sends one beat of cmd header to the pump_in, then pump_in decodes it and sends N beats of cmds to the L2 adapter FSM.

Songchun-Li:
There are 2 changes.
1. Replaced the cnt_o with stream_addr_o for the pump_in module like we discussed above. With this change, UCE and cce_to_cache will not need to repeatedly calculate the address, especially the address is demanding in multiple scenarios to fill the cache. It decouples the pump_in from fsm in a better way.
2. Changed the logic to get last_cnt for all three modules mentioned above, except the dff_en_bypass one as there is no special condition for the num_stream in the updated design any more
These changes passed cache_hammer, riscv_testlist and beebs, and it is ready for review.

bp_me/src/v/wormhole/bp_stream_pump_in.v
dpetrisko: Makes sense that there should always be 1:1 correspondence between the FSM and the stream side.  Then I don't understand what you mean by having 1:N for reading.  You can't send more data into the stream than you get from the FSM...

Songchun-Li: 1:N for read and N:1 for write write response are about the header/addr and have nothing to do with data, and they happen only at the client side(cce_to_cache). 
For cache read, we have only 1 beat of command with critical address from the stream side, then pump_in decodes this command into 8 sub-block address (with 64 bit cache data width) and send it to cce_to_cache FSM, so we have 1:8 for read. As for write response, the pump_out waits for 8 response from FSM, and only sends out 1 response to the stream side, when all 8 responses is receive, so we have 8:1. As you can see, neither of them involves data.
I understand this maybe misleading and try to get a better way to describe it.

Songchun-Li: I added some new comments in the latest commit.

bp_me/src/v/wormhole/bp_stream_pump_out.v
```diff
@@ -0,0 +1,173 @@
...
+ import bsg_cache_pkg::*;
+ #(parameter bp_params_e bp_params_p = e_bp_default_cfg
+   `declare_bp_proc_params(bp_params_p)
+
+   , parameter cache_data_width_p = dword_width_p
+   // The way it handles block load/store
+   // 1 : master: read:  receives the only 1 header from FSM, and sends the 1 block load command to the stream side
+   //             write: receives the header and N sub-blocks of data from dirty_data_r with cnt_o, and sends N sub-block store command to the stream side.
+   // 0 : client: read:  receives N sub-block of data from bsg_cache and sends N sub-block load responses to the stream side
+   //             write: combines N sub-block store responses from bsg_cache into 1 block store response, and sends it to stream side
```
dpetrisko, line 15: bsg_cache should only be sending 1 store response for a burst store, per the stream specs

bp_me/src/v/wormhole/bp_stream_pump_in.v
```diff
@@ -0,0 +1,205 @@
...
+   , parameter cache_data_width_p = dword_width_p
+   // The way it handles block load/store
+   // 1 : master: read:  receives N sub-block load responses from stream side and sends N sub-blocks of data to FSM together with corresponding address
+   //             write: receives 1 block store response from stream side and sends the only 1 header to the FSM
+   // 0 : client: read:  decodes the only 1 block load command from the stream side into N sub-block addresses, and sents them to the cce_to_cache FSM
+   //             write: receives N sub-blocks store command from stream side, and sends N sub-blocks of data to FSM together with corresponding address
+   , parameter logic master_p = 1
+
+   `declare_bp_mem_if_widths(paddr_width_p, cache_data_width_p, lce_id_width_p, lce_assoc_p, mem)
+   , localparam block_offset_width_lp = `BSG_SAFE_CLOG2(cce_block_width_p >> 3)
```
dpetrisko, line 19: This should be a parameter, not just cce_block_width_p

bp_me/src/v/wormhole/bp_stream_pump_in.v
```diff
@@ -0,0 +1,205 @@
+module bp_stream_pump_in
+ import bp_cce_pkg::*;
+ import bp_common_pkg::*;
+ import bp_common_aviary_pkg::*;
+ import bp_me_pkg::*;
+ import bsg_cache_pkg::*;
+ #(parameter bp_params_e bp_params_p = e_bp_default_cfg
+   `declare_bp_proc_params(bp_params_p)
+   
+   , parameter cache_data_width_p = dword_width_p
```
dpetrisko, line 10: Rename, not specific to cache

bp_me/src/v/wormhole/bp_stream_pump_in.v
```diff
@@ -0,0 +1,205 @@
...
+   // The way it handles block load/store
+   // 1 : master: read:  receives N sub-block load responses from stream side and sends N sub-blocks of data to FSM together with corresponding address
+   //             write: receives 1 block store response from stream side and sends the only 1 header to the FSM
+   // 0 : client: read:  decodes the only 1 block load command from the stream side into N sub-block addresses, and sents them to the cce_to_cache FSM
+   //             write: receives N sub-blocks store command from stream side, and sends N sub-blocks of data to FSM together with corresponding address
+   , parameter logic master_p = 1
```
dpetrisko, line 16: Take a look at how bp_burst_to_lite et al handle the specialization.  It's more flexible and less confusing than this.   I think your master/client terminology is a bit off, but this should fix it anyway

bp_me/src/v/wormhole/bp_stream_pump_in.v
```diff
@@ -0,0 +1,205 @@
...
+
+  // bus side
+  , input        [mem_msg_header_width_lp-1:0] mem_header_i
+  , input        [cache_data_width_p-1:0]      mem_data_i
+  , input                                      mem_v_i
+  , input                                      mem_lock_i
+  , output logic                               mem_ready_o
+  
+  // FSM side
+  , output logic [mem_msg_header_width_lp-1:0] header_o
```
dpetrisko, line 36: I'd name these fsm_* for clarity

bp_me/src/v/wormhole/bp_stream_pump_in.v
```diff
@@ -0,0 +1,205 @@
...
+  , output logic [cache_data_width_p-1:0]      data_o
+  , output logic                               v_o
+  , input                                      yumi_i
+
+  // control signals
+  , output logic [paddr_width_p-1:0]           stream_addr_o
```
dpetrisko, line 42: fsm_addr_o

bp_me/src/v/wormhole/bp_stream_pump_in.v
```diff
@@ -0,0 +1,205 @@
...
+                cnt_up = new_o;
+                mem_yumi_li = ~(is_read_op & ~single_data_beat) & yumi_i;
+                state_n = new_o 
+                          ? (is_read_op & ~single_data_beat) 
+                            ? e_decode_stream 
+                            : e_pass_stream
+                          : e_single;
+              end
+          end
+        e_pass_stream:
```
dpetrisko, line 161: Comment what each of these states are doing.  I think pass_stream and decode_stream can be combined, since the only difference seems to be whether an ack is sent

bp_me/src/v/wormhole/bp_stream_pump_out.v
```diff
@@ -0,0 +1,173 @@
+module bp_stream_pump_out
```
dpetrisko, line 1: Similar comments to stream_pump_in

Songchun-Li:
The pump modules are changed with payload_mask_p and the pass/decode, pass/combine stream states are integrated. It passed cache_hammer, riscv and beebs testlist, and is ready for review.

bp_me/src/v/wormhole/bp_stream_pump_in.v
```diff
@@ -0,0 +1,171 @@
...
+      ,.data_i({mem_lock_i, mem_header_i, mem_data_i})
+      ,.v_i(mem_v_i)
+      ,.ready_o(mem_ready_o)
+
+      ,.data_o({mem_lock_lo, mem_header_lo, mem_data_lo})
+      ,.v_o(mem_v_lo)
+      ,.yumi_i(mem_yumi_li)
+      );
+
+  wire is_read_op  = mem_header_lo.msg_type inside {e_bedrock_mem_uc_rd, e_bedrock_mem_rd};
```
dpetrisko, line 71: is_read / is_write should be at best irrelevant and at worst incorrect with payload_mask_p

bp_me/src/v/wormhole/bp_lite_to_stream.v
```diff
@@ -89,30 +89,45 @@ module bp_lite_to_stream
...
+  assign first_cnt = mem_cast_i.header.addr[stream_offset_width_lp+:data_len_width_lp];
+  
+  always_comb 
+    begin
+      if (num_stream_cmds == data_len_width_lp'(1))
+        last_cnt  = first_cnt;
+      else 
+        last_cnt  = first_cnt - 1'b1;
+    end
+  bsg_dff_en
```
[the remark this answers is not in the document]
dpetrisko: Bump for this change

bp_me/src/v/wormhole/bp_stream_pump_out.v
```diff
@@ -0,0 +1,148 @@
...
+  );
+
+  `declare_bp_bedrock_mem_if(paddr_width_p, stream_data_width_p, lce_id_width_p, lce_assoc_p, xce);
+  
+  `bp_cast_i(bp_bedrock_xce_mem_msg_header_s, fsm_header);
+  `bp_cast_o(bp_bedrock_xce_mem_msg_header_s, mem_header);
+
+  enum logic [1:0] {e_reset, e_single, e_stream, e_combine_stream} state_n, state_r;
+  
+  wire is_read_op  = fsm_header_cast_i.msg_type inside {e_bedrock_mem_uc_rd, e_bedrock_mem_rd};
```
dpetrisko, line 51: same here, replace just with payload mask

bp_me/src/v/wormhole/bp_stream_pump_out.v
```diff
@@ -0,0 +1,148 @@
...
+  , output logic [data_len_width_lp-1:0]       cnt_o
+  , output logic                               done_o
+  );
+
+  `declare_bp_bedrock_mem_if(paddr_width_p, stream_data_width_p, lce_id_width_p, lce_assoc_p, xce);
+  
+  `bp_cast_i(bp_bedrock_xce_mem_msg_header_s, fsm_header);
+  `bp_cast_o(bp_bedrock_xce_mem_msg_header_s, mem_header);
+
+  enum logic [1:0] {e_reset, e_single, e_stream, e_combine_stream} state_n, state_r;
```
dpetrisko, line 49: remove unused combine_stream state

bp_me/src/v/wormhole/bp_stream_pump_out.v
```diff
@@ -0,0 +1,148 @@
...
+      cnt_up = '0; 
+      done_o = '0;
+
+      state_n = state_r;
+      case (state_r)
+        e_reset:
+          begin
+            state_n = e_single;
+          end
+        e_single:
```
dpetrisko, line 95: I actually don't think you need this FSM at all.  The degenerate case of stream is a single data beat which doesn't have any special properties other than being length 1

bp_me/src/v/wormhole/bp_stream_pump_in.v
```diff
@@ -0,0 +1,171 @@
...
+      cnt_up = '0;
+      done_o = '0;
+
+      state_n = state_r;
+      case (state_r)
+        e_reset:
+          begin
+            state_n = e_single;
+          end
+        e_single:
```
dpetrisko, line 129: Same here, I don't think we need an FSM, and instead reduce the single_beat case to be a stream of length 1

bp_me/src/v/wormhole/bp_stream_pump_in.v
```diff
@@ -0,0 +1,171 @@
...
+  , output logic                                               mem_ready_o
+  
+  // FSM side
+  , output logic [bp_bedrock_xce_mem_msg_header_width_lp-1:0] fsm_header_o
+  , output logic [stream_data_width_p-1:0]                    fsm_data_o
+  , output logic                                              fsm_v_o
+  , input                                                     fsm_yumi_i
+
+  // control signals
+  , output logic [paddr_width_p-1:0]           fsm_addr_o
```
dpetrisko, line 41: put with the other FSM signals

bp_me/src/v/wormhole/bp_stream_pump_out.v
```diff
@@ -0,0 +1,148 @@
...
+
+  // bus side
+  , output logic [bp_bedrock_xce_mem_msg_header_width_lp-1:0] mem_header_o
+  , output logic [stream_data_width_p-1:0]                    mem_data_o
+  , output logic                                              mem_v_o
+  , output logic                                              mem_lock_o
+  , input                                                     mem_yumi_i
+  
+  // FSM side
+  , input        [bp_bedrock_xce_mem_msg_header_width_lp-1:0] fsm_header_i
```
dpetrisko, line 34: fsm_base_header

bp_me/src/v/wormhole/bp_stream_pump_in.v
```diff
@@ -0,0 +1,171 @@
...
+
+  // bus side
+  , input         [bp_bedrock_xce_mem_msg_header_width_lp-1:0] mem_header_i
+  , input         [stream_data_width_p-1:0]                    mem_data_i
+  , input                                                      mem_v_i
+  , input                                                      mem_lock_i
+  , output logic                                               mem_ready_o
+  
+  // FSM side
+  , output logic [bp_bedrock_xce_mem_msg_header_width_lp-1:0] fsm_header_o
```
dpetrisko, line 35: fsm_base_header

dpetrisko (changes requested):

dpetrisko:
Gave more feedback to simplify stream pumps

Songchun-Li:
Updates are made based on the above suggestions and it is ready for review.

Songchun-Li:
Found some issues in the pump_in when it is attached to UCE and got them fixed. The updated pump modules are tested.

bp_me/src/v/wormhole/bp_stream_pump_in.v
```diff
@@ -0,0 +1,152 @@
...
+    ,.clear_i(done_o)
+    ,.data_o(streaming_r)
+    );
+
+  always_comb
+    begin
+      fsm_base_header_cast_o = mem_header_lo;
+      fsm_data_o = mem_data_lo;
+      fsm_v_o = mem_v_lo;
+      if (single_data_beat | (is_master & ~has_data))
```
dpetrisko, line 124: I don't think you need this if statement

bp_me/src/v/wormhole/bp_stream_pump_out.v
```diff
@@ -0,0 +1,129 @@
...
+  assign cnt_o = set_cnt ? first_cnt : current_cnt;
+  assign is_last_cnt = (cnt_o == last_cnt);
+
+  always_comb 
+    begin
+      mem_header_cast_o = fsm_base_header_cast_i;
+      mem_data_o = fsm_data_i;
+      mem_v_o = fsm_v_i;
+
+      if (single_data_beat | (is_master & ~has_data))
```
dpetrisko, line 93: Same here, I think this is unnecessary

bp_me/src/v/wormhole/bp_stream_pump_out.v
```diff
@@ -0,0 +1,129 @@
...
+              mem_lock_o = ~is_last_cnt;
+
+              cnt_up     = mem_yumi_i;
+              fsm_yumi_o = mem_yumi_i;
+            end
+          else
+            begin
+              // handle message size > stream_data_width_p w/o data payload (combines write responses into one)
+              mem_v_o = is_last_cnt & fsm_v_i;
+
```
dpetrisko, line 120: mem_lock_o is undriven in this case

bp_me/src/v/wormhole/bp_stream_pump_out.v
```diff
@@ -0,0 +1,129 @@
...
+   )
+  ( input clk_i
+  , input reset_i
+
+  // bus side
+  , output logic [bp_bedrock_xce_mem_msg_header_width_lp-1:0] mem_header_o
+  , output logic [stream_data_width_p-1:0]                    mem_data_o
+  , output logic                                              mem_v_o
+  , output logic                                              mem_lock_o
+  , input                                                     mem_yumi_i
```
dpetrisko, line 31: We should convert this to mem_ready_and_i and fsm_ready_and_o.  That will be much more flexible

bp_me/src/v/wormhole/bp_stream_pump_in.v
```diff
@@ -0,0 +1,152 @@
...
+   )
+  ( input clk_i
+  , input reset_i
+
+  // bus side
+  , input         [bp_bedrock_xce_mem_msg_header_width_lp-1:0] mem_header_i
+  , input         [stream_data_width_p-1:0]                    mem_data_i
+  , input                                                      mem_v_i
+  , input                                                      mem_lock_i
+  , output logic                                               mem_ready_o
```
dpetrisko, line 32: We should convert this to mem_ready_and_o and fsm_ready_and_i. That will be much more flexible

bp_me/src/v/wormhole/bp_stream_pump_out.v
```diff
@@ -0,0 +1,129 @@
...
+  
+  wire is_master = (payload_mask_p == mem_cmd_payload_mask_gp);
+  wire has_data = payload_mask_p[fsm_base_header_cast_i.msg_type];
+  wire [data_len_width_lp-1:0] num_stream = `BSG_MAX((1'b1 << fsm_base_header_cast_i.size) / (stream_data_width_p / 8), 1'b1);
+  wire single_data_beat = (num_stream == data_len_width_lp'(1));
+
+  logic ready_r, cnt_up, is_last_cnt;
+  logic [data_len_width_lp-1:0] first_cnt, last_cnt, current_cnt;
+
+  bsg_dff_reset_en
```
dpetrisko, line 59: bsg_dff_reset_set_clear

bp_me/src/v/wormhole/bp_stream_pump_in.v
```diff
@@ -0,0 +1,152 @@
...
+  
+  // FSM side
+  , output logic [bp_bedrock_xce_mem_msg_header_width_lp-1:0] fsm_base_header_o
+  , output logic [paddr_width_p-1:0]                          fsm_addr_o
+  , output logic [stream_data_width_p-1:0]                    fsm_data_o
+  , output logic                                              fsm_v_o
+  , input                                                     fsm_yumi_i
+
+  // control signals
+  , output logic                               new_o
```
dpetrisko, line 42: Maybe stream_new_o, stream_done_o is more clear?  new_o seems like it should be high even for single beat transactions

bp_me/src/v/wormhole/bp_stream_pump_in.v
```diff
@@ -0,0 +1,152 @@
...
+      fsm_data_o = mem_data_lo;
+      fsm_v_o = mem_v_lo;
+      if (single_data_beat | (is_master & ~has_data))
+          begin
+            // handle message size < stream_data_width_p & write response w/o data payload
+            fsm_addr_o = '0;
+            
+            mem_yumi_li = fsm_yumi_i;
+
+            new_o = '0;
```
dpetrisko, line 131: Seems like there's a lot of overlap between new_o, done_o and lock.  

Could we just do fsm_lock_o, and then the transition would happen on raising and lowering of lock? I guess the issue is the last beat.  This is the problem with not having a "last" signal in the interface, I guess

bp_me/src/v/wormhole/bp_stream_pump_in.v
```diff
@@ -0,0 +1,152 @@
...
+  
+  logic [data_len_width_lp-1:0] cnt_o;
+  assign cnt_o = new_o ? first_cnt : current_cnt;
+  assign is_last_cnt = (cnt_o == last_cnt);
+
+  logic [block_offset_width_lp-1:0] critical_addr, critical_addr_r;
+  assign critical_addr = mem_header_lo.addr[0+:block_offset_width_lp];
+
+  // store this addr for stream state
+  bsg_dff_en_bypass
```
dpetrisko, line 98: Do you need to save this separately? Isn't mem_header_lo valid for the entire stream transaction?

bp_me/src/v/wormhole/bp_stream_pump_in.v
```diff
@@ -0,0 +1,152 @@
...
+  logic [data_len_width_lp-1:0] first_cnt, last_cnt, current_cnt;
+  bsg_counter_set_en
+   #(.max_val_p(stream_words_lp-1), .reset_val_p(0))
+   data_counter
+    (.clk_i(clk_i)
+     ,.reset_i(reset_i)
+
+     ,.set_i(new_o) 
+     ,.en_i(cnt_up)
+     ,.val_i(first_cnt + 1'b1)
```
dpetrisko, line 84: I think you can simplify a lot of logic here with .val_i(first_cnt + cnt_up).  Then first_cnt == last_cnt for single beat transactions.  You'll still need the cnt_o bypass logic, though

bp_me/src/v/wormhole/bp_stream_pump_in.v
```diff
@@ -0,0 +1,152 @@
...
+  bsg_dff_reset_set_clear
+   #(.width_p(1)
+   ,.clear_over_set_p(1))
+   streaming_reg
+    (.clk_i(clk_i)
+    ,.reset_i(reset_i)
+    ,.set_i(cnt_up)
+    ,.clear_i(done_o)
+    ,.data_o(streaming_r)
+    );
```
dpetrisko, line 117: I think streaming is more simply == mem_v_lo & ~done_o?

bp_me/src/v/wormhole/bp_stream_pump_out.v
```diff
@@ -0,0 +1,129 @@
...
+
+  bsg_counter_set_en
+   #(.max_val_p(stream_words_lp-1), .reset_val_p(0))
+   data_counter
+    (.clk_i(clk_i)
+     ,.reset_i(reset_i)
+
+     ,.set_i(set_cnt) 
+     ,.en_i(cnt_up)
+     ,.val_i(first_cnt + 1'b1)
```
dpetrisko, line 78: Similar comments to stream_pump_in

bp_me/src/v/wormhole/bp_stream_pump_out.v
```diff
@@ -0,0 +1,129 @@
...
+  );
+
+  `declare_bp_bedrock_mem_if(paddr_width_p, stream_data_width_p, lce_id_width_p, lce_assoc_p, xce);
+  
+  `bp_cast_i(bp_bedrock_xce_mem_msg_header_s, fsm_base_header);
+  `bp_cast_o(bp_bedrock_xce_mem_msg_header_s, mem_header);
+
+  enum logic [1:0] {e_reset, e_single, e_stream} state_n, state_r;
+  
+  wire is_master = (payload_mask_p == mem_cmd_payload_mask_gp);
```
dpetrisko, line 51: get rid of is_master

bp_me/src/v/wormhole/bp_stream_pump_in.v
```diff
@@ -0,0 +1,152 @@
...
+      ,.data_i({mem_lock_i, mem_header_i, mem_data_i})
+      ,.v_i(mem_v_i)
+      ,.ready_o(mem_ready_o)
+
+      ,.data_o({mem_lock_lo, mem_header_lo, mem_data_lo})
+      ,.v_o(mem_v_lo)
+      ,.yumi_i(mem_yumi_li)
+      );
+
+  wire is_master = (payload_mask_p == mem_resp_payload_mask_gp);
```
dpetrisko, line 69: get rid of is_master

dpetrisko (changes requested):

bp_me/src/v/wormhole/bp_stream_pump_in.v
Songchun-Li: The tricky part is that the mem_header_lo is different betweem read commands and write commands at the L2 adapter side. For read commands, we receive only one cmd from stream side, so we can take it as the fsm_base_header_o directly. However, for write commands, we receive N cmds from the stream sides, so the critical_addr is saved to keep fsm_base_header_o the same. Otherwise, the addr in the fsm_base_header_o keeps incrementing.

bp_me/src/v/wormhole/bp_stream_pump_in.v
Songchun-Li: new_o is uesd in cce_to_cache as a transiting condition of jumping to stream state from ready state, so I think it should not be high for single beat transactions

bp_me/src/v/wormhole/bp_stream_pump_in.v
Songchun-Li: The reason I use a register to hold this is because the new_o = fsm_yumi_i & ~streaming_r, is high only for the first cycle, so it can be used to set the counter and the registers holding the base addr. If we make streaming = 
 mem_v_lo & ~done_o, it will alyways be high before the done signal

bp_me/src/v/wormhole/bp_stream_pump_out.v
Songchun-Li: mem_yumi_i is converted to mem_ready_and_i, but not sure if I did the right conversion for fsm_yumi_i because unlike other modules with ready_and_o I seen, there is no fifo in the pump out, so the fsm side has to be demanding

Songchun-Li:
Made the change of replacing lock signal with last signal and changes based on the feedback. Passed cache_hammer and riscv tests and is ready for review.

bp_me/src/v/wormhole/bp_stream_pump_in.v
dpetrisko: makes sense!

dpetrisko: Oh, new_o = mem_v_lo & (stream_cnt == first_cnt) is lower hardware overhead.  It shouldn't depend on the fsm handshake i.e. new_o is 'helpful'

bp_me/src/v/wormhole/bp_stream_pump_out.v
```diff
@@ -0,0 +1,133 @@
...
+   bsg_dff_reset_set_clear
+   #(.width_p(1)
+   ,.clear_over_set_p(0))
+   streaming_reg
+    (.clk_i(clk_i)
+    ,.reset_i(reset_i)
+    ,.set_i(done_o)
+    ,.clear_i(cnt_up & ~done_o)
+    ,.data_o(ready_r)
+  );
```
dpetrisko, line 67: Can get rid of this register, ready if cnt == '0

bp_me/src/v/wormhole/bp_stream_pump_in.v
```diff
@@ -0,0 +1,147 @@
...
+          
+          stream_new_o = '0;
+          cnt_up = '0;
+          done_o = fsm_ready_and_i & fsm_v_o; // used for UCE to send credits return
+        end
+      else
+        begin
+          // handle message size > stream_data_width_p w/ data payload or commands reading more than data than stream_data_width_p
+          fsm_base_header_cast_o.addr[0+:block_offset_width_lp] = critical_addr_r; // keep the address to be the critical word address
+          fsm_addr_o = { mem_header_lo.addr[paddr_width_p-1:stream_offset_width_lp+data_len_width_lp]
```
dpetrisko, line 134: Does this work with fill_width == 512? I think this concatenation would break because stream_cnt should be 0 bits in that case

bp_me/src/v/wormhole/bp_stream_pump_in.v
```diff
@@ -0,0 +1,147 @@
...
+  // FSM side
+  , output logic [bp_bedrock_xce_mem_msg_header_width_lp-1:0] fsm_base_header_o
+  , output logic [paddr_width_p-1:0]                          fsm_addr_o
+  , output logic [stream_data_width_p-1:0]                    fsm_data_o
+  , output logic                                              fsm_v_o
+  , input                                                     fsm_ready_and_i
+
+  // control signals
+  , output logic                                              stream_new_o
+  , output logic                                              done_o
```
dpetrisko, line 44: minor; stream_done_o

bp_me/src/v/wormhole/bp_stream_pump_out.v
```diff
@@ -0,0 +1,133 @@
...
+  , input                                                     mem_ready_and_i
+  
+  // FSM side
+  , input        [bp_bedrock_xce_mem_msg_header_width_lp-1:0] fsm_base_header_i
+  , input        [stream_data_width_p-1:0]                    fsm_data_i
+  , input                                                     fsm_v_i
+  , output logic                                              fsm_ready_and_o
+
+  // control signals
+  , output logic [data_len_width_lp-1:0]       cnt_o
```
dpetrisko, line 41: minor; stream_cnt_o, stream_done_o

bp_me/src/v/wormhole/bp_stream_pump_out.v
Songchun-Li: Making ready = (cnt == '0) does not work in some cases when the critical word address is not aligned to the block, because cnt value after streaming is done can be non-zero. In addition, with a wrapped-around sequence, it is possible that cnt == 0 at the middle. In this case, stream_cnt will be set back to first_cnt because it rise the set_cnt signal.

bp_me/src/v/wormhole/bp_stream_pump_in.v
Songchun-Li: I've make the stream_new_o to be helpful, but the given logic may not work because stream_cnt = stream_new_o ? first_cnt : current_cnt; so there is a loop here. Therefore, I also try to reset the counter with stream_done, so that we can have new_o = fsm_v_o & (current_cnt == '0), like your suggestion below for the ready signal. However, the issue prevents me from doing this.

bp_me/src/v/wormhole/bp_stream_pump_in.v
Songchun-Li: Yep, it does not work with fill_width == 512. I was planning to make it support 64 bit for the first version and add the variable fill width support in the other PR, the bp_stream with locking arbiter one.

dpetrisko: I'd rather have it support both here so we can merge it in without breaking anything.  I think it should be pretty straightforward using this: https://verificationacademy.com/forums/systemverilog/zero-bit-width-array

bp_me/src/v/wormhole/bp_stream_pump_out.v
```diff
@@ -0,0 +1,133 @@
...
+  always_comb 
+    begin
+      mem_header_cast_o = fsm_base_header_cast_i;
+      mem_data_o = fsm_data_i;
+      mem_v_o = fsm_v_i;
+
+      if (single_beat)
+        begin
+          // handle message size < stream_data_width_p & read command w/o data payload
+          mem_last_o = 1'b1;
```
dpetrisko, line 95: this whole clause is the same as has_data, can be combined

bp_me/src/v/wormhole/bp_stream_pump_out.v
dpetrisko: I see your point.  Doesn't this start in a ~ready state?

bp_me/src/v/wormhole/bp_stream_pump_in.v
dpetrisko: Hm, let me think about this.  The 1-bit FSM might make more sense after all

bp_me/src/v/wormhole/bp_stream_pump_in.v
```diff
@@ -0,0 +1,147 @@
...
+
+  always_comb
+    begin
+      fsm_base_header_cast_o = mem_header_lo;
+      fsm_data_o = mem_data_lo;
+      fsm_v_o = mem_v_lo;
+      if (single_beat)
+        begin
+          // handle message size < stream_data_width_p & write response w/o data payload
+          fsm_addr_o = '0;
```
dpetrisko, line 125: I think this can be combined with the below case

bp_me/src/v/wormhole/bp_stream_pump_in.v
```diff
@@ -0,0 +1,147 @@
...
+        end
+      else
+        begin
+          // handle message size > stream_data_width_p w/ data payload or commands reading more than data than stream_data_width_p
+          fsm_base_header_cast_o.addr[0+:block_offset_width_lp] = critical_addr_r; // keep the address to be the critical word address
+          fsm_addr_o = { mem_header_lo.addr[paddr_width_p-1:stream_offset_width_lp+data_len_width_lp]
+                       , stream_cnt
+                       , mem_header_lo.addr[0+:stream_offset_width_lp]};        
+
+          cnt_up = fsm_ready_and_i & fsm_v_o;
```
dpetrisko, line 140: To combine, I think cnt_up = fsm_ready_and_i & fsm_v_o & ~is_last

bp_me/src/v/wormhole/bp_stream_pump_out.v
Songchun-Li: ready_r was used in pump_out and streaming_r is used in pump_in. They are for the same purpose, but in opposite presentation, so I changed it to streaming_r in pump_out to make them consist in the last commit.And I agree to use a 1-bit FSM which should make more sense while solving this register issue.

bp_me/src/v/wormhole/bp_stream_pump_in.v
Songchun-Li: This may not be a good idea. After running some tests, I found out when there is a wrapped around sequence, the stream_cnt can be any value after a stream_done and so does the last_cnt of the next sequence. If stream_cnt == last_cnt at the beginning, the counter will not cnt_up and mistakenly send out stream_done.

Songchun-Li:
Combine signals in the pump modules. Update the pump_in with 1-bit FSM. Any additional comments on the pump_in module? Still working on combining signasl and fsm update  in pump_out.

bp_me/src/v/wormhole/bp_stream_pump_out.v
```diff
@@ -0,0 +1,142 @@
...
+          fsm_ready_and_o = is_last_cnt ? (mem_ready_and_i & mem_v_o) : fsm_v_i;
+
+          cnt_up  = fsm_v_i;
+          // set_cnt = fsm_v_i & ~streaming_r;
+        end
+      
+      set_cnt = '0;
+
+      state_n = state_r;
+      unique case (state_r)
```
Songchun-Li, line 122: It seems using 1-bit fsm in pump_out may not be a good idea sinc the 2 clauses are kind of parallel. @dpetrisko Do you have any suggestions?

Songchun-Li:
Add variable stream_data_width support including stream_data_width == 512 bit. Discard the 1 bit fsm change because using 1-bit reigster can be more flexible to work with stream_data_width from 64bit to 512 bit.

dpetrisko (approved):
Awesome!! Thanks for the iterations on this, the code looks great now

PR #623, opened 2020-09-23
new BedRock interface

muwyse:
This PR reworks the LCE-CCE and CCE-MEM interfaces into a more generic BedRock interface that is compatible with the proposed BP Lite/Burst/Stream network protocol changes. The goal is to make the BedRock messages easier to pass through generic/parameterized BP Lite/Burst/Stream protocol networks and converter modules (i.e., Lite->Burst, Burst->Stream), and to enable the removal of serialization/deserialization steps throughout the LCE-CCE-MEM networks.

The main change is to create a generic BedRock message struct, similar to the existing CCE-MEM interface struct with {data, payload, size, addr, msg_type}
--- msg_type is a union of the LCE Req, Cmd, Resp message type enums and Memory message type enum that are standardized to 4-bits wide.

Functionally, the LCE Req, Cmd, and Resp and Memory Cmd/Resp messages do not change. All the fields still exist for each message class, and the meanings of the messages remain unchanged.


dpetrisko:
Looks very promising.  What do you think about including the mem channels in the bedrock unions? That would enforce the size restrictions more strongly.

Only concern is naming.  We could do _lce_cmd and _mem_cmd or come up with something unique?

dpetrisko:
>modify the size field to include a 0B entry in the enum. This may not be needed if it is easy to specify a parameterized check on the message type to use/ignore the size field in the converters.

Yeah, I think we should be able to do this by passing a list of zero payload message types to the converter modules.  I'll check out if the syntax works in Verilator

muwyse:
> Looks very promising. What do you think about including the mem channels in the bedrock unions? That would enforce the size restrictions more strongly.
> 
> Only concern is naming. We could do _lce_cmd and _mem_cmd or come up with something unique?

I think we could include them. The hierarchy would be BedRock is our logical level protocol and has 5 channels/instances (lce_req, lce_cmd, lce_resp, mem_cmd, mem_resp). It would keep the size field consistent as all of BedRock would live in a single header file. I think lce_* and mem_* naming is sufficient and clear.

muwyse:
> > modify the size field to include a 0B entry in the enum. This may not be needed if it is easy to specify a parameterized check on the message type to use/ignore the size field in the converters.
> 
> Yeah, I think we should be able to do this by passing a list of zero payload message types to the converter modules. I'll check out if the syntax works in Verilator

Cool. If SV and Verilator support it, my vote would be to take that approach and leave the message size enum starting at 1B.

I have changed my opinion here. 0B as a size isn't all that useful because, e.g., a mem_read needs to specify the size to tell memory how much data it wants returned. So the modules can't only look at the size field to determine if the message has data.

muwyse:
Updated to merge CCE-Mem and LCE-CCE interfaces into single BedRock interface

dpetrisko:
So, we currently have the ability to name the structs for mem_cmd/mem_resp to e.g. uce_cmd/uce_resp dram_cmd/dram_resp. The only thing that's really parameterizable here is the data width.  I propose that we get rid of that naming and replace all msg_s formats with separate header/data fields.  This would allow us to declare all 5 BedRock networks in a single macro, rather than separating them out

muwyse:
> So, we currently have the ability to name the structs for mem_cmd/mem_resp to e.g. uce_cmd/uce_resp dram_cmd/dram_resp. The only thing that's really parameterizable here is the data width. I propose that we get rid of that naming and replace all msg_s formats with separate header/data fields. This would allow us to declare all 5 BedRock networks in a single macro, rather than separating them out

I'm not sure I follow here. I can create a macro that does all five at once, but we also commonly only use the LCE-CCE or Mem-CCE network definitions in modules so I think there is value in there being separate macros, too.

dpetrisko:
>I'm not sure I follow here. I can create a macro that does all five at once, but we also commonly only use the LCE-CCE or Mem-CCE network definitions in modules so I think there is value in there being separate macros, too.

Yeah, I see your point.  The preprocessor is "free", so I don't think it makes real a difference either way.  We can leave it for now

muwyse:
Status update: ci scripts for single core riscv_tests and multicore run with vcs both pass on kk9

bp_top/src/v/bp_cfg.v
```diff
@@ -40,17 +40,17 @@ module bp_cfg
    );
 
   `declare_bp_cfg_bus_s(vaddr_width_p, core_id_width_p, cce_id_width_p, lce_id_width_p, cce_pc_width_p, cce_instr_width_p);
-  `declare_bp_mem_if(paddr_width_p, dword_width_p, lce_id_width_p, lce_assoc_p, xce_mem);
+  `declare_bp_bedrock_mem_if(paddr_width_p, cce_block_width_p, lce_id_width_p, lce_assoc_p, xce);
```
muwyse, line 43: @dpetrisko was this mismatch in size purposeful? The interface width definitions above used cce_block_width, but the struct declarations used dword_width.

I changed dword_width to cce_block_width so that the width and struct macros were consistent.

bp_top/src/v/bp_clint_slice.v
```diff
@@ -32,14 +32,14 @@ module bp_clint_slice
    , output                                             external_irq_o
    );
 
-`declare_bp_mem_if(paddr_width_p, dword_width_p, lce_id_width_p, lce_assoc_p, xce_mem);
+`declare_bp_bedrock_mem_if(paddr_width_p, dword_width_p, lce_id_width_p, lce_assoc_p, xce);
```
muwyse, line 35: @dpetrisko same question here about data width fields. Although looks like I did not make the change here.

bp_top/src/v/bp_cfg.v
dpetrisko: I can't remember the exact reasoning but it's something along the lines of:
- for partial fill, we support 64-512 bits on the mem port of the XCE
- these devices (clint, cfg) only ever use 64 bits of the data
- For the arbitration in unicore or bp_tile, we slice 64 bits from the mem port of the XCE to match the port width

dpetrisko: Does it cause problems to have dword width instead of cce_block_width?

muwyse: Okay, that was what I was vaguely remembering. Do those extra wires get synthesized away? I see using dword_width for the struct makes the FIFO narrower since we use the $bits() function on the struct, instead of the width_lp from the width parameters macro.

It doesn't seem to cause build issues because of how the data bits field is arranged in the structs as the high-order bits ({data, header}). My guess is the linters might complain.

dpetrisko: They do get synthesized away, but it's probably better practice to be explicit

muwyse: I reverted so this change does not modify the data_width passed to the macro. It can be adjusted or made more explicit in a separate change if desired.

PR #625, opened 2020-09-25
Adding I$ miss metadata to fe-be interface

dpetrisko:
This is an intermediate PR as part of a larger change towards non-speculative I$ misses. The idea here is that we pass the I$ miss metadata to the backend, where it is forwarded back upon commit, as an "exception".

This allows for future expansion of multiple outstanding misses, multithreads, etc.


dpetrisko:
Closing, going in a different direction for non-speculative miss

PR #628, opened 2020-10-05
Rudimentary critical word first support

dpetrisko:
...in the UCE and LCE.  This PR simply raises critical signal when the word is being written in the data_mem_pkt.

(PR is made to FE dev because it is first in a series of related PRs)

### DIFF
--- a/bp_me/src/v/cce/bp_uce.v
+++ b/bp_me/src/v/cce/bp_uce.v
@@ -248,19 +248,19 @@ module bp_uce
 
   // We can do a little better by sending the read_request before the writeback
   enum logic [3:0] {e_reset, e_clear, e_flush_read, e_flush_scan, e_flush_write, e_flush_fence, e_ready, e_send_critical, e_writeback_evict, e_writeback_read_req, e_writeback_write_req, e_write_wait, e_read_req, e_uc_read_wait} state_n, state_r;
-  wire is_reset         = (state_r == e_reset);
-  wire is_clear         = (state_r == e_clear);
-  wire is_flush_read    = (state_r == e_flush_read);
-  wire is_flush_scan    = (state_r == e_flush_scan);
-  wire is_flush_write   = (state_r == e_flush_write);
-  wire is_flush_fence   = (state_r == e_flush_fence);
-  wire is_ready         = (state_r == e_ready);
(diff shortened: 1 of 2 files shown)


muwyse:
Does the UCE only support critical word first, and is that documented somewhere for other users?

Similarly for the LCE, will it only support critical word first or will this be parameterized?

muwyse (approved):
LGTM

See two non-blocking questions in comment of PR.

dpetrisko:
>Does the UCE only support critical word first, and is that documented somewhere for other users?
>Similarly for the LCE, will it only support critical word first or will this be parameterized?

Not sure I understand the question.  Is the alternative not supporting critical word first or supporting some other scheme, like early restart?

muwyse:
Yes, the alternative is supporting multiple data packets but not critical word first, i.e., receiving [a b c d] in four packets and in that order. Adding early restart on top of that is also possible, if, e.g., the critical word was "c" and the cache can send c to the core as soon as it arrives.

I think this is the correct definition: critical word first is when the critical word returns from memory to cache in the first data packet and is sent to the processor immediately. Early restart is whether the cache provides the critical word to the processor so it can resume before the cache has all the data (in particular when the cache block arrives in multiple data packets, but not necessarily critical word first order), as soon as that word arrives at the cache. Critical word first is effectively the most optimized form of early restart.

My question is what all do we support, and what do we want to require from users connecting to our core+cache / UCE / CCE, etc.?

dpetrisko:
Yep, that all sounds correct to me.  Great question.

>My question is what all do we support, and what do we want to require from users connecting to our core+cache / UCE / CCE, etc.?

From interface requirements:
- The cache engine doesn't need to support critical word _first_, just critical word. Meaning that cache_req_critical goes high when the critical word is being written. It could also be an early restart scheme and (interestingly) the cache can't tell the difference.
- The cache itself doesn't need to support critical words at all. For instance, the replay mechanism we had would allow for the cache to simply re-read when cache_req_complete goes high and ignore all other hints.

So more abstractly, cache_req_critical, cache_req_safe and cache_req_complete are all hints. They can be ignored by the consumer (cache), but must be corrected asserted by the producer (cache engine)

PR #636, opened 2020-10-14
Fix Bootrom flow with Checkpointing

sripathi-muralitharan:
This PR - 
1. Fixes the bootrom flow for checkpointing
2. Adds a new CI target to check checkpointing flow

### DIFF
--- a/ci/checkpoint.sh
+++ b/ci/checkpoint.sh
@@ -30,13 +30,13 @@ let JOBS=${#cfgs[@]}
 let CORES_PER_JOB=${N}/${JOBS}+1
 
 # The base command to append the configuration to
-cmd_base="make -j ${CORES_PER_JOB} -C bp_top/syn build.v sim_sample.v SUITE=beebs PROG=aha-compress CHECKPOINT_P=1 SAMPLE_START_P=1000 SAMPLE_MEMSIZE=64"
+cmd_base="make -C bp_top/syn build.v sim_sample.v SUITE=beebs PROG=aha-compress CHECKPOINT_P=1 SAMPLE_START_P=1000 SAMPLE_MEMSIZE=64"
 
 # Any setup needed for the job
 make -C bp_top/syn clean.${SUFFIX}
 
 # Run the regression in parallel on each configuration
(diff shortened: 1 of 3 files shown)


dpetrisko (approved):
Awesome! Thanks for the fix!

ci/checkpoint.sh
```diff
@@ -0,0 +1,44 @@
+#!/bin/bash
```
dpetrisko, line 1: You've got to add a job to the gitlab_ci.yml so that it actually runs

ci/checkpoint.sh
```diff
@@ -0,0 +1,44 @@
...
+    "e_bp_unicore_cfg"
+    "e_bp_multicore_1_cfg"
+    "e_bp_multicore_1_cce_ucode_cfg"
+    )
+
+let JOBS=${#cfgs[@]}
+let CORES_PER_JOB=${N}/${JOBS}+1
+
+# The base command to append the configuration to
+cmd_base="make -j ${CORES_PER_JOB} -C bp_top/syn build.v sim_sample.v SUITE=beebs PROG=aha-compress CHECKPOINT_P=1 SAMPLE_START_P=1000 SAMPLE_MEMSIZE=64"
```
dpetrisko, line 33: This won't work for multithreading since it'll trying sim-ing before build.  I would remove CORES_PER_JOB as a variable

dpetrisko (changes requested):

ci/checkpoint.sh
```diff
@@ -0,0 +1,44 @@
...
+    "e_bp_unicore_cfg"
+    "e_bp_multicore_1_cfg"
+    "e_bp_multicore_1_cce_ucode_cfg"
+    )
+
+let JOBS=${#cfgs[@]}
+let CORES_PER_JOB=${N}/${JOBS}+1
+
+# The base command to append the configuration to
+cmd_base="make -C bp_top/syn build.v sim_sample.v SUITE=beebs PROG=aha-compress CHECKPOINT_P=1 SAMPLE_START_P=1000 SAMPLE_MEMSIZE=64"
```
dpetrisko, line 33: don't think build.v will run verilator :)

Use SUFFIX

PR #638, opened 2020-10-15
override bootrom by default for bp_top sim

muwyse:
Make OVERRIDE_BOOTROM parameter default to 1 so bootroms build for tests in bp_top


dpetrisko:
I think this breaks tire_kick? Maybe just add OVERRIDE_BOOTROM to that target?

muwyse:
> I think this breaks tire_kick? Maybe just add OVERRIDE_BOOTROM=0 to that target?

Ah, okay. Is that because the programs are assumed to not have been built for the tire_kick so it relies on a prebuilt bootrom for hello world?

dpetrisko:
I'll merge once CI passes (which includes a tire_kick target)

PR #649, opened 2020-10-26
Adding writeback packet to memory pipe

dpetrisko:
- Adds writeback packet to memory pipe
- Rolls back on D$ miss, but still dequeues
- Adds rudimentary critical word first and data replication to support this
- Minor change to commit packet to clarify difference between commit and trap

Fixes non-idempotency of uncached loads/atomics

### DIFF
--- a/bp_be/src/v/bp_be_mem/bp_be_dcache/bp_be_dcache.v
+++ b/bp_be/src/v/bp_be_mem/bp_be_dcache/bp_be_dcache.v
@@ -391,10 +391,6 @@ module bp_be_dcache
   wire amomaxu_req = v_tv_r & decode_tv_r.amomaxu_op;
   wire l2_amo_req  = v_tv_r & decode_tv_r.l2_op;
 
-  // Uncached and L2 atomic refactor
-  logic uncached_load_data_v_r;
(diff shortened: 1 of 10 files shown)


bp_be/src/v/bp_be_mem/bp_be_dcache/bp_be_dcache.v
```diff
@@ -909,15 +894,35 @@ module bp_be_dcache
...
+     ,.data_o(data_dm_n)
+     );
+
+  assign dm_we = v_tv_r;
   always_ff @(posedge clk_i) begin
     if (reset_i) begin
       v_dm_r <= '0;
     end else begin
-      v_dm_r <= dm_we & ~flush_i;
+      v_dm_r <= (dm_we & early_v_o & ~flush_i) | cache_req_critical_i;
```
Shashank-Vijay, line 922: Ok, so we're using this same stage for both final data calculation on a normal hit and also for the snooping for data returned on cache misses.

dpetrisko: yep

Shashank-Vijay (approved):
LGTM!

PR #656, opened 2020-10-30
Update dev from top_dev

dpetrisko:
This is a partial port of the wormhole streamer changes, and 
- makes bp_unicore and bp_tile bp_burst
- makes cache slice bp_burst 
- changes burst2lite2burst modules to ready_and_o

No impact on cycles, since it removes the DMA serialization and adds burst_to_lite serialization.  But should make other changes easier


bp_top/src/v/bp_unicore.v
```diff
@@ -11,7 +11,9 @@ module bp_unicore
  import bsg_noc_pkg::*;
  #(parameter bp_params_e bp_params_p = e_bp_default_cfg
    `declare_bp_proc_params(bp_params_p)
-  
+ 
+   , localparam dram_data_width_lp = dword_width_p
```
muwyse, line 15: I think this localparam should be removed. The signals it is used for rely on the width being dword_width_p in other modules, so it shouldn't be exposed as something that could be set to a different value, unless that value is then explicitly passed into those modules (i.e., bp_me_cache_slice or bp_lite_to_burst)

muwyse (approved):
LGTM.

One question on a potentially spurious localparam.

bp_top/src/v/bp_unicore.v
dpetrisko: yeah that's a fair point.  My justification is that we'll want to adjust it in the future, but don't currently support it. The localparam vs param indicates it can't be changed flippantly

muwyse: That is fine, too, but if it is left in place then we need to make sure that any module instantiated here and using signals defined to be dram_data_width_lp in size are passed that parameter explicitly so they all stay consistently sized if we ever change the parameter.

Otherwise, we will end up with a bug like one you just fixed with the memory networks where the flit width wasn't being set properly (i.e., using some global parameter instead of a local flit_width_p)

dpetrisko: good point.  I'll remove it

muwyse (approved):
LGTM

PR #658, opened 2020-11-01
Fixing Weird Cache Configs

sripathi-muralitharan:
This PR adds a fix for the l1_medium and l1_small cache configs that were found to be broken. A new l1_tiny config will follow in a future PR as well. 

The main reason for these configs to break was that we currently strongly require all caches - I$, D$ and A$ (Accelerator cache) to have the same parameters (the blockwidth specifically).

### DIFF
--- a/bp_common/src/include/bp_common_aviary_pkg.vh
+++ b/bp_common/src/include/bp_common_aviary_pkg.vh
@@ -121,14 +121,18 @@ package bp_common_aviary_pkg;
                         );
 
   localparam bp_proc_param_s bp_unicore_l1_medium_override_p =
-    '{icache_sets         : 64
+    '{icache_sets         : 128
       ,icache_assoc       : 4
       ,icache_block_width : 256
       ,icache_fill_width  : 256
-      ,dcache_sets        : 64
+      ,dcache_sets        : 128
       ,dcache_assoc       : 4
       ,dcache_block_width : 256
       ,dcache_fill_width  : 256
+      ,acache_sets        : 128
+      ,acache_assoc       : 4
+      ,acache_block_width : 256
+      ,acache_fill_width  : 256
       ,default : "inv"
       };
   `bp_aviary_derive_cfg(bp_unicore_l1_medium_cfg_p
@@ -137,14 +141,18 @@ package bp_common_aviary_pkg;
                         );
 
   localparam bp_proc_param_s bp_unicore_l1_small_override_p =
-    '{icache_sets         : 64
+    '{icache_sets         : 256
       ,icache_assoc       : 2
       ,icache_block_width : 128
       ,icache_fill_width  : 128
-      ,dcache_sets        : 64
+      ,dcache_sets        : 256
       ,dcache_assoc       : 2
       ,dcache_block_width : 128
       ,dcache_fill_width  : 128
+      ,acache_sets        : 256
+      ,acache_assoc       : 2
+      ,acache_block_width : 128
+      ,acache_fill_width  : 128
       ,default : "inv"
       };
   `bp_aviary_derive_cfg(bp_unicore_l1_small_cfg_p
@@ -184,14 +192,18 @@ package bp_common_aviary_pkg;
                         );
 
   localparam bp_proc_param_s bp_multicore_1_l1_medium_override_p =
-    '{icache_sets         : 64
+    '{icache_sets         : 128
       ,icache_assoc       : 4
       ,icache_block_width : 256
       ,icache_fill_width  : 256
-      ,dcache_sets        : 64
+      ,dcache_sets        : 128
       ,dcache_assoc       : 4
       ,dcache_block_width : 256
       ,dcache_fill_width  : 256
+      ,acache_sets        : 128
+      ,acache_assoc       : 4
+      ,acache_block_width : 256
+      ,acache_fill_width  : 256
       ,default : "inv"
       };
   `bp_aviary_derive_cfg(bp_multicore_1_l1_medium_cfg_p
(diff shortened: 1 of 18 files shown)


sripathi-muralitharan:
Verilator still complains that some code is unsupported in basejump STL. The error I see is - 
```
%Error-UNSUPPORTED: /mnt/users/ssd1/homes/sripathi/BlackParrot/black-parrot/external/basejump_stl/testing/bsg_dmc/lpddr_verilog_model/mobile_ddr.v:583:9: Unsupported: timing control statement in this location
                                                                                                                                                        : ... Suggest have one timing control statement per procedure, at the top of the procedure
  583 |         @ (posedge diff_ck) begin
      |         ^
```
A detailed log might even show up on the CI reports

dpetrisko:
>%Error-UNSUPPORTED: /mnt/users/ssd1/homes/sripathi/BlackParrot/black-parrot/external/basejump_stl/testing/bsg_dmc/lpddr_verilog_model/mobile_ddr.v:583:9: Unsupported: timing control statement in this location

During what command? I can execute build.sc fine

dpetrisko:
Can we also get a bp_unicore_l1_hetero_override_p with I$=8k, D$=16k? (+test)

dpetrisko:
>The main reason for these configs to break was that we currently strongly require all caches - I$, D$ and A$ (Accelerator cache) to have the same parameters (the blockwidth specifically).

This is only true for the multicore configs, we support different blockwidths for all caches in unicore

sripathi-muralitharan:
> During what command? I can execute build.sc fine
It was during the build.sc. But I guess it's some issue I have locally because CI (tirekick) went fine.

bp_top/src/v/bp_tile.v
```diff
@@ -95,24 +97,48 @@ bp_bedrock_cce_mem_msg_s       cce_mem_resp_li;
...
+bp_bedrock_xce_mem_msg_s       loopback_mem_cmd;
 logic                  loopback_mem_cmd_v_li, loopback_mem_cmd_ready_lo;
 bp_bedrock_cce_mem_msg_s       loopback_mem_resp_lo;
+bp_bedrock_xce_mem_msg_s       loopback_mem_resp;
 logic                  loopback_mem_resp_v_lo, loopback_mem_resp_yumi_li;
+assign loopback_mem_cmd = '{header: loopback_mem_cmd_li.header
+                           ,data: loopback_mem_cmd_li.data[0+:dword_width_p]
+                           };
+assign loopback_mem_resp_lo = '{header: loopback_mem_resp.header
+                               ,data: {'0, loopback_mem_resp.data}
```
dpetrisko, line 109: This should be replicated, not zero

bp_common/src/include/bp_common_aviary_pkg.vh
```diff
@@ -152,6 +152,22 @@ package bp_common_aviary_pkg;
                         ,bp_unicore_cfg_p
                         );
 
+  localparam bp_proc_param_s bp_unicore_l1_hetero_override_p =
```
dpetrisko, line 155: We also need a config for a wide line config e.g. 512 bit line on a 16kB cache

bp_fe/src/v/bp_fe_icache.v
```diff
@@ -333,15 +332,19 @@ module bp_fe_icache
   assign tag_mem_pkt = tag_mem_pkt_i;
   bp_icache_stat_mem_pkt_s stat_mem_pkt;
   assign stat_mem_pkt = stat_mem_pkt_i;
-
+  
+  // Weird Calculation to find correct request size instead of adding another
+  // mux
+  bp_cache_req_size_e req_size = bp_cache_req_size_e'(`BSG_SAFE_CLOG2(icache_block_width_p) - 3);
```
dpetrisko, line 338: Will not synthesize into ASIC!!!!! This is an initialization statement.

Need to use ```localparam bp_cache_req_size_e req_size = ```

Also, this is a constant, so using the "mux" is better since there's no actual overhead.  On the extremely unlikely event that we change the enum values, this will break.

bp_me/src/v/cache/bp_me_cce_to_cache.v
```diff
@@ -276,11 +276,33 @@ module bp_me_cce_to_cache
      ,.data_i(mem_cmd_lo.header)
      ,.data_o(mem_resp_cast_o.header)
      );
+  
+  localparam num_bytes_lp = cce_block_width_p/8;
```
dpetrisko, line 280: This seems unnecessary, but I'll take a closer look

bp_top/src/v/bp_tile.v
sripathi-muralitharan: I fixed this and everywhere else this was used (clint, cfg, etc. and unicore/tile)

bp_common/src/include/bp_common_aviary_pkg.vh
sripathi-muralitharan: Added a config l1_wide where there are 2 16KB caches with 512-bit block widths.

bp_fe/src/v/bp_fe_icache.v
sripathi-muralitharan: Updated this. Thanks for letting me know about this synthesis bug (I guess this is what you mentioned in the meeting as well).

dpetrisko: Thanks! I added a check in DC so that the check_design target will now catch this

dpetrisko (approved):
LGTM! Just waiting on CI

PR #659, opened 2020-11-01
Adding bp_finish at the end of test

sammy17:
The test is hanging at the end of runtime due to missing   bp_finish(0);. Adding this will make it terminate correctly.


dpetrisko:
LGTM

PR #660, opened 2020-11-03
Non-replay D$

dpetrisko:
This PR adds D$ operation to the scoreboard, so that cache misses do not block the pipeline. This is nice for performance, but necessary for non-idempotent remote atomic or uncached operations

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -1,5 +1,5 @@
 variables:
-  GIT_STRATEGY: fetch
+  GIT_STRATEGY: clone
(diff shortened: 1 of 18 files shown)


dpetrisko:
Outstanding problems include:
- need a "true miss" output on D$, to distinguish between D$ failure and a miss which was actually sent out
- write buffer arbitration (need to write store miss data after miss resolves, but other data before miss resolves)

Moonshot is to tie to #724, probably another PR

dpetrisko:
Depends on #845

PR #663, opened 2020-11-06
Adding full tag info to the cache service interface

dpetrisko:
- Adds tag info to the cache service struct
- Updates this struct in cache engines

Not used for anything right now, but enables extra state (coherence or otherwise) to be stored in the tag, useful for certain classes of cache engine. For instance, a system where true tags are stored in L1 rather than shadow tags.


muwyse (approved):
LGTM

PR #665, opened 2020-11-07
Fixing up synchronization sequence in LCEs

dpetrisko:
This patch lets the LCE synchronization sequence guarantee to complete before sending out cached requests on the network.  It also reorders the bootrom to support this.

There's also an unrelated change to custom configs, allowing for a user-defined base config, that I'm sneaking in here.

This is tested up to 16 cores


muwyse (approved):
LGTM.

Is the bootrom sequence documented anywhere? We should add a note somewhere saying that CCE mode should be set before cache modes to ensure proper operation.

dpetrisko:
No, not really. I’ll add comments to bootrom here and an issue to more document it. The sequence only matters during this bootstrapping, during the default NBF load it doesn’t matter at all

PR #669, opened 2020-11-08
Tiny Instruction Cache

sripathi-muralitharan:
This PR adds support for a tiny cache - 4KB and a corresponding tiny config (reflects only the tiny icache changes) that is supported only in the UCE. 

A similar change for the D$ will soon follow in a future PR.

### DIFF
--- a/bp_common/src/include/bp_common_cache_engine_if.svh
+++ b/bp_common/src/include/bp_common_cache_engine_if.svh
@@ -156,12 +156,12 @@ typedef enum logic [1:0] {
 
 `define declare_bp_cache_stat_info_s(ways_mp, cache_name_mp)  \
   typedef struct packed {                 \
-    logic [ways_mp-2:0] lru;              \
+    logic [`BSG_MAX(ways_mp-2, 0):0] lru; \
     logic [ways_mp-1:0] dirty;            \
   } bp_``cache_name_mp``_stat_info_s
 
 `define bp_cache_stat_info_width(ways_mp) \
-  (2*ways_mp-1)
+  (`BSG_MAX(2, (2*ways_mp-1)))
 
 `define declare_bp_cache_engine_if(addr_width_mp, tag_width_mp, sets_mp, ways_mp, req_data_width_mp, block_data_width_mp, fill_width_mp, cache_name_mp) \
   `declare_bp_cache_req_s(req_data_width_mp, addr_width_mp, cache_name_mp);                              \
--- a/bp_common/src/include/bp_common_cache_engine_if.svh
+++ b/bp_common/src/include/bp_common_cache_engine_if.svh
@@ -156,12 +156,13 @@ typedef enum logic [1:0] {
 
 `define declare_bp_cache_stat_info_s(ways_mp, cache_name_mp)  \
   typedef struct packed {                 \
-    logic [`BSG_MAX(ways_mp-2, 0):0] lru; \
+    logic [`BSG_SAFE_MINUS(ways_mp, 2):0] lru; \
     logic [ways_mp-1:0] dirty;            \
   } bp_``cache_name_mp``_stat_info_s
 
+// Direct mapped caches need 2-bits in the stat info
 `define bp_cache_stat_info_width(ways_mp) \
-  (`BSG_MAX(2, (2*ways_mp-1)))
+  (`BSG_MAX(2,2*ways_mp-1))
 
 `define declare_bp_cache_engine_if(addr_width_mp, tag_width_mp, sets_mp, ways_mp, req_data_width_mp, block_data_width_mp, fill_width_mp, cache_name_mp) \
   `declare_bp_cache_req_s(req_data_width_mp, addr_width_mp, cache_name_mp);                              \
--- a/bp_fe/src/v/bp_fe_icache.sv
+++ b/bp_fe/src/v/bp_fe_icache.sv
@@ -38,9 +38,9 @@ module bp_fe_icache
     , localparam num_words_per_bank_lp = bank_width_lp / word_width_p
     , localparam data_mem_mask_width_lp=(bank_width_lp >> 3)
     , localparam byte_offset_width_lp=`BSG_SAFE_CLOG2(bank_width_lp >> 3)
-    , localparam bank_offset_width_lp = $clog2(icache_assoc_p)
+    , localparam bank_offset_width_lp = `BSG_SAFE_CLOG2(icache_assoc_p)
     , localparam index_width_lp=`BSG_SAFE_CLOG2(icache_sets_p)
-    , localparam block_offset_width_lp = (bank_offset_width_lp+byte_offset_width_lp)
+    , localparam block_offset_width_lp = (icache_assoc_p > 1) ? (bank_offset_width_lp+byte_offset_width_lp) : byte_offset_width_lp
     , localparam block_size_in_fill_lp = icache_block_width_p / icache_fill_width_p
     , localparam fill_size_in_bank_lp = icache_fill_width_p / bank_width_lp
     )
@@ -102,9 +102,9 @@ module bp_fe_icache
   bp_fe_icache_pkt_s icache_pkt;
   assign icache_pkt = icache_pkt_i;
 
-  logic [vtag_width_p-1:0]                               vaddr_vtag;
-  logic [index_width_lp-1:0]                             vaddr_index;
-  logic [`BSG_SAFE_MINUS(bank_offset_width_lp,1):0]      vaddr_offset;
+  logic [vtag_width_p-1:0]              vaddr_vtag;
+  logic [index_width_lp-1:0]            vaddr_index;
+  logic [bank_offset_width_lp-1:0]      vaddr_offset;
 
   logic [icache_assoc_p-1:0]            way_v_tv_r; // valid bits of each way
   logic [lg_icache_assoc_lp-1:0]        way_invalid_index; // first invalid way
@@ -114,7 +114,7 @@ module bp_fe_icache
   logic fencei_req;
 
   assign vaddr_index      = icache_pkt.vaddr[block_offset_width_lp+:index_width_lp];
-  assign vaddr_offset     = (icache_assoc_p > 1) ? icache_pkt.vaddr[byte_offset_width_lp+:`BSG_MAX(bank_offset_width_lp,1)] : '0;
+  assign vaddr_offset     = icache_pkt.vaddr[byte_offset_width_lp+:bank_offset_width_lp];
   assign vaddr_vtag       = icache_pkt.vaddr[block_offset_width_lp+index_width_lp+:vtag_width_p];
 
   // TL stage
@@ -173,9 +173,10 @@ module bp_fe_icache
   end
 
   // data memory
+  localparam data_mem_addr_width_lp = (icache_assoc_p > 1) ? (index_width_lp+bank_offset_width_lp) : index_width_lp;
   logic [icache_assoc_p-1:0]                                           data_mem_v_li;
   logic                                                                data_mem_w_li;
-  logic [icache_assoc_p-1:0][index_width_lp+bank_offset_width_lp-1:0]  data_mem_addr_li;
+  logic [icache_assoc_p-1:0][data_mem_addr_width_lp-1:0]               data_mem_addr_li;
   logic [icache_assoc_p-1:0][bank_width_lp-1:0]                        data_mem_data_li;
   logic [icache_assoc_p-1:0][data_mem_mask_width_lp-1:0]               data_mem_w_mask_li;
   logic [icache_assoc_p-1:0][bank_width_lp-1:0]                        data_mem_data_lo;
@@ -198,18 +199,18 @@ module bp_fe_icache
     );
   end
 
-  logic [ptag_width_p-1:0]                          addr_tag_tl;
-  logic [`BSG_SAFE_MINUS(bank_offset_width_lp,1):0] addr_bank_offset_tl;
-  logic [icache_assoc_p-1:0]                        addr_bank_offset_dec_tl;
-  logic [icache_assoc_p-1:0]                        hit_v_tl;
-  logic [paddr_width_p-1:0]                         addr_tl;
-  logic [icache_assoc_p-1:0]                        way_v_tl;
-  logic [index_width_lp-1:0]                        vaddr_index_tl;
-  logic [vtag_width_p-1:0]                          vaddr_vtag_tl;
+  logic [ptag_width_p-1:0]         addr_tag_tl;
+  logic [bank_offset_width_lp-1:0] addr_bank_offset_tl;
+  logic [icache_assoc_p-1:0]       addr_bank_offset_dec_tl;
+  logic [icache_assoc_p-1:0]       hit_v_tl;
+  logic [paddr_width_p-1:0]        addr_tl;
+  logic [icache_assoc_p-1:0]       way_v_tl;
+  logic [index_width_lp-1:0]       vaddr_index_tl;
+  logic [vtag_width_p-1:0]         vaddr_vtag_tl;
    
   assign addr_tl = {ptag_i, vaddr_tl_r[0+:bp_page_offset_width_gp]};
   assign addr_tag_tl = addr_tl[block_offset_width_lp+index_width_lp+:ptag_width_p];
-  assign addr_bank_offset_tl = (icache_assoc_p > 1) ? addr_tl[byte_offset_width_lp+:`BSG_MAX(bank_offset_width_lp,1)] : '0;
+  assign addr_bank_offset_tl = addr_tl[byte_offset_width_lp+:bank_offset_width_lp];
 
   assign vaddr_index_tl = vaddr_tl_r[block_offset_width_lp+:index_width_lp];
   assign vaddr_vtag_tl = vaddr_tl_r[block_offset_width_lp+index_width_lp+:vtag_width_p];
@@ -289,10 +290,8 @@ module bp_fe_icache
   bp_icache_stat_info_s          stat_mem_mask_li;
   bp_icache_stat_info_s          stat_mem_data_lo;
 
-  logic [lg_icache_assoc_lp-1:0] lru_encode;
-
   bsg_mem_1rw_sync_mask_write_bit #(
-    .width_p(`BSG_MAX(2, icache_assoc_p)-1)
+    .width_p(icache_assoc_p-1)
     ,.els_p(icache_sets_p)
   ) stat_mem (
     .clk_i(clk_i)
@@ -304,6 +303,8 @@ module bp_fe_icache
     ,.w_i(stat_mem_w_li)
     ,.data_o(stat_mem_data_lo.lru)
   );
+
+  logic [lg_icache_assoc_lp-1:0] lru_encode;
 
   bsg_lru_pseudo_tree_encode #(
     .ways_p(icache_assoc_p)
@@ -372,12 +373,7 @@ module bp_fe_icache
      );
 
   // invalid way takes priority over LRU way
-  // For direct-mapped cache, we default to 0 since there is no concept of way
-  assign cache_req_metadata_cast_lo.repl_way = (icache_assoc_p > 1)
-                                                 ? invalid_exist
-                                                   ? way_invalid_index
-                                                   : lru_encode
-                                                 : '0;
+  assign cache_req_metadata_cast_lo.repl_way = invalid_exist ? way_invalid_index : lru_encode;
   assign cache_req_metadata_cast_lo.dirty = '0;
 
   // Cache Miss Tracker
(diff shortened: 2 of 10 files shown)


taylor-bsg:
Nice work!

This is a lot of extra code to support a corner case.  Lot of it has to do with signal width underflow kind of conditions. Unfortunately, this means that meta programming is only partially working, and our code gets multiplied up exponentially with all of these cases.

My suggestion is that as much as possible you should try to push the underflow down into the leaf modules, so that they are more robust in general to this sort of thing, for example the LRU module or the stat ram. Then when we write other code that has underflow, it will just work out of the box. A simple example of this is the variable #input mux module, and making it handle the case of only one incoming item. In general we have created the macro BSG_SAFE_CLOG2 for underflow cases where only one item is there, and the CLOG2 function returns 0, results in a zero width signal. In this case we may need something like BSG_SAFE_MINUS.

M

dpetrisko:
^ took the words right out of my mouth

sripathi-muralitharan:
Opened a PR in Basejump-STL with this macro. PR 352 - https://github.com/bespoke-silicon-group/basejump_stl/pull/352. I hope this is sufficient to solve the meta programming problem here. 

I haven't made changes to the code in BP yet (Waiting for the one in Basejump-STL to be merged in)

dpetrisko:
You should make the changes, point the submodule to your basejump stl branch, and then write “depends on #xyz” as a comment here. That’s the best way to track these things

bp_fe/src/v/bp_fe_icache.v
```diff
@@ -95,7 +98,7 @@ module bp_fe_icache
   bp_cfg_bus_s cfg_bus_cast_i;
   assign cfg_bus_cast_i = cfg_bus_i;
 
-  `declare_bp_cache_service_if(paddr_width_p, ptag_width_p, icache_sets_p, icache_assoc_p, dword_width_p, icache_block_width_p, icache_fill_width_p, icache);
+  `declare_bp_cache_service_if(paddr_width_p, ptag_width_p, icache_sets_p, `BSG_MAX(2, icache_assoc_p), dword_width_p, icache_block_width_p, icache_fill_width_p, icache);
```
dpetrisko, line 101: do the max(2) in the macro

bp_fe/src/v/bp_fe_icache.v
```diff
@@ -29,7 +29,7 @@ module bp_fe_icache
   import bp_fe_icache_pkg::*;
   #(parameter bp_params_e bp_params_p = e_bp_default_cfg
     `declare_bp_proc_params(bp_params_p)
-    `declare_bp_cache_service_if_widths(paddr_width_p, ptag_width_p, icache_sets_p, icache_assoc_p, dword_width_p, icache_block_width_p, icache_fill_width_p, icache)
+    `declare_bp_cache_service_if_widths(paddr_width_p, ptag_width_p, icache_sets_p, `BSG_MAX(2, icache_assoc_p), dword_width_p, icache_block_width_p, icache_fill_width_p, icache)
```
dpetrisko, line 32: max(2) in macro

bp_fe/src/v/bp_fe_icache.v
```diff
@@ -38,9 +38,12 @@ module bp_fe_icache
     , localparam num_words_per_bank_lp = bank_width_lp / word_width_p
     , localparam data_mem_mask_width_lp=(bank_width_lp >> 3)
     , localparam byte_offset_width_lp=`BSG_SAFE_CLOG2(bank_width_lp >> 3)
+    // Two kinds of bank offsets to handle indexing and memory instantion respectively for a direct-mapped cache. 
+    // They boil down to the same thing for a set associative cache
```
dpetrisko, line 42: This comment is confusing -- instead, I would draw an ascii diagram

bp_fe/src/v/bp_fe_icache.v
```diff
@@ -38,9 +38,12 @@ module bp_fe_icache
...
     , localparam bank_offset_width_lp = `BSG_SAFE_CLOG2(icache_assoc_p)
+    , localparam mem_bank_offset_width_lp = $clog2(icache_assoc_p)
```
dpetrisko, line 44: having both bank_offset_width and mem_bank_offset_width is very confusing.  Why not just set bank_offset_width correctly?

bp_fe/src/v/bp_fe_icache.v
```diff
@@ -306,10 +311,10 @@ module bp_fe_icache
     ,.data_o(stat_mem_data_lo.lru)
   );
 
-  logic [lg_icache_assoc_lp-1:0] lru_encode;
-
+  // For direct-mapped the LRU bits are unused. Constant propagation will
```
dpetrisko, line 314: These comments about direct-mapped cache are unnecessary.  It's a very niche configuration

bp_fe/src/v/bp_fe_icache.v
```diff
@@ -292,8 +295,10 @@ module bp_fe_icache
   bp_icache_stat_info_s          stat_mem_mask_li;
   bp_icache_stat_info_s          stat_mem_data_lo;
 
+  logic [lg_icache_assoc_lp-1:0] lru_encode;
+
   bsg_mem_1rw_sync_mask_write_bit #(
```
dpetrisko, line 300: @taylor-bsg Crazy idea, but does it make sense to support 0-width memories as well?

dpetrisko:
@taylor-bsg .  Most of the other corner cases are related to 0-width concatenations e.g. {index, offset}, where offset is 0 bits. I can't think of a good way to get around that

taylor-bsg:
Yeah I suggested to Sripathi that we support 0-width memories rather than a
separate generate statement.

BSG_SAFE_CONCAT ?



M

dpetrisko:
> BSG_SAFE_CONCAT

Hm, we need to signal to the macro if the field is actually a 0-width signal disguised as a 1-width signal `BSG_SAFE_CONCAT(index, offset) makes it ambiguous if offset is actually 0 or 1 bits.

`BSG_SAFE_CONCAT(index, offset, $clog2(offset)) would solve this problem but is a little messy.

bp_fe/src/v/bp_fe_icache.v
sripathi-muralitharan: That's because when defining widths, I wanted them to be one so that I don't get bit vectors with negative indices. But for places where I want to index using the +:, I wanted a zero to be correct.

sripathi-muralitharan:
For BSG_SAFE_CONCAT, is it okay to enforce ordering in the parameters? Like the first parameter has to be something non-zero and the second one could be 0 or non-zero?

bp_fe/src/v/bp_fe_icache.v
sripathi-muralitharan: Got rid of one of them and reverted back to the previous state of only 1 offset parameter by leveraging BSG_SAFE_MINUS

sripathi-muralitharan:
Managed to solve the other 0-width concatenations issue by using a conditional concatenation feature in Verilog. Ready for review again.

bp_me/src/v/cce/bp_uce.v
```diff
@@ -285,18 +285,24 @@ module bp_uce
...
 
+  // For direct-mapped caches, the fill width should be equal to block width
+  // since bank width = block width and fill width can't be less than bank
+  // width
   if (fill_size_in_bank_lp == 1)
     begin
       assign bank_index =  mem_cmd_cnt;
-      assign fill_index_shift = mem_resp_cast_i.header.addr[byte_offset_width_lp+:bank_offset_width_lp];
+      // Setting fill_index_shift to zero for direct-mapped caches since there aren't any opportunities for iterative filling as well as no data memory banks
+      assign fill_index_shift = (assoc_p > 1)
```
dpetrisko, line 298: I think you can use the zero width concatentation here

bp_me/src/v/cce/bp_uce.v
```diff
@@ -285,18 +285,24 @@ module bp_uce
...
+      // Setting fill_index_shift to zero for direct-mapped caches since there aren't any opportunities for iterative filling as well as no data memory banks
+      assign fill_index_shift = (assoc_p > 1) 
+                                  ? mem_resp_cast_i.header.addr[byte_offset_width_lp+:`BSG_MAX(bank_offset_width_lp,1)]
+                                  : '0;
     end
   else
     begin
       assign bank_index = mem_cmd_cnt << bank_sub_offset_width_lp;
-      assign fill_index_shift = mem_resp_cast_i.header.addr[byte_offset_width_lp+:bank_offset_width_lp] >> bank_sub_offset_width_lp;
+      assign fill_index_shift = mem_resp_cast_i.header.addr[byte_offset_width_lp+:`BSG_MAX(bank_offset_width_lp,1)] >> bank_sub_offset_width_lp;
```
dpetrisko, line 305: Do you need the max here? If fill_size_in_bank == 1 for direct mapped, then this clause will never get called

bp_me/src/v/cce/bp_uce.v
```diff
@@ -285,18 +285,24 @@ module bp_uce
   // to write back dirty data to L2 in the size of fill_width.
   logic [fill_cnt_width_lp-1:0] mem_cmd_cnt;
   logic [block_size_in_fill_lp-1:0] fill_index_shift;
-  logic [bank_offset_width_lp-1:0] bank_index;
+  logic [`BSG_SAFE_MINUS(bank_offset_width_lp,1):0] bank_index;
   logic [paddr_width_p-1:0] critical_addr;
 
+  // For direct-mapped caches, the fill width should be equal to block width
```
dpetrisko, line 291: Is bank_width == block_width? I thought there was just an inequality there

bp_me/src/v/cce/bp_uce.v
```diff
@@ -23,7 +23,7 @@ module bp_uce
     , localparam num_dwords_per_bank_lp = bank_width_lp / dword_width_p
     , localparam byte_offset_width_lp  = `BSG_SAFE_CLOG2(bank_width_lp>>3)
     // Words per line == associativity
-    , localparam bank_offset_width_lp  = `BSG_SAFE_CLOG2(assoc_p)
+    , localparam bank_offset_width_lp  = $clog2(assoc_p)
```
dpetrisko, line 26: If you keep this as SAFE_CLOG2 and use zero width concatentation, I believe that will get rid of all the edge cases below

bp_fe/src/v/bp_fe_top.v
```diff
@@ -20,10 +20,10 @@ module bp_fe_top
    , localparam num_dwords_per_bank_lp = bank_width_lp / dword_width_p
```
dpetrisko, line 20: I don't think any of these parameters are used?

bp_fe/src/v/bp_fe_icache.v
```diff
@@ -624,8 +638,8 @@ module bp_fe_icache
       stat_mem_data_li.lru = lru_decode_data_lo;
       stat_mem_mask_li.lru = lru_decode_mask_lo;
     end else begin
-      stat_mem_data_li.lru = {(icache_assoc_p-1){1'b0}};
-      stat_mem_mask_li.lru = {(icache_assoc_p-1){1'b1}};
+      stat_mem_data_li.lru = {(`BSG_MAX(2, icache_assoc_p)-1){1'b0}};
```
dpetrisko, line 641: can just use '0 and '1

bp_fe/src/v/bp_fe_icache.v
```diff
@@ -597,8 +611,8 @@ module bp_fe_icache
     ? addr_index_tv
     : stat_mem_pkt.index;
 
-  logic [icache_assoc_p-2:0] lru_decode_data_lo;
-  logic [icache_assoc_p-2:0] lru_decode_mask_lo;
+  logic [`BSG_MAX(2, icache_assoc_p)-2:0] lru_decode_data_lo;
```
dpetrisko, line 614: bsg_safe_minus?

bp_fe/src/v/bp_fe_icache.v
```diff
@@ -494,11 +499,20 @@ module bp_fe_icache
...
-    assign data_mem_addr_li[i] = tl_we
-      ? {vaddr_index, vaddr_offset}
-      : {data_mem_pkt.index, data_mem_pkt_offset};
+        assign data_mem_addr_li[i] = tl_we
+          ? {vaddr_index, vaddr_offset}
+          : {data_mem_pkt.index, data_mem_pkt_offset};
+      end
+    else
+      begin: dmap
+        assign data_mem_addr_li[i] = tl_we
```
dpetrisko, line 512: zero width concat

bp_fe/src/v/bp_fe_icache.v
```diff
@@ -372,7 +372,12 @@ module bp_fe_icache
      );
 
   // invalid way takes priority over LRU way
-  assign cache_req_metadata_cast_lo.repl_way = invalid_exist ? way_invalid_index : lru_encode;
+  // For direct-mapped cache, we default to 0 since there is no concept of way
+  assign cache_req_metadata_cast_lo.repl_way = (icache_assoc_p > 1)
```
dpetrisko, line 376: goes away with zero width memory

bp_common/src/include/bp_common_cache_engine_if.vh
```diff
@@ -156,12 +156,12 @@ typedef enum logic [1:0] {
 
 `define declare_bp_cache_stat_info_s(ways_mp, cache_name_mp)  \
   typedef struct packed {                 \
-    logic [ways_mp-2:0] lru;              \
+    logic [`BSG_MAX(ways_mp-2, 0):0] lru; \
```
dpetrisko, line 159: safe minus

bp_me/src/v/cce/bp_uce.v
sripathi-muralitharan: The general inequality is bank_width <= block_width. For the direct-mapped case, it becomes bank_width = block_width

bp_fe/src/v/bp_fe_top.v
sripathi-muralitharan: Yes, there are some redundant parameters. I can get rid of them.

bp_fe/src/v/bp_fe_icache.v
sripathi-muralitharan: Not so sure about this. Seems to me that the priority encoder from where the `invalid_exist` and `way_invalid_index` are generated end up throwing out X's when width_p = 1. Seems to me that these corner conditions haven't been handled fully/correctly. This means, that will end up putting an X on the metadata way. I am guessing it may not become a problem since its all direct mapped and there is no concept of a way, but I am worried about the X-propagation that might cause issues down the line? What do you think about this?

bp_me/src/v/cce/bp_uce.v
sripathi-muralitharan: I believe that will not calculate some of the other parameters correctly. So, I think its best to keep it the way it is. But exactly which edge cases are you talking about? The ones with the indexed-part selects (eg. line 305 in `bp_uce.v`)?

bp_me/src/v/cce/bp_uce.v
sripathi-muralitharan: Yup. Fixed that in the latest commit.

bp_fe/src/v/bp_fe_icache.v
dpetrisko: that means those leaf modules do not handle zero width signals correctly and should be fixed

bp_common/src/include/bp_common_cache_engine_if.svh
```diff
@@ -156,12 +156,12 @@ typedef enum logic [1:0] {
...
 `define declare_bp_cache_stat_info_s(ways_mp, cache_name_mp)  \
   typedef struct packed {                 \
-    logic [ways_mp-2:0] lru;              \
+    logic [`BSG_SAFE_MINUS(ways_mp, 2):0] lru; \
     logic [ways_mp-1:0] dirty;            \
   } bp_``cache_name_mp``_stat_info_s
 
 `define bp_cache_stat_info_width(ways_mp) \
-  (2*ways_mp-1)
+  (`BSG_MAX(2, (2*ways_mp-1)))
```
dpetrisko, line 164: nit, but do (safe_minus() + ways_p) to be explicit

bp_me/src/v/cce/bp_uce.v
dpetrisko: pretty much everywhere bank_offset_width_lp is used below, it is with BSG_SAFE_MINUS or BSG_MAX.  If it's set to 1 up here, then those will go away.  

Better to adjust in the parameter calculation than in the datapath, for clarity

bp_me/src/v/cce/bp_uce.v
dpetrisko: This comment is kind of out of place, doesn't really say much about the below code.  I'd combine it with the below comment or just get rid of it

bp_fe/src/v/bp_fe_icache.v
sripathi-muralitharan: Fixed in basejump and opened a PR - https://github.com/bespoke-silicon-group/basejump_stl/pull/357

bp_common/src/include/bp_common_cache_engine_if.svh
sripathi-muralitharan: Tried to do it the way that was suggested, but doing a safe_minus returns a 0 which is not very useful for width calculations. To make things more explicit, I added in a ?: to switch between ways = 1 and ways != 1.

dpetrisko: how it was originally was fine

bp_me/src/v/cce/bp_uce.sv
```diff
@@ -291,7 +291,8 @@ module bp_uce
   if (fill_size_in_bank_lp == 1)
     begin
       assign bank_index =  mem_cmd_cnt;
-      assign fill_index_shift = mem_resp_cast_i.header.addr[byte_offset_width_lp+:bank_offset_width_lp];
+      // Setting fill_index_shift to zero for direct-mapped caches since there aren't any opportunities for iterative filling as well as no data memory banks
```
dpetrisko, line 294: // fill_index_shift corresponds to the single bank in a direct-mapped cache

bp_me/src/v/cce/bp_uce.sv
```diff
@@ -24,7 +24,7 @@ module bp_uce
     , localparam byte_offset_width_lp  = `BSG_SAFE_CLOG2(bank_width_lp>>3)
```
dpetrisko, line 24: Nice! I think this module is minimally modified

bp_fe/src/v/bp_fe_icache.sv
```diff
@@ -688,7 +697,7 @@ module bp_fe_icache
 
   // LCE: stat_mem
   // Stub out dirty bits in icache
-  assign stat_mem_o = {stat_mem_data_lo.lru, icache_assoc_p'(0)};
+  assign stat_mem_o = (icache_assoc_p > 1) ? {stat_mem_data_lo.lru, icache_assoc_p'(0)} : '0;
```
dpetrisko, line 700: don't need

bp_fe/src/v/bp_fe_icache.sv
```diff
@@ -494,11 +503,11 @@ module bp_fe_icache
     & (data_mem_pkt.opcode == e_cache_data_mem_write);
 
   for (genvar i = 0; i < icache_assoc_p; i++) begin : rof1
-    wire [bank_offset_width_lp-1:0] data_mem_pkt_offset = (bank_offset_width_lp'(i) - data_mem_pkt.way_id);
+    wire [`BSG_SAFE_MINUS(bank_offset_width_lp, 1):0] data_mem_pkt_offset = (`BSG_MAX(1, bank_offset_width_lp)'(i) - data_mem_pkt.way_id);
```
dpetrisko, line 506: I don't think you actually need this: a [-1:0] slice is perfectly legal, and with get removed below

bp_fe/src/v/bp_fe_icache.sv
```diff
@@ -372,7 +376,12 @@ module bp_fe_icache
      );
 
   // invalid way takes priority over LRU way
-  assign cache_req_metadata_cast_lo.repl_way = invalid_exist ? way_invalid_index : lru_encode;
+  // For direct-mapped cache, we default to 0 since there is no concept of way
+  assign cache_req_metadata_cast_lo.repl_way = (icache_assoc_p > 1)
```
dpetrisko, line 380: don't need conditional

bp_fe/src/v/bp_fe_icache.sv
```diff
@@ -114,7 +114,11 @@ module bp_fe_icache
   logic fencei_req;
 
   assign vaddr_index      = icache_pkt.vaddr[block_offset_width_lp+:index_width_lp];
-  assign vaddr_offset     = icache_pkt.vaddr[byte_offset_width_lp+:bank_offset_width_lp];
+  // A BSG_MAX is required for the indexed part-select here even though in the
+  // direct-mapped case this clause is not invoked, because this is a ternary
+  // operator-based generate which is not treated the same as a regular if
+  // generate by the simulator which ends up throwing a syntax error. 
+  assign vaddr_offset     = (icache_assoc_p > 1) ? icache_pkt.vaddr[byte_offset_width_lp+:`BSG_MAX(bank_offset_width_lp,1)] : '0;
```
dpetrisko, line 121: I also don't think you need the clause here since the signal is unused in direct mapped

bp_common/src/include/bp_common_cache_engine_if.svh
sripathi-muralitharan: Added in a comment to remove future confusions and reverted back to the older version.

sripathi-muralitharan:
In addition to the suggested comments, I also ended up removing a lot of the earlier changes from the I$ code. I think right now it is at the most minimally changed version.

bp_fe/src/v/bp_fe_icache.sv
```diff
@@ -173,9 +173,10 @@ module bp_fe_icache
   end
 
   // data memory
+  localparam data_mem_addr_width_lp = (icache_assoc_p > 1) ? (index_width_lp+bank_offset_width_lp) : index_width_lp;
```
dpetrisko, line 176: nice!

bp_me/src/v/cce/bp_uce.sv
```diff
@@ -291,7 +291,8 @@ module bp_uce
   if (fill_size_in_bank_lp == 1)
```
dpetrisko, line 291: Technically unrelated change, but I think it fits here. Can you change bank_sub_offset_width_lp to $clog2 instead of SAFE_CLOG2 and get rid of this if statement?

dpetrisko (changes requested):
LGTM other than the last change requested and basejump going back to master once all PRs are merged.  Good work!

bp_me/src/v/cce/bp_uce.sv
sripathi-muralitharan: Done.

bp_fe/src/v/bp_fe_icache.sv
sripathi-muralitharan: Thanks!

dpetrisko (approved):

PR #670, opened 2020-11-09
Minor renames for interfaces

dpetrisko:
Big LOC change, but all clarity, no functionality 

- Cache Service Interface -> Cache Engine Interface
- Adds cache_req_credits_empty / cache_req_credits_full to Cache Engine Interface
- Adds back bp_core_minimal, which turns out to be a useful wrapper
- Some misc cleanup, deleting unused mocks

PR #672, opened 2020-11-10
Bump dev to me_dev

dpetrisko:
* Adding credits signals to cache_req interface

* Renaming cache service interface and core if

* Update interface specifications

* Fixing up coherent accelerator for interface changes

* Fixing up cache wrappers for new interface


dpetrisko:
Essentially same as #670 , just waiting on CI

PR #673, opened 2020-11-10
Add CCE PMA to enable uncached access to cacheable memory

muwyse:
This change adds a PMA module to the CCE and enables uncached access to cacheable DRAM memory. The CCE PMA considers DRAM memory cacheable, and all other memory cacheable. This feature allows the CCE to keep all of memory coherent when both cached and uncached accesses occur to DRAM memory. Thus the following accesses are now supported by the CCE:

1. cached access to DRAM (cacheable memory)
2. uncached access to DRAM (cacheable memory)
3. uncached access to non-DRAM (uncacheable memory)

Cached access to uncacheable memory is not currently supported, and issuing such an access has undefined behavior.

Some preliminary support for L2 (i.e., beyond the CCE) atomics is also added as placeholder in the CCE, but this is not implemented.


bp_me/src/v/cce/bp_cce_msg.v
```diff
@@ -246,6 +243,16 @@ module bp_cce_msg
      ,.o(pe_lce_id_one_hot)
      );
 
+  // CCE coherence PMA - Mem responses
+  logic resp_pma_coherent_lo;
+  bp_cce_pma
+    #(.bp_params_p(bp_params_p)
+      )
+    resp_pma
```
dpetrisko, line 251: may want to gate this with v_i, otherwise X-prop could be a problem (it was in the LCE pma)

bp_me/src/v/cce/bp_cce_pma.v
```diff
@@ -0,0 +1,38 @@
...
+ */
+
+module bp_cce_pma
+  import bp_common_pkg::*;
+  import bp_common_aviary_pkg::*;
+  #(parameter bp_params_e bp_params_p = e_bp_default_cfg
+    `declare_bp_proc_params(bp_params_p)
+  )
+  (input [paddr_width_p-1:0]                           paddr_i
+   , output logic                                      coherent_o
```
dpetrisko, line 28: Maybe cached, rather than coherent? There's the whole "PGAS is trivially coherent" thing

bp_me/src/v/cce/bp_cce_reg.v
```diff
@@ -109,6 +109,16 @@ module bp_cce_reg
   assign coh_state_o    = coh_state_r;
   assign auto_fwd_msg_o = auto_fwd_msg_r;
 
+  // CCE coherence PMA - LCE requests
+  logic req_pma_coherent_lo;
+  bp_cce_pma
+    #(.bp_params_p(bp_params_p)
+      )
+    req_pma
```
dpetrisko, line 117: may want to gate this with v_i, otherwise X-prop could be a problem (it was in the LCE pma)

bp_me/src/v/cce/bp_cce_msg.v
```diff
@@ -246,6 +243,16 @@ module bp_cce_msg
      ,.o(pe_lce_id_one_hot)
      );
 
+  // CCE coherence PMA - Mem responses
+  logic resp_pma_coherent_lo;
```
dpetrisko, line 247: Could we use the mem_msg type here rather than checking PMAs again? Since the mem_cmd has the correct PMA I think it's implicitly the same

dpetrisko:
>Cached access to uncacheable memory is not currently supported, and issuing such an access has undefined behavior.

Is there a good place to add an assertion for this?

bp_me/src/v/cce/bp_cce_pma.v
muwyse: Right, some people consider uncacheable memory trivially coherent and others consider it incoherent because it does not explicitly participate in the coherence protocol....

For us, I guess we can go either way. I usually fall on the side of considering uncacheable memory to be incoherent is clearer because it doesn't imply that something is explicitly enforcing coherence.

bp_me/src/v/cce/bp_cce_reg.v
muwyse: will do

bp_me/src/v/cce/bp_cce_msg.v
muwyse: I don't think so, because an uncached response can be to either cacheable or uncacheable memory

bp_me/src/v/cce/bp_cce_pma.v
muwyse: Adjusted to cacheable_addr_o instead of coherent_o

muwyse:
> > Cached access to uncacheable memory is not currently supported, and issuing such an access has undefined behavior.
> 
> Is there a good place to add an assertion for this?

Yes, I think there is a place. I will add.

PR #674, opened 2020-11-12
write bp_top/nonsynth output to "global" file; separate files for each core; option of choosing either

hchenji:
Writing to (0x0010_1000+core_id) writes to stdout.core_id. Writing to (0x0010_3000) writes to stdoutglobal.out in addition to simulation terminal out. Could be useful when debugging multicore programs.

Tested only with VCS and bubblesort-demo, on devtoolset-8. 

Ignore all changes in the request except the ones to bp_nonsynth_host.v!


bp_top/test/common/bp_nonsynth_host.v
```diff
@@ -116,21 +132,29 @@ module bp_nonsynth_host
...
+      putch_core_data_cmd_v = 0;
+
       unique
-      casez (io_cmd_lo.header.addr)
+      casez (io_cmd_lo.header.addr & 64'hffff_ffff_ffff_f000)
         putchar_base_addr_gp: putchar_data_cmd_v = io_cmd_v_lo;
         getchar_base_addr_gp: getchar_data_cmd_v = io_cmd_v_lo;
         finish_base_addr_gp : finish_data_cmd_v = io_cmd_v_lo;
         bootrom_base_addr_gp: bootrom_data_cmd_v = io_cmd_v_lo;
+        putch_core_base_addr_gp:  begin
```
dpetrisko, line 154: You can use ??? in the case statement to match all but the lower bits

dpetrisko:
Hi, thanks for the patch!

Can you remove the unnecessary parts so we can merge in the host changes only?

hchenji:
forgot to branch before PR. replacing this with #697.

PR #677, opened 2020-11-15
fix accelerator tests

zaazad:


dpetrisko:
Thanks! Can you:
- rebase on dev to fix the conflicts
- Add a ci/accelerator.sh test with the accelerator config / test
- Add a job to the .gitlab_ci to run the accelerator test

dpetrisko:
Can you make separate PR for the zipline stuff? I would like to fix the current accelerator tests first, since including zipline in the flow will need more thought

PR #684, opened 2020-11-21
adding new axi wrappers and converters

tanglingshu:
Creating new pull request because I just can't update my obsolete black-parrot top_dev branch correctly


bp_me/src/v/axi/bp_axi4_to_axi4_lite_converter.v
```diff
@@ -0,0 +1,230 @@
+`include "bsg_defines.v"
```
dpetrisko, line 1: we want a bp_lite to axi_lite converter, not to convert to axi and then axi-lite.  If we need to convert axi to axi lite, we can typically use existing FPGA IP

bp_me/src/v/axi/bp_axi4_to_axi4_lite_converter.v
```diff
@@ -0,0 +1,230 @@
...
+
+module bp_axi4_to_axi4_lite_converter
+
+ #(// AXI WRITE DATA CHANNEL PARAMS
+   parameter  axi_data_width_p             = 64 //or 32, has to be either 32 or 64 for axi-lite
+  , localparam axi_strb_width_lp            = axi_data_width_p/8
+  , localparam lg_axi_data_width_in_byte_lp = `BSG_SAFE_CLOG2(axi_data_width_p/8)
+
+  // AXI WRITE/READ ADDRESS CHANNEL PARAMS  
+  , parameter  axi_addr_width_p             = 64 //or 32, has to be either 32 or 64 for axi-lite
```
dpetrisko, line 11: assertion?

dpetrisko (changes requested):
Nice, looks pretty good.  We need the wrapper in this PR so that these modules are tested

bp_me/src/v/axi/bp_lite_to_axi_lite_master.sv
```diff
@@ -0,0 +1,257 @@
...
+        m_axi_lite_wvalid_o = io_cmd_v_i;
+
+        write_strb          = write_strb << (axi_strb_width_lp - 2**io_cmd_header_r.size);
+        write_strb          = write_strb >> (axi_strb_width_lp - 2**io_cmd_header_r.size);
+        m_axi_lite_wstrb_o  = ((2**io_cmd_header_r.size) < axi_strb_width_lp)
+                            ? write_strb
+                            : '1;
+
+        if (m_axi_lite_bvalid_i) begin
+          io_resp_cast_o.data[31:0] = 32'hFFFF_FFFF;
```
dpetrisko, line 210: what's this write doing?

bp_me/src/v/axi/bp_lite_to_axi_lite_master.sv
```diff
@@ -0,0 +1,257 @@
...
+        m_axi_lite_awvalid_o = io_cmd_v_i;
+        if (m_axi_lite_awready_i) begin
+          state_n = e_write_data_tx;
+        end
+      end
+
+      e_write_data_tx : begin
+        io_cmd_ready_o      = m_axi_lite_wready_i;
+
+        m_axi_lite_wdata_o  = io_cmd_cast_i.data[31:0];
```
dpetrisko, line 200: axi_data_width_p?

dpetrisko: (throughout)

bp_me/src/v/axi/bp_lite_to_axi_lite_master.sv
tanglingshu: If you're asking about "io_resp_cast_o.data[31:0] = 32'hFFFF_FFFF;", originally I wanted to return all 0s for io_resp's data field on I/O writes, but I looked at the waveform and 32'hFFFF_FFFF was returned as the io_resp data (in blue circle below)
image
If the data field of io_resp on I/O writes should always return 0s, let me know. 

The write strobe is supposed to mask every 8 bits of data (with 1 bit of the strobe). For instance:
data (32bits)= 0xDEADBEEF
write_strb (4bits) = 0x0011
client side should manipulate the data so that it receives 0x0000BEEF. 

I used the io_cmd's header's size field to help me determine what the write strobe should be. So for instance, if the  

msg_size | write_strb
-- | --
1byte | 0x0000_0001
2bytes | 0x0000_0011
4bytes | 0x0000_1111
8bytes | 0x1111_1111


I used shifting bits for now, but I spoke Tommy and he suggested a solution that uses "bsg_decode" and "bsg_expand_bitmask" for that. That's something I'd like to implement later

bp_me/src/v/axi/bp_lite_to_axi_lite_master.sv
tanglingshu: Is this referring to what the axi_data_width_p should be? or why does it exist at all? Most of the axi-lite devices use 32bit (at least on genesys 2 board). If BP only supports 64bits for IO communication, then maybe a serdes (PISO/SIPO) is required.

If the above is not what you're asking, may you clarify your original question? Thank you

bp_me/src/v/axi/axi_lite_to_bp_lite_client.sv
```diff
@@ -0,0 +1,204 @@
...
+    endcase
+
+    case (state_r) 
+      e_wait : begin
+        s_axi_lite_awready_o = io_cmd_yumi_i;
+        s_axi_lite_wready_o  = io_cmd_yumi_i;
+        s_axi_lite_arready_o = io_cmd_yumi_i;
+
+        if (s_axi_lite_arvalid_i) begin
+          io_cmd_cast_o.header.payload  = bp_out_mem_payload_width_lp'('h0010);
```
dpetrisko, line 145: What's this magic number?

bp_me/src/v/axi/axi_lite_to_bp_lite_client.sv
```diff
@@ -0,0 +1,204 @@
...
+  `declare_bp_bedrock_mem_if(paddr_width_p, cce_block_width_p, lce_id_width_p, lce_assoc_p, bp_out);
+
+  bp_bedrock_bp_out_mem_msg_s io_cmd_cast_o, io_resp_cast_i;
+  assign io_cmd_o       = io_cmd_cast_o;
+  assign io_resp_cast_i = io_resp_i;
+
+  // intermediate values for io_cmd_data_o resulted from a generate block
+  logic [axi_data_width_p-1:0] io_cmd_data_lo;
+  for (genvar i = 0; i < axi_strb_width_lp; i++) begin : io_cmd_data
+    assign io_cmd_data_lo[i*8+:8] = (s_axi_lite_wstrb_i[i])
```
dpetrisko, line 84: We don't support arbitrary strobes, just aligned accesses. So this doesn't really do much. We should add an assertion that the strobe is correct, though.

Should use bsg_bus_pack to replicate data across the io_cmd.data

bp_me/src/v/axi/axi_lite_to_bp_lite_client.sv
```diff
@@ -0,0 +1,204 @@
...
+
+   // AXI WRITE DATA CHANNEL PARAMS
+   , parameter  axi_data_width_p             = 64
+   , localparam axi_strb_width_lp            = axi_data_width_p/8
+
+   // AXI WRITE/READ ADDRESS CHANNEL PARAMS  
+   , parameter axi_addr_width_p              = 64 
+
+   // BP messeage size
+   , localparam msg_size_lp                  = `BSG_SAFE_CLOG2(axi_data_width_p/8)
```
dpetrisko, line 24: unused

bp_me/src/v/axi/axi_lite_to_bp_lite_client.sv
```diff
@@ -0,0 +1,204 @@
...
+   , input                                      io_cmd_yumi_i
+
+   , input [bp_out_mem_msg_width_lp-1:0]        io_resp_i
+   , input                                      io_resp_v_i
+   , output logic                               io_resp_ready_o
+
+   //====================== AXI-4 LITE =========================
+   // WRITE ADDRESS CHANNEL SIGNALS
+   , input [axi_addr_width_p-1:0]               s_axi_lite_awaddr_i
+   , input [2:0]                                s_axi_lite_awprot_i
```
dpetrisko, line 43: are these enums? If so, declare them?

bp_me/src/v/axi/axi_lite_to_bp_lite_client.sv
```diff
@@ -0,0 +1,204 @@
...
+   , output logic                               s_axi_lite_awready_o
+
+   // WRITE DATA CHANNEL SIGNALS
+   , input [axi_data_width_p-1:0]               s_axi_lite_wdata_i
+   , input [axi_strb_width_lp-1:0]              s_axi_lite_wstrb_i
+   , input                                      s_axi_lite_wvalid_i
+   , output logic                               s_axi_lite_wready_o
+
+   // WRITE RESPONSE CHANNEL SIGNALS
+   , output logic [1:0]                         s_axi_lite_bresp_o
```
dpetrisko, line 54: What are the 2 bits for?

bp_me/src/v/axi/axi_lite_to_bp_lite_client.sv
```diff
@@ -0,0 +1,204 @@
...
+   , output logic                               s_axi_lite_arready_o
+
+   // READ DATA CHANNEL SIGNALS
+   , output logic [axi_data_width_p-1:0]        s_axi_lite_rdata_o
+   , output logic [1:0]                         s_axi_lite_rresp_o
+   , output logic                               s_axi_lite_rvalid_o
+   , input                                      s_axi_lite_rready_i
+  );
+
+  wire [2:0] unused_0 = s_axi_lite_awprot_i;
```
dpetrisko, line 71: Are these optional ports, or do we only support a single protocol?

If optional, remove. If unsupported, add assertion

bp_top/src/v/bp_unicore_with_axi_wrapper.sv
```diff
@@ -0,0 +1,263 @@
...
+
+  // dram interface wrapper
+  bp_mem_to_axi_master
+   #(.bp_params_p(bp_params_p)
+   	,.axi_data_width_p(axi_full_data_width_p)
+   	,.axi_addr_width_p(axi_full_addr_width_p)
+   	,.axi_burst_type_p(axi_full_burst_type_p)
+   	,.axi_id_width_p(axi_full_id_width_p)
+   	)
+   mem_interface_axi_full_wrapper
```
dpetrisko, line 239: nit. These are converters, not wrappers. Adjust comments and rename

mem2axi, io2axil, axil2io

bp_top/src/v/bp_unicore_with_axi_wrapper.sv
```diff
@@ -0,0 +1,263 @@
...
+  bp_bedrock_uce_mem_msg_header_s mem_resp_header_li;
+  logic mem_resp_header_v_li, mem_resp_header_yumi_lo;
+  logic [dword_width_lp-1:0] mem_resp_data_li;
+  logic mem_resp_data_v_li, mem_resp_data_yumi_lo;
+
+  bp_unicore
+   #(.bp_params_p(bp_params_p))
+   unicore
+   (.clk_i(clk_i)
+   	,.reset_i(reset_i)
```
dpetrisko, line 151: nit, weird alignment

bp_top/src/v/bp_unicore_with_axi_wrapper.sv
```diff
@@ -0,0 +1,263 @@
...
+   , parameter axi_lite_data_width_p   = 32
+   , localparam axi_lite_strb_width_lp = axi_lite_data_width_p/8
+   
+   // AXI4-FULL PARAMS
+   , parameter axi_full_addr_width_p   = 64
+   , parameter axi_full_data_width_p   = 64
+   , parameter axi_full_id_width_p     = 1
+   , parameter axi_full_burst_type_p   = 2'b01 //INCR type
+   , localparam axi_full_strb_width_lp = axi_full_data_width_p/8
+   , localparam dword_width_lp         = axi_full_data_width_p
```
dpetrisko, line 31: dword is 64 bits, shouldn't be calculated. If only one parameterization is supported, add assertion

bp_me/src/v/axi/axi_lite_to_bp_lite_client.sv
```diff
@@ -0,0 +1,204 @@
...
+          io_cmd_cast_o.header.msg_type = e_bedrock_mem_uc_rd;
+          io_cmd_v_o                    = axi_lite_r_v;
+          
+          state_n                       = (io_cmd_yumi_i & axi_lite_r_v)
+                                        ? e_read_data_tx
+                                        : state_r;
+        end
+
+        else if (s_axi_lite_awvalid_i) begin
+          io_cmd_cast_o.header.payload             = bp_out_mem_payload_width_lp'('h0010);
```
dpetrisko, line 158: magic number?

bp_me/src/v/axi/axi_lite_to_bp_lite_client.sv
```diff
@@ -0,0 +1,204 @@
...
+  #(parameter bp_params_e bp_params_p = e_bp_default_cfg
+   `declare_bp_proc_params(bp_params_p)
+   `declare_bp_bedrock_mem_if_widths(paddr_width_p, cce_block_width_p, lce_id_width_p, lce_assoc_p, bp_out)
+
+   // AXI WRITE DATA CHANNEL PARAMS
+   , parameter  axi_data_width_p             = 64
+   , localparam axi_strb_width_lp            = axi_data_width_p/8
+
+   // AXI WRITE/READ ADDRESS CHANNEL PARAMS  
+   , parameter axi_addr_width_p              = 64
```
dpetrisko, line 21: Are the restrictions on allowable size here? Should we match to paddr_width_p?

bp_me/src/v/axi/bp_lite_to_axi_lite_master.sv
```diff
@@ -0,0 +1,209 @@
...
+    ,.reset_i(~aresetn_i)
+    ,.en_i(io_cmd_v_i)
+    ,.data_i(io_cmd_cast_i.header)
+    ,.data_o(io_cmd_header_r)
+    );
+
+  // io cmd read/write validity
+  logic io_cmd_w_v, io_cmd_r_v;
+
+  assign io_cmd_w_v = io_cmd_v_i & (io_cmd_cast_i.header.msg_type == e_bedrock_mem_uc_wr);
```
dpetrisko, line 93: nit, use ```wire io_cmd_w_v = ``` syntax

bp_me/src/v/axi/bp_lite_to_axi_lite_master.sv
```diff
@@ -0,0 +1,209 @@
...
+  typedef enum logic [1:0] {
+    e_wait           = 2'b00
+    ,e_read_data_tx  = 2'b01
+    ,e_write_resp_rx = 2'b10
+  } state_e;
+
+  state_e state_r, state_n;
+
+  // combinational Logic
+  always_comb begin
```
dpetrisko, line 106: nit, begin / end on their own line

bp_me/src/v/axi/bp_lite_to_axi_lite_master.sv
```diff
@@ -0,0 +1,209 @@
...
+    );
+
+  // io cmd read/write validity
+  logic io_cmd_w_v, io_cmd_r_v;
+
+  assign io_cmd_w_v = io_cmd_v_i & (io_cmd_cast_i.header.msg_type == e_bedrock_mem_uc_wr);
+  assign io_cmd_r_v = io_cmd_v_i & (io_cmd_cast_i.header.msg_type == e_bedrock_mem_uc_rd);
+
+  // declaring all possible states
+  typedef enum logic [1:0] {
```
dpetrisko, line 97: don't waste a typedef, just

enum logic [1:0] {} state_n, state_r;

bp_me/src/v/axi/bp_lite_to_axi_lite_master.sv
```diff
@@ -0,0 +1,209 @@
...
+        io_resp_cast_o.data[axi_data_width_p-1:0] = {(axi_data_width_p-1){1'b0}};
+        io_resp_v_o                               = m_axi_lite_bvalid_i;
+        state_n                                   = m_axi_lite_bvalid_i
+                                                  ? e_wait
+                                                  : state_r;
+      end
+
+     e_read_data_tx : begin
+       io_resp_v_o                               = m_axi_lite_rvalid_i;	
+       io_resp_cast_o.data[axi_data_width_p-1:0] = (m_axi_lite_rresp_i == '0)
```
dpetrisko, line 175: don't need to gate the data, we're not looking at it if it's not valid

dpetrisko: applies throughout

dpetrisko (changes requested):

bp_me/src/v/axi/axi_lite_to_bp_lite_client.sv
tanglingshu: There is a restriction on the allowable size here. It's either 32 or 64bits. Since the paddr_width_p is 40, I think we can just assign the 40-bit address to the 32/64 bit axi address. AXI32 would be header.addr[31:0] and AXI64 would be header.addr with sign extension I believe.

bp_me/src/v/axi/axi_lite_to_bp_lite_client.sv
tanglingshu: will remove in the latest commit

bp_me/src/v/axi/axi_lite_to_bp_lite_client.sv
tanglingshu: The awprot signal exists for axi-lite but it's typically ignored by the client device. The protection encoding is as follows. I think unsupported assertion is more appropriate here 
image

bp_me/src/v/axi/axi_lite_to_bp_lite_client.sv
tanglingshu: The encoding for both bresp and rresp is as follows. I'm defaulting the bresp (write response) to OKAY since I'm not sure if other errors are supported by BP.
image

bp_me/src/v/axi/axi_lite_to_bp_lite_client.sv
tanglingshu: yes they are enums but since awprot is unsupported, should I still declare them as enum?

bp_me/src/v/axi/axi_lite_to_bp_lite_client.sv
tanglingshu: what kind of strobe would be considered a correct strobe?

bp_me/src/v/axi/axi_lite_to_bp_lite_client.sv
tanglingshu: This is the number that I see in the header's payload field in a simulation. I originally set it to 0s but it wasn't passing the test. I do wish to know if there is a significance to it or it is irrelevant in most cases.

bp_me/src/v/axi/axi_lite_to_bp_lite_client.sv
tanglingshu: same as above

bp_me/src/v/axi/bp_lite_to_axi_lite_master.sv
tanglingshu: will correct in the next PR

bp_me/src/v/axi/bp_lite_to_axi_lite_master.sv
tanglingshu: will correct in the next PR

bp_me/src/v/axi/bp_lite_to_axi_lite_master.sv
tanglingshu: this gate is used for valid but erroneous data. In this case, I say that if the client device responded with anything other than "OKAY", then send all 0s to io_resp's data field. I think you're right that most of the time it will be valid and correct data. So if an error response is not of concern, then I can remove this gate.

bp_me/src/v/axi/bp_lite_to_axi_lite_master.sv
tanglingshu: will correct in the next PR

bp_top/src/v/bp_unicore_with_axi_wrapper.sv
tanglingshu: will remove in the next PR

bp_top/src/v/bp_unicore_with_axi_wrapper.sv
```diff
@@ -0,0 +1,263 @@
...
+  logic mem_cmd_header_v_lo, mem_cmd_header_ready_li;
+  logic [dword_width_lp-1:0] mem_cmd_data_lo;
+  logic mem_cmd_data_v_lo, mem_cmd_data_ready_li;
+
+  bp_bedrock_uce_mem_msg_header_s mem_resp_header_li;
+  logic mem_resp_header_v_li, mem_resp_header_yumi_lo;
+  logic [dword_width_lp-1:0] mem_resp_data_li;
+  logic mem_resp_data_v_li, mem_resp_data_yumi_lo;
+
+  bp_unicore
```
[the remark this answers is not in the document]
tanglingshu: will add multicore option in the next PR. Though multicore doesn't have the same bp_burst interface as unicore. What do I do with that?

bp_top/src/v/bp_unicore_with_axi_wrapper.sv
tanglingshu: will correct in the next PR

bp_top/src/v/bp_unicore_with_axi_wrapper.sv
tanglingshu: will correct in the next PR

bp_top/src/v/bp_unicore_with_axi_wrapper.sv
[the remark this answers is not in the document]
dpetrisko: https://github.com/black-parrot/black-parrot/blob/master/bp_top/test/tb/bp_tethered/wrapper.sv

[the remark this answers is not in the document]
dpetrisko: Rather have them both in this PR, since this file would change name and purpose in another one

bp_me/src/v/axi/axi_lite_to_bp_lite_client.sv
dpetrisko: yes, look into that... The payload is a struct, which fields are being set?

bp_me/src/v/axi/axi_lite_to_bp_lite_client.sv
dpetrisko: I guess I'm not sure what's legal in the AXI spec. This block looks like it implies any arbitrary set of bytes can be set / unset.

We support contiguous bytes
-----------X
----------XX
-------XXXX
and their shifted variants

bsg_bus_pack is needed for replication

bp_me/src/v/axi/axi_lite_to_bp_lite_client.sv
dpetrisko: make an enum

bp_me/src/v/axi/axi_lite_to_bp_lite_client.sv
dpetrisko: yes

bp_me/src/v/axi/axi_lite_to_bp_lite_client.sv
dpetrisko: just push the assertion on this PR

bp_me/src/v/axi/axi_lite_to_bp_lite_client.sv
dpetrisko: Cool, 64 bits is fine then, add an assertion.

>AXI64 would be header.addr with sign extension I believe.

Sign extension seems weird to me, I would think zero extension is more likely because then you could actually use the entire 64-bit space if you had other peripherals

bp_me/src/v/axi/bp_lite_to_axi_lite_master.sv
dpetrisko: we don't support errors on the bus, so there's no difference between a zero and invalid response. Would rather save the hardware. Should loudly fail in simulation if an error comes back though

bp_me/src/v/axi/axi_lite_to_bp_lite_client.sv
tanglingshu: I see, sorry about that. The payload is a memory payload that has:
coherence state: invalid (3'h0)
way_id: 3'h0
lce_id: 4'h2
prefetch: 1'b0
uncached: 1'b0
speculative; 1'b0

I guess this sequence of payload is valid only for the IO coming out of the nbf_loader. Reason why I set this value is that I attached this wrapper to an nbf_loader for simulation. Practically, if any device initiates a transaction with BP, is there a way to figure out the payload field? like a default value?

bp_me/src/v/axi/axi_lite_to_bp_lite_client.sv
tanglingshu: According to the AXI spec, any combination of write strobe is legal. Each bit of the strobe masks 8 bits of the input data. For example:
Input: DEAD_BEEF
wstrb: 0101
Output: 00AD_00EF

I guess we could either do:
1) bsg_bus_pack where we only support limited strobes and make an assertion for unsupported strobes
2) supports arbitrary strobes and write 0s for invalid data

dpetrisko: >bsg_bus_pack where we only support limited strobes and make an assertion for unsupported strobes

Yeah, that's what I'm suggesting

bp_me/src/v/axi/bp_mem_to_axi_master.sv
```diff
@@ -0,0 +1,309 @@
...
+
+  // memory command read/write validity
+  wire mem_cmd_header_read_v  = mem_cmd_header_v_i & mem_cmd_header_cast_i.msg_type inside {e_bedrock_mem_rd, e_bedrock_mem_uc_rd, e_bedrock_mem_pre};
+  wire mem_cmd_header_write_v = mem_cmd_header_v_i & mem_cmd_header_cast_i.msg_type inside {e_bedrock_mem_wr, e_bedrock_mem_uc_wr};
+
+  // Write counters and burst logic
+  logic [7:0] awburst_cnt_r, awburst_cnt_n;
+  logic rready_r; 
+
+  function automatic [7:0] len_calc;
```
dpetrisko, line 138: @taylor-bsg I know functions are intended to synthesize by spec, but how is tool support?

bp_top/src/v/bp_unicore_with_axi_wrapper.sv
```diff
@@ -0,0 +1,262 @@
...
+  , input [axi_full_data_width_p-1:0]         m_axi_rdata_i 
+  , input [1:0]                               m_axi_rresp_i 
+  , input                                     m_axi_rlast_i 
+  , input                                     m_axi_rvalid_i
+  , output logic                              m_axi_rready_o
+  );
+
+  if (multicore_p)
+    begin : multicore
+    // multi-core + wrapper implementation
```
dpetrisko, line 125: In the interest of getting this in, just put an $error here that multicore axi top is not currently supported. I'll make an issue to fix it up in the near future

tanglingshu: will make all of the necessary changes throughout this week. Thank you for your patience

dpetrisko: Thanks Ling!

taylor-bsg:
@tanglingshu Hi Ling! I hope you are well -- how are we doing in terms of finalizing this?

PR #703, opened 2020-12-12
FE cleanup refactor

dpetrisko:
- Adds BP MMU
- Separates out redirects from attaboy in FE
- Clean up I$
- Fix a few FE predictor bugs, increase in performance


taylor-bsg:
Interesting what were the predictor bugs and how much perf increase?

dpetrisko:
Redirection would clobber the branch sites, so metadata for branches following mispredicts would get updated based on the wrong index. Similar issue with global history.

Looks like 1% on coremark and 3% on Linux boot, but haven't looked too close.  The Linux boot in particular seems too good to be true...

PR #706, opened 2020-12-14
FPGA Signal Alignment Bug

farzamgl:
On Vivado, this bug causes the fraction part of the result to be all-zeros when converting between single and double precision floating-point numbers.


dpetrisko (changes requested):
Weird bug.  Can we get a test reproducing it? And a comment about which Vivado version can't handle the alternate syntax

dpetrisko:
@taylor-bsg I forget, what was the style guide ruling on this? Explicitly cast width of expressions?

taylor-bsg:
Yeah, or a concatenation is okay as well — the width is not determined by
the LHS

dpetrisko (approved):
Yeah I think we can merge this then, end result was addition to the style guide to use explicit sizing in struct field assignment

PR #707, opened 2020-12-15
Update CCE interfaces to BedRock Burst

muwyse:
This PR updates the CCEs to use the BedRock Burst interface. This is a large change and part of a series that will convert all XCE modules to use BedRock Burst and, hopefully, simplify the memory end networks.

The primary changes are:

1. change CCE interface to BedRock Burst
2. update mock LCE to use BedRock Lite
3. add some new sipo/piso modules that use last signals
4. update bsg_wormhole_stream_control with has_data_o and last_o signals (useful for both Burst and Stream interfaces)
5. add/update BedRock interface converters
6. add Bedrock Burst-wormhole converters that leverage sipo/piso with last
7. update bp_tile to use new CCE and new interface converters
8. some random fixes in various testbenches/modules reported by linter
9. remove bp_cce_fsm_top - no longer needed since CCE interface is now ready&valid


muwyse:
I did not make major adjustments to the tiles, so it is possible there is more buffering that can still be removed.

bp_me/src/v/cce/bp_bedrock_size_to_len.sv
```diff
@@ -0,0 +1,51 @@
+/**
```
dpetrisko, line 1: ah, I see this answers my question on the other PR

bp_me/src/v/cce/bp_cce.sv
```diff
@@ -274,6 +279,51 @@ module bp_cce
    * Decode/Execute Stage
    */
 
+  // lce request header buffer
```
dpetrisko, line 282: Generally better to put the buffers outside the module. In this case, aren't we connecting to wormhole routers which are helpful?

bp_me/src/v/cce/bp_cce_inst_stall.sv
```diff
@@ -49,12 +49,12 @@ module bp_cce_inst_stall
 
   // Note: these signals exist to suppress a warning about the valid/ready
   // needing to be inout signals for this module when building with VCS.
```
dpetrisko, line 51: what's the warning? I've never seen that before

bp_me/src/v/cce/bp_io_cce.sv
```diff
@@ -20,15 +20,15 @@ module bp_io_cce
```
dpetrisko, line 20: Not necessarily part of this PR, but can this be converted to ready_and on both sides? I think so

bp_me/src/v/lce/bp_lce.sv
```diff
@@ -86,12 +86,12 @@ module bp_lce
     // Req: ready->valid
     , output logic [lce_req_msg_width_lp-1:0]        lce_req_o
     , output logic                                   lce_req_v_o
-    , input                                          lce_req_ready_i
```
dpetrisko, line 89: I think this can fairly easily be converted to ready_and since we're stalling in the FSM anyway

dpetrisko (commented):
This is awesome. I haven't looked at the tile stuff yet, I'd believe there's some optimization to be done.

I think at a high level, this should be two PRs.  First, we should do as much ready_and conversion as we can in these modules before converting to burst. That should ensure that we have the fewest required buffers during the next step.  (I'd also be in favor of putting the bugfixes and testing changes in PR 1 just for simplicity)

bp_me/src/v/cce/bp_bedrock_size_to_len.sv
muwyse: yes, I changed this to be a zero-based length on the output to align with the WH network stuff and the other PR.

bp_me/src/v/cce/bp_cce.sv
muwyse: I think I agree in general, not sure about this case. The ucode CCE functionality requires a demanding interface with the inbound message headers. I do think it is nice to have the CCE interface be r&v across the board, though.

I could push these into the cce_wrapper module, but I am not convinced pushing them out beyond that is better. It is nice to be able to take either the ucode or FSM CCE outside the wrapper and use it directly with all r&v interfaces.

bp_me/src/v/cce/bp_cce_inst_stall.sv
muwyse: It (at least used to) comes up with VCS where it thinks these input signals should be in/out instead of just input. I think it has to do with how they get used since the stall_o is basically an OR tree drawn from the input signals. I can try removing and rechecking the compile/warnings/synth checks to see if it still complains.

bp_me/src/v/cce/bp_io_cce.sv
muwyse: I'm sure it can, yes, and likely without too much effort.

bp_me/src/v/lce/bp_lce.sv
muwyse: Likely, yes. I'll take a look.

muwyse:
> I think at a high level, this should be two PRs.

This might only be partially possible. It could be doable for things like the LCE, but doing it for CCE might be more effort than it is worth to first change up interfaces then apply burst.

muwyse:
@dpetrisko I think this is ready for re-review. Single core RISCV tests, multicore tests, ME tests, and check_design.syn CI script all look good. I'll need to do another pass on the changes, but feedback would be great if you can spare a few minutes. Thanks!

bp_me/src/v/lce/bp_lce.sv
muwyse: filed issue: https://github.com/black-parrot/black-parrot/issues/846

bp_me/src/v/cce/bp_cce.sv
```diff
@@ -276,6 +281,51 @@ module bp_cce
...
+      ,.ready_o(lce_req_header_ready_and_o)
+      ,.data_i(lce_req_header_i)
+      ,.v_i(lce_req_header_v_i)
+      ,.v_o(lce_req_v)
+      ,.data_o(lce_req)
+      ,.yumi_i(lce_req_yumi)
+      );
+
+  // lce response header buffer
+  logic lce_resp_v, lce_resp_yumi;
```
dpetrisko, line 300: Related to above, but if we do need fifos here, they should probably be twofers so that they don't have a bubble?

bp_me/src/v/cce/bp_uce.sv
```diff
@@ -595,7 +595,7 @@ module bp_uce
           end
         e_ready:
           begin
-            if ((uc_store_v_li && (~uc_hit_v_li || (l1_writethrough_p == 1))) || wt_store_v_li)
+            if (((uc_store_v_li || uc_amo_v_li) && (~uc_hit_v_li || (l1_writethrough_p == 1))) || wt_store_v_li)
```
dpetrisko, line 598: Undo this change, I believe it's a bug from rebase

bp_top/src/v/bp_tile.sv
```diff
@@ -366,73 +399,204 @@ module bp_tile
...
-    ,.packet_o(cce_lce_resp_packet_li)
-    ,.v_o(cce_lce_resp_v_li)
-    ,.yumi_i(cce_lce_resp_yumi_lo)
+    ,.data_o(cce_lce_resp_data)
+    ,.data_v_o(cce_lce_resp_data_v)
+    ,.data_ready_and_i(cce_lce_resp_data_ready_and)
     );
-  assign cce_lce_resp_li = '{header: cce_lce_resp_packet_li.header.msg_hdr, data: cce_lce_resp_packet_li.data};
+
+  // Mem Command
```
dpetrisko, line 539: We won't want to merge this part in, will want to incorporate Songchun's stream cache slice

bp_me/src/v/cce/bp_cce_msg.sv
```diff
@@ -179,24 +205,94 @@ module bp_cce_msg
...
+      ,.yumi_i(mem_resp_header_yumi_o)
       ,.count_o(mem_credit_count_lo)
       );
 
   wire mem_credits_empty = (mem_credit_count_lo == mem_noc_max_credits_p);
   wire mem_credits_full = (mem_credit_count_lo == 0);
   assign mem_credits_empty_o = mem_credits_empty;
 
+  // Memory response data packets
+  wire [counter_width_lp-1:0] mem_resp_size_in_packets =
```
dpetrisko, line 227: I know we try to keep hardcoded enum values to a minimum, but we have these types of expressions everywhere in the code. Can we just enshrine that msg_size == 2**mem_resp.size? That makes the math here (and everywhere into just divides / shifts

bp_top/src/v/bp_tile.sv
```diff
@@ -514,12 +678,18 @@ module bp_tile
   wire is_cache_cmd        = ~local_cmd_li || (local_cmd_li & (device_cmd_li == cache_dev_gp));
   wire is_loopback_cmd     = local_cmd_li & ~is_cfg_cmd & ~is_clint_cmd & ~is_cache_cmd;
 
+  // TODO: fix me!
```
dpetrisko, line 681: The bug here is that cfg, clint, cache, or loopback are ready->valid rather than ready-valid&?

dpetrisko (changes requested):
Okay, so this is a doozy.  In general, it looks like there are some unrelated changes in here for ME testbenches, handshakes, etc. If we can merge those in first that would help.

Major thing is we need to start pulling Songchun's stuff in, which I'll try to do over the next week.  The actual burst code I think is reasonable, although we may want to combine with Songchun's stream_pump modules to make burst_pump modules. This would avoid having repetition for interfacing between an FSM and a bp burst interface which we're going to be doing all over the place

muwyse:
> Why does the ucode CCE require demanding? Seems like the data buffering policy would be the same for both, but maybe I'm missing something

Basically, the ucode cce can be stalled by various internal events or conditions. I can look into this more, but as it stands, it is easier to reason about it if it gets to be demanding (valid->yumi) as opposed to helpful (ready&valid).

bp_me/src/v/cce/bp_cce.sv
dpetrisko: discussed offline

bp_me/src/v/cce/bp_cce.sv
dpetrisko: discussed offline. mem_resp needs to be helpful due to arbitration policies.  Other FIFOs can most likely be removed

bp_me/src/v/cce/bp_bedrock_size_to_len.sv
```diff
@@ -0,0 +1,51 @@
...
+ *
+ *   The len_o output is equal to flits-1 (i.e., it is zero based)
+ *
+ *   width_p is the width to use for the output length integer
+ *
+ *   (1 << e_bedrock_msg_size_X) gives the number of bytes for the message size enum value
+ *
+ */
+
+module bp_bedrock_size_to_len
```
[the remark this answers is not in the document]
muwyse: The intent is this module computes the number of data packets that would be required if data exists for the message. The module that consumes the len_o output from here is responsible for examining the message type (or other handshake info, e.g., has_data signal in BedRock Burst) to determine if data exists or not for the message being processed.

bp_me/src/v/cce/bp_cce.sv
muwyse: Not addressing this in the PR. These are only buffering the headers (about 64-bits a piece) so the cost isn't huge.

muwyse: opened #869 to track

bp_me/src/v/cce/bp_cce_msg.sv
muwyse: Yes, I am good with that for the BedRock message size enum. I see no reason why we would want to change the values. I will add a comment to the BedRock package definitions.

bp_top/src/v/bp_tile.sv
muwyse: This will be separate PR. Opened #870 to track

muwyse:
@dpetrisko this should be ready for review and merge to me_dev

dpetrisko:
Awesome. It looks great at a glance, but I probably won't have time to do a thorough review until next week.  Have you done any performance measurements?

dpetrisko (approved):
Sorry for the delay. Minor clarifications and an offline comment.

I think before merging into dev, we should delete the _to_lite / _from_lite modules and replace with burst_to_stream anywhere applicable. But we can merge this to me_dev first, no problem.

I would also like to update songchun's stream cache controller and merge that in before  going to dev.

Overall, awesome!

bp_top/src/v/bp_tile.sv
```diff
@@ -366,73 +424,191 @@ module bp_tile
...
+     ,.len_width_p(coh_noc_len_width_p)
+     ,.cid_width_p(coh_noc_cid_width_p)
+     ,.pr_hdr_width_p(lce_resp_msg_header_width_lp)
+     ,.pr_data_width_p(dword_width_gp)
+     ,.pr_len_width_p(pr_len_width_lp)
      )
-   cmd_adapter_in
+   cce_lce_resp_wh_to_burst
+   (.clk_i(clk_i)
+    ,.reset_i(reset_i)
```
dpetrisko, line 530: These all need to be reset_r, not reset_i

bp_me/src/v/cce/bp_cce_inst_decode.sv
```diff
@@ -416,7 +416,7 @@ module bp_cce_inst_decode
               decoded_inst_o.src_a_sel = e_src_sel_flag;
               decoded_inst_o.src_a.flag = op_type_u.rtype.src_a.flag;
               decoded_inst_o.src_b_sel = e_src_sel_flag;
-              decoded_inst_o.src_b.flag = op_type_u.rtype.src_a.flag;
+              decoded_inst_o.src_b.flag = op_type_u.rtype.src_b.flag;
```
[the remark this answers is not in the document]
muwyse: correct. These particular instructions do not appear to be used in the current microcode

bp_common/src/v/bsg_serial_in_parallel_out_passthrough_dynamic_last.v
```diff
@@ -0,0 +1,129 @@
...
+ *
+ * The last_i signal must be raised early with v_i when the last serial word
+ * is available on the input and the input will not fill the SIPO (i.e., less
+ * than max_els_p are being deserialized to the parallel output).
+ *
+ * If max_els_p == 1 then this module simply passes through the signals.
+ * Otherwise, the output becomes valid in the cycle that the last input
+ * arrives.
+ *
+ * This module passes throught he last input data word directly to the output
```
muwyse, line 16: typo

bp_common/src/v/bsg_serial_in_parallel_out_passthrough_dynamic_last.v
```diff
@@ -0,0 +1,129 @@
...
+      ,.set_i(last_word_received)
+      ,.clear_i(transaction_done)
+      ,.data_o(last_r)
+      );
+
+    // data becomes valid when last_i is raised as last word is simply passed
+    // through directly to output or when the last word has been captured into
+    // the data_dff if the input message has fewer than max_els_p words.
+    // v_o is also raised if last element is valid on input
+    assign v_o = last_i | last_r | (v_i & data_en_li[max_els_p-1]);
```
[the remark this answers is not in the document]
muwyse: No. The way I implemented it does not require last_i to be raised on the last serial word if the SIPO is being completely filled.

bp_me/src/v/wormhole/bp_burst_to_wormhole.sv
```diff
@@ -0,0 +1,215 @@
+/**
```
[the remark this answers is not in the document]
muwyse: They are very similar, yes. I think it is beneficial to have one specialized for BedRock Burst and another for BedRock Stream protocols. They differ mainly in the input interface. The wormhole streamer takes in pr_data_beats_i to specify the length of the message whereas this uses the has_data and last signals from Burst to determine the number of data beats required.

The question is for a "generic" protocol X <-> wormhole, what are the signals required on the X side. In my opinion, pr_data_beats_i doesn't make sense for Burst or Stream because they already contain the required flow control info with the has_data/last signaling. Adding logic to back-compute pr_data_beats_i seems silly. However, we could decide that the bsg_wormhole_stream_in/out talks to something that is effectively BedRock Burst or Stream. Then, we could wrap the stream_in/out to easily create burst-wormhole and stream-wormhole converters for BP.

muwyse:
In response to this question in bp_cce_msg.sv:
> I think we can just cast here then without the muxing?

I don't think it can be a direct cast. The message size is in log2(bytes) but we need to convert to dword_width_gp packets. I think the mux could potentially be transformed to some shifts and maybe a subtract by one (or I convert the FSM logic to handle the off by one thing to remove the dynamic subtract).

bp_me/src/v/cce/bp_cce_msg.sv
muwyse: Okay, I think I have this simplified now. A similar approach could be used in the bp_bedrock_size_to_len module.

muwyse:
>I think before merging into dev, we should delete the _to_lite / _from_lite modules and replace with burst_to_stream anywhere > applicable. But we can merge this to me_dev first, no problem.
>
> I would also like to update songchun's stream cache controller and merge that in before going to dev.

I think both of these are reasonable. I have already started working to change interfaces on the "uncore" modules in the tile so that everything uses burst or stream

bp_common/src/v/bsg_serial_in_parallel_out_passthrough_dynamic_last.v
```diff
@@ -0,0 +1,129 @@
...
+        ,.en_i   (data_en_li[i])
+        ,.data_o (data_o    [i])
+        );
+      end
+    // final word of full message is passed through directly to output
+    assign data_o[max_els_p-1] = data_i;
+
+    // data valid register
+    // output data becomes available the cycle following the last input word
+    logic last_r;
```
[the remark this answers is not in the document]
muwyse: If it weren't for the passthrough and last behavior I think I would agree. However, since the last serial input is never buffered AND there is no requirement that the SIPO fill up for a transaction (i.e., last can go high with any serial input word), I don't think the logic would be simplified.

PR #710, opened 2020-12-19
Adding domain and sac register read in cfg

Shashank-Vijay:
This PR adds support for reading the domain and sac registers in the CFG module.


dpetrisko:
LGTM. Let's bump bp_tests to master to include those new tests in this PR

Issue #718, opened 2020-12-23
Support non-blocking loads


dpetrisko:
We already support separate commit / writeback. Adding a simple in-order scoreboard should allow us to track dependencies for these long-latency instructions.

Open Questions:
- Do we support non-point-to-point loads?
- If so, how do we track the requests? Adding reg_id payload?
- Does cache engine or cache track the outstanding requests?
- Does this follow through the network? Or do we track the return address?

taylor-bsg:
The manycore is a good guide for how this should be done..

dpetrisko:
We'll definitely look at the manycore. The scoreboarding approach will most likely be identical.  However, there are distinct differences in the system e.g. cache coherence, caches at all ..., that complicate the implementation

dpetrisko:
Closing; supported

PR #724, opened 2020-12-27
Adding scoreboards for idiv and fdiv

dpetrisko:
Adds an integer and floating point scoreboard. For now, only used for IDIV and FDIV.  Expected minor performance improvements.

Before merging, need to figure out how to adjust cosim module to see in-order writebacks.  Most like need to reorder them in the non-synth module.

Partially in support of #718


taylor-bsg:
@tommydcjung

PR #732, opened 2021-01-15
Hotfixes for cosim

dpetrisko:
- MPRVEN initial value should be 0, since we want to disable translation during debug mode
- disable mprv-based translation during debug mode
- Also small fix to watchdog timer, which should better detect "hardware idle loops"


farzamgl (approved):
Looks good. Do we need to keep mprven hardwired though?

dpetrisko:
Oh yeah that’s a good point, guess we don’t. I’ll leave it with a comment for now, but we should look into more debug mode behavior. I think riscv-tests has a decent suite (even if there are some SiFive specifics)

PR #733, opened 2021-01-18
Adding mstatus checking to dromajo

dpetrisko:
Datapath implementation is suboptimal but will get improved with upcoming CSR updates

### DIFF
--- a/bp_be/src/v/bp_be_calculator/bp_be_csr.sv
+++ b/bp_be/src/v/bp_be_calculator/bp_be_csr.sv
@@ -57,6 +57,7 @@ module bp_be_csr
   bp_cfg_bus_s cfg_bus_cast_i;
   bp_be_csr_cmd_s csr_cmd;
   bp_be_exception_s exception;
+  rv64_instr_s exception_instr;
 
   bp_be_wb_pkt_s wb_pkt_cast_i;
   bp_be_commit_pkt_s commit_pkt_cast_o;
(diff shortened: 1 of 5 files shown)


bp_be/src/v/bp_be_calculator/bp_be_csr.sv
```diff
@@ -661,7 +663,16 @@ module bp_be_csr
       fcsr_li.fflags |= fflags_acc_i;
 
       // Set FS to dirty if: fflags set, frf written, fcsr written
-      mstatus_li.fs |= {2{|fflags_acc_i | frf_w_v_i | (csr_cmd_v_i & (csr_cmd.csr_addr == `CSR_ADDR_FCSR))}};
+      // TODO: Should pre_decode this write, but requires multiple changes to the datapath
+      //   For now, a few comparators will do
+      mstatus_li.fs |= {2{(csr_cmd_v_i & (csr_cmd.csr_addr == `CSR_ADDR_FCSR))}}
+                       || {2{exception_v_i & exception_instr.t.rtype.opcode inside {`RV64_FLOAD_OP
```
farzamgl, line 669: Why are we writing this on FP exceptions and not FP instructions?

dpetrisko: it's a bad name -- exception_v_i is high on any committed instruction

dpetrisko:
So there's an interesting problem here. Basically, we do a check based on the "next" value of the CSR. But read-only fields of the CSR are generated from the latched version. Probably the best way to fix is to latch the commit signals before they go into the fifo, that way they will be synchronized with the csr after-effects

Issue #715, opened 2020-12-22
package / macro issue


taylor-bsg:
https://github.com/black-parrot/black-parrot/blob/4e829d9a91786bbf99af03fc9d1837540189d4f7/bp_me/src/include/v/bp_cce_pkg.v#L10

packages do not contain macros, so this works just because the order of systemverilog files is being specified in a way that this should be read in. should be a .vh file?

PR #737, opened 2021-01-25
Breaking reliance on `define order

### LINKED ISSUE #714: package should actually be include file
https://github.com/black-parrot/black-parrot/blob/4e829d9a91786bbf99af03fc9d1837540189d4f7/bp_me/src/include/v/bp_cce_pkg.v#L10

packages do not contain macros, so this works just because the order of systemverilog files is being specified in a way that this should be read in. should be a .vh file?

dpetrisko:
Fixes #714 #715

Should improve tool compatibility, no longer relies on compile order.

- All defines go into top level defines: common_defines, fe_defines, be_defines, me_defines, subdefines are `included there
- Only bp_common_pkg, bp_fe_pkg, bp_be_pkg, bp_me_pkg top level packages remain
- What was before a subpackage is now a "_pkgdef" or a set of parameters and typedefs which need to be `included into a "_pkg"
- ~It's possible we'll also need to at least `include bp_common_defines on every file~ 
- At least verilator and vcs are fine with this right now. Other tool users can file bug reports

@muwyse this required some renaming in the ME, mostly macros->$bits().  Let me know if there's anything controversial here


dpetrisko:
Lumping in a ton of cleanup changes here. All parameters which are implicitly encoded into macros are now _gp instead of _p (or should be). All modules have their own `includes generated. It's possible there were some spurious includes generated by my sed.

I also split off some of the includes in bp_common.  None of these changes should break anything (except some FPGA tcl scripts)

muwyse (approved):
ME looks fine to me.

PR #745, opened 2021-01-31
[WIP] PTW Injection

dpetrisko:
- Injecting PTW exceptions into pipeline instead of out-of-band
- Injecting interrupts into pipeline instead of out-of-band
- Decouples CSR illegal access from trap
- Moves CSR access into EX1 (saves a bunch of unnecessary pipe registers)
- splits BE redirects into exceptions and special instructions


dpetrisko:
Superceded by #766

PR #748, opened 2021-02-05
[WIP] Adding amo ops to D$

dpetrisko:
- Adds optional L1 amo op support to D$
- Adds L1 atomic config to aviary


dpetrisko:
@farzamgl This doesn't work with COSIM, because dromajo expects a trap whenever we encounter an atomic instruction. Can you add a flag to dromajo which will enable / disable trapping on atomics? And then in Makefile.cfgs, set that flag when CFG=e_bp_unicore_l1_atomic_cfg?

farzamgl:
@dpetrisko added the flag. Can you confirm if it fixes the cosim issues?

dpetrisko:
works perfectly!

PR #749, opened 2021-02-05
WFI

dpetrisko:
This PR adds support for the WFI instruction.  Upon WFI, it freezes the BE sets the FE to the wait state, until an interrupt is taken.

Dependent on https://github.com/black-parrot/bp_tests/pull/7

This PR is mostly functional. However, we also need to add a case for an interrupt occuring while interrupts are globally disabled. This should wake the processor, despite it being frozen. I'm holding off adding that bit until #745 is merged, because interrupt behavior is changing substantially.

@zaazad It should be sufficient to run accelerator tests, provided you don't rely on this behavior

### DIFF
--- a/bp_be/src/include/bp_be_defines.svh
+++ b/bp_be/src/include/bp_be_defines.svh
@@ -129,6 +129,7 @@
       logic                           fencei;                                                      \
       logic                           sfence;                                                      \
       logic                           satp;                                                        \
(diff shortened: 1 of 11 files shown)


dpetrisko:
@zaazad The resume on disabled interrupt feature is now functional.  So you could modify the interrupt handler for accelerators, just wait for the disabled interrupt

PR #755, opened 2021-02-07
Minimizing cache tags

dpetrisko:
Currently, we store the entire physical tag in the caches. This is wasteful, because the upper bits of address are used for uncacheable memory. (Thanks to @Shashank-Vijay for pointing this out!)

- [X] Add new ctag parameter
- [X] Shrink cache tags to the cacheable memory size (248->184, -25.8%)
- [X] Shrink directory stores to cacheable memory size (496 -> 368, -25.8%)
- [X] Shrink L2 tag size (232->168, -27.6%)

Larger savings on larger (uncacheable) address space!


bp_me/src/v/lce/bp_lce_cmd.sv
```diff
@@ -39,7 +39,7 @@ module bp_lce_cmd
     , localparam lg_block_size_in_bytes_lp = `BSG_SAFE_CLOG2(block_size_in_bytes_lp)
 
    `declare_bp_bedrock_lce_if_widths(paddr_width_p, cce_block_width_p, lce_id_width_p, cce_id_width_p, lce_assoc_p, lce)
-   `declare_bp_cache_engine_if_widths(paddr_width_p, ptag_width_p, sets_p, assoc_p, dword_width_gp, block_width_p, fill_width_p, cache)
+   `declare_bp_cache_engine_if_widths(paddr_width_p, ctag_width_p, sets_p, assoc_p, dword_width_gp, block_width_p, fill_width_p, cache)
```
muwyse, line 42: I think the LCE will need more changes than just using ctag_width_p in the macros. It likely does something with the actual address to/from cache commands and LCE req/cmd/resp messages.

muwyse (changes requested):

bp_me/src/v/lce/bp_lce_cmd.sv
dpetrisko: The outgoing messages are all the full address - the tag parameter is only used for tag_mem_pkts. Or at least, that’s what is intended :)

bp_common/src/include/bp_common_addr_pkgdef.svh
```diff
@@ -41,5 +41,8 @@
   localparam coproc_base_addr_gp       = 40'h10_0000_0000;
   localparam global_base_addr_gp       = 40'h20_0000_0000;
 
+  // 4 GB
```
taylor-bsg, line 44: feels like this should be in a machine config struct? all of these other addresses more or less are located on chip and will be self-consistent with the verilog (right?), but feels like people may be surprised to learn that their 64-bit processor can only do 4GB

dpetrisko: Yeah that's a good point, will update

bp_me/src/v/lce/bp_lce_cmd.sv
dpetrisko: Looking through the lce_cmd code nothing stands out to me as depending on particular address bits, and I've confirmed in the cache_engine macros tag_width is only used in tag_mem_pkt_s and tag_info_s. Do you see any other problematic cases?

PR #756, opened 2021-02-11
Update README.md

taylor-bsg:


dpetrisko:
Maybe it makes sense to have the actual project details in an issue and link to it? Or have the list in a Google doc? That would make it so we don’t have to update this README every time we get a new project

taylor-bsg:
I think it’s good to put the projects front and center in the readme. So they see it while they are setting up BP and say, oh hey, that sounds pretty easy, I could do that. Linking to details in a github issue sounds good. Not all issues are good projects, so I think the curation is good.

dpetrisko (approved):

PR #759, opened 2021-02-12
Blocks interrupts on pending cache replay

dpetrisko:
This PR blocks interrupts while there's a cache replay being performed. This is necessary for making sure that uncached operations are not repeated, which can break things in non-idempotent remote memory space.  Slight penalty for interrupt responsiveness, definitely a bandaid fix until we get non-blocking loads

### DIFF
--- a/bp_be/src/v/bp_be_checker/bp_be_detector.sv
+++ b/bp_be/src/v/bp_be_checker/bp_be_detector.sv
@@ -41,10 +41,11 @@ module bp_be_detector
    , input                             mem_ready_i
    , input                             ptw_busy_i
    , input                             irq_pending_i
+   , input                             replay_pending_i
 
    // Pipeline control signals from the checker to the calculator
-   , output                            chk_dispatch_v_o
-   , output                            chk_interrupt_v_o
+   , output logic                      chk_dispatch_v_o
+   , output logic                      chk_interrupt_v_o
    , input [dispatch_pkt_width_lp-1:0] dispatch_pkt_i
    , input [commit_pkt_width_lp-1:0]   commit_pkt_i
    , input [wb_pkt_width_lp-1:0]       iwb_pkt_i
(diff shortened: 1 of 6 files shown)


bp_be/src/v/bp_be_dcache/bp_be_dcache.sv
```diff
@@ -522,7 +523,7 @@ module bp_be_dcache
     ? (sc_success != 1'b1)
     : early_data;
 
-  assign early_v_o = v_tv_r & ( (uncached_op_tv_r & (decode_tv_r.load_op & uncached_load_data_v_r))
+  assign early_v_o = v_tv_r & ( (uncached_op_tv_r & (decode_tv_r.load_op & uncached_pending_r))
```
Shashank-Vijay, line 526: For this, are we using the "pending" signal even though it will be 1 for the entire miss time because the other ones are sure to be 0 during that time?

bp_be/src/v/bp_be_checker/bp_be_detector.sv
```diff
@@ -287,7 +288,7 @@ module bp_be_detector
   // Generate calculator control signals
   assign chk_dispatch_v_o  = ~(control_haz_v | data_haz_v | struct_haz_v) & ~irq_pending_i;
   // TODO: Inject into pipeline rather than "happening" at EX3
-  assign chk_interrupt_v_o = irq_pending_i & ~cfg_bus_cast_i.freeze & ~ptw_busy_i & ~dep_status_r[2].instr_v;
+  assign chk_interrupt_v_o = irq_pending_i & ~replay_pending_i & ~cfg_bus_cast_i.freeze & ~ptw_busy_i & ~dep_status_r[2].instr_v;
```
Shashank-Vijay, line 291: Didn't know I could use this signal back when I wrote the hack fix for ParrotPiton use :D

bp_be/src/v/bp_be_dcache/bp_be_dcache.sv
dpetrisko: Yeah, an uncached load will flush the pipeline, so the replay is guaranteed to be the next instruction

Issue #730, opened 2021-01-12
Remove buffering in lite to stream modules


dpetrisko:
https://github.com/black-parrot/black-parrot/blob/master/bp_me/src/v/wormhole/bp_lite_to_stream.sv#L72
https://github.com/black-parrot/black-parrot/blob/master/bp_me/src/v/wormhole/bp_stream_to_lite.sv#L73

Not sure we have a configuration that uses this yet, but can be made to use passthrough piso/sipos
Depends on https://github.com/bespoke-silicon-group/basejump_stl/pull/232

dpetrisko:
https://github.com/black-parrot/black-parrot/blob/me_dev_streams/bp_me/src/v/wormhole/bp_stream_to_lite.v#L45

This is super inefficient right now. It'll introduce a bubble after every stream beat.  Instead, it should latch the base stream address and compose the outgoing header based on that address (handle wraparound, etc)

dpetrisko:
Fixed in #760

PR #760, opened 2021-02-13
Remove Buffering in lite to stream modules

Songchun-Li:
This PR is aimed at fixing issue #730.

In current design of lite to stream modules, header and data buffers account for a majority of area consumption. Also, a bubble is introduced between consecutive requests, which is very inefficient, especially for uncached requests.

This PR replaces the header fifo with bypass registers, and make use of the sipo/piso_passthrough_dynamic to hold the data. These changes not only reduce the area consumption of the conversion modules, but also eliminate the bubbles between consecutive requests and improve the performance.

### DIFF
--- a/bp_me/src/v/cache/bp_me_cache_slice.sv
+++ b/bp_me/src/v/cache/bp_me_cache_slice.sv
@@ -91,7 +91,7 @@ module bp_me_cache_slice
 
     ,.in_msg_header_i(mem_header_li)
     ,.in_msg_data_i(mem_data_li)
-    ,.in_msg_v_i(mem_ready_lo & mem_v_li)
+    ,.in_msg_v_i(mem_v_li)
     ,.in_msg_ready_and_o(mem_ready_lo)
     ,.in_msg_last_i(mem_last_li)
 
--- a/bp_me/src/v/wormhole/bp_lite_to_stream.sv
+++ b/bp_me/src/v/wormhole/bp_lite_to_stream.sv
@@ -1,5 +1,5 @@
 
-module bp_lite_to_stream
+module bp_lite_to_stream_bypass
  import bp_common_pkg::*;
  import bp_common_aviary_pkg::*;
  import bp_me_pkg::*;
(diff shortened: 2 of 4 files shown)


bp_me/src/v/cache/bp_me_cache_slice.sv
```diff
@@ -91,7 +91,7 @@ module bp_me_cache_slice
 
     ,.in_msg_header_i(mem_header_li)
     ,.in_msg_data_i(mem_data_li)
-    ,.in_msg_v_i(mem_ready_lo & mem_v_li)
+    ,.in_msg_v_i(mem_v_li)
```
Songchun-Li, line 94: `mem_v_li` from stream_pump_out is helpful. By removing the mem_ready_lo here, we do not need to edit the sipo_passthrough_dynamic to register the last data packet.

dpetrisko:
Does this rely on the passthrough PISO handshake fix we discussed?

Songchun-Li:
> Does this rely on the passthrough PISO handshake fix we discussed?

I think we actually do not need to fix that handshake, with the commented change I added above

dpetrisko:
>I think we actually do not need to fix that handshake, with the commented change I added above. The conversion modules works perfectly now and passes regression test with varying fill_width

Nice, that's great. I think it's still worth fixing, just to make it so future users don't have to think about this case

Songchun-Li:
The work of fixing the handshale issue of sipo_passthrough_dynamic is [here](https://github.com/bespoke-silicon-group/basejump_stl/pull/399). It is tested within the stream_to_lite. Is there any other way to test this module?

dpetrisko:
There are testbenches for bsg_sipo_passthrough in basejump_stl/testing/bsg_dataflow/... , if you could copy that testbench and extend it to _dynamic, that would be very helpful

bp_me/src/v/wormhole/bp_lite_to_stream.sv
```diff
@@ -40,71 +40,90 @@ module bp_lite_to_stream
...
-
-     // We rely on the header fifo to handle ready/valid handshaking
-     ,.len_v_o(/* Unused */)
-     ,.ready_o(/* Unused */)
-     );
+    ,.reset_i(reset_i)
+
+    // data_i and v_i should hold during the entire transcation
+    ,.data_i(in_msg_data_lo)
+    ,.v_i(in_msg_v_i | streaming_r)
```
dpetrisko, line 101: what if the input changes during a stream? Wouldn't we load new data accidentally?

Songchun-Li: That's a good catch. I made some changes in both modules to prevent non-intended data update on the input side.

bp_me/src/v/wormhole/bp_lite_to_stream.sv
```diff
@@ -40,71 +40,90 @@ module bp_lite_to_stream
...
+  bp_bedrock_in_msg_header_s in_msg_header_lo;
+  logic [in_data_width_p-1:0] in_msg_data_lo;
+  // Hold the data and header for multi-cycle streaming
+  bsg_dff_en_bypass
    #(.width_p($bits(bp_bedrock_in_msg_header_s)))
-   header_fifo
+   header_reg
     (.clk_i(clk_i)
-     ,.reset_i(reset_i)
+    ,.en_i(in_msg_ready_and_o & in_msg_v_i)
```
Songchun-Li, line 57: in_msg_ready_and_o is low during streaming, so input header and data will not be updated accidently during streaming

bp_me/src/v/wormhole/bp_stream_to_lite.sv
```diff
@@ -44,57 +44,59 @@ module bp_stream_to_lite
...
-
-     // We use the sipo ready/valid
-     ,.ready_o(/* Unused */)
-     ,.v_o(/* Unused */)
-     );
-
-  bp_bedrock_in_msg_header_s msg_header_cast_i;
-  assign msg_header_cast_i = in_msg_header_i;
-  wire has_data = payload_mask_p[msg_header_cast_i.msg_type];
+    ,.en_i(in_msg_ready_and_o & in_msg_v_i & ~streaming_r)
```
Songchun-Li, line 56: With ~streaming_r, it latches the first stream header with the critical word address as the lite header.

dpetrisko (approved):

Issue #736, opened 2021-01-24
Support Iterative Filling in Direct Mapped Cache Configurations


sripathi-muralitharan:
Currently, the direct-mapped caches are capable of supporting fills of size 512. This means that the downstream serialization/deserialization/buffering logic would have to be 512-bit wide. 

We should support iterative filling in direct-mapped caches because - 
1. We are not fully parameterizable until we do.
2. It has a large area overhead which is not conducive to creating a Tiny BlackParrot.

One possible fix could be to bank the data memories to have 8 banks of 64-bit widths instead of 1 bank with 512-bit width. Modifying the banking scheme this way also has the benefit of creating square memories which have good area and power properties.

dpetrisko:
Disagree on the shape of the rams being better for 64x64, they’re pretty well optimized at 512x64. If we really wanted to bank them, we could physically bank the hardened swap rams. 

I agree that it doesn’t make much sense to have a 512-bit cache line with a direct-mapped cache. Parameterization is good for supporting different use-cases, but parameterization for parameterization’s sake just makes it harder to maintain. Shrinking the line size and matching the fill size is better for TinyParrot, no? That would minimize the number of data registers in the cache. 

There’s a prefetching argument for keeping to say, 128-bit lines rather than 64. But that’s obviously workload dependent and you could effectively do the same thing with a simple next-line prefetcher. 

In general, I don’t hate the idea. It would require modifying the Cache Engine Interface. But the hardware cost of changing the fill mask to a sub-bank mask is not terrible, assuming we enforce alignment of sub-banks.

sripathi-muralitharan:
Shrinking the line size and matching to fill size would be fine from an area perspective (provided we physically depth-bank the tall memory if we still want a 4KB cache) but will have a large performance hit based on workload. Given that the tiny core is expected to be Linux capable, can't we assume the workloads might be such that a fill width = line width cache makes the core terrible? (Not sure what other tiny linux capable cores have...)

As I see it now, we are restricted to have bank widths to be atleast fill width size and having sub-block filling would be pretty useful if we want to take full advantage of the stream and burst protocols right?

dpetrisko:
> Shrinking the line size and matching to fill size would be fine from an area perspective

Yeah, the exercise is to make the tiniest BlackParrot possible. Performance tuning comes after we profile the workload.
[...]

PR #761, opened 2021-02-15
Sub-Bank Masking in the D$

sripathi-muralitharan:
Currently, the iterative filling strategy supports fill widths of size equal to the bank width or higher. This PR adds the capability to partially fill banks upto a minimum width of 64-bits in the unicore configuration.

Other fixes: The D$ testbench was outdated with recent changes so there are some fixes for that as well. 

Testing: 
1. Tested with 64-bit fill widths, 128-bit bank width and 512-bit block width on the RISC-V regression suite.
2. Tested with the D$ testbench regression suite at the same config above. 
3. I added a config and included that config to ci/weird_config.sh script for CI testing.


bp_top/test/common/bp_nonsynth_if_verif.sv
```diff
@@ -72,8 +72,8 @@ module bp_nonsynth_if_verif
     $fatal("Error: Writethrough with coherent_l1 is unsupported");
   if ((icache_fill_width_p > icache_block_width_p) || (dcache_fill_width_p > dcache_block_width_p))
     $fatal("Error: Cache fill width should be less or equal to L1 cache block width");
-  if ((icache_fill_width_p % (icache_block_width_p/icache_assoc_p) != 0) || (dcache_fill_width_p % (dcache_block_width_p / dcache_assoc_p) != 0))
-    $fatal("Error: Cache fill width should be a multiple of cache bank width");
+  if ((icache_fill_width_p < 64) || (dcache_fill_width_p < 64))
```
dpetrisko, line 75: This is a less stringent constraint -- we don't support 65 bit fill, for example

bp_be/src/v/bp_be_dcache/bp_be_dcache.sv
```diff
@@ -178,6 +178,11 @@ module bp_be_dcache
   localparam block_offset_width_lp    = (assoc_p > 1)
     ? (bindex_width_lp+byte_offset_width_lp)
     : byte_offset_width_lp;
+  localparam sub_fill_size_in_bank_lp = bank_width_lp / fill_width_p;
```
dpetrisko, line 181: this seems backwards? bank size in fill?

sripathi-muralitharan: That's right. I take it to mean number of fills required to fill a bank. I can rename it to say `bank_size_in_fill_lp`

bp_top/test/common/bp_nonsynth_if_verif.sv
sripathi-muralitharan: Hmm, I can maybe write it as - `BSG_SAFE_CLOG2(*cache_fill_width_p) != 8, 16, 32 or 64`

bp_common/src/include/bp_common_cache_engine_if.svh
```diff
@@ -65,12 +65,14 @@
       logic [`BSG_SAFE_CLOG2(ways_mp)-1:0]            way_id;                                 \
       logic [fill_width_mp-1:0]                       data;                                   \
       logic [(block_data_width_mp/fill_width_mp)-1:0] fill_index;                             \
+      logic [`BSG_MAX(1, `BSG_SAFE_CLOG2((block_data_width_mp/fill_width_mp)/ways_mp))-1:0] sub_fill_index; \
```
dpetrisko, line 68: I think just fill_index is sufficient? block_data_width/fill_width should give the correct index, even if fill_width < bank size

bp_me/src/v/cce/bp_uce.sv
```diff
@@ -286,10 +288,17 @@ module bp_uce
   logic [fill_cnt_width_lp-1:0] mem_cmd_cnt;
   logic [block_size_in_fill_lp-1:0] fill_index_shift;
   logic [bank_offset_width_lp-1:0] bank_index;
+  logic [byte_offset_width_lp-1:0] sub_bank_index;
   logic [paddr_width_p-1:0] critical_addr;
 
-  assign bank_index = mem_cmd_cnt << bank_sub_offset_width_lp;
-  // fill_index_shift corresponds to a single data bank in a direct-mapped cache
+  assign bank_index = (fill_size_in_bank_lp != 0)
```
dpetrisko, line 294: Yeah I think this will be cleaner just having a single fill_index. The cache itself can slice of the subbank index

bp_be/src/v/bp_be_dcache/bp_be_dcache.sv
```diff
@@ -937,12 +942,44 @@ module bp_be_dcache
   ///////////////////////////
   // Data Mem Control
   ///////////////////////////
-  logic [block_size_in_fill_lp-1:0][fill_size_in_bank_lp-1:0] data_mem_pkt_fill_mask_expanded;
-  for (genvar i = 0; i < block_size_in_fill_lp; i++)
+  logic [banks_per_block_lp-1:0][`BSG_MAX(fill_size_in_bank_lp, 1)-1:0] data_mem_pkt_fill_mask_expanded;
```
dpetrisko, line 945: looks like this didn't get updated, fill_size_in_bank

dpetrisko: or below

sripathi-muralitharan: No, here I want to use `fill_size_in_bank` (as before) to indicate which bank should be the one I fill to. I generate a separate bank mask below that will get ANDed with this one.

bp_be/src/v/bp_be_dcache/bp_be_dcache.sv
```diff
@@ -937,12 +942,44 @@ module bp_be_dcache
...
+  logic [data_mem_mask_width_lp-1:0] bank_mask;
+  if (fill_size_in_bank_lp != 0)
+   begin : no_sub_bank
+     assign bank_mask = {data_mem_mask_width_lp{1'b1}};
+   end
+  else 
+   begin : sub_bank
+    logic [bank_size_in_fill_lp-1:0] sub_fill_index_dec;
+    bsg_decode
+      #(.num_out_p(bank_size_in_fill_lp))
```
dpetrisko, line 967: if bsg_decode has width 0, is o== 1'b1? what about width 1?

sripathi-muralitharan: For width 1, o = 1'b1. There doesn't seem to be zero width support for it now. If num_out_p was 0, it would actually create a 2-bit output because of the range becoming [-1:0]

bp_be/src/v/bp_be_dcache/bp_be_dcache.sv
dpetrisko: oh I didn't know both variables existed

bp_be/src/v/bp_be_dcache/bp_be_dcache.sv
dpetrisko: so if we make fill_size_in_bank and bank_size_in_fill minimum 1 I think both these special cases go away? Does that cause problems elsewhere?

bp_common/src/include/bp_common_cache_engine_if.svh
sripathi-muralitharan: Block width / fill width gives the number of fills required to fill that block. Earlier when we had fill widths >= bank widths this could sufficiently address the case of filling a single bank or multiple banks. Now we want to have a way to address locations within a bank as well where the sub fill index helps. 

fill index --> which bank
sub fill index --> location within bank

Maybe combine both into a 2-D array like [bank _index - 1: 0][within_bank_index-1:0] although assignment might become slightly complicated (I think)

bp_me/src/v/cce/bp_uce.sv
sripathi-muralitharan: Same reasoning as above. I don't think we can have just one variable tracking both the physical bank and the logical bank. 

Example: 
512-bit block, 128-bit bank, 64-bit fills. 4 physical banks, 8 logical banks.

bp_be/src/v/bp_be_dcache/bp_be_dcache.sv
sripathi-muralitharan: There shouldn't be any problems. A case where both become 1 is when we have 512-bit blocks and 64-bit fills with 64-bit banks. In that case, the conditional statement defaults to what we had before.

bp_common/src/include/bp_common_cache_engine_if.svh
dpetrisko: block_width = 512
fill_width = 64
direct mapped.


fill_index ranges from 0-7 but we have only 1 bank...

bp_be/src/v/bp_be_dcache/bp_be_dcache.sv
sripathi-muralitharan: The minimums for both of them is actually 0 since clog2(0) is 0. Even BSG_SAFE_CLOG2(0) is 0. Don't know if we want to be safe wrt zero as well but I think it should be reasonable.

bp_me/src/v/cce/bp_uce.sv
dpetrisko: physical bank is just a bit slice of logical bank

dpetrisko: which is what I'm calling fill_index :)

taylor-bsg (commented):
May I suggest that you pull out the partial SRAM use into a cleanly abstracted module in BaseJump STL? I think that thinking about that interface will make the logic cleaner here, and also result in a reusable module. 

See for example, this module, https://github.com/bespoke-silicon-group/basejump_stl/blob/master/bsg_mem/bsg_mem_1rw_sync_mask_write_bit_banked.v, which does the opposite. :-)

dpetrisko:
That’s a great idea! Would solve the byte vs non-byte masked parameterization in the icache, as well

bp_common/src/include/bp_common_cache_engine_if.svh
sripathi-muralitharan: When assoc_p is 1, fill index is set to 0. So no, it won't. Whereas sub_fill_index will do what you just said.

bp_me/src/v/cce/bp_uce.sv
sripathi-muralitharan: Isn't it the other way around. For the sub-bank fill case, we have smaller but more logical banks. So logical banks become bit slices of physical banks right?

dpetrisko: What I mean is that having more logical banks means the logical bank index is larger. The physical bank index is a bit slice within the logical bank address space

bp_common/src/include/bp_common_cache_engine_if.svh
dpetrisko: discussed offline

bp_common/src/include/bp_common_cache_engine_if.svh
```diff
@@ -65,12 +65,14 @@
       logic [`BSG_SAFE_CLOG2(ways_mp)-1:0]            way_id;                                 \
       logic [fill_width_mp-1:0]                       data;                                   \
       logic [(block_data_width_mp/fill_width_mp)-1:0] fill_index;                             \
```
dpetrisko, line 67: I think this should be `BSG_SAFE_CLOG2(block_data_width_mp/fill_width_mp)-1:0], incidentally

dpetrisko:
Made an issue for the subbanked ram: https://github.com/bespoke-silicon-group/basejump_stl/issues/400

dpetrisko:
fixes #736

bp_me/src/v/cce/bp_uce.sv
sripathi-muralitharan: Have a look at what's there now. I think its cleaner.

bp_common/src/include/bp_common_cache_engine_if.svh
sripathi-muralitharan: Yup, refactored to do this.

bp_common/src/include/bp_common_aviary_pkgdef.svh
```diff
@@ -389,6 +389,22 @@
                         ,bp_unicore_cfg_p
                         );
 
+  localparam bp_proc_param_s bp_unicore_l1_sub_bank_mask_override_p =
```
dpetrisko, line 392: nit;

bp_unicore_l1_subbank_override_p

bp_fe/src/v/bp_fe_icache.sv
```diff
@@ -571,12 +574,27 @@ module bp_fe_icache
...
+
+  logic [(2**bank_fill_index_width_lp)-1:0] bank_fill_index_dec;
+    bsg_decode
+      #(.num_out_p(2**bank_fill_index_width_lp))
+      bank_fill_index_decoder
+      (.i(bank_fill_index)
+      ,.o(bank_fill_index_dec)
+      );
+
+  logic [(2**bank_fill_index_width_lp)-1:0][`BSG_MAX(fill_size_in_bank_lp, 1)-1:0] data_mem_pkt_fill_mask_expanded;
```
dpetrisko, line 589: logic [assoc_p-1:0]?

bp_fe/src/v/bp_fe_icache.sv
```diff
@@ -571,12 +574,27 @@ module bp_fe_icache
...
+  bsg_expand_bitmask
+    #(.in_width_p(2**bank_fill_index_width_lp)
+    ,.expand_p(`BSG_MAX(1, fill_size_in_bank_lp))
```
dpetrisko, line 592: Probably should modify bsg_expand_bitmask so that .expand_p(0) also acts as passthrough (like 1)

bp_fe/src/v/bp_fe_icache.sv
```diff
@@ -571,12 +574,27 @@ module bp_fe_icache
...
 
-  logic [block_size_in_fill_lp-1:0][fill_size_in_bank_lp-1:0] data_mem_pkt_fill_mask_expanded;
-  // use fill_index to generate bank write mask
-  for (genvar i = 0; i < block_size_in_fill_lp; i++)
-    begin : fill_mask
-      assign data_mem_pkt_fill_mask_expanded[i] = {fill_size_in_bank_lp{data_mem_pkt_cast_i.fill_index[i]}};
-    end
+  // Using BSG_MAX gets rid of illegal part select errors
+  logic [`BSG_SAFE_MINUS(bank_fill_index_width_lp, 1):0] bank_fill_index;
+  assign bank_fill_index  = (assoc_p > 1) ? data_mem_pkt_cast_i.fill_index[lg_num_subbanks_lp+:`BSG_MAX(bank_fill_index_width_lp, 1)] : '0;
```
dpetrisko, line 579: bsg_decode .num_out_p(1) should always be 1, so no need to make zero explictly

bp_be/src/v/bp_be_dcache/bp_be_dcache.sv
```diff
@@ -980,12 +983,57 @@ module bp_be_dcache
   ///////////////////////////
   // Data Mem Control
   ///////////////////////////
-  logic [block_size_in_fill_lp-1:0][fill_size_in_bank_lp-1:0] data_mem_pkt_fill_mask_expanded;
-  for (genvar i = 0; i < block_size_in_fill_lp; i++)
-    begin : fill_mask
-      assign data_mem_pkt_fill_mask_expanded[i] = {fill_size_in_bank_lp{data_mem_pkt_cast_i.fill_index[i]}};
-    end
+  // Using BSG_MAX gets rid of illegal part select errors
```
dpetrisko, line 986: Same comments as I$ apply

bp_fe/src/v/bp_fe_icache.sv
```diff
@@ -121,6 +121,9 @@ module bp_fe_icache
     : byte_offset_width_lp;
   localparam block_size_in_fill_lp  = block_width_p / fill_width_p;
   localparam fill_size_in_bank_lp   = fill_width_p / bank_width_lp;
+  localparam num_subbanks_lp          = (fill_size_in_bank_lp != 0) ? 0 : block_size_in_fill_lp / assoc_p;
+  localparam lg_num_subbanks_lp       = `BSG_SAFE_CLOG2(num_subbanks_lp);
```
dpetrisko, line 125: I don't think BSG_SAFE_CLOG2(0) does what you think... $clog2(0)

bp_fe/src/v/bp_fe_icache.sv
```diff
@@ -121,6 +121,9 @@ module bp_fe_icache
...
+  localparam bank_fill_index_width_lp = $clog2(block_size_in_fill_lp) - lg_num_subbanks_lp;
```
dpetrisko, line 126: `BSG_SAFE_CLOG2(block_width_p/fill_width_p)?

dpetrisko: or $clog2, as needed

dpetrisko (changes requested):

bp_fe/src/v/bp_fe_icache.sv
sripathi-muralitharan: Yeah, I want it to return 0 when num_subbanks_lp is 0.

bp_fe/src/v/bp_fe_icache.sv
sripathi-muralitharan: Yeah makes sense. Will get that in, in the next commit.

bp_fe/src/v/bp_fe_icache.sv
sripathi-muralitharan: Alright, will open a PR in basejump for that.

bp_fe/src/v/bp_fe_icache.sv
sripathi-muralitharan: Gotcha. My intention was actually to get rid of a build warning that says "Index out of bounds". But even with the change I have now, it doesn't go away. I need the expression to be the way it is to be general. This comes up when the cache is direct-mapped with 512 bit banks and 64 bit fills.

bp_fe/src/v/bp_fe_icache.sv
sripathi-muralitharan: But the argument to $clog2() as given is block_size_in_fill_lp which is exactly that. Is your intention to go for more clarity?

bp_be/src/v/bp_be_dcache/bp_be_dcache.sv
sripathi-muralitharan: Should I get rid of the BSG_MAX comment? Is it obvious why that macro was used?

bp_fe/src/v/bp_fe_icache.sv
sripathi-muralitharan: https://github.com/bespoke-silicon-group/basejump_stl/pull/423

bp_fe/src/v/bp_fe_icache.sv
dpetrisko: yeah

dpetrisko:
Just noting to myself, this PR needs to solve the critical word snooping problem (32-bit fill), resume on 64 rather than during the first word

dpetrisko:
Closing for staleness; someone else could pick this up some day

PR #765, opened 2021-02-18
Adding AMO subop types and cache req metadata updates

dpetrisko:
This PR adds AMO subop types and cache req metadata changes needed for L2 atomic support. I believe it has no functional changes as currently written.

### DIFF
--- a/bp_common/src/include/bp_common_bedrock_if.svh
+++ b/bp_common/src/include/bp_common_bedrock_if.svh
@@ -60,6 +60,7 @@
       logic [payload_width_mp-1:0]                 payload;                                 \
       bp_bedrock_msg_size_e                        size;                                    \
       logic [addr_width_mp-1:0]                    addr;                                    \
+      bp_bedrock_wr_subop_e                        subop;                                   \
       bp_bedrock_msg_u                             msg_type;                                \
     } bp_bedrock_``name_mp``_msg_header_s;                                                  \
                                                                                             \
@@ -157,7 +158,7 @@
    */
 
   `define bp_bedrock_msg_header_width(addr_width_mp, payload_width_mp) \
-    ($bits(bp_bedrock_msg_u)+addr_width_mp+$bits(bp_bedrock_msg_size_e)+payload_width_mp)
+    ($bits(bp_bedrock_msg_u)+$bits(bp_bedrock_wr_subop_e)+addr_width_mp+$bits(bp_bedrock_msg_size_e)+payload_width_mp)
 
   `define bp_bedrock_msg_width(addr_width_mp, payload_width_mp, data_width_mp) \
     (`bp_bedrock_msg_header_width(addr_width_mp, payload_width_mp)+data_width_mp)
--- a/bp_common/src/include/bp_common_bedrock_if.svh
+++ b/bp_common/src/include/bp_common_bedrock_if.svh
@@ -88,6 +88,7 @@
     {                                                                                  \
       logic [`BSG_SAFE_CLOG2(lce_assoc_mp)-1:0]    lru_way_id;                         \
       bp_bedrock_req_non_excl_e                    non_exclusive;                      \
+      bp_bedrock_wr_subop_e                        subop;                              \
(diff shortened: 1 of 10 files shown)


bp_common/src/include/bp_common_bedrock_if.svh
```diff
@@ -88,6 +88,7 @@
     {                                                                                  \
       logic [`BSG_SAFE_CLOG2(lce_assoc_mp)-1:0]    lru_way_id;                         \
       bp_bedrock_req_non_excl_e                    non_exclusive;                      \
+      bp_bedrock_wr_subop_e                        subop;                              \
```
muwyse, line 91: The subop type needs to be added to the mem message struct, too, since we want to communicate to the L2/Mem endpoint what amo op needs to occur. Adding it to the mem_payload_s would break the abstraction that the payload is opaque to the endpoint though. But perhaps this is acceptable? Otherwise, the bp_bedrock_mem_type_e needs to have the amo types added. They would fit in the current sized enum, but leave no room for future codes.

dpetrisko: That’s a great point. I think that abstraction is valuable

muwyse: From the ucode CCE perspective, there should be at least one bit we could free up in the instruction encodings to support all of the bedrock message types being 1 bit wider, which would give room for expansion in the mem commands after adding amo

dpetrisko: Without the subop, we’ll need to have a noreturn bit, though

dpetrisko: How is the mem_cmd payload generated in the CCE ucode? Is it all from metadata or is some of it from the instruction itself?

dpetrisko: Coming back this, I guess this is a weird divergence from the LCE bedrock messages. There, the payload is exposed to the endpoint. Here, we expect it's not

muwyse: Right. Maybe we just say the payload is exposed. In the mem messages then the only thing that matters to the endpoint is the subop and they just leave the rest unchanged?

If we truly wanted the payload information to be hidden from the networks or endpoints, then it really ought to be "cached" by the CCE and not even exposed externally.

I think for now it is okay to have the mix of subop is exposed while the rest should not be used. The contract is still that the memory will return the payload as-is to the CCE though.

muwyse: > How is the mem_cmd payload generated in the CCE ucode? Is it all from metadata or is some of it from the instruction itself?

I'll need to double check but I think it is primarily generated by the CCE hardware (i.e., it is based on the results of the directory read and processing and not necessarily encoded in a ucode instruction field)

dpetrisko: >I think for now it is okay to have the mix of subop is exposed while the rest should not be used. The contract is still that the memory will return the payload as-is to the CCE though.

I think that's perfect. The caveat on not using the payload is really more about preventing future incompatibility than anything else.

> I'll need to double check but I think it is primarily generated by the CCE hardware (i.e., it is based on the results of the directory read and processing and not necessarily encoded in a ucode instruction field)

Here's a proposal: wr_subop_flag register. When we send a mem_cmd from the CCE, it uses the subop from this flag. When a new LCE request is popped off the queue, it auto-populates the wr_subop_flag with the subop in the LCE request.  It can also be manually read/written by ucode.

This would mean 0-cycle overhead for LCE->MEM AMOs, and only 1 cycle for custom messages.

What do you think?

muwyse: Works for me. We can add register to CCE so ucode can read/write the subop if needed for custom messages.

dpetrisko: Great!

Issue #729, opened 2021-01-12
Remove buffering in lite to burst modules


dpetrisko:
https://github.com/black-parrot/black-parrot/blob/master/bp_me/src/v/wormhole/bp_lite_to_burst.sv#L71
https://github.com/black-parrot/black-parrot/blob/master/bp_me/src/v/wormhole/bp_burst_to_lite.sv#L81

Can make use of passthrough pisos and sipos

Depends on https://github.com/bespoke-silicon-group/basejump_stl/pull/232

dpetrisko:
Fixed in #769

PR #769, opened 2021-02-23
Remove Buffering in lite to burst modules

Songchun-Li:
This PR is aimed at fixing issue #729.

In current design of lite to burst conversion modules, header and data buffers account for a majority of area consumption. Also, a bubble is introduced between consecutive no-data request, which is inefficient.

This PR replaces the header fifo with bypass registers, and make use of the sipo/piso_passthrough_dynamic to hold the data. These changes not only reduce the area consumption of the conversion modules, but also pass through the header and data from the input to output at the same cycle.

This change passes weird configure regression.


bp_me/src/v/wormhole/bp_lite_to_burst.sv
```diff
@@ -48,45 +48,84 @@ module bp_lite_to_burst
...
+    ,.data_i(in_msg_cast_i.header)
+    ,.data_o(header_lo) 
+    );
+  
+  // header valid logic
+  logic header_v_r, header_clear;
+  bsg_dff_reset_set_clear
+   #(.width_p(1)
+   ,.clear_over_set_p(1))
+    header_v_loeg
```
dpetrisko, line 67: header_v_reg

(search and replace? :))

bp_me/src/v/wormhole/bp_lite_to_burst.sv
```diff
@@ -48,45 +48,84 @@ module bp_lite_to_burst
...
+  wire has_data = payload_mask_p[header_lo.msg_type];
   localparam data_len_width_lp = `BSG_SAFE_CLOG2(burst_words_lp);
-  wire [data_len_width_lp-1:0] num_burst_cmds = `BSG_MAX(1, (1'b1 << in_msg_cast_i.header.size) / out_data_bytes_lp);
-  logic [out_data_width_p-1:0] data_lo;
-  bsg_parallel_in_serial_out_dynamic
-   #(.width_p(out_data_width_p), .max_els_p(burst_words_lp))
-   piso
+  wire [data_len_width_lp-1:0] num_burst_cmds = `BSG_MAX(1, (1'b1 << header_lo.size) / out_data_bytes_lp);
+
+  logic data_bursting_r, data_burst_clear, data_ready_and_lo;
```
dpetrisko, line 82: So you wouldn't have to hold the data here as long as the input module here is also a ready_and_valid. May require upstream changes to convert from ready-then-valid to ready-valid-and, but I think it's worth it

dpetrisko (changes requested):
burst to lite looks great. lite to burst can remove the data regs, with some care

bp_me/src/v/wormhole/bp_lite_to_burst.sv
Songchun-Li: I went through the modules using the lite to burst module, the main barrier is in bp_unicore_no_l2, where lite2burst is connected to the cmd arbiter, a ready_then_valid handshake. But the good thing is lite2burst will be replaced with stream2burst later with the stream_pump attached to the UCE side. 

Therefore, my idea is adding a temporary fifo between the arbiter and the lite2burst to keep it working, before the stream2burst replaces it?

dpetrisko: Yeah I think that's a good solution

Songchun-Li:
The header and data reg in lite_to_burst are removed, and the handshake connection of the conversion modules are checked.

dpetrisko (approved):

PR #770, opened 2021-02-23
Fixes non-idempotency in the I$

dpetrisko:
- Adds fetch and fill ops to the I$
- Adds a nonidempotency PMA (tied to off-chip space currently)
- Adds a mode where all I$ misses are treated as speculative
- Minor modification to LCE to support both modes

### DIFF
--- a/bp_be/src/v/bp_be_calculator/bp_be_pipe_mem.sv
+++ b/bp_be/src/v/bp_be_calculator/bp_be_pipe_mem.sv
@@ -178,6 +178,7 @@ module bp_be_pipe_mem
      ,.sum_i(trans_info.mstatus_sum)
      ,.trans_en_i(trans_info.translation_en)
      ,.uncached_mode_i((cfg_bus.dcache_mode == e_lce_mode_uncached))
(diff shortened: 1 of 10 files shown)


dpetrisko:
@muwyse can you take a quick look here? The idea is that we don't want the I$ to speculatively fetch into uncached space, ever.  And sometimes, we want the I$ to forgo speculatively misses altogether

muwyse:
> @muwyse can you take a quick look here? The idea is that we don't want the I$ to speculatively fetch into uncached space, ever. And sometimes, we want the I$ to forgo speculatively misses altogether

Okay, so the idea is three modes for the I$:
1. uncached only, non-speculative only
2. normal (uncached or cached, must be non-speculative to uncached)
3. non-speculative (uncached or cached, must be non-speculative for all)

In regard to the LCE, it looks like its behavior doesn't really change other than it sends normal requests if in either normal or nonspec modes? This means that the FE or icache itself is responsible for not issuing an incorrect miss request to the LCE, correct? If answer to both of those questions is yes then I don't see any issues.

dpetrisko:
Yep exactly

PR #783, opened 2021-02-26
Select the correct mem_resp on multiple valid sources

farzamgl:
The `bsg_mux_one_hot` module doesn't generate a valid output when multiple select lines are high. So we have to do the selection based on the yumi signals that are one-hot.

Also fixes the multicore Linux boot.


dpetrisko:
Nice! looks like this was fixed in the unicore at some point, but the change didn't make it over

dpetrisko (approved):

PR #791, opened 2021-03-05
Changing Memory Interface to bsg_cache_dma

dpetrisko:
This PR changes the L2 dma interface to purely cache dma, rather than mem_cmd/mem_resp.

Currently this is only done with the unicore interface, multicore is WIP


dpetrisko:
This should be ready for review, but depends on https://github.com/bespoke-silicon-group/basejump_stl/pull/410

bp_fe/syn/Makefile
```diff
@@ -14,6 +14,7 @@ export TOUCH_PATH   := $(BP_FE_DIR)/syn/touchfiles
 
 export TB    ?= bp_fe_icache
 export CFG   ?= e_bp_unicore_half_cfg
+export DRAM  ?= dramsim3
```
farzamgl, line 17: Maybe the default should be the AXI for faster regressions?

bp_me/test/common/bp_nonsynth_dram.sv
```diff
@@ -0,0 +1,338 @@
...
+
+         ,.axi_rid_o(axi_rid)
+         ,.axi_rdata_o(axi_rdata)
+         ,.axi_rresp_o(axi_rresp)
+         ,.axi_rlast_o(axi_rlast)
+         ,.axi_rvalid_o(axi_rvalid)
+         ,.axi_rready_i(axi_rready)
+         );
+
+      if (preload_mem_p)
```
farzamgl, line 327: I think we can support preloading for AXI right?

bp_me/test/common/bp_nonsynth_mem.sv
```diff
@@ -0,0 +1,144 @@
+
+/**
+ * bp_nonsynth_dram.v
```
farzamgl, line 3: Fix comment.

bp_top/src/v/bp_unicore.sv
```diff
@@ -512,92 +510,79 @@ module bp_unicore
...
+     ,.data_i(cache_data_lo)
+     ,.v_i(cache_data_v_lo)
+     ,.yumi_o(cache_data_yumi_li)
+     );
+  
+  bsg_cache
+   #(.addr_width_p(caddr_width_p)
+     ,.data_width_p(dword_width_gp)
+     ,.block_size_in_words_p(cce_block_width_p/dword_width_gp)
+     ,.sets_p(l2_en_p ? l2_sets_p : 2)
```
farzamgl, line 546: So we still have a small cache here even if L2 is disabled? Is it needed for L2 atomics?

bp_top/test/tb/bp_tethered/Makefile.vcs
```diff
@@ -69,9 +72,7 @@ $(SIM_DIR)/cce_ucode.bin: $(SIM_DIR)/cce_ucode.mem
 	cd $(@D); python3 $(UCODE2BOOT) --ucode=$< --path=$(@D) | $(XXD) -r -p > $@
 
 NBF_INPUTS ?= --ncpus=$(NCPUS) --ucode=cce_ucode.mem
-ifeq ($(PRELOAD_MEM_P), 0)
```
farzamgl, line 72: We need preloading to get reasonable simulation time for parallel cosim of Linux. Why are we removing it totally?

bp_me/test/common/bp_nonsynth_dram.sv
```diff
@@ -0,0 +1,338 @@
...
+       logic [`dram_pkg::channel_addr_width_p-1:0] dram_ch_addr_li;
+       logic dram_write_not_read_li, dram_v_li, dram_yumi_lo;
+       logic [`dram_pkg::data_width_p-1:0] dram_data_li;
+       logic dram_data_v_li, dram_data_yumi_lo;
+       logic [`dram_pkg::data_width_p-1:0] dram_data_lo;
+       logic dram_data_v_lo;
+
+       localparam cache_block_size_in_words_lp = cce_block_width_p/dword_width_gp;
+       localparam cache_bank_addr_width_lp = `BSG_SAFE_CLOG2(dram_max_size_p/num_cce_p);
+       bsg_cache_to_test_dram
```
farzamgl, line 82: I recommend using https://github.com/black-parrot/bp_tests/commit/8965f28669de22d1c9f13a8f1aed84f4fdab876b to stress test the various dram options.

bp_top/src/v/bp_unicore.sv
muwyse: I'm curious about this, too. Unless needed for atomics (or something) it seems like we could get away with a simple adapter or converter.

bp_me/test/tb/bp_cce/Makefile.params
```diff
@@ -47,8 +44,7 @@ export TB_PARAMS  = \
 
 export DUT_DEFINES =
 export TB_DEFINES  = +define+BP_SIM_CLK_PERIOD=$(BP_SIM_CLK_PERIOD) \
-                     +define+dram_pkg=$(DRAM_PKG) \
-                     +define+$(DRAM_MODEL)
+					 +define+den2048Mb+sg5+x16+FULL_MEM
```
muwyse, line 47: do we not want to be able to change the DRAM_MODEL ever? (question applies to all testbenches)

docs/software_guide.md
```diff
@@ -28,11 +28,11 @@ To add a new test suite, add a submodule containing the suite to bp_common/test/
 ## Building a Checkpoint Test
 BlackParrot can use Dromajo to generate checkpoints for certain tests. It runs the test on Dromajo for a certain number of instructions and then generates a memory image and a bootrom image which loads the internal architectural state of the cores(PC, registers, CSRs, privilege mode, ...). We use the `sim_sample` target to create a checkpoint at `SAMPLE_START_P=<n>` for RTL simulation. We have to run it with `NBF_CONFIG_P=1` to load the processor configuration using the nbf loader and also use a bootrom configuration to load the internal state of the cores.
 
-Optionally we can use `PRELOAD_MEM_P=1` to preload the memory image instead of writing it using the nbf loader and `SAMPLE_MEMSIZE=<k in MB>` to specify the size of memory image(default is 128MB).
+Optionally we can use `SAMPLE_MEMSIZE=<k in MB>` to specify the size of memory image(default is 128MB).
```
muwyse, line 31: It seems like memory is always loaded by NBF loader now? (excepting solution for Farzam's question regarding Linux parallel cosim)

bp_me/test/common/bp_nonsynth_dram.sv
dpetrisko: I can look into it, it's not as easy because it's not byte width and there's no dpi function.  But it should be possible

bp_fe/syn/Makefile
dpetrisko: It needs to support preloading, so if we can get the axi preload working then yeah I agree

bp_top/src/v/bp_unicore.sv
dpetrisko: It's actually needed for all uncached requests. The cache_dma operation works on the granularity of a cacheline.  So picture an uncached store of 64-bits.  You can't just write a full cacheline because you'll clobber whatever is in memory.  You would need to do a read-modify-write operation. Add in different size loads / stores and atomics and you've basically got a mini-cache.

Added an issue for it: https://github.com/black-parrot/black-parrot/issues/794

bp_top/test/tb/bp_tethered/Makefile.vcs
dpetrisko: mistake, reverted

bp_me/test/common/bp_nonsynth_dram.sv
dpetrisko:
```
Error-[ICPD] Illegal combination of drivers
/mnt/users/ssd0/no_backup/petrisko/pre-alpha-release/external/basejump_stl/bsg_test/bsg_nonsynth_axi_mem.v, 48
  Illegal combination of procedural drivers
  Variable "ram" is driven by an invalid combination of procedural drivers.
  Variables written on left-hand of "always_ff" cannot be written to by any
  other processes, including other "always_ff" processes.
  This variable is declared at
  "/mnt/users/ssd0/no_backup/petrisko/pre-alpha-release/external/basejump_stl/bsg_test/bsg_nonsynth_axi_mem.v",
  48: logic [(axi_data_width_p - 1):0] ram[(mem_els_ ...
  The first driver is at
  "/mnt/users/ssd0/no_backup/petrisko/pre-alpha-release/bp_me/test/common/bp_nonsynth_dram.sv",
  339: bp_nonsynth_dram.axi.axi_mem.ram[i] = axi.preload_mem[i];
  The second driver is at
  "/mnt/users/ssd0/no_backup/petrisko/pre-alpha-release/external/basejump_stl/bsg_test/bsg_nonsynth_axi_mem.v",
  188: always_ff @(posedge clk_i) begin
  if (reset_i) begin
```

bp_me/test/common/bp_nonsynth_dram.sv
dpetrisko: will switch on merge to bp_tests/master

bp_me/test/tb/bp_cce/Makefile.params
dpetrisko: I think we only have the one micron model

docs/software_guide.md
dpetrisko: Oh yeah let me revert this, originally I thought we had to get rid of the preloading

PR #795, opened 2021-03-09
Refactoring repos for sdk hdk and tools

dpetrisko:
Major movement and reorg, very few functional changes

Breaks BlackParrot into several subrepos and organizations
- black-parrot contains the RTL (and all the following)
- black-parrot-tools contains the tools needed to build and simulate a BlackParrot program
- black-parrot-sdk contains the tools and software to build BP programs
- black-parrot-hdk will contain the hardware for external integrations (Arty A7, OpenPiton, LiteX, etc.)

TODO:
- [x] Fix up some weird RTL diffs
- [x] Update documentation with new changes
- [x] Update getting started docs for new changes
- [x] Port outstanding branches to their new home (can be done after merge as well
- [x] Fix up CI
- [x] Fix up tire_kick


muwyse:
What is the rationale behind making -sdk and -hdk their own organizations? I'm not convinced this is a good idea because I think we want everything easily accessible from a single place, namely the black-parrot organization. From a "getting started" point of view I think it makes more sense to say "there is one black-parrot org, it has these 4 super-repos, and the rest of the repos fall under those" compared to "we have different orgs for BP, sdk, hdk".

We can still pin the black-parrot, -sdk, -hdk, and -tools "super-repos" at the top of the org (like black-parrot is now) to make these four visible.

dpetrisko:
@taylor-bsg wanted to do it this way to emulate the subrepository feature of bitbucket, which lets you group together sets of related repos. I think having a mono-org and pinning is totally reasonable too. I don't feel strongly either way

muwyse:
> @taylor-bsg wanted to do it this way to emulate the subrepository feature of bitbucket, which lets you group together sets of related repos. I think having a mono-org and pinning is totally reasonable too. I don't feel strongly either way

I see. I think mono-org is much better because it provides a "one-stop shop" for users (and means there is only 1 org to manage)

dpetrisko:
Okay this is ready to go, waiting for some RTL changes to get merged into dev first

PR #796, opened 2021-03-11
[WIP] Added Cpp feature tests

farzamgl:
Added Cpp feature tests
Perch: Making bp_utils functions visible to C++ linker
Perch: Added linker script and updated start code
Demos: cleaup and added lfs demo
Bump coremark
Bump bp_tests
Bump riscv-gnu-toolchain


bp_common/test/src/cpp-tests/src/constructor.cpp
```diff
@@ -0,0 +1,27 @@
+#include <iostream>
+using namespace std;
+
+class parent {
+private:
+    int id;
+public:
+    parent(int i) {id = i; cout << "parent constructor " << id <<endl;}
+    ~parent() {cout << "parnet destructor " << id <<endl;}
```
dpetrisko, line 9: parnet

bp_common/test/src/cpp-tests/src/constructor.cpp
```diff
@@ -0,0 +1,27 @@
+#include <iostream>
+using namespace std;
```
dpetrisko, line 2: Comment on what this test is checking?  Is there a way to verify it in code or is it just manual inspection?

dpetrisko: Applies for all tests

bp_common/test/src/demos/Makefile
```diff
@@ -1,16 +1,27 @@
...
 all: $(addsuffix .riscv, $(BP_DEMOS))
 
 %.riscv:
-	$(RISCV_GCC) $(RISCV_LINK) -o $@ src/$*.c -lperch
+	$(RISCV_GCC) -o $@ $(wildcard src/$*/*.c) $(RISCV_GCC_OPTS) $(RISCV_LINK_OPTS)
+
+lfs.riscv: $(BP_TEST_DIR)/src/perch/start.S ./src/lfs/lfs.c
+	$(RISCV_GCC) -o $@ $^ $(wildcard src/lfs/*.c) -D_DRAMFS $(RISCV_GCC_OPTS) $(RISCV_LINK_OPTS)
+
+%/lfs.c:
```
dpetrisko, line 20: Maybe a makefile function that takes a filelist instead? We could add it to our Makefile.tests and then other suites could use it

 .txt is a little dangerous because it could grab README.txt or something like that

bp_common/test/src/perch/riscv.ld
```diff
@@ -0,0 +1,146 @@
+OUTPUT_ARCH( "riscv" )
```
dpetrisko, line 1: Just to make sure, this handles both C and C++ programs? And the extra C++ stuff just gets trimmed when we link it to our C bare-metal stuff?

bp_common/test/src/perch/start.S
```diff
@@ -1,73 +1,80 @@
...
+  addi x25, x0, 0
+  addi x26, x0, 0
+  addi x27, x0, 0
+  addi x28, x0, 0
+  addi x29, x0, 0
+  addi x30, x0, 0
+  addi x31, x0, 0
+
+  la      a0, __libc_fini_array # Register global termination functions
+  call    atexit                #  to be called upon exit
```
dpetrisko, line 67: Where is this defined? newlib?

bp_common/test/src/perch/start.S
```diff
@@ -1,73 +1,80 @@
...
+  addi x24, x0, 0
+  addi x25, x0, 0
+  addi x26, x0, 0
+  addi x27, x0, 0
+  addi x28, x0, 0
+  addi x29, x0, 0
+  addi x30, x0, 0
+  addi x31, x0, 0
+
+  la      a0, __libc_fini_array # Register global termination functions
```
dpetrisko, line 66: Should this be _CPP or something?

bp_common/test/src/perch/start.S
```diff
@@ -1,73 +1,80 @@
...
+  call    __libc_init_array     # Run global initialization functions
+
+#ifdef _DRAMFS
+  call dramfs_init
+#endif
 
-/* jump to main */
-    jal main
+  call main
+  tail bp_finish
```
dpetrisko, line 75: Good change, so our benchmarks can return instead of needing to use bp_finish

dpetrisko (commented):
Clarification requested, but very cool! I like breaking out the linker script

bp_common/test/src/cpp-tests/src/constructor.cpp
farzamgl: Good point. I'll add some additional checks for some of the tests.

bp_common/test/src/perch/start.S
farzamgl: Yes it's in the common newlib code.

bp_common/test/src/perch/riscv.ld
farzamgl: Yes. From my understanding, these init calls call a series of constructor initializers in the C++ case. If there are no constructors then it basically doesn't do anything.

bp_common/test/src/perch/start.S
farzamgl: I think it's general code. If there's no need for initialization the .init_array section will be empty.

farzamgl: You can also see the call in the default newlib crt0.S

dpetrisko:
@farzamgl , the repos have been updated on https://github.com/black-parrot/black-parrot/tree/refactor_repos. Can you port these changes to:
https://github.com/black-parrot-sdk/bp-tests/
https://github.com/black-parrot-sdk/bp-demos/
https://github.com/black-parrot-sdk/perch
https://github.com/black-parrot-sdk/black-parrot-sdk/tree/master/linker

as appropriate?

dpetrisko:
Is this PR obsolete then?

farzamgl:
Yes. It is.

PR #797, opened 2021-03-12
Fixing replay pending bug

dpetrisko:
Previously, dispatch would be blocked if there was an interrupt while a cache request was being replayed.

RTL is good to go, but needs a test.  Maybe a bunch of L1 amoadds with an interrupt timer going off (edited)


dpetrisko:
Bugs are fixed, tests are added to bp_tests, but master doesn't build because of changes to libperch. Will bump submodule when available

PR #800, opened 2021-03-14
Update top dev

dpetrisko:
- Adds domain mask to config read
- Changes memory interface to bsg_cache_dma #791 
- Adds more L2 configuration parameters
- Reworks DRAM system to use more standard modules


dpetrisko:
Dependent on https://github.com/bespoke-silicon-group/basejump_stl/pull/417/files

PR #805, opened 2021-03-16
Minimizing arbitration near cache engines

dpetrisko:
The response fifos here are unnecessary. This PR removes them with a slight modification to how the arbiter is used


farzamgl (approved):
Looks good to me.

PR #815, opened 2021-03-25
bp_common_defines.svh: remove no-op self-include

WasabiFan:
Noticed this as I was messing with includes elsewhere. `bp_common_defines.svh` includes itself, which doesn't have an effect anyway due to the header-guard. Just a minor cleanup.


dpetrisko (approved):

dpetrisko:
kudos on the username

Issue #798, opened 2021-03-14
Enable FST generation in Verilator


dpetrisko:
GTKWave works better / faster with FST files.  It should be pretty easy to enable this flow, with substantial runtime benefits

developandplay:
I looked into this issue to familiarize myself with the workflow of the project for GSoC.
Now I did manage to enable `--trace-fst` but after running into a few linking issues I realized it is incompatible with Systemc.
From the manual:
> How do I generate FST waveforms (aka dumps or traces) in SystemC?
> The FST library from GTKWave does not currently support SystemC; use VCD format instead.

Using the `CONNECTING TO C ++` section of the manual I was able to add the undefined `sc_time_stamp()`. However now I'm stuck on the remaining `undefined reference to 'main'`. Is there a simple way to switch from Systemc to C++ or is this a more involved task?

dpetrisko:
> I realized it is incompatible with Systemc.

Yep.

> Is there a simple way to switch from Systemc to C++ or is this a more involved task?

Our test_bp.cpp is pretty close to the example SystemC/C++ testbench Verilator provides. The most notable change is that we need to manage both the core clock and DRAM clock. I think without the SystemC clock class, this involves creating two clock threads. Nothing too crazy, but non-trivial

taylor-bsg:
Not sure it helpful or different but Dustin has the dpi based clock
generator classes.

dpetrisko:
Oh yeah that might be useful:

https://github.com/bespoke-silicon-group/basejump_stl/blob/master/bsg_test/bsg_nonsynth_dpi_clock_gen.cpp

Using this would mean that test_bp would have an ifdef VERILATOR selecting between the clock gens, and then test_bp.cpp would be instantiating these dpi clocks

developandplay:
Ok so I made some progress but I'm stuck on the dpi methods. Unfortunately they are resulting in undefined reference erros.
Now I did find the [testing/bsg_test/bsg_nonsynth_dpi/](https://github.com/bespoke-silicon-group/basejump_stl/blob/7556378810412be47e118b08910c02883936a64c/testing/bsg_test/bsg_nonsynth_dpi/) as a reference. This helped me fix quite a few prior errors but I'm not sure which flag/command ensures that dpi methods are included.

dpetrisko:
Seems it's defined there:

[...]

PR #820, opened 2021-03-26
Switch from vcd to fst dumps

developandplay:
As mentioned in #798


bp_top/test/tb/bp_tethered/testbench.sv
```diff
@@ -571,5 +611,15 @@ module testbench
    #(.bp_params_p(bp_params_p))
    if_verif
     ();
+  
+  initial
+    begin
+      `ifndef VERILATOR
```
dpetrisko, line 617: Put the ifndef outside of the initial block

bp_top/test/tb/bp_tethered/test_bp.cpp
```diff
@@ -1,55 +1,51 @@
...
-  reset = 1;
-
-  std::cout << "Raising reset" << std::endl;
-  for (int i = 0; i < 20; i++) {
-    sc_start(std::max(sim_period, dram_period), SC_PS);
+  while(tb->reset_i == 1) {
+    bsg_timekeeper::next();
+    tb->eval();
+    #if VM_TRACE_FST
+      wf->dump(sc_time_stamp());
```
dpetrisko, line 37: You have to dump on every time step? For vcd you just opened the wave and it dumped on every time step automatically

developandplay: I think that was possible because we were using systemc previously.
From the _How do I generate waveforms (traces) in C++?_ section in the [verilator manual](https://www.veripool.org/wiki/verilator/Manual-verilator):
> Create a VerilatedVcdC object, and in your main loop call "trace_object->dump(time)" every time step, and finally call "trace_object->close()".

bp_top/test/tb/bp_tethered/testbench.sv
developandplay: Yep makes sense. Changed it now.

bp_top/test/tb/bp_tethered/test_bp.cpp
dpetrisko: Option A seems to not require it. Option B is only needed for “finer-grained control”. Seems like the equivalent of autoflush in VCS

developandplay: I was sticking to B because that's what we had previously.
Thanks for the suggestion. Now I switched to A and the code became much simpler.

bp_top/test/tb/bp_tethered/testbench.sv
```diff
@@ -612,14 +612,17 @@ module testbench
    if_verif
     ();
   
-  `ifndef VERILATOR
-    initial
-      begin      
+  initial
```
developandplay, line 615: Note I changed your first suggestion back since now both verilator and vcs need initialization.

dpetrisko: Hm, I think opening the waveform with wf->open is equivalent to $dumpfile; $dumpvars. The problem with this approach is if you compile without tracing, doesn't it cause an error? In VCS that's the case

developandplay: I also assumed that some additional guard is required but the manual doesn't mention anything like that.
E.g. `make -C bp_top/syn build.sc sim.sc SUITE=riscv_tests PROG=towers -j 8` finished without any errors on the console.

dpetrisko:
I just downloaded and ran 

```
make clean build_dump.sc sim_dump.sc
```

and got thousands of these:

```
%Warning: previous dump at t=2040245000, requesting t=2040245000, dump call ignored
%Warning: previous dump at t=2040247500, requesting t=2040247500, dump call ignored
%Warning: previous dump at t=2040250000, requesting t=2040250000, dump call ignored
%Warning: previous dump at t=2040252500, requesting t=2040252500, dump call ignored
%Warning: previous dump at t=2040255000, requesting t=2040255000, dump call ignored
```

developandplay:
How do I turn on warnings?
But yes with the changes I can't open the dump.fst for hello_world anymore.

The dumps for aha_compress and towers on the other hand I am able to open.
Is there anything in hello_world that could interfere with dumping?

dpetrisko:
I didn’t do anything except checkout your branch and run that command, so not sure about the warnings. 

No, hello world is like 20 lines of assembly — as simple as it gets

dpetrisko:
If the dump call in test_bp.cpp like you had before works, we can just do that. It's not worth digging too deep into this...

developandplay:
Oh yes sorry I didn't push the dependency update for basejump_stl. I'm not an expert on git submodules but I think I can't by design. Just use the `fix_dpi_clock_gen` branch for the basejump submodule.
That's why you are getting those warnings...

Now that being said the inability to open the dump.fst for hello_world is independent of that.

dpetrisko:
oh of course! FYI, you can update submodules. They are simply pointers to other git repo commit hashes (or branches). The process is:

```
cd ./external/basejump_stl
git checkout fix_dpi_clock_gen
cd ../../ # To black-parrot
git add external/basejump_stl
git commit -m "Bump basejump"
```

dpetrisko:
I see this behavior for hello world as well as other short programs rv64ui-p-add, for example. My guess is the flush is periodic and doesn't happen with short programs

developandplay:
I had a similar suspicion...
Just an observation: If I switch to vcd (with $dumpfile and $dumpvars) then I'm able to open the file without problems.
Maybe vcd is uncompressed so it's not as sensitive if the file get's closed to early or improperly.

dpetrisko:
Found a fix, can you try this and see if it works for you?

[patch.txt](https://github.com/black-parrot/black-parrot/files/6215035/patch.txt)

developandplay:
Just ran hello world and towers on it.
Unfortunately in both cases if you view the file in gtkwave it only contains the wires but not the signals. It's just a few kB after all...
Let's go with the option B for now and open a bug report for option A on verilator.

dpetrisko:
Ah I didn’t actually look at the signals haha. Sounds like a plan

developandplay:
Now all testbenches are updated. The default test for `syn/me` runs without errors and I can open the dump.fst files for it.
@dpetrisko Since the `test_bp.sv` top levels are now removed I also removed their references from the `Makefile.vcs`'s. So please check if those work.

dpetrisko:
LGTM. Thanks, this solves a big problem we've had with using BP in classes!

dpetrisko (approved):

PR #833, opened 2021-04-09
Be more specific on Ubuntu prereq 

dpetrisko:
And Cmake and OS versions


taylor-bsg (approved):
Nice! But I also suggest checking in the makefile — nobody follows docs that well.

PR #834, opened 2021-04-12
Remove systemc submodule - no longer used

### LINKED ISSUE #366: Move testbenches to pure C++
We don't actually use any SystemC features (confirm @muwyse?), instead just wrapping our testbenches with a thin clock/reset driver.  But we're stuck with the overhead of compiling in systemc libraries. Small rewrites could easily port these testbenches to C++.

muwyse:
Removes systemc submodule since it is no longer used, and removes one deprecated header file in ME related to old systemc testbench.

Closes #366


dpetrisko:
Good chance this breaks CI, but we can refresh the caches. LGTM

dpetrisko (approved):

PR #837, opened 2021-04-14
remove BedRock assembler files from BP repo - exists in sdk/bedrock

muwyse:
Remove CCE microcode and assembler files that now exist in sdk/bedrock. All changes to microcode and CCE assembler should be made in sdk/bedrock.


muwyse:
CI looks good: https://gitlab.com/black-parrot/black-parrot/-/pipelines/285816860

PR #839, opened 2021-04-15
Adding NBF read

dpetrisko:
- Adds NBF.py read 
- Adds nbf.v read-validate capability
- Adds some additional bp_cfg read capabilities so we can we read them. Should have minimal PPA impact

This is really just a demo of the capability


taylor-bsg (approved):

bp_common/software/py/nbf.py
```diff
@@ -3,6 +3,21 @@
...
+#   opcode map:
+#     8'h00: 1 byte write 
+#     8'h01: 2 byte write
+#     8'h02: 4 byte write
+#     8'h03: 8 byte write
+#
+#     8'h04: 1 byte read
+#     8'h05: 2 byte read
+#     8'h06: 4 byte read
+#     8'h07: 8 byte read
```
muwyse, line 17: Thoughts on having these be 8'h10 through 8'h13? Would provide nice symmetry with the write commands in that the size is still specified by the lower four bits as 2^N.

(this would also then align with what I selected for the ArtyParrot project for read commands, but it is very easy for us to change the choice of opcode, too)

bp_common/software/py/nbf.py
```diff
@@ -217,8 +246,10 @@ def dump(self):
   parser.add_argument("--checkpoint", dest='checkpoint_file', metavar='sample.nbf',help='checkpoint nbf file')
   parser.add_argument('--skip_zeros', dest='skip_zeros', action='store_true', help='skip zero DRAM entries')
   parser.add_argument('--addr_width', type=int, default=40, help='Physical address width')
+  parser.add_argument('--data_width', type=int, default=64, help='Data width')
 
   args = parser.parse_args()
 
-  converter = NBF(args.ncpus, args.ucode_file, args.mem_file, args.checkpoint_file, args.config, args.skip_zeros, args.addr_width)
+  converter = NBF(args.ncpus, args.ucode_file, args.mem_file, args.checkpoint_file, args.config,
+          args.skip_zeros, args.addr_width, args.data_width)
```
muwyse, line 254: Not necessarily part of this change, but for configurations that don't use the ucode CCE, does the generated NBF still contain commands to load the microcode? It looks like this script would skip that step if ucode_file is not provided, but IIRC the rest of the build flow may always generate the ucode file for simulation, so it may always be included.

bp_top/test/common/bp_nonsynth_nbf_loader.sv
```diff
@@ -70,83 +109,63 @@ module bp_nonsynth_nbf_loader
...
+      e_send        : state_n = is_fence_packet
+                                ? e_fence
+                                : is_finish_packet
+                                  ? e_done
+                                  : (is_read_packet & io_cmd_yumi_i)
+                                    ? e_read
+                                    : e_send;
+      e_read        : state_n = read_return ? e_send : e_read;
+      e_fence       : state_n = credits_empty_lo ? e_send : e_fence;
+      e_done        : state_n = e_done;
```
muwyse, line 148: My only concern with the finish command putting the NBF loader into DONE is that no future reads can be performed. Does this change support uncached reads to any memory location, or only special locations like the config registers?

muwyse (commented):
Overall, looks good! A few questions which may not need to be addressed in this PR.

bp_common/software/py/nbf.py
dpetrisko: fine by me, since that still leaves room for fence and finish

bp_common/software/py/nbf.py
dpetrisko: Yep, you're right: https://github.com/black-parrot/black-parrot/issues/840

bp_top/test/common/bp_nonsynth_nbf_loader.sv
dpetrisko: Not that I necessarily disagree, but what else would the finish packet do? I believe the done signal is used by some testbenches to basically start operations in configs that don't have a freeze register.

This change supports uncached reads to any memory location, but since the read is actually a "verify" in the nbf.v file, I just use the configs to check the functionality. How I see this is the testbench in bp_tethered is a straight shot, pass/fail without interactivity. The FPGA stuff should be interactive, but that'll use a different NBF loader. Or even the riscv-debug module once we have that

muwyse: > Not that I necessarily disagree, but what else would the finish packet do? I believe the done signal is used by some testbenches to basically start operations in configs that don't have a freeze register.

Right, it basically becomes useless lol.

> This change supports uncached reads to any memory location, but since the read is actually a "verify" in the nbf.v file, I just use the configs to check the functionality. How I see this is the testbench in bp_tethered is a straight shot, pass/fail without interactivity. The FPGA stuff should be interactive, but that'll use a different NBF loader. Or even the riscv-debug module once we have that

Great!

Yes, FPGA has slightly different requirements for interactivity. For example, ArtyParrot basically makes the finish command useless because the FPGA Host simply loops it back to the PC.

No change needed here.

bp_common/software/py/nbf.py
muwyse: Cool. I think update these and then LGTM

PR #841, opened 2021-04-17
fix me_cce_to_cache bug - prevent sending responses until clear tag done

muwyse:
This change fixes a bug in bp_me_cce_to_cache module to disallow sending mem_resp messages until after the CLEAR_TAG routine completes in the command FSM. Prior to this, a valid memory command could arrive and trigger a memory response, regardless of the mem_cmd_ready_o status. So while consumption of the incoming mem_cmd obeyed handshaking rules, a valid command arriving before tag clear was done would trigger the sending of invalid memory response.

### DIFF
--- a/bp_me/src/v/cache/bp_me_cce_to_cache.sv
+++ b/bp_me/src/v/cache/bp_me_cce_to_cache.sv
@@ -44,7 +44,7 @@ module bp_me_cce_to_cache
     , output logic                        v_o
     , input                               ready_i
 
-    , input [l2_data_width_p-1:0]          data_i
+    , input [l2_data_width_p-1:0]         data_i
     , input                               v_i
     , output logic                        yumi_o
   );
@@ -320,7 +320,10 @@ module bp_me_cce_to_cache
 
     case (resp_state_r)
       RESP_RESET: begin
-        resp_state_n = RESP_READY;
+        // hold in RESP_RESET until command FSM finishes clearing cache tags
+        resp_state_n = (cmd_state_n == READY) ? RESP_READY : RESP_RESET;
+        // cache "acks" TAGST commands with zero-data responses
+        yumi_o = v_i;
       end
       RESP_READY: begin
         is_resp_ready = 1'b1;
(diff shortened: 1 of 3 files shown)


muwyse:
Fun side note: this only manifested after properly naming and using handshake signals between modules used in the icache testbench. The renaming of signals revealed one connection improperly using ready&valid signals (in this case, &'ing them together when not required by the connection). Fixing the connection then resulted in this bug appearing because it was being hidden by the &'ing of the ready&valid signals on an input.

bp_me/src/v/cache/bp_me_cce_to_cache.sv
```diff
@@ -74,6 +74,20 @@ module bp_me_cce_to_cache
   logic [counter_width_lp-1:0] cmd_counter_r, cmd_counter_n;
   logic [counter_width_lp-1:0] cmd_max_count_r, cmd_max_count_n;
 
+  logic cmd_clear_tag_done;
```
dpetrisko, line 77: How about instead:

```
      RESP_RESET: begin
-        resp_state_n = RESP_READY;
+        resp_state_n = (cmd_state_n == READY) ? RESP_READY : RESP_RESET;
      end
```

dpetrisko (changes requested):

bp_me/src/v/cache/bp_me_cce_to_cache.sv
muwyse: `yumi_o` is only toggled if `resp_state_r` is in the `RESP_READY` state, so this solution would actually prevent the module from completing the tag clear procedure.

muwyse: I had tried a similar fix earlier, because I thought (I'm guessing similar to what you thought) that the response FSM was only handling the memory response port, but it actually is responsible for both mem_resp_i and v_i (from cache)

muwyse: I suppose the above along with adding `yumi_o = v_i` to the `RESP_RESET` state could work though

dpetrisko: Yeah I think that's better / more clear.  I didn't know the tag store required yumis

muwyse: > Yeah I think that's better / more clear.

okay. I will implement our combo solution.

>  I didn't know the tag store required yumis

I thought it didn't either, until the other day when I was looking at something else related to BP's startup process and noticed that the cache sends back data "acks" with all zero data in response to the tag clear (TAGST) commands.

PR #845, opened 2021-04-19
Add tag and data critical hints

dpetrisko:
This PR adds distinct tag and data critical signals in the cache engine interface (the critical signal was previous implicitly data only). These signals can be used for snooping or early restart of the cache pipeline. Should be no functional change, just the new signals and the doc update


muwyse (approved):
LGTM

Issue #846, opened 2021-04-21
Convert LCEs to use ready_and instead of ready_then


muwyse:
Since the LCEs connect to networks in BP they should really use ready_and interfaces for the request, command, and response networks.

dpetrisko:
Yep, I have a branch somewhere I started to work on that. Shouldn’t be too bad to implement (and no cost, since the handshakes are in an FSM anyway)

dpetrisko:
Buggy first try here: https://github.com/black-parrot/black-parrot/tree/top_dev_fix_lcehelpful

muwyse:
This will be fixed when LCE is updated to use BedRock Burst

PR #851, opened 2021-04-26
[WIP] Making LCE modules helpful

dpetrisko:
This PR makes the LCEs as helpful as possible. There's still 1 demanding interface: the incoming lce_cmd interface. This must be demanding since the cache engine interface *_mem_pkts are demanding. I suppose those could also be made to be helpful, but that would require a rework of the cache pipeline as well, and this PR is already decently sized

### DIFF
--- a/bp_me/src/v/lce/bp_lce_cmd.sv
+++ b/bp_me/src/v/lce/bp_lce_cmd.sv
@@ -105,7 +105,7 @@ module bp_lce_cmd
     // Resp: ready->valid
     , output logic [lce_resp_msg_width_lp-1:0]       lce_resp_o
     , output logic                                   lce_resp_v_o
-    , input                                          lce_resp_ready_then_i
+    , input                                          lce_resp_ready_and_i
 
     // CCE-LCE interface
     // Cmd_i: valid->yumi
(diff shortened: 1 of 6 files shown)


bp_me/src/v/lce/bp_lce_cmd.sv
```diff
@@ -379,14 +379,14 @@ module bp_lce_cmd
               tag_mem_pkt.opcode = e_cache_tag_mem_set_state;
               tag_mem_pkt_v_o = lce_cmd_v_i;
 
-              lce_resp_v_o = tag_mem_pkt_yumi_i & lce_resp_ready_then_i;
```
dpetrisko, line 382: Actually, I think this is a bug. Going to look at this closer

bp_top/test/tb/bp_tethered/testbench.sv
```diff
@@ -547,64 +547,64 @@ module testbench
...
-              ,.lce_resp_v_i(lce_resp_v_o)
-              ,.lce_resp_ready_then_i(lce_resp_ready_then_i)
-              ,.lce_cmd_i(lce_cmd_i)
-              ,.lce_cmd_v_i(lce_cmd_v_i)
-              ,.lce_cmd_yumi_i(lce_cmd_yumi_o)
-              ,.lce_cmd_o_i(lce_cmd_o)
-              ,.lce_cmd_o_v_i(lce_cmd_v_o)
-              ,.lce_cmd_o_ready_then_i(lce_cmd_ready_then_i)
-              );
+          //bind bp_cce_wrapper
```
dpetrisko, line 550: This obviously needs to be fixed before merge, but waiting on figuring out the functional changes

dpetrisko:
In pursuit of this: https://github.com/black-parrot/black-parrot/issues/846

PR #855, opened 2021-04-29
add config variable for ucode loading in nbf

### LINKED ISSUE #840: Only load ucode on ucode CCE variants
Currently the nbf loader always loads ucode https://github.com/black-parrot/black-parrot/blob/master/bp_top/test/tb/bp_tethered/Makefile.vcs#L69. It should only send it if we're in a ucode cce variant. This can be detected using the Makefile.cfg structure: https://github.com/black-parrot/black-parrot/blob/master/bp_top/test/tb/bp_tethered/Makefile.cfgs#L1 switching on: https://github.com/black-parrot/black-parrot/blob/master/bp_common/src/include/bp_common_aviary_pkgdef.svh#L132

muwyse:
This PR adds a config variable in the bp_tethered testbench to detect whether the nbf file should load CCE ucode. The CCE ucode is only included in the nbf (and copied to the sim folder) if the config name contains the string "cce_ucode", as all standard configs with the ucode CCE do.

Note: a custom config that uses the ucode CCE by setting the cce_ucode field (https://github.com/black-parrot/black-parrot/blob/master/bp_common/src/include/bp_common_aviary_pkgdef.svh#L132) of the params struct to 1 will fail to load the ucode with this approach unless the config name can be set to include the string "cce_ucode".

Closes #840  

Also removes a deprecated variable in the makefiles related to CCE ucode file.

### DIFF
--- a/bp_me/test/tb/bp_cce/Makefile.params
+++ b/bp_me/test/tb/bp_cce/Makefile.params
@@ -19,7 +19,6 @@ export LCE_MODE_P ?= 0
 export ME_TEST_P ?= 0
 
 COH_PROTO   ?= mesi
-CCE_MEM_PATH = $(BP_ME_DIR)/src/asm/roms
 CCE_MEM      = $(COH_PROTO).mem
 
 NUMS = $(shell seq 0 `expr $(NUM_LCE_P) - 1`)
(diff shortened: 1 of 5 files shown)


dpetrisko (approved):
Cool!

muwyse:
@dpetrisko what do you think about the second commit here? It conditionally sets the ucode file prereq so that if the config is not a ucode CCE config then the ucode doesn't need to exist in sdk/ucode. Are you aware of a better way to implement a conditional prereq? I couldn't find a better approach.

dpetrisko:
Seems fine to me

PR #861, opened 2021-05-11
Repo reorg v2

dpetrisko:
This reorganizes the repos slightly such that BlackParrot mostly contains RTL + simulation/synthesis testbenches. The main highlights here are:
- SDK and HDK are no longer contained within BlackParrot. This prevents circular dependencies with HDK toplevels, as well as preventing long recursive clone times for external BP integration.
- New developers should clone https://github.com/black-parrot/black-parrot-sim to tag SDK + HDK versions.
- This also "inverts the dependency" such that BlackParrot is the bottom level repo. External users can tag a version of BlackParrot that works with that specific version of their system without fear of breakage.  And it gives us some leeway on updating not-easily tested environments

This should be a pretty non-invasive change. Existing users can simply move sdk up a directory and continue work as before.

PM for a link to the reorg presentation for further details...

NOTE: this will severely break CI when merged in, but I will fix it in sw_dev before moving to dev


muwyse (approved):

Makefile.tools
```diff
@@ -20,9 +24,18 @@ verilator_build:
 	$(MAKE) -C $(verilator_dir)
 	$(MAKE) -C $(verilator_dir) install
 
+dromajo_build:
+	mkdir -p $(dromajo_dir)/build
+	$(CMAKE) -S $(dromajo_dir) -B $(dromajo_dir)/build
```
farzamgl, line 29: `$(CMAKE) -S $(dromajo_dir) -B $(dromajo_dir)/build -DCMAKE_BUILD_TYPE=Release`
Small change to make dromajo with -O3.

farzamgl (approved):
Looks good!

Makefile.tools
dpetrisko: cool!

PR #866, opened 2021-05-17
CCE Dynamic Freezing of uCode Instruction Fetching

Kellen-Hartnett:
# CCE Dynamic Freezing of uCode Instruction Fetching
### By: Kellen Hartnett and Hans Gaensbauer, ECE 470, University of Washington
This feature modifies the stalling behavior of the CCE uCode instruction fetch to freeze instead of polling to reduce power consumption. This freezing will replace long stalls while the CCE waits for new requests or responses from the LCE.
## Current Behavior
Currently, when the CCE ucode processor stalls, it continuously fetches the next instruction, which does not get used. This leads to the CCE processor consuming power while doing nothing of importance. Shown below is the amount of instruction fetches that are stalls compared to fetches for program execution:

image

From our testing, unnecessary fetches for stalls make up 64.366% of instruction fetches for the CCE.
## Implemented Solution
To reduce power consumption while stalling, the change implemented in this branch assigns the read enable input of the RAM block to be enabled to depend on a new logic variable, `mem_freeze`, along with the current memory functionality:
`assign memory_read_enable = ucode_v_i | (ram_v_li & !mem_freeze);`
The variable `mem_freeze` is informed by:

- `stall_i`: An existing input to `bp_cce_inst_ram.sv` that is enabled when the CCE processor stalls. `mem_freeze` only is enabled if the processor is currently stalling (`stall_i` is 1)
- `traffic_i`: An input `bp_cce_inst_ram.sv` added for this feature that is enabled when the CCE is receiving data from the LCE. `mem_freeze` only is enabled if the CCE is not receiving a request or response from the LCE (`traffic_i` is 0)
- `stall_counter`: An internal logic variable added for this feature that counts the number of instructions that have been stalled consecutively. `mem_freeze` only is enabled if the amount of consecutive stalled instructions is >= the local parameter STALL_THRES (default: 3)

If `mem_freeze` is enabled, the CCE will not fetch instructions from the ucode instruction RAM, it will instead output the previous instruction used.
## Testing Methodology and Results
To test the effectiveness of this feature, we traced the instructions to find where the CCE stalls to measure how many instructions are fetched while stalling. This test was run on a variety of programs that utilize the CCE. 

Shown below is the data results of this testing:

image

This feature leads to an average of 64.247% decrease in fetched instructions across tested programs. This correlates to significant power savings for the CCE.

This feature was developed while in contact with @muwyse via email.


bp_me/src/v/cce/bp_cce_inst_ram.sv
```diff
@@ -71,19 +73,45 @@ module bp_cce_inst_ram
...
+  
+  bp_cce_inst_s ram_out;
+  bp_cce_inst_s last_ram_data;
+  
+  logic [1:0] stall_counter;
+  logic mem_freeze, memory_read_enable, last_re;
+  
   bsg_mem_1rw_sync
     #(.width_p(cce_instr_width_gp)
       ,.els_p(num_cce_instr_ram_els_p)
```
dpetrisko, line 85: RAMs generally have the ability to keep the last read data built in.  You can use latch_last_read_p here to take advantage of that

bp_me/src/v/cce/bp_cce_inst_ram.sv
```diff
@@ -71,19 +73,45 @@ module bp_cce_inst_ram
...
       ,.w_i(ucode_w_i)
-      ,.data_o(inst_o)
+      ,.data_o(ram_out)
       );
+  
+  always_ff @(posedge clk_i) begin
+    if(memory_read_enable) last_ram_data <= ram_out;
+    last_re <= memory_read_enable;
+    
+      if(reset_i) begin
```
dpetrisko, line 101: Use bsg_counter_clear_up here for clarity

dpetrisko:
Cool, thanks for the contribution! Couple questions for you or @muwyse 
- How critical is the stall path in the CCE generally?
> Currently, when the CCE ucode processor stalls, it continuously fetches the next instruction
- Is this through a busy loop or just rereading in RTL?
- Do we need the stall counter to initialize this? Does the ucode know what instructions will wait for traffic?
- Why is lce_cmd not included in the resume condition? Are there no unexpected lce_cmds in the protocol?
- Does this impact the ability of the CCE to multithread work for different requests? In this specific case, we're always waiting on the next request/response to come back. However, you could imagine an opportunistic CCE which does something like calculate prefetch data or statistics while waiting for requests. In this case, we'd probably want finer grained control over this freezing behavior

bp_me/src/v/cce/bp_cce_inst_ram.sv
Kellen-Hartnett: Fixed in d3c9752

bp_me/src/v/cce/bp_cce_inst_ram.sv
Kellen-Hartnett: Fixed in d3c9752

hansgaensbauer:
> Cool, thanks for the contribution! Couple questions for you or @muwyse
> 
> * Is this stall signal separate from the normal stall path?
> * How critical is the stall path in the CCE generally?
> 
> > Currently, when the CCE ucode processor stalls, it continuously fetches the next instruction
> 
> * Is this through a busy loop or just rereading in RTL?
> * Do we need the stall counter to initialize this? Does the ucode know what instructions will wait for traffic?
> * Why is lce_cmd not included in the resume condition? Are there no unexpected lce_cmds in the protocol?
> * Does this impact the ability of the CCE to multithread work for different requests? In this specific case, we're always waiting on the next request/response to come back. However, you could imagine an opportunistic CCE which does something like calculate prefetch data or statistics while waiting for requests. In this case, we'd probably want finer grained control over this freezing behavior

I can answer a few of these. The stall_i signal is connected directly to stall_lo from bp_cce, but I'm not sure how critical that path is. We can register that signal at the expense of another cycle required to wake up. Our changes just deassert the v_i signal on the instruction RAM after a certain number of stalls, since the instruction fetch FSM in the bp_cce_inst_ram won't change while the CCE is stalled (so we're only saving power by not reading from the instruction ram). The wfq and poph instructions cause the processor to stall when there are no lce requests or responses, so the uCode does know when stalls can happen. Our use of the stall counter was originally motivated by the possibility that there might be a larger cost to disabling the RAM, although at the moment there is only the single lost cycle when the RAM is not read from. It would also be possible to do away with the counter, and disable the memory any time there are two or more consecutive stalls. The majority of wfq/poph instructions don't cause stalls, so the memory should stay enabled at least for very short (0 or 1 cycle) stalls. 

It is my understanding that the lce commands are generated by the CCE, so that when the CCE is stalled no lce commands can be generated. Finally, yes, this would prevent the CCE from multithreading as the instruction RAM would be made unavailable whenever the currently running thread encountered a wfq or poph instruction. As I understand it, a future version of the CCE that does implement threading will need to eliminate the wfq/poph polling behavior another way, since at the moment the CCE cannot move to another instruction until it receives an lce request or response.

Thanks for everything!

README.md
```diff
@@ -1,6 +1,6 @@
...
+# BlackParrot: A Linux-Capable Accelerator Host RISC-V Multicore [![License](https://img.shields.io/badge/License-BSD%203--Clause-blue.svg)](https://opensource.org/licenses/BSD-3-Clause) [![Gitlab CI](https://gitlab.com/black-parrot/black-parrot/badges/master/pipeline.svg)](https://gitlab.com/black-parrot/black-parrot/pipelines) [![Contributers](https://img.shields.io/github/contributors/black-parrot/black-parrot.svg?style=flat)]() [![Twitter Follow](https://img.shields.io/twitter/follow/BlackParrotCore.svg?style=social)](https://twitter.com/BlackParrotCore) [![Join the chat at https://gitter.im/black-parrot/rtl](https://badges.gitter.im/black-parrot/rtl.svg)](https://gitter.im/black-parrot/rtl?utm_source=badge&utm_medium=badge&utm_campaign=pr-badge&utm_content=badge)
```
muwyse, line 3: I think the changes in this file are being captured from somewhere unrelated. Try rebasing the changes onto the latest me_dev and make sure that only your changes are present.

bp_me/src/v/cce/bp_cce_inst_ram.sv
```diff
@@ -25,6 +25,7 @@ module bp_cce_inst_ram
     `declare_bp_proc_params(bp_params_p)
     // Derived parameters
     , localparam cfg_bus_width_lp = `bp_cfg_bus_width(domain_width_p, core_id_width_p, cce_id_width_p, lce_id_width_p)
+    , parameter stall_threshold = 3
```
muwyse, line 28: parameters to modules should be named with _p suffix

bp_me/src/v/cce/bp_cce_inst_ram.sv
```diff
@@ -59,36 +61,55 @@ module bp_cce_inst_ram
   end
   //synopsys translate_on
 
-  `declare_bp_cfg_bus_s(domain_width_p, core_id_width_p, cce_id_width_p, lce_id_width_p);
+`declare_bp_cfg_bus_s(domain_width_p, core_id_width_p, cce_id_width_p, lce_id_width_p);
   bp_cfg_bus_s cfg_bus_cast_i;
   assign cfg_bus_cast_i = cfg_bus_i;
-
+
```
muwyse, line 67: get rid of added whitespace

muwyse (changes requested):
Thanks for contributing! See comments attached to code for some specific items, but my higher level feedback is I think the traffic_i signal is unnecessary. The stall_i signal will indicate if the instruction in the execute stage of the CCE was able to execute or not. I think it is sufficient to have a counter that counts up when `stall_i == 1 && stall_cnt < threshold` and clears on `stall_i == 0`. The stall on the instruction ram read would be active once `stall_cnt == threshold` . This avoids having to figure out what should unblock the stall and instead relies on the existing stall detection to unstall as soon as the currently executing instruction actually executes.

README.md
muwyse: This comment applies to one or two other files in the PR.

dpetrisko:
>I think it is sufficient to have a counter that counts up when stall_i == 1 && stall_cnt < threshold and clears on stall_i == 0. The stall on the instruction ram read would be active once stall_cnt == threshold

If that's the case, just
```
assign mem_read_enable = ucode_v_i | (ram_v_li & ~stall_v_i)
```
?

README.md
Kellen-Hartnett: Fixed! (I think)

hansgaensbauer:
> > I think it is sufficient to have a counter that counts up when stall_i == 1 && stall_cnt < threshold and clears on stall_i == 0. The stall on the instruction ram read would be active once stall_cnt == threshold
> 
> If that's the case, just
> 
> ```
> assign mem_read_enable = ucode_v_i | (ram_v_li & ~stall_v_i)
> ```
> 
> ?

I've tried just using the stall_i signal directly (just tried the statement above) but that causes the memory to freeze when the processor is stalled for another reason, and in the tests the CCE never starts executing code and the simulation never ends. Having the stall threshold allows the memory to stay enabled for data and control hazards, but I can also just register the stall signal (that is, essentially using a single bit counter). Would that be better than the parameterized bsg_counter_clear_up? That said, it's very possible I've missed something.

bp_me/src/v/cce/bp_cce_inst_ram.sv
Kellen-Hartnett: Fixed

bp_me/src/v/cce/bp_cce_inst_ram.sv
Kellen-Hartnett: Fixed

Kellen-Hartnett:
For some reason the commits show changes made before this branch from `me_dev`, doesn't actually need to change the files for those commits. Not sure what to do about that.

muwyse:
> For some reason the commits show changes made before this branch from `me_dev`, doesn't actually need to change the files for those commits. Not sure what to do about that.

The current diff looks good. What you can do now to remove the extra commits is do a soft reset to me_dev (from the HEAD of your branch), and then add and commit the changes to the single file as a single commit. Then force push your branch to your repo and the PR should auto-update.

muwyse:
> I've tried just using the stall_i signal directly (just tried the statement above) but that causes the memory to freeze when the processor is stalled for another reason, and in the tests the CCE never starts executing code and the simulation never ends. Having the stall threshold allows the memory to stay enabled for data and control hazards, but I can also just register the stall signal (that is, essentially using a single bit counter). Would that be better than the parameterized bsg_counter_clear_up? That said, it's very possible I've missed something.

This sounds like the stall threshold is just masking a bug then. The stall signal should disable a new read, but the RAM can be configured to hold the last read and then when stall_i goes low, it simply sends the last read to decode the next cycle and does a RAM read. I will try to take a look at a waveform today with the change to use stall_i directly and see if I can spot something.

bp_me/src/v/cce/bp_cce_inst_ram.sv
```diff
@@ -71,20 +73,42 @@ module bp_cce_inst_ram
...
+      ,.v_i(memory_read_enable)
       ,.data_i(ucode_data_i)
       ,.addr_i(ucode_v_i ? ucode_addr_i : fetch_pc_n)
       ,.w_i(ucode_w_i)
       ,.data_o(inst_o)
       );
-
+  
+  
+  //Disable instruction RAM after stalling for stall_threshold_p cycles
```
dpetrisko, line 96: Comment explaining 
- Benefits
- Reason we need a counter
- Default counter value justification

bp_me/src/v/cce/bp_cce_inst_ram.sv
```diff
@@ -25,6 +25,8 @@ module bp_cce_inst_ram
     `declare_bp_proc_params(bp_params_p)
     // Derived parameters
     , localparam cfg_bus_width_lp = `bp_cfg_bus_width(hio_width_p, core_id_width_p, cce_id_width_p, lce_id_width_p)
+    , localparam cfg_bus_width_lp = `bp_cfg_bus_width(domain_width_p, core_id_width_p, cce_id_width_p, lce_id_width_p)
```
dpetrisko, line 28: get rid of cfg_bus diff, it's residual

bp_me/src/v/cce/bp_cce_inst_ram.sv
```diff
@@ -25,6 +25,8 @@ module bp_cce_inst_ram
...
+    , parameter stall_threshold_p = 3
```
dpetrisko, line 29: What's the advantage of having this as more than 1?

muwyse:
Using stall_i directly with a minor change in the stall unit to ensure no erroneous stalls seems to work for me: https://github.com/black-parrot/black-parrot/pull/868

dpetrisko:
New PR LGTM, could you add @Kellen-Hartnett as a coauthor? @Kellen-Hartnett if your partner would like co-authorship can you give us their username?

https://docs.github.com/en/github/committing-changes-to-your-project/creating-a-commit-with-multiple-authors

muwyse:
I added both Kellen and Hans as co-authors in the other PR. I followed the directions exactly here (https://docs.github.com/en/github/committing-changes-to-your-project/creating-a-commit-with-multiple-authors) for the command line approach.

muwyse:
Closing due to merge of #868

PR #867, opened 2021-05-18
Update CFG addresses

dpetrisko:
Non 64-bit alignment causes issues with external buses which don't support wrapping behavior. This patch increases the space allocated to registers within the CFG bus. It'll break external software, but the internal nbf loader has been updated to support this. Docs have been updated to reflect the changes

A future patch should create a header which unifies the NBF loader along with the RTL so that they stay in sync automatically. Further changes to these addresses before that QoL enhancement are discouraged.

### DIFF
--- a/bp_common/software/py/nbf.py
+++ b/bp_common/software/py/nbf.py
@@ -26,14 +26,47 @@ import math
 import os
 import subprocess
 
+#  // The overall memory map of the config link is:
+#  //   16'h0000 - 16'h01ff: chip level config
+#  //   16'h0200 - 16'h03ff: fe config
+#  //   16'h0400 - 16'h05ff: be config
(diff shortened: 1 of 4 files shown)


muwyse:
By non 64-bit alignment do you mean that we were issuing 64-bit requests to misaligned addresses? Or that the external buses weren't supporting aligned 16 or 32-bit requests?

dpetrisko:
>By non 64-bit alignment do you mean that we were issuing 64-bit requests to misaligned addresses? Or that the external buses weren't supporting aligned 16 or 32-bit requests?

The former. The freeze register was at 0x2000_0001 and was technically a 64-bit register (etc, etc.)

muwyse (approved):
LGTM

PR #868, opened 2021-05-18
ME: skip CCE instruction RAM read on stall

muwyse:
This change blocks the CCE instruction RAM read when the currently executing instruction is stalling. The RAM is parameterized to latch the last read to keep the next instruction valid and ready for execution once the stall resolves. The stall logic is also slightly modified to ensure that stall is only raised for valid instructions.

This change supersedes #866  

ME, single core RISCV, and multicore regressions pass


muwyse:
@Kellen-Hartnett or @hansgaensbauer, any comments on this? I would be curious to see the results from running your stall profiling with these changes applied.

hansgaensbauer:
The results from the stall profiler are the same for when we set the stall_threshold_p to 0, that is, zero instructions fetches during stalls (as you'd expect). I tried it with mc_sanity_2 but the graphs aren't particularly interesting since it's so effective that there is no data to display. This is a much better solution.

PR #877, opened 2021-05-24
Fix cache tracer associativity

dpetrisko:
The tracers should use the overridable "assoc_p" variables instead of "icache_assoc_p", etc. Otherwise there is a mismatch between the bound parameters and the instantiated module parameters in the case of targeted testbenches


muwyse (commented):
LGTM

PR #883, opened 2021-05-29
Multi-element Return Address Stack

WasabiFan:
I implemented this for Oskin's architecture class and figured I would open a PR to see if there's any interest in upstreaming it. I won't be hurt if the answer is "no" 🙂 

---

This PR introduces a "real" multi-element RAS, with checkpointing support to restore the top-of-stack pointer ~~and topmost element~~ on misspeculation. Based on the benchmarking I've done, it almost perfects the possible performance one can get out of an RAS, up to the depth of the stack itself.

I default all configurations to have four entries. The overhead of this change, in addition to the handful of registers in the RAS itself and associated logic, is roughly doubling the width of a branch predictor "forward" message due to the whole top-of-stack PC included. _(edit: this has been removed and the alternative version used, see comment below. The difference turned out to be dramatically lower than I suggested.)_ There's an alternative version of this which only checkpoints the stack pointer which, relative to the statistics in the table below, would be at most two percentage points worse on CoreMark return hit percentage and 20 percentage points worse for the same on the "custom" test case. These are not exact numbers and I could benchmark such an implementation more precisely if there's interest.

---

The performance numbers are as follows. Note that the absolute values are slightly different (slightly _worse_) on latest fe_dev compared to the commit I originally branched off of and benchmarked on.

The "custom" test program represented in the rightmost column is a recursive function called in a loop; the function has two recursive call sites, which it alternates between. It recurses up to five calls deep. In other words, it's intentionally adversarial to the existing branch predictor -- confuses the BTB with alternating return targets and goes too deep for the existing one-element RAS.

Also note that the vast majority of CoreMark's returns are from invoking a small CRC function that makes no other calls.

image

For reference, on the latest fe_dev commit which this is now based on, the baseline is 1873217 cycles for CoreMark and this change brings it down to 1870327 cycles. It's approximately the same % change as in the table above but the numbers are higher. MPKI is also worse. I don't know what caused the drop in baseline performance in the last two weeks.

---

I'll let y'all decide if this is a worthwhile addition. The first real question is whether it's something you're interested in, and if so, I'm happy to take review feedback and update accordingly.

This PR, if it's going to be merged, should be squash-merged -- my full results for the class in question involve benchmarks of a bunch of different variants along the way, and as such the commit history is a mess.


bp_fe/src/include/bp_fe_defines.svh
```diff
@@ -43,6 +43,8 @@
       logic [bht_idx_width_mp-1:0]    bht_idx;                                                    \
       logic [bht_row_width_mp-1:0]    bht_row;                                                    \
       logic [ghist_width_mp-1:0]      ghist;                                                      \
+      logic [vaddr_width_mp-1:0]      ras_top_pc;                                                 \
+      logic [`BSG_WIDTH(ras_num_entries_mp-1)-1:0] ras_top_ptr;                                   \
```
WasabiFan, line 47: Some notes/questions on this:
- Should I be using `BSG_WIDTH` in these disparate places or pulling it into a constant? Perhaps since this is required to be a power of two anyway the parameter could just be the log-size?
- `BSG_WIDTH` does an addition before the `clog2`, such that it'll return one more bit than necessary if what you give it is the intended number of permutations. For this reason, I subtract one. However, I see other places which use it bare and provide a number of desired elements -- in other words, they seem to declare one more bit than necessary. I'd be curious to hear what the intent of `BSG_WIDTH` is.

bp_top/test/common/bp_nonsynth_branch_profiler.sv
```diff
@@ -119,9 +123,10 @@ module bp_nonsynth_branch_profiler
   final
     begin
       $fwrite(file, "Branch statistics\n");
-      $fwrite(file, "MPKI: %d\n", redirect_cnt / (instr_cnt / 1000));
-      $fwrite(file, "BTB hit%%: %d\n", (btb_hit_cnt * 100) / (attaboy_cnt + redirect_cnt));
-      $fwrite(file, "BHT hit%%: %d\n", (bht_hit_cnt * 100) / (br_cnt));
+      $fwrite(file, "MPKI: \t\t%.2f\n", redirect_cnt / (instr_cnt / 1000.0));
```
WasabiFan, line 126: MPKI could previously be `x` if fewer than 1k instructions were executed. I made all these floating-point, both to fix that issue and to provide more fine-grained statistics. I also added a counter for returns. I think these changes would be worthwhile on their own even if this PR doesn't go through.

bp_top/test/common/bp_nonsynth_branch_profiler.sv
```diff
@@ -81,10 +83,11 @@ module bp_nonsynth_branch_profiler
             br_cnt      <= br_cnt + branch_metadata.is_br;
             jal_cnt     <= jal_cnt + branch_metadata.is_jal;
             jalr_cnt    <= jalr_cnt + branch_metadata.is_jalr;
+            ret_cnt     <= ret_cnt + branch_metadata.is_ret;
 
             btb_hit_cnt <= btb_hit_cnt + branch_metadata.src_btb;
-            ras_hit_cnt <= ras_hit_cnt + branch_metadata.src_ret;
```
WasabiFan, line 86: Counting the RAS usages in this way is broken, based on my understanding of the pc gen logic.

Unlike the `src_btb` and similar flags, with the way the pipelining in the pc_gen module is done, `src_ras` is "1" on the instruction _fetched as a result of the prediction_ -- not the return which was predicted. This means it's two cycles later due to the squashed IF1 instruction. There's no attaboy nor redirect issued for the instruction _returned to_ (by design), and thus these counters only count when you happen to return directly to a jump.

I think it might be sensible to change the timing of these indicators in the pc_gen module, but have not done so.

bp_fe/src/include/bp_fe_defines.svh
dpetrisko: In general ras_idx_width_p is preferred as a parameter. This automatically handles the NPOT assertion below, and keeps all the signals tidy. The only place that needs to expand 2**ras_idx_width_p is the RAS itself

dpetrisko: To directly answer your question:
`BSG_SAFE_CLOG2(els_p) == address length for els_p
`BSG_WIDTH(els_p) == number of bits to represent the value els_p

They have different usages, but for indices `BSG_SAFE_CLOG2 is more appropriate

WasabiFan:
I've just pushed a change to remove checkpointing of the top value on the RAS. It now restores the top pointer to fix alignment but does not protect against a speculated pop _and_ push, which would overwrite the top element. The performance change is negligible.

For CoreMark cycle count:
```
Baseline:    1873217 
Original PR: 1870327
Modified:    1870398
```

It's only a handful of cycles worse. That translates to a drop to 99.15% return hit rate, from 99.31%.

Surprisingly to me, for my "custom" program, it had no effect whatsoever on return hit%.

bp_fe/src/include/bp_fe_defines.svh
WasabiFan: Thanks for the clarification; that makes sense. I've updated this to use a `ras_idx_width_p` as the configuration parameter. It's a lot cleaner now!

bp_top/test/common/bp_nonsynth_branch_profiler.sv
WasabiFan: It just occurred to me I still have a simulation open in which I was investigating this; here's a screenshot for reference, with gratuitous red Snipping Tool circles for ~~clarity~~ maximum confusion. Note how the behavior differs between the BTB and the RAS (`src_btb` and `src_ras`).

image

WasabiFan: Oh yeah, and returns are counted as BTB fetches for the purposes of these flags, since `ovr_ret` corresponds to IF1 but is being used in (among other places) IF2. Same behavior, different symptom.

dpetrisko: I am 100% willing to believe this profiler is broken. Maybe raise an issue? I don't think it should block this PR (although I wouldn't complain if it got fixed as part of it)

WasabiFan: Opened: https://github.com/black-parrot/black-parrot/issues/884. I don't have the bandwidth to look into it now but may be able to in future weeks.

bp_fe/src/include/bp_fe_defines.svh
```diff
@@ -43,6 +43,7 @@
       logic [bht_idx_width_mp-1:0]    bht_idx;                                                    \
       logic [bht_row_width_mp-1:0]    bht_row;                                                    \
       logic [ghist_width_mp-1:0]      ghist;                                                      \
+      logic [ras_idx_width_mp-1:0] ras_top_ptr;                                                   \
```
dpetrisko, line 46: ras_idx

bp_fe/src/include/bp_fe_defines.svh
```diff
@@ -28,7 +28,7 @@
       logic [vaddr_width_mp-1:0] imm;        \
     }  bp_fe_instr_scan_s;
 
-  `define declare_bp_fe_branch_metadata_fwd_s(btb_tag_width_mp, btb_idx_width_mp, bht_idx_width_mp, ghist_width_mp, bht_row_width_mp) \
+  `define declare_bp_fe_branch_metadata_fwd_s(btb_tag_width_mp, btb_idx_width_mp, bht_idx_width_mp, ghist_width_mp, bht_row_width_mp, ras_idx_width_mp, vaddr_width_mp) \
```
dpetrisko, line 31: remove vaddr_width as parameter

bp_fe/src/v/bp_fe_ras.sv
```diff
@@ -0,0 +1,139 @@
...
+      );
+
+  wire is_pop  = pop_pc_en_i;
+  wire is_push = push_pc_en_i;
+
+  // RAS memory
+  logic mem_w_v_li;
+  logic [ras_idx_width_p-1:0] mem_w_addr_li;
+  logic [vaddr_width_p-1:0] mem_w_data_li;
+  bsg_mem_1r1w
```
dpetrisko, line 85: This RAM can't realistically be hardened (async and small), so probably better to explicitly make it flops. Will get rid of the initialization logic as well, since the registers can be explicitly cleared

dpetrisko:
Sorry for the delay.  If you're still interested in working on this, I gave some feedback. Looks great in general! I'm sure we'll get this in eventually

taylor-bsg:
When we have the spec regression setup, we can get a more comprehensive
performance analysis and dial in on recommended default parameters!

M

bp_fe/src/v/bp_fe_ras.sv
WasabiFan: :+1: updated. Removed the init logic and replaced it with a direct initialization of the regs.

WasabiFan:
Updated for latest `fe_dev` and did some cursory re-validation. The commit history is a mess, so if this is merged it should be squashed.

dpetrisko:
Closing; included with #1179

PR #888, opened 2021-06-06
Breaking critical uncached path in cache engines

dpetrisko:
This PR does a few things.

- The most important is the critical path fix in the UCE and LCE.  Currently, we stream out non-blocking requests (uc, wt, l2 amo) in a single cycle. This leads to a path from the cache AMO unit through the cache engine to routers, accelerators, etc.  Instead, this PR delays the output of the request by a single cycle and accepts a new request if the non-blocking request is sent out in the same cycle. In this way, we maintain 100% throughput for non-blocking requests, so there are no cycle-level performance implications.
- The second is adding a tag mem read register in the LCE. This is currently unused because the shadow tags are not needed by the CCE which maintains the true tags. However, I tested it and synthesis rips this unused register out anyway, so I think it's good to have just for future proofing.
- The third is a helper module bsg_dff_sync_read, which maintains a synchronous read buffer. This is purely aesthetic, but there are a bunch of other places in the code where this pattern is used, so we can replace at our leisure.

@muwyse I'm going to pull this into the tapeout branch, so I'd appreciate a quick review. Fine with merging to me_dev after the burst changes if you foresee any conflicts.

### DIFF
--- a/bp_be/syn/flist.vcs
+++ b/bp_be/syn/flist.vcs
@@ -287,6 +287,8 @@ $BP_TOP_DIR/src/v/bp_tile_node.sv
 
 # BSG Staging area
 $BP_COMMON_DIR/src/v/bsg_async_noc_link.sv
-$BP_COMMON_DIR/src/v/bsg_dff_reset_half.v
 $BP_COMMON_DIR/src/v/bsg_cache_dma_to_wormhole.v
+$BP_COMMON_DIR/src/v/bsg_dff_reset_half.v
+$BP_COMMON_DIR/src/v/bsg_dff_sync_read.v
 $BP_COMMON_DIR/src/v/bsg_wormhole_to_cache_dma_fanout.v
+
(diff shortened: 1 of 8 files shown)


bp_top/src/v/bp_unicore_complex.sv
```diff
@@ -0,0 +1,187 @@
...
+         ,.routing_matrix_p(routing_matrix_p)
+         )
+       io_router
+        (.clk_i(clk_i)
+         ,.reset_i(reset_i)
+    
+         ,.my_cord_i()
+    
+         ,.link_i(
+         ,.link_o(
```
muwyse, line 182: This looks incomplete?

muwyse: Or perhaps this new file isn't actually part of this PR?

muwyse (changes requested):
LCE/UCE changes look good.

Might be an extra file added that isn't part of PR?

I would like to merge this to me_dev after CCE burst because those changes have been outstanding for so long. I don't see any conflicts between the two changes.

bp_top/src/v/bp_unicore_complex.sv
dpetrisko: Whoops! Yes I'll remove this

dpetrisko:
good to go?

muwyse (approved):
LGTM

PR #890, opened 2021-06-09
Adding bedrock register and stream peripherals

dpetrisko:
- Converts the tile peripherals to single-beat stream in prep for future
- Adds bp_me_bedrock_register which acts as a Bedrock->GPIO converter. This makes things a little bit easier to connect low-performance peripherals to a tile. Can be expanded as needed for bursting requirements

Depends on #888

### DIFF
--- a/bp_me/src/v/cache/bp_me_bedrock_register.sv
+++ b/bp_me/src/v/cache/bp_me_bedrock_register.sv
@@ -2,6 +2,13 @@
 `include "bp_common_defines.svh"
 `include "bp_me_defines.svh"
 
+//
+// This module is used to interface a BP Stream interface to a general-purpose
+//   register read/write interface. The data is stored externally so that
+//   control/status registers can be controlled by this interface while
+//   retaining special semantics. Registers are assumed to be synchronous
+//   read/write which is compatible (although suboptimal) for asynchronous
+//   registers.
 module bp_me_bedrock_register
  import bp_common_pkg::*;
  import bp_me_pkg::*;
@@ -9,14 +16,22 @@ module bp_me_bedrock_register
    `declare_bp_proc_params(bp_params_p)
    `declare_bp_bedrock_mem_if_widths(paddr_width_p, dword_width_gp, lce_id_width_p, lce_assoc_p, xce)
 
+   // The width of the registers. Currently, must all be the same.
    , parameter reg_width_p = dword_width_gp
+   // The address width of the registers. For addresses less than paddr_width_p,
+   //   the upper bits of the paddr are ignored for matching purposes
    , parameter reg_addr_width_p = paddr_width_p
+   // The number of registers to control
    , parameter els_p = 1
+   // An unpacked array of integer register base addresses
+   //   e.g. localparam integer base_addr_lp [1:0] = '{0xf00bad, 0x00cafe}
+   // Can also accept pattern matches such as 0x8???
    , parameter integer base_addr_p [els_p-1:0] = '{0}
    )
   (input                                            clk_i
    , input                                          reset_i
 
+   // Network-side BP-Stream interface
    , input [xce_mem_msg_header_width_lp-1:0]        mem_cmd_header_i
    , input [dword_width_gp-1:0]                     mem_cmd_data_i
    , input                                          mem_cmd_v_i
@@ -30,7 +45,7 @@ module bp_me_bedrock_register
    , output logic                                   mem_resp_last_o
 
 
-   // Synchronous read/write interface.
+   // Synchronous register read/write interface.
    // Actually 1rw, but expose both ports to prevent unnecessary and gates
    , output logic [els_p-1:0]                       r_v_o
    , output logic [els_p-1:0]                       w_v_o
@@ -48,7 +63,7 @@ module bp_me_bedrock_register
   logic mem_cmd_v_li, mem_cmd_yumi_li;
   bsg_one_fifo
    #(.width_p($bits(bp_bedrock_xce_mem_msg_s)))
-   header_fifo
+   cmd_fifo
     (.clk_i(clk_i)
      ,.reset_i(reset_i)
   
--- a/bp_me/src/v/cache/bp_me_bedrock_register.sv
+++ b/bp_me/src/v/cache/bp_me_bedrock_register.sv
@@ -1,0 +1,94 @@
+
+`include "bp_common_defines.svh"
+`include "bp_me_defines.svh"
+
+module bp_me_bedrock_register
+ import bp_common_pkg::*;
+ import bp_me_pkg::*;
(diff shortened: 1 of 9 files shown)


bp_me/src/v/cache/bp_me_bedrock_register.sv
```diff
@@ -0,0 +1,94 @@
...
+
+   , input [xce_mem_msg_header_width_lp-1:0]        mem_cmd_header_i
+   , input [dword_width_gp-1:0]                     mem_cmd_data_i
+   , input                                          mem_cmd_v_i
+   , output logic                                   mem_cmd_ready_and_o
+
+   , output logic [xce_mem_msg_header_width_lp-1:0] mem_resp_header_o
+   , output logic [dword_width_gp-1:0]              mem_resp_data_o
+   , output logic                                   mem_resp_v_o
+   , input                                          mem_resp_ready_and_i
```
muwyse, line 28: These should conform to BedRock Stream interface (https://github.com/black-parrot/black-parrot/blob/me_dev/docs/bedrock_guide.md#bedrock-stream) - i.e., they are missing the last signal. Or, if they will always be a single data beat then rename and add comment on interface ports saying they are BedRock Lite. If we want to support multi-beat messages in the future (or imply it could happen) we should make this BedRock Stream, add the last signal, and just tie it to 1 for now.

bp_top/src/v/bp_cfg.sv
```diff
@@ -13,114 +13,100 @@ module bp_cfg
...
+  (input                                            clk_i
+   , input                                          reset_i
 
-   , input [xce_mem_msg_width_lp-1:0]        mem_cmd_i
-   , input                                   mem_cmd_v_i
-   , output                                  mem_cmd_ready_and_o
+   , input [xce_mem_msg_header_width_lp-1:0]        mem_cmd_header_i
+   , input [dword_width_gp-1:0]                     mem_cmd_data_i
+   , input                                          mem_cmd_v_i
+   , output logic                                   mem_cmd_ready_and_o
```
muwyse, line 22: Same comment about interface as above.

bp_top/src/v/bp_clint_slice.sv
```diff
@@ -15,75 +15,47 @@ module bp_clint_slice
   (input                                                clk_i
    , input                                              reset_i
 
-   , input [xce_mem_msg_width_lp-1:0]                   mem_cmd_i
+   , input [xce_mem_msg_header_width_lp-1:0]            mem_cmd_header_i
```
muwyse, line 18: Same comment about interface as above.

bp_top/src/v/bp_loopback.sv
```diff
@@ -13,41 +13,36 @@ module bp_cce_loopback
...
-    , input                                         mem_cmd_v_i
-    , output                                        mem_cmd_ready_and_o
-
-    , output [cce_mem_msg_width_lp-1:0]             mem_resp_o
-    , output                                        mem_resp_v_o
-    , input                                         mem_resp_yumi_i
+   (input                                            clk_i
+    , input                                          reset_i
+
+    , input [cce_mem_msg_header_width_lp-1:0]        mem_cmd_header_i
```
muwyse, line 19: Same comment about interface as above.

bp_top/src/v/bp_cfg.sv
```diff
@@ -13,114 +13,100 @@ module bp_cfg
...
-   hio_mask_reg
-    (.clk_i(clk_i)
-     ,.reset_i(reset_i)
-     ,.en_i(hio_w_v_li)
-     ,.data_i(hio_li)
-     ,.data_o(hio_mask_r)
-     );
+  assign cce_ucode_v_o    = cce_ucode_r_v_li | cce_ucode_w_v_li;
+  assign cce_ucode_w_o    = cce_ucode_w_v_li;
+  assign cce_ucode_addr_o = addr_lo[3+:cce_pc_width_p];
```
muwyse, line 108: the 3 maybe shouldn't be a magic number, for clarity. Can it be derived from somewhere?

bp_me/src/v/cache/bp_me_bedrock_register.sv
```diff
@@ -0,0 +1,94 @@
+
+`include "bp_common_defines.svh"
+`include "bp_me_defines.svh"
+
+module bp_me_bedrock_register
```
muwyse, line 5: A comment or two about this module could be helpful. I initially thought, based on the name, that this module would be holding the registered data internally. But, if I understand correctly, it is actually generating the read/write signals for some els_p set of registers, each reg_width_p wide. For read requests it generates r_v_o output and the next cycle the data is valid on data_i, which is then formed into the mem_resp_data_o, correct?

bp_me/src/v/cache/bp_me_bedrock_register.sv
```diff
@@ -0,0 +1,94 @@
...
+   );
+
+  `declare_bp_bedrock_mem_if(paddr_width_p, dword_width_gp, lce_id_width_p, lce_assoc_p, xce);
+
+  bp_bedrock_xce_mem_msg_header_s mem_cmd_header_li;
+  logic [dword_width_gp-1:0] mem_cmd_data_li;
+  logic mem_cmd_v_li, mem_cmd_yumi_li;
+  bsg_one_fifo
+   #(.width_p($bits(bp_bedrock_xce_mem_msg_s)))
+   header_fifo
```
muwyse, line 47: rename to cmd_fifo? Buffers header and a data word.

muwyse (changes requested):
Looks good overall. A few small questions to clarify.

bp_me/src/v/cache/bp_me_bedrock_register.sv
dpetrisko: Yeah I'm fine with that. I figured users would just hardcode the signals outside of the module, but it's more clear to stub it inside

dpetrisko: It actually does support multi-beat messages implicitly, since the autoincrement happens and last is ignored

muwyse: > It actually does support multi-beat messages implicitly, since the autoincrement happens and last is ignored

lol, fair.

> Yeah I'm fine with that. I figured users would just hardcode the signals outside of the module, but it's more clear to stub it inside

Agreed. Stubbing the signals in this module makes it clearer.

bp_me/src/v/cache/bp_me_bedrock_register.sv
dpetrisko: added comments!

dpetrisko:
Comments addressed, thanks!

muwyse (approved):
LGTM!

PR #893, opened 2021-07-01
Update ME components to use Stream Interfaces

muwyse:


bp_top/src/v/bp_l2e_tile.sv
```diff
@@ -473,32 +473,86 @@ module bp_l2e_tile
   assign cfg_mem_resp_yumi_li      = cce_mem_resp_yumi_lo & cfg_mem_resp_grant_li;
   assign loopback_mem_resp_yumi_li = cce_mem_resp_yumi_lo & loopback_mem_resp_grant_li;
 
+  bp_bedrock_cce_mem_msg_header_s cache_mem_cmd_header_lo;
```
dpetrisko, line 476: note: make I/O cce stream to get rid of these

bp_common/src/include/bp_common_aviary_pkgdef.svh
```diff
@@ -244,15 +244,15 @@
       ,dcache_sets          : 64
       ,dcache_assoc         : 8
       ,dcache_block_width   : 512
```
dpetrisko, line 246: @muwyse @taylor-bsg Let's discuss which defaults make sense offline

bp_me/src/v/dev/bp_me_cce_to_cache.sv
```diff
@@ -65,312 +79,332 @@ module bp_me_cce_to_cache
...
+            // send ready_and signal back to pump_out
+            mem_cmd_yumi_li = cache_pkt_ready_i & cache_pkt_v_o;
+
+            //cmd_state_n = (mem_cmd_new_lo & mem_cmd_yumi_li) ? STREAM : READY;
           end
-      end
-      RESP_SEND: begin
-        mem_resp_v_o = 1'b1;
-        if (mem_resp_yumi_i)
+        STREAM:
```
dpetrisko, line 367: This state is unused; let's remove

bp_me/src/v/network/bp_me_burst_to_stream.sv
```diff
@@ -0,0 +1,122 @@
...
+    ,.set_i(in_msg_header_v_i)
+    ,.clear_i(header_clear)
+    ,.data_o(header_v_r)
+    );
+  assign header_v_lo  = in_msg_header_v_i | header_v_r;
+  assign header_clear = out_msg_ready_and_i & out_msg_v_o & out_msg_last_o;
+  // Accept new header only if there is no header currently buffered
+  assign in_msg_header_ready_and_o = ~header_v_r;
+
+  // Stream pump to handle all the details of properly forming the output BedRock stream message
```
dpetrisko, line 87: This is a super interesting use of stream pump!

bp_me/src/v/network/bp_me_xbar_burst_bidir.sv
```diff
@@ -0,0 +1,136 @@
+/**
```
dpetrisko, line 1: I tend towards getting rid of this code, since it's a straight passthrough wrapper. bsg style is to eliminate wrappers unless they contain some kind of complex logic

bp_me/src/v/network/bp_me_xbar_stream_bidir.sv
```diff
@@ -0,0 +1,112 @@
+/**
```
dpetrisko, line 1: Same as above -- unjustified wrapper

bp_top/src/v/bp_tile.sv
```diff
@@ -680,51 +407,268 @@ module bp_tile
...
 
+
+  //synopsys translate_off
+  // This assertion guarantees that the data width of CCE-Mem interface is consistent in the system
+  // for all devices. Clint, CFG, ME Loopback and CCE all use dword_width_gp for the data width, but
+  // bp_me_cce_to_cache uses l2_data_width_p. This assertion could be removed if an adapter was
+  // attached to the mem_cmd/resp ports of bp_me_cce_to_cache, or the module itself was parameterized
+  // to support a different data width on the mem_cmd/resp interface.
+  initial begin
+    assert(l2_data_width_p == dword_width_gp) else
```
dpetrisko, line 666: Is the issue here just the zero extending / truncating of the device signals? Admittedly there's not a super clean way I know of, but the extra crossbar logic should synthesize out. We wouldn't want to add a PISO/SIPO to the cache adapter to solve what is ostensibly a Verilog issue

bp_top/src/v/bp_tile.sv
```diff
@@ -17,13 +17,16 @@ module bp_tile
...
+
+   , localparam cfg_bus_width_lp        = `bp_cfg_bus_width(hio_width_p, core_id_width_p, cce_id_width_p, lce_id_width_p)
 
-    , localparam cfg_bus_width_lp        = `bp_cfg_bus_width(hio_width_p, core_id_width_p, cce_id_width_p, lce_id_width_p)
    // Wormhole parameters
    , localparam coh_noc_ral_link_width_lp = `bsg_ready_and_link_sif_width(coh_noc_flit_width_p)
    , localparam mem_noc_ral_link_width_lp = `bsg_ready_and_link_sif_width(mem_noc_flit_width_p)
+
+   // LCE-CCE network wormhole adapter length input/output field size
+   , localparam pr_len_width_lp = 8
```
dpetrisko, line 29: localparam should go where they're first used unless they're needed for port widths

bp_top/src/v/bp_unicore.sv
```diff
@@ -53,75 +76,79 @@ module bp_unicore
...
 
-  bp_bedrock_uce_mem_msg_s mem_cmd_lo;
-  logic mem_cmd_v_lo, mem_cmd_ready_and_li;
-  bp_bedrock_uce_mem_msg_s mem_resp_li;
-  logic mem_resp_v_li, mem_resp_yumi_lo;
+  //synopsys translate_off
+  // This assertion guarantees that the data widths of bp_unicore_lite and bp_me_cce_to_cache match.
+  // bp_me_cce_to_cache use l2_data_width_p while bp_unicore_lite uses uce_mem_data_width_lp.
+  initial begin
+    assert(l2_data_width_p == uce_mem_data_width_lp) else
```
dpetrisko, line 85: I believe we already have this assertion in the nonsynth_if_verif (icache_fill_width == dcache_fill_width == l2_data_width). If not, it belongs there

bp_me/src/v/network/bp_me_stream_pump_out.sv
```diff
@@ -0,0 +1,226 @@
...
+ *   Generates a BedRock Stream protocol output message from an FSM that provides
+ *   a base header and, if required, data words. The base header is held constant
+ *   by the FSM throughout the transaction.
+ *
+ */
+
+`include "bp_common_defines.svh"
+`include "bp_me_defines.svh"
+
+module bp_me_stream_pump_out
```
dpetrisko, line 16: We should have an optional input buffer for stream pump out

bp_me/src/v/network/bp_me_stream_pump_in.sv
```diff
@@ -0,0 +1,245 @@
...
+   //    cache arriving on the BedRock Stream input can be decomposed into a stream of
+   //    8B reads on the FSM output port.
+   // 3. Message types set in both will have N:N beats. Every beat on the input
+   //    will produce a beat on the output. This is commonly used for all messages
+   //    with data payloads.
+   // Constructed as (1 << e_rd/wr_msg | 1 << e_uc_rd/wr_msg)
+   , parameter msg_stream_mask_p = 0
+   , parameter fsm_stream_mask_p = msg_stream_mask_p
+
+   , parameter buffer_els_p = 2
```
dpetrisko, line 45: buffer_els should be optionally 0, we will have some cases of stream pumps which do not need the input buffers. Unfortunately, that means we should revert the fsm_yumi_i change to fsm_ready_and_i (no functional change, but slight clarity regression).

bp_top/src/v/bp_l2e_tile.sv
dpetrisko: created #909, resolving

bp_top/src/v/bp_tile.sv
dpetrisko: Perhaps we can extend the xbar_stream to handle heterogenous widths with a parameter array...

bp_me/src/v/network/bp_me_xbar_burst_bidir.sv
dpetrisko: See page 26 here: https://docs.google.com/document/d/1xA5XUzBtz_D6aSyIBQUwFk_kSUdckrfxa2uzGjMgmCU/edit#heading=h.9c85j16kz5tc

bp_me/src/v/cce/bp_cce.sv
```diff
@@ -32,6 +32,10 @@ module bp_cce
     , localparam cfg_bus_width_lp          = `bp_cfg_bus_width(hio_width_p, core_id_width_p, cce_id_width_p, lce_id_width_p)
     `declare_bp_bedrock_lce_if_widths(paddr_width_p, cce_block_width_p, lce_id_width_p, cce_id_width_p, lce_assoc_p, lce)
     `declare_bp_bedrock_mem_if_widths(paddr_width_p, cce_block_width_p, lce_id_width_p, lce_assoc_p, cce)
+
+    // stream pump
+    , localparam stream_words_lp = cce_block_width_p / dword_width_gp
```
dpetrisko, line 37: define at use

bp_me/src/v/cce/bp_cce.sv
```diff
@@ -287,6 +296,74 @@ module bp_cce
...
+  // Memory Response Stream Pump
+  bp_me_stream_pump_in
+    #(.bp_params_p(bp_params_p)
+      ,.stream_data_width_p(dword_width_gp)
+      ,.block_width_p(cce_block_width_p)
+      ,.payload_width_p(cce_mem_payload_width_lp)
+      ,.msg_stream_mask_p(mem_resp_payload_mask_gp)
+      ,.fsm_stream_mask_p(mem_resp_payload_mask_gp)
+      // provide buffer space for two 8-beat long data carrying messages
+      ,.buffer_els_p(16)
```
dpetrisko, line 317: derive, not magic

bp_me/src/v/network/bp_me_stream_pump_in.sv
```diff
@@ -0,0 +1,245 @@
...
+
+  `declare_bp_bedrock_if(paddr_width_p, payload_width_p, stream_data_width_p, lce_id_width_p, lce_assoc_p, xce);
+
+  `bp_cast_o(bp_bedrock_xce_msg_header_s, fsm_base_header);
+
+  bp_bedrock_xce_msg_header_s msg_header_lo;
+  logic [stream_data_width_p-1:0] msg_data_lo;
+  logic msg_v_lo, msg_yumi_li, msg_last_lo;
+
+  bsg_fifo_1r1w_small
```
dpetrisko, line 88: Header is repeated here N times, we don't want to store it each time or it'll be a huge waste. We should have a data buffer and a smaller header buffer

bp_me/src/v/cce/bp_cce.sv
dpetrisko: also, 16 seems excessive?

dpetrisko: And last, we should think about whether it makes more sense to buffer at CCE or at the cache pump. I think this a knob we shouldn't turn until we have profile information

bp_top/src/v/bp_tile.sv
muwyse: The issue is me_cce_to_cache uses l2_data_width_p to define its mem_cmd/resp interface data width and the xbar uses dword_width_gp, I suppose. (and also, the rest of the devices use dword_width_p for the data width, but perhaps these synthesize out). So, if l2_data_width_p is increased then it is highly likely that data will be dropped each beat by the narrower xbar.

bp_me/src/v/cce/bp_cce.sv
muwyse: I will double check the value, but the coherence protocol requires some buffering on the memory response channel to guarantee safety (i.e., to not livelock). This was always set as two messages worth, which is where the 16 came from (= 2 * 8 beats).

The max beats per message can be derived, the 2 is a constant from the protocol.

muwyse: > And last, we should think about whether it makes more sense to buffer at CCE or at the cache pump. I think this a knob we shouldn't turn until we have profile information

Not sure what you mean here. The cache pumps are in the CCE and connect directly to the mem_cmd/resp interfaces of the CCE. If we are going to have buffering ability in the stream pumps then it doesn't make sense to set the buffer sizes to 0 and add buffers external to the stream pumps inside the CCE.

bp_me/src/v/network/bp_me_stream_pump_out.sv
muwyse: I think this should be added after this PR as there is no current requirement for it in the system.

bp_me/src/v/network/bp_me_stream_pump_in.sv
muwyse: buffers have been reworked, with support for 0 elements (no buffering), and split header and data buffers. This did not require adjusting the interface on the FSM side.

bp_top/src/v/bp_tile.sv
muwyse: I moved this assertion to nonsynth_if_verif. Allowing l2_data_width to be larger than dword_width is likely fine for the devices (should synthesize out), but there is also a mismatch then with the xbar and the CCE. We could make the xbar use l2_data_width, but then the muxes will get larger as l2_data_width increases, but this doesn't then solve a width mismatch between xbar and CCE since the CCE expects dword_width sized data beats. Adjusting the CCE would take some serious thought and possible redesign.

bp_me/src/v/network/bp_me_stream_pump_out.sv
muwyse: I added a placeholder for the buffering. After doing that, my opinion is that it is much easier to buffer the output, even though this would mean buffering header+data beats. Buffering on the input would require duplicating a lot of the stream pump logic to compute fsm_done_o and fsm_count_o, which must be tied to the FSM input interface,. The signals that these ports rely on are currently located at the fsm-msg connection point (which would be the output side of a potential buffer).

bp_me/src/v/network/bp_me_stream_pump_in.sv
```diff
@@ -0,0 +1,311 @@
...
+  logic [stream_data_width_p-1:0] msg_data_lo;
+  logic msg_v_lo, msg_yumi_li, msg_last_lo;
+
+  if (buffer_els_p == 0)
+    begin: passthrough
+      assign msg_header_lo = msg_header_cast_i;
+      assign msg_data_lo = msg_data_i;
+      assign msg_v_lo = msg_v_i;
+      assign msg_last_lo = msg_last_i;
+      assign msg_ready_and_o = msg_yumi_li;
```
dpetrisko, line 96: can't connect ready_and_o to yumi_i

muwyse: fixed. all users of this module still connect a yumi to the port, but the stream pump's interface is now ready_and

bp_me/src/v/cce/bp_cce.sv
dpetrisko: I mean at the pumps in the CCE vs the pumps in the L2. Effectively the difference is whether the buffers come before or after the crossbar

bp_me/src/v/network/bp_me_stream_pump_out.sv
dpetrisko: works for me

PR #895, opened 2021-07-14
update FSM CCE to use BedRock Stream on memory ports

muwyse:
This PR updates the FSM CCE to use BedRock Stream on its memory ports.

For now, this is a draft for viewing the diffs.

### DIFF
--- a/bp_me/src/v/cce/bp_cce_fsm.sv
+++ b/bp_me/src/v/cce/bp_cce_fsm.sv
@@ -16,9 +16,6 @@
  *
  *   Atomics in L2/Mem will be supported in a future change.
  *
- *   TODO items:
- *   - should CCE align addresses for memory commands, lce commands, etc.?
- *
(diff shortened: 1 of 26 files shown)


bp_me/src/v/cce/bp_cce_fsm.sv
```diff
@@ -254,7 +254,7 @@ module bp_cce_fsm
   // MSHR
   `declare_bp_cce_mshr_s(lce_id_width_p, lce_assoc_p, paddr_width_p);
   bp_cce_mshr_s mshr_r, mshr_n;
-  logic [paddr_width_p-1:0] mshr_r_paddr_aligned = (mshr_r.paddr >> lg_block_size_in_bytes_lp) << lg_block_size_in_bytes_lp;
```
dpetrisko: oh good catch, that's scary

Issue #784, opened 2021-03-02
Fix retiming for Vivado FPGA synthesis (FMA + FAUX)


dpetrisko:
We have RTL for an FMA unit which requires 5 cycle retiming [fma](https://github.com/black-parrot/black-parrot/blob/master/bp_be/src/v/bp_be_calculator/bp_be_pipe_fma.sv#L202)

When we use the retiming flag, we only see a few logic levels moved around. (DC balances these pipelines well). Logic levels for the entire design are ~20.

```
INFO: [Synth 8-5816] Retiming module `wrapper_GT0_tempName`
    Effective logic levels on critical path before retiming is: 60
    Total number of crtical paths = 198
    Optimizing at the module level
    Optimizing locally to improve critical paths(may not reduce worst delay)
    Effective logic levels on critical path after retiming is: 49
    Total number of crtical paths = 3127
    Numbers of forward move = 190, and backward move = 125
```

It seems that Rocket has also encountered this issue and given up citing backwards retiming being a strategy for ASICs rather than FPGA:
[issue 1](https://github.com/ucb-bar/fpga-zynq/issues/30)
[issue 2](https://github.com/ucb-bar/fpga-zynq/issues/57)

Maybe there exists some high-effort retiming or targeted retiming flags to help out? This is our current synthesis command:

```
synth_design -top wrapper -part $PART -include_dirs $dir_list -flatten_hierarchy rebuilt -retiming
```

dpetrisko:
Fixed with #899

PR #899, opened 2021-07-17
Parameterized manual retiming for Zynq 7020

mysoreanoop:
time. A few details to check:

## Issue Fixed
Does this PR address a BlackParrot GitHub issue? Link it here:
dpetrisko: https://github.com/black-parrot/black-parrot/issues/784

## Area
Module/tool/makefile/script in question.
bp_be_pipe_fma.v, Vivado, Zynq 7020

Why is this change required?
For devices/tools that do not do the necessary retiming automatically

Next steps, downstream modules affected, etc.
Once HardFloat changes are pulled, we can set manual latency distribution; otherwise unchanged.
[Ref](https://github.com/bsg-external/HardFloat/pull/8)

## Analysis
None as is
dpetrisko: expected frequency boost from 25->50MHz in FPGA. Minimal area impact in ASIC/FPGA and no performance impact.

## Verification
None, open to suggestions
dpetrisko: 
- [x] beebs suite
- [x] FPGA synthesis
- [ ] ASIC synthesis

## Additional Context

### DIFF
--- a/bp_be/src/v/bp_be_calculator/bp_be_pipe_fma.sv
+++ b/bp_be/src/v/bp_be_calculator/bp_be_pipe_fma.sv
@@ -10,29 +10,18 @@
  *   This module relies on cross-boundary flattening and retiming to achieve
  *     good QoR
  *   
- *   pipeline_stages_p is used to insert pipelining registers in rawMul for some devices, 
- *   where automatic backwards retiming from the retiming_chain doesn't help. 
- *     - No pipelining (default)   : '{0,0}
- *     - 64 bit, Zynq 7020 @ 50 MHz: '{3,1} (distributed as in the below figure)
+ *   ASIC tools prefer to have retiming chains be pure register chains at the end of
+ *   a combinational logic cloud, whereas FPGA tools prefer explicitly instantiated registers.
+ *   With this FPGA optimization, we've achieved 50MHz on a Zynq 7020
  *
  *   This module:
  *            ...
- *            fma         -> takes pipeline_p[0]
- *           /   \
- *     pre_round  \       -> takes pipeline_p[1]
- *      round_*p   \      -> ideally, some x should cross here
- *         |        |
- *         x        y     -> remaining pipeline (retiming_chain)
- *     fma_out   imul_out
+ *            fma 4 cycles     reservation
+ *           /   \                 |
+ *        round  imul_out      meta_info
+ *          |
+ *       fma_out
  *
- *   fma: (internally)
- *         preMul
- *          /  \
- *         0    0        -> 1 pipeline stage
- *         | (mulAdd)    -> Integer MulAdd (uses DSP on Zynq 7020)
- *         1    1        -> 2 pipeline stages
- *          \  /
- *         postMul
  */
 `include "bp_common_defines.svh"
 `include "bp_be_defines.svh"
@@ -42,10 +31,6 @@ module bp_be_pipe_fma
  import bp_be_pkg::*;
  #(parameter bp_params_e bp_params_p = e_bp_default_cfg
    `declare_bp_proc_params(bp_params_p)
-   , parameter int pipeline_stages_p[0:1] = {0,0},
-   , parameter imul_latency_p = "inv"
-   , parameter fma_latency_p  = "inv"
-
    , localparam dispatch_pkt_width_lp = `bp_be_dispatch_pkt_width(vaddr_width_p)
    )
   (input                               clk_i
@@ -75,17 +60,15 @@ module bp_be_pipe_fma
   assign frs1 = reservation.rs1;
   assign frs2 = reservation.rs2;
   assign frs3 = reservation.imm;
-  wire [dword_width_gp-1:0] rs1 = reservation.rs1[0+:dword_width_gp];
-  wire [dword_width_gp-1:0] rs2 = reservation.rs2[0+:dword_width_gp];
+  wire [dword_width_gp-1:0] rs1 = decode.opw_v ? (frs1 << word_width_gp) : frs1;
+  wire [dword_width_gp-1:0] rs2 = frs2;
 
   //
   // Control bits for the FPU
   //   The control bits control tininess, which is fixed in RISC-V
   rv64_frm_e frm_li;
-  rv64_frm_e frm_r;
   assign frm_li = (instr.t.fmatype.rm == e_dyn) ? frm_dyn_i : rv64_frm_e'(instr.t.fmatype.rm);
   wire [`floatControlWidth-1:0] control_li = `flControl_default;
-  wire [`floatControlWidth-1:0] control_r;
 
   wire is_fadd_li    = (decode.fu_op == e_fma_op_fadd);
   wire is_fsub_li    = (decode.fu_op == e_fma_op_fsub);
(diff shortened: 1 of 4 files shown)


dpetrisko:
Hi, thanks for the PR. You can embed these retiming registers at the input/output of the FMA unit and that should get the same QoR as having them in bp_be_pipe_fma, but will be a lot cleaner (as long as cross-boundary retiming / ungrouping is enabled for the FPU).

bp_be/src/v/bp_be_calculator/bp_be_pipe_fma.sv
```diff
@@ -9,7 +9,25 @@
  * Notes:
  *   This module relies on cross-boundary flattening and retiming to achieve
  *     good QoR
- *
+ *   FPGA retiming parameter array is used to distribute the latency manually
```
dpetrisko, line 12: A diagram like this would be useful in the hardfloat module as well

bp_be/src/v/bp_be_calculator/bp_be_pipe_fma.sv
```diff
@@ -19,7 +37,7 @@ module bp_be_pipe_fma
  import bp_be_pkg::*;
  #(parameter bp_params_e bp_params_p = e_bp_default_cfg
    `declare_bp_proc_params(bp_params_p)
-
+   , parameter integer latency_dstr_p[0:2] = {0,0,0} //Default = all retiming DFFs at the terminal DFF chain
```
dpetrisko, line 40: What are these set to in the zynq?

bp_be/src/v/bp_be_calculator/bp_be_pipe_fma.sv
```diff
@@ -105,13 +124,21 @@ module bp_be_pipe_fma
...
+
+  logic invalid_exc_n, is_nan_n, is_inf_n, is_zero_n;
+  logic fma_out_sign_n;
+  logic [dp_exp_width_gp+1:0] fma_out_sexp_n;
+  logic [dp_sig_width_gp+2:0] fma_out_sig_n;
+
   mulAddRecFNToRaw
    #(.expWidth(dp_exp_width_gp)
      ,.sigWidth(dp_sig_width_gp)
+     ,.latencyDstr({latency_dstr_p[0], latency_dstr_p[1]})
```
dpetrisko, line 136: {1, 0}?

bp_be/src/v/bp_be_calculator/bp_be_pipe_fma.sv
dpetrisko: Also, rather than having each independent, it's usually cleaner to do have it aggregate such as https://github.com/bespoke-silicon-group/basejump_stl/blob/master/bsg_noc/bsg_wormhole_router.v#L16 does it. Then you can use the latencies directly in most places, rather than needing to sum each

bp_be/src/v/bp_be_calculator/bp_be_pipe_fma.sv
```diff
@@ -105,13 +124,21 @@ module bp_be_pipe_fma
   logic [dp_exp_width_gp+1:0] fma_out_sexp;
   logic [dp_sig_width_gp+2:0] fma_out_sig;
   logic [dword_width_gp-1:0] imul_out;
+
+  logic invalid_exc_n, is_nan_n, is_inf_n, is_zero_n;
```
dpetrisko, line 128: These should get absorbed based on my high level comment, but in general:
basejump_stl convention is _r for registered values, and _n for next values. Even if they are latched in the module, from the perspective of these wires they are _r.

bp_be/src/v/bp_be_calculator/bp_be_pipe_fma.sv
```diff
@@ -186,24 +241,27 @@ module bp_be_pipe_fma
...
+   imul_retiming_chain
     (.clk_i(clk_i)
 
      ,.data_i({imul_v_li, imul_result})
      ,.data_o({imul_v_o, imul_data_o})
      );
 
-  wire fma_v_li = reservation.v & reservation.decode.pipe_fma_v;
+  wire fma_v_li = reservation_v_fma_n & decode_pipe_fma_v_n;
   bsg_dff_chain
```
dpetrisko, line 262: Does bsg_dff_chain handle the zero case?

dpetrisko (changes requested):
Thanks! changes requested, but mostly in hardfloat

bp_be/src/v/bp_be_calculator/bp_be_pipe_fma.sv
mysoreanoop: Yes, it becomes a wire assignment essentially; also, should we hook up the flush connection to a "reset-capable" bsg_dff_chain?

mysoreanoop:
> Hi, thanks for the PR. You can embed these retiming registers at the input/output of the FMA unit and that should get the same QoR as having them in bp_be_pipe_fma, but will be a lot cleaner (as long as cross-boundary retiming / ungrouping is enabled for the FPU).

I tried using the `-retiming` and `-flatten_hierarchy rebuilt/full` options, but couldn't replicate the results from when having this manually placed retiming DFF. Could I be using the tool wrongly maybe?

taylor-bsg:
Dan is saying that we can remove the retiming registers from the BP code, and just feed in a different set of retiming parameters for ASIC versus Zynq. 

Dan, seems like we would scoop the *= operator into a bsg_mul_add module which takes a pipeline parameter and then use /hard to vector to FPGA versus ASIC implementation? That seems like a reasonable way to abstract it in a more general way?

mysoreanoop:
> Dan is saying that we can remove the retiming registers from the BP code, and just feed in a different set of retiming parameters for ASIC versus Zynq.

I meant without manually distributing some of the trailing retiming_chain registers, I'm unable to get Vivado to automatically retime the bp_be_pipe_fma module just as well. Either way, it's a difference of ~3 ns, and none of the paths are even close to violating timing at 50 MHz, so I'll go ahead and remove them.

> Dan, seems like we would scoop the *= operator into a bsg_mul_add module which takes a pipeline parameter and then use /hard to vector to FPGA versus ASIC implementation? That seems like a reasonable way to abstract it in a more general way?

I have the suggested changes integrated with HardFloat, and I can get this done too if you decide to have it that way.

dpetrisko:
>Dan, seems like we would scoop the *= operator into a bsg_mul_add module which takes a pipeline parameter and then use /hard to vector to FPGA versus ASIC implementation? That seems like a reasonable way to abstract it in a more general way?

I think that's reasonable, but the latency distribution parameter is still needed to handle the swap, to know how many cycles to pause after the muladd

dpetrisko:
>Could I be using the tool wrongly maybe?

It seems like you're doing the right thing. Ungrouping / flattening should mean that having the flops on either side of the boundary is identical to the other. However, another student has also observed seemingly random limitations in Vivado, making me think the internal representations are not what you'd expect.

>Either way, it's a difference of ~3 ns, and none of the paths are even close to violating timing at 50 MHz, so I'll go ahead and remove them.

This could just be noise if there are other tighter paths

bp_be/src/v/bp_be_calculator/bp_be_pipe_fma.sv
dpetrisko: cool! It's fine (and more efficient) without reset, because it's a datapath. So during a flush the control signals are zero'd and these are don't cared

bp_be/src/v/bp_be_calculator/bp_be_pipe_fma.sv
```diff
@@ -19,7 +42,7 @@ module bp_be_pipe_fma
  import bp_be_pkg::*;
  #(parameter bp_params_e bp_params_p = e_bp_default_cfg
    `declare_bp_proc_params(bp_params_p)
-
+   , parameter int pipeline_stages_p[0:1] = {0,0},
```
taylor-bsg, line 45: @dpetrisko default should be 3,1 for ASIC and FPGA right? it can just be a local parameter?

dpetrisko: Yeah I’m going to push through asic tools to verify that the rounding flop doesn’t impact qor and then we should just use the FMA and IMUL latency directly. I’m wary of parameterizations that break functionality for any but default value

dpetrisko (approved):
Great work!

PR #912, opened 2021-08-04
Fixes L2 hashing problem

### LINKED ISSUE #910: Reduce L2 tag width based on hash
## Area
bp_me_cce_to_cache, bsg_cache, memory system

## Current behavior
Currently, we send the entire address to the L2 cache. However, because we have already hashed to distribute to the L2s, this is unnecessary. 

## Desired behavior
We can trim the lower tag bits and re-append them at the DRAM interface (this must be kept in sync with the hashing algorithm).

## Proposed solution
Modify the bp_me_cce_to_cache adapter and instantiate a smaller bsg_cache. Also need to modify the test dram interface to interact with the cache ids, so that addresses are correctly reassembled.

## Analysis
Mild reduction in tag size (lg_num_cores), wire reduction in dma packet, no impact on wormhole size because data > header for that interface. Small buffer removal.

## Verification
Standard regression, paying particular attention to non-zero / large numbers of cores.

## Additional context
Need to update downstream tapeout repos to generate the correct memories

dpetrisko:
## Issue Fixed
#910 , context and justification there

## Area
bsg_cache_to_axi, test memory system

Changes to bsg_fpga, width of L2 cache tag RAMs

## Verification
Standard regression

### DIFF
--- a/bp_be/syn/flist.vcs
+++ b/bp_be/syn/flist.vcs
@@ -218,7 +218,7 @@ $BP_ME_DIR/src/v/lce/bp_lce.sv
 $BP_ME_DIR/src/v/lce/bp_lce_req.sv
 $BP_ME_DIR/src/v/lce/bp_lce_cmd.sv
 # Cache
-$BP_ME_DIR/src/v/cache/bp_me_cce_to_cache.sv
+$BP_ME_DIR/src/v/dev/bp_me_cce_to_cache.sv
 # CCE
 $BP_ME_DIR/src/v/cce/bp_cce.sv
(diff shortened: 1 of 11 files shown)


taylor-bsg (commented):
How about a regression test to check for this bug in the future?

dpetrisko:
Not sure how, it’s invisible to software

dpetrisko:
I suppose we could have a hardware monitor check for cache misses

taylor-bsg:
Yep, or a bind statement

PR #919, opened 2021-09-01
Refactor L2/dram address swizzle into bp_me_dram_hash module

muwyse:
## Summary
This PR refactors the code used to swizzle memory access addresses headed to L2/DRAM into the bp_me_dram_hash module. This swizzling is required to allow a core utilize all of the L2 cache in multicore designs and is applied to the L2 cache packets issued by bp_me_cce_to_cache.

In our simulation setup, the swizzling is reverted in bp_nonsynth_dram so the access to DRAM uses the original physical address issued toward memory, to help with debugging.

Clarity and maintainability.

## Analysis
Small additional logic in nonsynth_dram to revert the swizzle applied by bp_me_cce_to_cache. Should be no change in logic to bp_me_cce_to_cache.

## Verification
Regressions (RISCV_TESTLIST, multicore, ME) passed locally.


bp_me/src/v/dev/bp_me_dram_hash.sv
```diff
@@ -0,0 +1,61 @@
...
+ *   Encode takes an address of form {A, b...bb, c...cc, D} and outputs
+ *   address of {A, c...cc, b...bb, D}. The bit widths of b and c bits may differ and are
+ *   specified by the offset_widths_p parameter.
+ *   Decode reverses the swizzle operation.
+ *
+ */
+
+`include "bp_common_defines.svh"
+
+module bp_me_dram_hash
```
dpetrisko, line 19: I'd recommend splitting into bp_me_dram_hash_encode and _decode, since they are disjoint, but not required.

Can we put a comment for offset_widths_p for what each field means?

bp_me/test/common/bp_nonsynth_dram.sv
```diff
@@ -40,6 +40,32 @@ module bp_nonsynth_dram
...
+  assign dma_pkt_li = dma_pkt_i;
+  localparam block_offset_lp = `BSG_SAFE_CLOG2(cce_block_width_p/8);
+  localparam lg_lce_sets_lp = `BSG_SAFE_CLOG2(lce_sets_p);
+  localparam lg_num_cce_lp = `BSG_SAFE_CLOG2(num_cce_p);
+  genvar i;
+  for (i = 0; i < num_dma_p; i++) begin : address_hash
+    logic [daddr_width_p-1:0] addr_lo;
+    bp_me_dram_hash
+      #(.bp_params_p(bp_params_p)
+        ,.offset_widths_p('{(lg_lce_sets_lp-lg_num_cce_lp), lg_num_cce_lp, block_offset_lp})
```
dpetrisko, line 54: assign to a localparam for tool safety. Old versions of verilator didn't like this, not sure if it's been updated to fix

bp_me/src/v/dev/bp_me_dram_hash.sv
dpetrisko: Oh I see the comment in the header, that's fine

bp_me/src/v/dev/bp_me_dram_hash.sv
```diff
@@ -0,0 +1,61 @@
...
+   , localparam lg_num_cce_lp = `BSG_SAFE_CLOG2(num_cce_p)
+   , parameter int offset_widths_p[2:0] = '{ (lg_lce_sets_lp-lg_num_cce_lp), lg_num_cce_lp, block_offset_lp }
+   , localparam offset_width_lp = offset_widths_p[0] + offset_widths_p[1] + offset_widths_p[2]
+   , parameter addr_width_p = paddr_width_p
+   , parameter decode_p = 0
+   )
+  (input [addr_width_p-1:0]         addr_i
+   , output logic [addr_width_p-1:0] addr_o
+   );
+
```
dpetrisko, line 34: assign localparams here if they're not used for port sizing (which they're not)

dpetrisko (changes requested):
Minor changes requested, otherwise great!

bp_me/src/v/dev/bp_me_dram_hash.sv
muwyse: split modules.

bp_me/src/v/dev/bp_me_dram_hash.sv
muwyse: moved one down, but three still in param list at top to set reasonable defaults for offset_widths_p.

dpetrisko: Oh, didn't catch that's what they were for. My bad!

dpetrisko (approved):

PR #924, opened 2021-09-16
Breaking CI

dpetrisko:
This is a PR used to debug CI

### DIFF
--- a/bp_common/syn/Makefile.vcs
+++ b/bp_common/syn/Makefile.vcs
@@ -92,7 +92,7 @@ sim.v: SIM_ERROR  := $(REPORT_DIR)/$(TB).$(CFG).$(TAG).sim.$(SUITE).$(PROG).err
 $(SIM_COLLATERAL): | $(TOUCH_DIR) $(RESULTS_DIR) $(LOG_DIR) $(REPORT_DIR) $(SIM_DIR)
 $(SIM_DIR)/run_simv: | $(SIM_COLLATERAL)
 	cd $(@D); ./simv $(VCS_OPTIONS) $(VCS_PLUSARGS) 2>&1 | tee $(SIM_LOG)
-	-@grep "PASS" $(SIM_LOG) || echo "FAILED" > $(SIM_ERROR)
+	-@grep "FAIL" $(SIM_LOG) && echo "FAILED" > $(SIM_ERROR)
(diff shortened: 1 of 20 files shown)


bp_common/syn/Makefile.vcs
```diff
@@ -92,7 +92,7 @@ sim.v: SIM_ERROR  := $(REPORT_DIR)/$(TB).$(CFG).$(TAG).sim.$(SUITE).$(PROG).err
 $(SIM_COLLATERAL): | $(TOUCH_DIR) $(RESULTS_DIR) $(LOG_DIR) $(REPORT_DIR) $(SIM_DIR)
 $(SIM_DIR)/run_simv: | $(SIM_COLLATERAL)
 	cd $(@D); ./simv $(VCS_OPTIONS) $(VCS_PLUSARGS) 2>&1 | tee $(SIM_LOG)
-	-@grep "PASS" $(SIM_LOG) || echo "FAILED" > $(SIM_ERROR)
+	-@grep "FAIL" $(SIM_LOG) && echo "FAILED" > $(SIM_ERROR)
```
muwyse, line 95: This is good, but we might also want to add a check for the PASS not being present in the log. Assuming all of our tests will print PASS on success (or multiple PASS if multiple cores), I believe it would be possible for both PASS and FAIL to not be in the log, which should be considered a FAIL for the test.

bp_top/test/common/bp_nonsynth_watchdog.sv
```diff
@@ -26,45 +27,77 @@ module bp_nonsynth_watchdog
     , input                                   instret_i
     );
 
+  enum logic {e_run, e_halt} state_n, state_r;
+  wire is_run = (state_r == e_run);
+  wire is_halt = (state_r == e_halt);
+
```
muwyse, line 33: add finish_i signal and consider halt only if core is finished

bp_common/syn/Makefile.vcs
dpetrisko: Yeah makes sense. I may add a check for $finish to really be sure

dpetrisko: Okay, latest commit should be pretty solid. CHecks for PASS & !FAIL & finish

bp_top/test/common/bp_nonsynth_watchdog.sv
dpetrisko: Fixed!

PR #915, opened 2021-08-12
update ME interfaces to Stream or Burst

muwyse:
This PR significantly updates the memory end of BlackParrot. Major changes include:
- updating CCEs to use BedRock Stream protocol on the CCE-Mem networks
- updating CCEs to use BedRock Burst protocol on the LCE-CCE networks
- updating UCE to use BedRock Stream protocol on UCE-Mem interface
- updating cce_to_cache to use BedRock Stream interface
- refactoring BP tile and unicore to use new BedRock Stream crossbars
- ME documentation updates

This PR addresses the following open issues:
- #870 
- #260 
- #872 

CI is running, but local runs were passing

PR #931, opened 2021-09-22
Adding buffered crossbar

dpetrisko:
## Summary
This PR adds a new module, bp_me_stream_xbar_buffered, which wraps the xbar with twofers to break timing paths.

## Issue Fixed
This fixes the timing regression from #915, at the cost of additional buffers in the crossbar. 

## Area
The crossbars in bp_tile and bp_unicore_lite, notably the paths to/from the stream pumps.

While buffers are not functionally required, even for helpfulness, a medium-sized crossbar will require these fifos. Additionally, in the unicore these paths include I/O paths, which can be severely late-arriving.

## Analysis
Increased area and 2 additional cycles to L1 cache fill.

## Verification
Normal regression, as this is not a functional change.

### DIFF
--- a/bp_me/src/v/network/bp_me_xbar_stream_buffered.sv
+++ b/bp_me/src/v/network/bp_me_xbar_stream_buffered.sv
@@ -1,7 +1,7 @@
 /**
  *
  * Name:
- *   bp_me_xbar_stream.sv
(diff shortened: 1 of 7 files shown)


muwyse (approved):
LGTM!

PR #933, opened 2021-09-22
Making cosimulation fifo asynchronous for faster draining

dpetrisko:
This fixes a potential issue where cosimulation buffers fill up and overflow. The case that could cause this is if there are a large number of nonblocking OOO operations such as integer divides or cache misses, several completed entries could enter the cosimulation fifo. If the FIFO is committing at 1 IPC, then it can only "catch up" to the processor on processor stalls. If there aren't any catchup cycles, eventually the fifo may overrun.

This PR makes the cosimulation clock much faster than the core clock, enable it to drain as soon as dependencies are met. Additionally, this provides additional safety in the context of multicore cosimulation.

Last, it provides a hook to wait on coherence operations to complete. This is essential for ordering commits of stores in the context of a cosimulated SC multicore.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -26,6 +26,7 @@ stages:
     - me_dev
     - top_dev
     - sw_dev
+    - be_dev_fix_asynccosim
   before_script:
       - git submodule update --init --checkout --recursive external/
   artifacts:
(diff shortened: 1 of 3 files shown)


taylor-bsg:
How does it affect cosim performance?

dpetrisko:
Good question! I’ll run some tests

dpetrisko:
Seems to be negligible.  All tests I ran were +-1s, with the async cosim being faster in some cases

PR #936, opened 2021-09-29
Fix L2 != CCE block width hash function

dpetrisko:
time. A few details to check:

## Summary
This PR fixes a bug in the mapping of cache lines to L2 caches.

## Area
DRAM hash functions in bp_me_cce_to_cache and bp_me_nonsynth_dram

The DRAM hash function is intended to spread traffic to all L2 caches. This means alternating cache lines between L2 cache. However, we were using CCE block width for determining the offset of the hash. For a normal configuration of CCE block == L2 block, there is no problem. However for a configuration with a larger L2, this has aliasing issues.

## Analysis
Bugfix, no PPA or performance implementation for default configs.

## Verification
I tried a few non-default configs with small L1 and large L2 block sizes. Several beebs tests failed which pass with this fix. Manual inspection showed aliasing between L2 cache lines.

## Additional Context
The acache parameterization changes here expose the bug. Before, having an acache with 512b block size actually implied that CCE block width is always 512, same as L2. Now, we only use acache in the CCE block width computation if there is an acache enabled.

### DIFF
--- a/bp_me/src/v/dev/bp_me_cce_to_cache.sv
+++ b/bp_me/src/v/dev/bp_me_cce_to_cache.sv
@@ -312,7 +312,7 @@ module bp_me_cce_to_cache
 
   // Swizzle address bits for L2 cache command
   // Note: the upper paddr_width_p-daddr_width_p bits of the mem_cmd address are dropped
-  localparam block_offset_lp = `BSG_SAFE_CLOG2(cce_block_width_p/8);
+  localparam block_offset_lp = `BSG_SAFE_CLOG2(l2_block_width_p/8);
   localparam lg_lce_sets_lp = `BSG_SAFE_CLOG2(lce_sets_p);
   localparam lg_num_cce_lp = `BSG_SAFE_CLOG2(num_cce_p);
   localparam int hash_offset_widths_lp[2:0] = '{(lg_lce_sets_lp-lg_num_cce_lp), lg_num_cce_lp, block_offset_lp};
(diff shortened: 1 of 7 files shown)


bp_me/src/v/dev/bp_me_cce_to_cache.sv
```diff
@@ -312,8 +312,8 @@ module bp_me_cce_to_cache
 
   // Swizzle address bits for L2 cache command
   // Note: the upper paddr_width_p-daddr_width_p bits of the mem_cmd address are dropped
-  localparam block_offset_lp = `BSG_SAFE_CLOG2(cce_block_width_p/8);
-  localparam lg_lce_sets_lp = `BSG_SAFE_CLOG2(lce_sets_p);
+  localparam block_offset_lp = `BSG_SAFE_CLOG2(l2_block_width_p/8);
+  localparam lg_lce_sets_lp = `BSG_SAFE_CLOG2(l2_sets_p);
```
muwyse, line 316: rename the localparam to match the computation

bp_me/test/common/bp_nonsynth_dram.sv
```diff
@@ -43,8 +43,8 @@ module bp_nonsynth_dram
   `declare_bsg_cache_dma_pkt_s(daddr_width_p);
   bsg_cache_dma_pkt_s [num_dma_p-1:0] dma_pkt_li, dma_pkt;
   assign dma_pkt_li = dma_pkt_i;
-  localparam block_offset_lp = `BSG_SAFE_CLOG2(cce_block_width_p/8);
-  localparam lg_lce_sets_lp = `BSG_SAFE_CLOG2(lce_sets_p);
+  localparam block_offset_lp = `BSG_SAFE_CLOG2(l2_block_width_p/8);
+  localparam lg_lce_sets_lp = `BSG_SAFE_CLOG2(l2_sets_p);
```
muwyse, line 47: rename the localparam to match the computation

muwyse (changes requested):
Looks good, just need to fix some naming for clarity

dpetrisko:
Fixed names!

muwyse (approved):
LGTM!

PR #937, opened 2021-09-30
Adding wbuf handshake

muwyse:
DRAFT: Adds handshake on write buffer, helps with atomic accesses.

### DIFF
--- a/bp_be/src/v/bp_be_dcache/bp_be_dcache_wbuf.sv
+++ b/bp_be/src/v/bp_be_dcache/bp_be_dcache_wbuf.sv
@@ -47,7 +47,7 @@ module bp_be_dcache_wbuf
     (.clk_i(clk_i)
      ,.reset_i(reset_i)
 
-     ,.up_i(v_i)
+     ,.up_i(v_i & ready_and_o)
      ,.down_i(yumi_i)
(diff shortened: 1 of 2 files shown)


bp_be/src/v/bp_be_dcache/bp_be_dcache_wbuf.sv
```diff
@@ -174,6 +175,8 @@ module bp_be_dcache_wbuf
      ,.data_o(data_merged_o)
      );
 
+  assign ready_and_o = (num_els_r < 2'd2) || yumi_i;
```
muwyse, line 178: @dpetrisko I don't think this violates the v&r interface, but it appears to make the ready_and_o signal "late" instead of "early." Do you think this is okay regarding timing? I wasn't sure how critical the wbuf paths are.

dpetrisko: Yeah this is why I mentioned it’s probably not the final fix. This is part of a larger dcache refactor that deals with the half cycle paths more elegantly. 

But from a timing perspective, that yumi comes pretty early because it’s just based off incoming cache requests, which are early signals

dpetrisko:
superceded by #942

PR #942, opened 2021-10-04
[WIP] Updating D$ posedge->negedge handshakes

dpetrisko:
This is an alternative to #940, where the handshakes themselves are modified. The advantage of this approach is that the cache engines themselves can remain unmodified, which makes it easier to handle in the common case of having more cache engines than cache implementations. The major disadvantage is the (minor) use of latches.

This PR needs extensive verification before merging


bp_me/src/v/lce/bp_lce_req.sv
```diff
@@ -218,6 +220,8 @@ module bp_lce_req
       e_ready: begin
         ready_o = lce_req_ready_then_i & ((lce_mode_i == e_lce_mode_uncached) || sync_done_i);
 
+        // TODO: This could drop uncached stores, if we accept but lce is not ready
+        //   in the next cycle
```
muwyse, line 224: Seems dangerous...what's the fix?

bp_me/src/v/lce/bp_lce_req.sv
```diff
@@ -31,6 +31,7 @@ module bp_lce_req
     // issue non-exclusive read requests
     , parameter non_excl_reads_p = 0
 
+    , parameter req_invert_clk_p = 0
```
muwyse, line 34: Are any changes needed for the request FSM if the cache request is arriving on negedge? Is half a cycle enough time for the request to be generated and issued to the network?

bp_be/src/v/bp_be_dcache/bp_be_dcache.sv
```diff
@@ -193,12 +193,32 @@ module bp_be_dcache
...
+     ,.data_o({data_mem_pkt_yumi_o, tag_mem_pkt_yumi_o})
+     );
+
+  wire _flush_self = flush_i | sram_hazard_flush | miss_request_flush | engine_flush;
+  logic stat_mem_pkt_yumi_lo, cache_req_yumi_li, flush_self;
+  // Handshakes that are consumed at posedge are extended to the next negedge
+  bsg_dlatch
+   #(.width_p(3), .i_know_this_is_a_bad_idea_p(1))
+   negedge_extend
+    (.clk_i(~clk_i)
```
muwyse, line 218: I'm wondering if for clarity's sake it might be helpful to have a `wire negedge_clk = ~clk_i` signal (or similarly named) that is then used throughout the module for all negedge elements? Can't decide if my brain would find that or the current `~clk_i` more appealing. Or perhaps some updated comments in the header about what happens on the negedge would help?

muwyse (commented):
Looks great overall, a few questions to ponder.

bp_me/src/v/lce/bp_lce_req.sv
dpetrisko: Basically to emulate the behavior at https://github.com/black-parrot/black-parrot/blob/master/bp_me/src/v/cce/bp_uce.sv#L599. I'll make an issue for it

taylor-bsg: Can you add an assert?

PR #946, opened 2021-10-12
[WIP] Adding acquire-release semantics for atomics

### LINKED ISSUE #943: AQ RL semantics for L1/L2 atomics
## Area
D$ / Cache Engines

## Current behavior

RISC-V atomics define acquire and release semantics as:

```
To provide more efficient support for release consistency [5], each atomic instruction has two bits,
aq and rl, used to specify additional memory ordering constraints as viewed by other RISC-V harts.
The bits order accesses to one of the two address domains, memory or I/O, depending on which
address domain the atomic instruction is accessing. No ordering constraint is implied to accesses
to the other domain, and a FENCE instruction should be used to order across both domains.
If both bits are clear, no additional ordering constraints are imposed on the atomic memory operation.
If only the aq bit is set, the atomic memory operation is treated as an acquire access,
i.e., no following memory operations on this RISC-V hart can be observed to take place before the
acquire memory operation. If only the rl bit is set, the atomic memory operation is treated as a
47
48 Volume I: RISC-V Unprivileged ISA V20191214-draft
release access, i.e., the release memory operation cannot be observed to take place before any earlier
memory operations on this RISC-V hart. If both the aq and rl bits are set, the atomic memory
operation is sequentially consistent and cannot be observed to happen before any earlier memory
operations or
```

While BP is sequentially consistent, this is insufficient in the context of non-uniform memory access latencies. For instance:

If you have a  L1-uncached memory region, then outgoing writes could be much slower than the update to the lock; and then another core can observe that the lock is free, and then send a request to the L1-uncached memory region that beats the original set of writes

## Expected behavior
Atomic release should act as fence. Atomic acquire is more interesting and challenging. This requires stalling future memory operations until the atomic returns. In the case of non-blocking atomics, this either requires handling the acquire semantics in the cache engine or changing the cache engine interface to unlock.

## Verification
Tests verifying the consistency bug as well as the solution. Tests which intersperse L1 and L2 accesses to change latencies.

dpetrisko:
## Summary
This PR adds acquire-release semantics for atomics to the pipeline. This means when an atomic acquire instruction happens, memory instructions in the pipeline are stalled until it is completed (uncached stores return). When an atomic release instruction happens, the instruction acts as a fence.

## Issue Fixed
#943 

## Area
D$, cache engines, some minor pipeline changes.

This is a feature of RISC-V that we wish to support. We could also disable ordering between L2-atomic regions using PMAs, but that's no fun.

## Analysis
Performance hit on uncached atomic acquire operations which are guaranteed to be point to point ordered. We could theoretically only do the barrier for addresses in different regions. However, this information is only known after the TLB access, which is too late to stall. To roll back and stall would complicate the pipeline. Therefore, we do the low-performance, TINY thing.

There is additional complication but performance improvement by considering memory and I/O regions as separate.

## Verification
- [ ] Test interspersing cached atomic acquire / release and uncached operations


bp_be/src/v/bp_be_checker/bp_be_detector.sv
```diff
@@ -133,6 +133,17 @@ module bp_be_detector
      ,.rd_match_o(frd_match_lo)
      );
 
+  logic acquire_r;
```
dpetrisko, line 136: Actually need to acquire at dispatch to prevent back to back mem requests from succeeding

dpetrisko:
image

Reading closer into the spec, there is no ordering implied between the domains, so PTP ordering is sufficient. There remains a question about uncached atomic ordering, but that's a separate issue with a separate solution

PR #978, opened 2021-12-10
ME interface and data widths

muwyse:
Draft PR to review changes for LCE burst interface and cce data width > 64 bits

### DIFF
--- a/bp_common/src/include/bp_common_aviary_defines.svh
+++ b/bp_common/src/include/bp_common_aviary_defines.svh
@@ -112,6 +112,8 @@
     , localparam fe_queue_fifo_els_p = proc_param_lp.fe_queue_fifo_els                             \
     , localparam fe_cmd_fifo_els_p   = proc_param_lp.fe_cmd_fifo_els                               \
                                                                                                    \
+    , localparam bedrock_data_width_p   = proc_param_lp.bedrock_data_width                         \
+                                                                                                   \
     , localparam async_coh_clk_p        = proc_param_lp.async_coh_clk                              \
     , localparam coh_noc_max_credits_p  = proc_param_lp.coh_noc_max_credits                        \
     , localparam coh_noc_flit_width_p   = proc_param_lp.coh_noc_flit_width                         \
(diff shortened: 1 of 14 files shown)


bp_common/src/include/bp_common_aviary_defines.svh
```diff
@@ -96,7 +96,7 @@
     , localparam cce_pc_width_p             = proc_param_lp.cce_pc_width                           \
     , localparam num_cce_instr_ram_els_p    = 2**cce_pc_width_p                                    \
     , localparam cce_way_groups_p           =                                                      \
-        `BSG_MAX(dcache_sets_p, `BSG_MAX(icache_sets_p, num_cacc_p ? acache_sets_p : '0))          \
+        `BSG_MIN(dcache_sets_p, `BSG_MIN(icache_sets_p, num_cacc_p ? acache_sets_p : icache_sets_p)) \
```
dpetrisko, line 99: Can you explain this? My understanding was there was one way-group per set minimally

bp_common/src/include/bp_common_aviary_pkgdef.svh
```diff
@@ -169,6 +169,9 @@
     // Size of the cmd queue
     integer unsigned fe_cmd_fifo_els;
 
+    // width of LCE-CCE and CCE-Mem BedRock network data channels
+    integer unsigned bedrock_data_width;
```
dpetrisko, line 173: what's the relation between the various fill_widths and this?

bp_common/src/include/bp_common_aviary_defines.svh
muwyse: A way group serializes related coherence transactions. In this context, related means addresses that could map to the same set in *any* cache. If all caches have the same number of sets, then the number of way groups == number of sets. If caches have different number of sets, then we need to restrict the number of way groups to the minimum number of cache sets of any cache type.

This is because the CCE controls replacements/evictions. Consider a system with two cache types. Type A has S sets and type B has 2*S sets. Some block that maps to some set in cache A could map to one of two different sets in cache B. Consequently, if the number of way groups is the MAX of sets across cache types, two different requests could map to different way groups (equivalently, sets in B) but target the same set in A. It might be true that the way the directory currently handles requests, this would be okay, but the conservative way to be safe is to use MIN of sets to compute number of way groups.

bp_common/src/include/bp_common_aviary_pkgdef.svh
muwyse: As of now, in multicore, the $ fill widths must be the same and the fill width must match the bedrock data width. Additional logic would allow this to be relaxed, but likely requires handling some odd edge cases. I suppose it could be open to debate as to whether this param should be in aviary or just a param of the LCE and CCE that gets set to fill_width in multicore. I think having a separate parameter is cleaner because it makes it clear what the data width will be for all the LCE, CCE, xbar, etc. and then we impose a constraint on it due to what we have chosen to implement.

muwyse:
Closing draft; breaking into two PRs

PR #979, opened 2021-12-13
ME device tracer

muwyse:
Add a ME device tracer module (CLINT, CFG, loopback, cce_to_mem). Module traces only the BedRock Stream cmd/resp interface.

Instantiate for CLINT in multicore and unicore, controlled by `dev_trace_p` parameter in testbench.

### DIFF
--- a/bp_me/src/v/dev/bp_me_clint_slice.sv
+++ b/bp_me/src/v/dev/bp_me_clint_slice.sv
@@ -14,6 +14,8 @@ module bp_me_clint_slice
   (input                                                clk_i
    , input                                              reset_i
 
+   , input [core_id_width_p-1:0]                        id_i
+
    , input [xce_mem_header_width_lp-1:0]                mem_cmd_header_i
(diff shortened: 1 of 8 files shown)


dpetrisko:
Could we bind into bedrock_register instead? That would be a bit more scalable. Might require a hack to disambiguate the names though

muwyse:
> Might require a hack to disambiguate the names though

That was my concern. One way to disambiguate is to pass an id to bedrock register from the module above it. Currently, I think all instances would have a unique ID.

I see benefit binding to either the ME devices or bedrock register. Binding to devices allows this one tracer to attach to any of the current ME devices. The tradeoff is it isn't specialized for any of them.

muwyse:
At the moment, I lean toward the generic device tracer proposed here

dpetrisko:
Yeah that makes sense. Maybe let’s just do this and put a comment in the TB that we could go with the ID approach if it ends up being cluttered

PR #981, opened 2021-12-14
Add bedrock_data_width parameter to ME, update CCE

muwyse:
This PR adds a new param `bedrock_data_width` to the CCEs that controls the width of the LCE-CCE and CCE-MEM interface data widths. The data width of the LCE-CCE and CCE-MEM interface is required to be the same to eliminate unnecessary gearboxing inside the CCE. The physical networks carrying the messages can use an independently selected channel width, but are responsible for required gearboxing.


muwyse:
Requires merge of be_dev to dev to capture some aviary rom changes

PR #982, opened 2021-12-14
Make xbar stream a true crossbar

dpetrisko:
## Summary
This PR makes bp_me_xbar_stream into a true crossbar, rather than a glorified round robin arbiter. 

## Area
bp_me_xbar_stream

There is potential for deadlock if there are circular dependencies in the local crossbar. This can happen when an external device is writing to BP, while BP is polling the host. There are many ways to break this cycle, but preventing unnecessary blocking in this crossbar is low cost and provides nice guarantees for BP integrated in external systems

Need to modify xbar burst to also have this behavior.

## Analysis
Should have a minimal PPA impact. The data plane of the crossbar is unchanged, it's simply the 

## Verification
- [ ] Run this change through zynq-parrot to verify that https://github.com/black-parrot-hdk/zynq-parrot/issues/16 is fixed

## Additional Context
https://github.com/bespoke-silicon-group/basejump_stl/pull/536

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -26,6 +26,7 @@ stages:
     - me_dev
     - top_dev
     - sw_dev
+    - top_dev_fix_lockingxbar
   before_script:
       - git submodule update --init --checkout --recursive external/
   artifacts:
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -26,7 +26,6 @@ stages:
     - me_dev
     - top_dev
     - sw_dev
-    - top_dev_fix_lockingxbar
   before_script:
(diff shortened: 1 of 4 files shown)


muwyse (approved):
BP changes LGTM!

Is the bump of basejump a bump on the repos master branch? I suppose it is okay to allow an off-master bump for now, but for stability I don't think BP master should ever refer to non-master basejump.

dpetrisko:
Yeah the bump will get reverted when https://github.com/bespoke-silicon-group/basejump_stl/pull/536 is merged. I'd go farther and say dev probably should only have master basejump...

muwyse:
Sounds good. Merge when tests pass.

PR #985, opened 2021-12-23
Update core profiler

farzamgl:
This PR updates the core stall profiler and adds more distinction between various pipeline dependencies.

### DIFF
--- a/bp_top/test/common/bp_nonsynth_core_profiler.sv
+++ b/bp_top/test/common/bp_nonsynth_core_profiler.sv
@@ -3,9 +3,7 @@
   {
     logic fe_queue_full;
     logic fe_wait;
-    logic itlb_miss;
     logic icache_miss;
-    logic icache_rollback;
     logic branch_override;
     logic ret_override;
     logic fe_cmd;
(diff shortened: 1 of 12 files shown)


bp_top/test/common/bp_nonsynth_core_profiler.sv
```diff
@@ -172,96 +176,104 @@ module bp_nonsynth_core_profiler
...
+      stall_stage_n[3].mul_dep           |= mul_dep_i;
+      stall_stage_n[3].fma_dep           |= fma_dep_i;
+      stall_stage_n[3].sb_iraw_dep       |= sb_iraw_dep_i;
+      stall_stage_n[3].sb_fraw_dep       |= sb_fraw_dep_i;
+      stall_stage_n[3].sb_iwaw_dep       |= sb_iwaw_dep_i;
+      stall_stage_n[3].sb_fwaw_dep       |= sb_fwaw_dep_i;
+      stall_stage_n[3].struct_haz        |= struct_haz_i;
+      stall_stage_n[3].long_i_busy       |= long_busy_i & long_i_busy_i;
+      stall_stage_n[3].long_f_busy       |= long_busy_i & long_f_busy_i;
+      stall_stage_n[3].long_if_busy      |= long_busy_i & ~(long_i_busy_i | long_f_busy_i);
```
dpetrisko, line 220: If I'm reading this right, this is really a write port stall, not related to the FU itself

https://github.com/black-parrot/black-parrot/blob/master/bp_be/src/v/bp_be_calculator/bp_be_pipe_long.sv#L249

farzamgl: Those 3 stalls are for when we want to issue a long instruction but the long pipe is already processing another. The pipe ready signal seems to cover both the FU busy signals and the writeback delay. And the writeback stall is because of conflict on the regfile's write port?

farzamgl: I see what you mean now. The 3 currently mean:
long int waiting for long int
long float waiting for long float
long int waiting for float

But I think it probably makes sense to divide it only by the type of instruction being processed in the pipe, and add the writeback stall.

dpetrisko: Yeah stalls are only in terms of waiting for, not what’s waiting.

That said, the way the code is written may cause an extra cycle as we wait for the FU to clear before dispatching, which then takes one more cycle to get into the FU. I’d have to look at waves to verify

farzamgl: Yeah I think it should be change to:
`assign ready_o = fdiv_ready_lo & idiv_ready_and_lo & ~(v_li | (rd_w_v_r & ~(iwb_yumi_i | fwb_yumi_i)));`

bp_top/test/common/bp_nonsynth_core_profiler.sv
```diff
@@ -172,96 +176,104 @@ module bp_nonsynth_core_profiler
...
-      stall_stage_n[2].icache_rollback   |= icache_rollback;
-      stall_stage_n[2].icache_fence      |= icache_fence;
-      stall_stage_n[2].fe_cmd            |= fe_cmd;
-      stall_stage_n[2].dcache_rollback   |= dcache_rollback;
-      stall_stage_n[2].branch_override   |= branch_override;
-      stall_stage_n[2].ret_override      |= ret_override;
-      stall_stage_n[2].exception         |= exception;
-      stall_stage_n[2].eret              |= eret;
-      stall_stage_n[2]._interrupt        |= _interrupt;
+      stall_stage_n[3]                    = fe_queue_empty_i ? stall_stage_r[2] : '0;
```
dpetrisko, line 204: Were you observing false stalls emanating from this? The idea is that stalls from the backend take precedence over stalls from the front end, so if the fe_queue is full there should either not be a stall or a backend stall

farzamgl: I didn't explicitly check for it, but by doing this if we're not counting a backend stall we'll see it included in the unknown category instead of being masked by frontend stalls.

dpetrisko: makes sense

bp_top/test/common/bp_nonsynth_core_profiler.sv
```diff
@@ -172,96 +176,104 @@ module bp_nonsynth_core_profiler
 
   always_comb
     begin
-      // By default, move down the pipe
-      for (integer i = 0; i < num_stages_p; i++)
-        stall_stage_n[i] = (i == '0) ? '0 : stall_stage_r[i-1];
+      // IF0
+      stall_stage_n[0]                    = '0;
+      stall_stage_n[0].fe_wait           |= (fe_state_n_i == fe_e_wait);
```
dpetrisko, line 181: we should get these signals directly from fe, rather than reconstructing them here

farzamgl: Some of these signals aren't explicitly declared in those modules. Do we want to add debug signals to the actual hardware instead of generating them here?

PR #986, opened 2021-12-24
Draft: Support 32-bit instructions aligned to 16 bits (IALIGN=16)

WasabiFan:
## Summary

This PR eliminates "instruction misaligned" exceptions, and allows execution of 32-bit instructions aligned to any 16-bit boundary. This is a partial step toward "C" support.

I have also introduced a new profiler (#988), which outputs a trace of instructions through the frontend. It is similar in spirit to the core profiler, except it a) concretely identifies the instructions in each frontend stage and b) ignores the backend behavior, showing stalls in the frontend even if the backend is _also_ stalled. It identifies the source of each fetch and external effects influencing behavior.

This PR is marked as "Draft" because I have not yet finalized and validated handling of edge cases such as exceptions. I am opening the PR for review of the changes and will be validating the minutiae soon. I anticipate nontrivial structural feedback so wanted to give this some time to soak.

Changes here are nontrivial. Let me know if a live call or other more interactive review would be helpful.

### Design

This change introduces a "realigner" component which postprocesses fetches in the frontend. It observes the fetch sequence and I$ responses. When a fetch to an aligned address is returned, it forwards the result combinationally. When a fetch to a misaligned address is returned, it fuses its lower half with the upper half of the preceding fetch, outputs the combined result, and stashes the upper half for the following fetch. Validity is tracked appropriately.

Management of pipeline flow is dependent on a pattern I'm calling a "linear" fetch: a fetch to some pc _x_ immediately after fetching pc _x-4_. Aligned instructions support fully random access, while a misaligned instruction fetch is only valid if you've made at least one linear fetch to retrieve the lower _and_ upper halves of the instruction. The description below uses this term.

BTB and BHT entries are largely unchanged. Predictions are indexed by the instruction's aligned address, i.e. the address of the first fetch to contain that instruction. When this implementation is extended to support "C" instructions, having the branch prediction metadata "early" will be necessary for eagerly predicting compressed branches. For now, a misaligned instruction returns branch prediction metadata on its first half and is actually issue-able once we've retrieved its second.

To facilitate this fetch pattern, the following other changes are made:
- The pipeline tracks two distinct addresses: the PC of the instruction currently being fetched, and the address that is being requested from the I$. On nonlinear fetch of a misaligned address, we "hold back" the PC by one cycle, so that the fetch for the second half corresponds to the address of the first half. This means that the same PC has two consecutive entries in the pipeline when we jump/redirect to a misaligned instruction. 
- Overrides and BTB predictions are ignored for one cycle after making a "nonlinear" fetch to a misaligned instruction.
- Branch prediction data is "delayed" by one cycle when the current fetch is misaligned. For example, the "effective" prediction for processing BTB jumps is actually the output that the BTB produced for the previous fetch, now in IF2. This is necessary to ensure that jumps are only taken once the entirety of their preceding instruction has been fetched.

## Issue Fixed

No issue

## Area

Frontend and PC gen. Removal of "instruction fetch misaligned" exception.

We want to support the "C" extension. Part of the "C" requirement is to relax IALIGN to 16. This PR does so. Further, it is designed to be an effective bridge to full "C" support and lays the groundwork for our intended design.

Dependencies:
- Dromajo needs to support+allow misaligned instructions. It currently ties "C" support and IALIGN=16 together in the same flag. Enabling "C" support might be the easiest way forward for now, but separating the flags is also an option.
- New programs in bp-tests (https://github.com/black-parrot-sdk/bp-tests/pull/19)
- #988 

Not currently planned for this PR:
- Performance analysis and overall testing on "real" programs. This will require some work with the linker and compiler, or some clever script hacks.

## Analysis

The following are required for processor soundness:
- The modified frontend is able to correctly issue at least one instruction after a redirect
- Exceptions are issued at the correct time with appropriate attached pc

The first one has been verified with reasonable certainty. The latter will be tested before merging.

Additionally, we expect the performance impact to be as follows:
- Aligned instructions see no change to performance whatsoever
- Misaligned instructions incur a one-cycle penalty on each jump, override or mispredict to a misaligned instruction
  - Misaligned instructions are branch-predicted exactly the same as if they were aligned

The following are known potential areas for performance improvement which have not been implemented:
- This implementation chooses to make only aligned 32-bit requests to the I$, and reconstruct consecutive halfwords later in the pipeline. This incurs a one-cycle penalty when jumping to, or overriding a mispredict to, a misaligned instruction. The reconstruction logic will always be necessary, as there are inevitably cases where instructions cross cache lines. However, for the case that an instruction fetch is entirely contained within a cache line, there is performance to be gained by teaching the I$ how to return a misaligned 32-bit word in one cycle. This would improve cycle count at the expense of PPA.
- I have chosen to delay overrides in the same way that BTB predictions are delayed, so that they only happen once the full instruction has been fetched. This is because overrides require knowing whether an instruction is a branch, jump or return, and being able to compute relative jump targets. It _may_ be possible to identify some of this information with only the first half of an instruction and safe a cycle; I'm not sure.

## Verification

I have created two new programs in bp-tests which test jumps between aligned and misaligned code. I also introduced the new tracer in the testbench as described above.

Running these programs, I have verified that:
1. The number of cycles taken to run aligned programs is unchanged
2. The number of cycles to run my unaligned test programs increases by the expected amount (branch-to-unaligned penalty times number of anticipated instances)
3. The flow of instructions through the frontend, and their prediction dynamics, is equivalent between aligned and partially-misaligned versions of my test programs

For (2), the programs output a cycle count for the important section of the program. The overall execution cycle count is more widely variable due to slightly different cache behavior.

For (3), I have used the new tracer to textually diff traces of the unmodified BP core to my PR branch as well as diff the trace of executing an aligned program and the same program with partial unalignment. I have also manually stepped through executions in DVE, inspecting every branch misprediction in chosen sections.

This verification has, so far, been largely limited to the two bp-tests programs I have introduced plus the existing `loop_test`. I haven't done any in-depth verification of programs that aren't mine, though.

<details>
<summary>
Click for sample screenshot of output and diff
</summary>

image

</details>

Using these tracer diffs, I have identified one way in which my changes modify the behavior of BP for aligned instructions. Specifically, under I$ miss stall, I no longer restrict the branch predictors from issuing predictions. This is _entirely_ inert, as no writes will occur under stall and stalls are always followed by redirections. However, it _does_ mean that when viewed in a debugger or tracer the predictions are not always pc+4. This could be changed if desired.

I will be making a follow-up comment with a test plan to be followed for final pre-merge verification.

## Additional Context

Design docs:
- Frontend diagrams, including original BP frontend and modified variant: https://docs.google.com/presentation/d/17emA7YldfHZRiXkK10YWAQ5f8Ph9ewhbfcTWuy09Sfw/edit?usp=sharing
- Walkthrough of jump behavior: https://docs.google.com/presentation/d/1EYkD9GumgJtdrjmn6OrcXyb7EVIBGt5Vm9czJjgG_Sw/edit?usp=sharing
- Design for branch prediction of C instructions, which impacts choices in this PR: https://docs.google.com/document/d/1REslHNPOi9HuCZsFEUGp_jzLJO3ev0Tq0YJn6hT_XGw/edit?usp=sharing
- Notes on review of Rocket's relevant details: https://docs.google.com/document/d/1YEsNAi_voeqNBfp1ZKJj992SX2eQPDBjXe9g2wkmLh4/edit?usp=sharing

## Pending and open questions

- Exception handling needs changes and testing
- TODOs (will be making GH comments)
- `fence.i` handling. Does this need to be treated specially?


bp_be/src/include/bp_be_ctl_pkgdef.svh
```diff
@@ -236,7 +236,6 @@
     logic ebreak;
     logic illegal_instr;
     logic instr_access_fault;
-    logic instr_misaligned;
```
dpetrisko, line 239: Keep the instruction misaligned exception — we may want a front end variant which doesn’t support C

bp_fe/src/v/bp_fe_pc_gen.sv
```diff
@@ -29,16 +29,19 @@ module bp_fe_pc_gen
...
    , output logic                                    ovr_o
 
    , input [instr_width_gp-1:0]                      fetch_i
-   , input                                           fetch_instr_v_i
-   , input                                           fetch_exception_v_i
+   , input                                           fetch_v_i
    , output logic [branch_metadata_fwd_width_p-1:0]  fetch_br_metadata_fwd_o
    , output logic [vaddr_width_p-1:0]                fetch_pc_o
+   , output logic [instr_width_gp-1:0]               fetch_instr_o
+   , output                                          fetch_instr_v_o
```
dpetrisko, line 42: Output logic

bp_fe/src/v/bp_fe_top.sv
```diff
@@ -371,7 +382,7 @@ module bp_fe_top
       begin
         fe_queue_cast_o = '0;
         fe_queue_cast_o.msg_type                     = e_fe_exception;
-        fe_queue_cast_o.msg.exception.vaddr          = fetch_pc_lo;
```
dpetrisko, line 374: Second half

bp_fe/src/include/bp_fe_defines.svh
```diff
@@ -45,14 +45,17 @@
       logic [ghist_width_mp-1:0]      ghist;                                                      \
     }  bp_fe_branch_metadata_fwd_s;
 
-  `define declare_bp_fe_pc_gen_stage_s(vaddr_width_mp, ghist_width_mp, bht_row_width_mp) \
+  // TODO: consider splitting this into multiple structs. some are only used for IF1 pred mux. others are used in metadata fwd above.
+  `define declare_bp_fe_pred_s(vaddr_width_mp, ghist_width_mp, bht_row_width_mp) \
```
WasabiFan, line 49: This struct warrants some thought.

Since the new logic requires muxes which can select some of this information from multiple pipeline stages, I've included it here. For the most part, this works great.

However, not all stages need all fields. If we wanted to minimize surprises, we could split it into the fields needed in three different places (some may be single variables), at the expense of less readable mux logic. Happy to take thoughts here.

bp_fe/src/include/bp_fe_defines.svh
```diff
@@ -63,6 +66,13 @@
   `define bp_fe_pred_width(vaddr_width_mp, ghist_width_mp, bht_row_width_mp) \
     (5 + bht_row_width_mp + ghist_width_mp)
 
+  `define bp_addr_is_aligned(addr_mp, num_bytes_mp) \
+    (!(|{ addr_mp[$clog2(num_bytes_mp)-1:0] }))
+
+  `define bp_align_addr(addr_mp, vaddr_width_mp, num_bytes_mp) \
+    ({addr_mp[vaddr_width_mp-1:$clog2(num_bytes_mp)], 2'b0})
+    // ({addr_mp[vaddr_width_mp-1:$clog2(num_bytes_mp)], {($clog2(num_bytes_mp))'b0}})
```
WasabiFan, line 74: Can anyone help on the syntax here? I _want_ the `$clog2` used as the number of zeroes to add on the tail. But I haven't been able to coax VCS into accepting it, either in the `()'b0` bit-width-notation form or using the replication operator.

bp_fe/src/v/bp_fe_pc_gen.sv
```diff
@@ -62,29 +65,57 @@ module bp_fe_pc_gen
...
+          next_logical_fetch_address = br_tgt_lo;
+      else if (btb_taken)
+          next_logical_fetch_address = effective_pred_if1.btb_tgt;
+      else
+        begin
+          next_fetch_linear = 1'b1;
+          next_logical_fetch_address = last_fetch_plus4;
+        end
+    end
+
```
WasabiFan, line 105: Note: the implementation here is slightly different than my diagrams, in that I do the "effective PC" computation in IF0 rather than IF1. This is because the "fetch address" we output to the I$ wants to be aligned, while the "effective PC" computation requires knowing whether we jumped to the start or the middle of the word.

bp_fe/src/include/bp_fe_defines.svh
dpetrisko: Not sure what the intention here is, since you have 2’b0 this will always align to a fixed number of bytes anyway

bp_fe/src/v/bp_fe_top.sv
```diff
@@ -108,16 +108,19 @@ module bp_fe_top
...
      ,.fetch_i(fetch_li)
-     ,.fetch_instr_v_i(fetch_instr_v_li)
-     ,.fetch_exception_v_i(fetch_exception_v_li)
+     ,.fetch_v_i(fetch_v_li)
      ,.fetch_br_metadata_fwd_o(fetch_br_metadata_fwd_lo)
      ,.fetch_pc_o(fetch_pc_lo)
+     ,.fetch_instr_o(fetch_instr_lo)
+     ,.fetch_instr_v_o(fetch_instr_v_lo)
+     ,.fetch_instr_progress_o(fetch_instr_fill_v_lo)
+     ,.fetch_instr_ready_i(fetch_instr_ready_li)
```
WasabiFan, line 123: Just want to call attention to this.

I added a "ready" signal to the pc gen module. It doesn't actually stall the PC gen if the frontend isn't ready, this only controls validity.

I added the handshake because, previously, the I$ fetch validity was masked to be 0 when passed into PC gen if the frontend queue wasn't ready or we were otherwise going to stall. I decoupled the I$ validity from the stall behavior because it's now possible for a one-cycle queue backup to not affect the fetch validity if we are working on a misaligned instruction. However, we want to avoid some minor undesirable behavior surrounding RAS writes under stall _somehow_, so I chose to introduce an explicit ready signal.

If there's another convention preferred, let me know.

bp_fe/src/v/bp_fe_top.sv
```diff
@@ -371,7 +382,7 @@ module bp_fe_top
       begin
         fe_queue_cast_o = '0;
         fe_queue_cast_o.msg_type                     = e_fe_exception;
-        fe_queue_cast_o.msg.exception.vaddr          = fetch_pc_lo;
+        fe_queue_cast_o.msg.exception.vaddr          = fetch_pc_lo; // TODO: on exception triggered by second half, should this point to the start or the end?
```
WasabiFan, line 385: Still untested and TODO. Talked with Dan previously about the correct behavior here but haven't handled it.

bp_top/test/common/bp_nonsynth_core_profiler.sv
```diff
@@ -29,14 +30,15 @@
...
+    fe_queue_stall       = 5'd24
+    ,fe_wait_stall       = 5'd23
+    ,itlb_miss           = 5'd22
+    ,icache_miss         = 5'd21
+    ,icache_rollback     = 5'd20
+    ,icache_fence        = 5'd19
+    ,branch_override     = 5'd18
+    ,ret_override        = 5'd17
+    ,misaligned_fetch    = 5'd16
     ,fe_cmd              = 5'd15
```
WasabiFan, line 42: I added the "misaligned fetch" event here, but it turns out to not be very useful. The backend usually trumps any interesting output. And even if it weren't the backend, this tracer is not well-suited to this purpose, since "misaligned fetch" penalties are pretty much always followed by a successful fetch of the same PC, and outputs of this tracer are per-PC. So other events for the same instruction, such as the branch prediction which triggered the fetch, pretty much always show up in the output instead.

bp_fe/src/include/bp_fe_defines.svh
WasabiFan: I want to replace `2'b0` with `($clog2(num_bytes_mp))'b0` -- i.e., I want the number of zeroes to be `$clog2(num_bytes_mp)`. But either I don't understand the syntax or the SV language doesn't allow you to use tasks in constants.

dpetrisko: Gotcha. You can do zero-width replication which is the answer here. Take a look at the cache offsets to see examples of this. I can send you exact code snippets late, but I’m about to get on a plane

WasabiFan:
# Test plan

The below steps will be followed for verification after changes are made. I will be editing this comment to add more steps.

## "misaligned_instructions_basic_jumps" program in bp-tests

### Cases evaluated

- Aligned -> unaligned unconditional jump, mispredicted as not taken (overridden)
- Unaligned -> unaligned unconditional jump, mispredicted as not taken (overidden)
- Unaligned -> aligned unconditional jump, mispredicted as not taken (overidden)
- Unaligned -> aligned call, return back to unaligned code (call mispredicted as not taken)

### Procedure

For each of the cases below:
1. Compile bp-tests. In `sdk`:

    ```
    rm install/touchfiles/bp-tests.*; make bp-tests
    ```

1. Run the `misaligned_instructions_basic_jumps` program in simulation. From `black-parrot/bp_top/syn`:

   ```
   make clean.v build_dump.v sim_dump.v SUITE=bp-tests PROG=misaligned_instructions_basic_jumps
   ```

1. Verify that `Hello Misaligned World!` is printed and that the final output was a PASS.

1. Note the hexadecimal integer printed to stdout. This is the number of cycles taken to execute the
   relevant part of the program.

Run the above steps in full for each of the following cases:
- The latest version of the RTL and latest corresponding `bp-tests`
- The latest version of the RTL and latest corresponding `bp-tests`, with the `ENABLE_MISALIGNMENT`
  macro undefined
- The latest common ancestor RTL commit between the current branch and the target branch, with
  latest corresponding `bp-tests` and the `ENABLE_MISALIGNMENT` macro undefined

### Expected results

- `Hello Misaligned World!` printed and PASS result for all
- The cycle count for both non-misaligned executions is the same
- The cycle count for the misaligned execution is almost exactly 19 larger, one for each jump
  to misaligned code

## "misaligned_instructions_advanced_jumps" program in bp-tests

### Cases evaluated

- Trained predictor able to output "taken" predictions from the BTB
- Conditional branches with BHT taken state changing
- Chained function calls and returns

# Procedure

Same as "basic" version, but using the new program name.

### Expected results

- Two lines of "12" followed by two lines of "123" are printed, as follows::

  ```
  12
  12
  123
  123
  ```
- The cycle count for both non-misaligned executions is the same
- The cycle count for the misaligned execution is almost exactly 22 larger, one for each jump
  to misaligned code

bp_fe/src/include/bp_fe_defines.svh
dpetrisko: it's fine, synthesizers will optimize

bp_common/src/include/bp_common_pkg.sv
```diff
@@ -25,6 +25,7 @@ package bp_common_pkg;
   localparam half_width_gp        = 16;
```
dpetrisko, line 25: Can you link your Rocketchip analysis in the PR as well?

bp_fe/src/v/bp_fe_pc_gen.sv
```diff
@@ -287,5 +350,26 @@ module bp_fe_pc_gen
 
   assign init_done_o = bht_init_done_lo & btb_init_done_lo;
 
+  // Pred data for "IF3", used only for misaligned instructions to look up their prediction in IF2
```
dpetrisko, line 353: I think we want to associate the branch metadata with misaligned instructions second halves to avoid explicitly storing this. i.e.
```
[ a0 | x1 ] <- fetch X going out, metadata if for x1
[ b0 | a1 ] <- fetch A going out, metadata is for a1, jump to C
[ c1 | c0 ] <- fetch C going out, metadata is for c1
```

bp_fe/src/v/bp_fe_top.sv
```diff
@@ -347,22 +353,27 @@ module bp_fe_top
...
   assign icache_poison_tv = fe_exception_v | cmd_nonattaboy_v;
 
   assign fe_cmd_yumi_o = pc_gen_init_done_lo & (cmd_nonattaboy_v | attaboy_yumi_lo);
   assign next_pc_yumi_li = (state_n == e_run);
 
-  assign fetch_instr_v_li     = fe_queue_v_o & fe_instr_v;
   assign fetch_exception_v_li = fe_queue_v_o & fe_exception_v;
-  assign fetch_fail_v_li      = v_if2_r & ~fe_queue_v_o;
-  assign fetch_li             = icache_data_lo;
+  assign fetch_fail_v_li      = v_if2_r & ~fe_progress;
```
dpetrisko, line 373: The fail here is overly broad. Really there are only two conditions we can't fetch and need to replay -- I$ fail and queue fail. We should explicitly track those, and get rid of the fe_progress/fetch_instr_fill mechanism.

bp_fe/src/v/bp_fe_realigner.sv
```diff
@@ -0,0 +1,54 @@
...
+   , output                      fetch_instr_progress_o // is the current input aiding resolution of a coherent instruction fetch?
+   );
+
+  wire [instr_half_width_gp-1:0] icache_data_lower_half_li = fetch_data_i[instr_half_width_gp-1:0];
+  wire [instr_half_width_gp-1:0] icache_data_upper_half_li = fetch_data_i[instr_width_gp-1     :instr_half_width_gp];
+
+  logic [instr_half_width_gp-1:0] upper_half_buffer_n, upper_half_buffer_r;
+  logic upper_half_buffer_v_n, upper_half_buffer_v_r;
+  // For ease in catching errors only, zero out register when input is invalid
+  assign upper_half_buffer_n   = fetch_data_v_i ? icache_data_upper_half_li : '0;
```
dpetrisko, line 36: don't add hardware for error catching, use nonsynth methods

bp_fe/src/v/bp_fe_pc_gen.sv
dpetrisko: Another way to say it always the aligned 32-bit address

bp_fe/src/v/bp_fe_pc_gen.sv
dpetrisko: Can't do it earlier than IF1 because of compressed. I think rather than worrying about the alignment to I$ and branch predictors, having them just drop the appropriate bits is simpler

bp_fe/src/v/bp_fe_pc_gen.sv
```diff
@@ -62,29 +65,57 @@ module bp_fe_pc_gen
...
-    else if (ovr_ret)
-        next_pc_o = ras_tgt_lo;
-    else if (ovr_taken)
-        next_pc_o = br_tgt_lo;
-    else if (btb_taken)
-        next_pc_o = btb_br_tgt_lo;
+    // "hold back" the PC for one cycle while executing misaligned instructions, since fetching a halfword resolves the
+    // pc which started in the previous one.
+    if (last_fetch_misaligned && next_fetch_linear)
+      next_pc = fetch_addr_if1_r;
```
dpetrisko, line 112: I don't like having both PC and fetch_addr. I think realigner is responsible for generating the correct PC for the current fe_queue submission, and fetch_address is responsible for generating the next fetch_address i.e. PC

dpetrisko:
What happens during nonspec mode, where I$ misses need to be confirmed by the BE? Does the realigner pass through the exception? Does the second half of a misaligned nonspec access act as speculative or nonspeculative?

bp_fe/src/v/bp_fe_pc_gen.sv
```diff
@@ -29,16 +29,19 @@ module bp_fe_pc_gen
    , input                                           redirect_br_ntaken_i
    , input                                           redirect_br_nonbr_i
 
-   , output logic [vaddr_width_p-1:0]                next_pc_o
+   , output logic [vaddr_width_p-1:0]                next_fetch_o
    , input                                           next_pc_yumi_i
 
    , output logic                                    ovr_o
 
    , input [instr_width_gp-1:0]                      fetch_i
```
dpetrisko, line 37: At a high level, we want to try to modify this interface as little as possible. Really think about what information is necessary for the pc_gen to sync up with the rest of the front end. It looks like the contract for the FE is 

```
// IF0
// it should predict the next PC
next_pc_o
next_pc_yumi_i

// IF1
// it should say whether to kill the instruction in IF1
ovr_o
// it needs to know whether to use the NLP or do a linear fetch
aligned_i // New signal, this info is available from I$ in IF1

// IF2
// It seems like it in IF2, the core operation in pc_gen is to update the branch predictors
// From this perspective, it seems like we can keep IF2 the same, with the change being that it comes from the 
// realigner (now outside of pc_gen) to pc_gen
fetch_instr_v_i // now indicates whether a complete instruction is being sent to fe_queue
fetch_exception_v_i // don't really need this with the new paradigm, currently unused, can remove
fetch_br_metadata_fwd_o // the metadata associated with the last instruction sent to fe_queue
fetch_pc_o // the PC in IF2
```

This is me thinking out loud on how to simplify it, will have to look more deeply to see if there are problems by going the other way

bp_fe/src/v/bp_fe_pc_gen.sv
```diff
@@ -62,29 +65,57 @@ module bp_fe_pc_gen
...
+  logic next_fetch_linear;
   logic ovr_ret, ovr_taken, btb_taken;
   logic [vaddr_width_p-1:0] btb_br_tgt_lo;
   logic [vaddr_width_p-1:0] ras_tgt_lo;
   logic [vaddr_width_p-1:0] br_tgt_lo;
-  wire [vaddr_width_p-1:0] pc_plus4  = pc_if1_r + vaddr_width_p'(4);
+  wire [vaddr_width_p-1:0] last_fetch_plus4 = fetch_addr_if1_r + vaddr_width_p'(4);
+  always_comb
+    begin
+      next_fetch_linear = 1'b0;
```
dpetrisko, line 90: rather than tracking linear fetch explicitly, we can simply have a state bit in the realigner. We're suppressing branch prediction on misaligned first halves so this means that the realigner will always get the next valid instruction during the next I$ return. On I$ miss, it doesn't flush the realigner or redirect the PC.  The only way to flush the realigner is to have an fe_cmd, which should explicitly flush the state

bp_fe/src/v/bp_fe_top.sv
dpetrisko: see comment in header. In general I would say that this is a good workaround for needing to differentiate between icache return and fetch output. However, the only thing that needed icache return is the realigner (I believe), so extracting that and putting it in fe top should make the interface cleaner

bp_fe/src/v/bp_fe_pc_gen.sv
dpetrisko: Also, we actually track this at the fe_top level now as well: https://github.com/black-parrot/black-parrot/blob/dev/bp_fe/src/v/bp_fe_top.sv#L260. So this should be fairly easy to combine along with the imminent 'compressed' bit from I$ such that it appears in IF1 in pc_gen

WasabiFan: If we drop the bits in the I$ and such then we can indeed change this behavior, as far as I know. Will make the change.

bp_common/src/include/bp_common_pkg.sv
WasabiFan: Added to description

bp_fe/src/v/bp_fe_pc_gen.sv
WasabiFan: If the branch prediction data for a misaligned instruction is written into the BTB entry for its second half, that's its aligned 32-bit address plus four. Is storing branch prediction metadata there the intended behavior?

My implementation initially did this but I was asked to change it so that it stored with the first half and manually delayed the result. I recall the reasoning was around C instructions but I don't remember the details of why we came to that conclusion; it may have been in the context of another constraint we've since worked around. I'll take a look and see if it's feasible. I don't see any issues right now.

bp_fe/src/v/bp_fe_realigner.sv
WasabiFan: What kind of solution would be appropriate here? macro-defining the zeroes and/or some `'X` in? I'd love to put X in here when targeting simulation if that's appropriate.

bp_fe/src/v/bp_fe_pc_gen.sv
WasabiFan: I'm a bit confused here regarding this:

> it comes from the realigner (now outside of pc_gen) to pc_gen

You had previously asked me to move the realigner _into_ the pc_gen module. This is how it is diagrammed and implemented currently.

Is the ask to move it back out?

bp_fe/src/v/bp_fe_pc_gen.sv
WasabiFan: > rather than tracking linear fetch explicitly, we can simply have a state bit in the realigner.

What kind of state bit would this be?

If the thought is to monitor the address being fetched and recognize the nonlinearity there, I very much would like to do so if possible. I have avoided it because such a design would seem to require re-computing `addr+4` to check whether the fetch is the next address up or not. Is this not the case and/or is that acceptable?

> We're suppressing branch prediction on misaligned first halves so this means that the realigner will always get the next valid instruction during the next I$ return.

Every instruction second half also comes with an instruction first half. So jumps and branches also "flush" the realigner, in that sense -- what I called a "nonlinear" fetch. If you make a fetch that isn't x+4, and there's a misaligned instruction involved, the next realigner output is bogus.

Relatedly, the logic is written to only suppress jumps when _all_ of the following are true:
- The fetch is "nonlinear"
- The target is a misaligned instruction
In such a case, it's impossible that following a jump would be appropriate.

If either of those is not the case, it's fine to follow a jump immediately. A fetch being misaligned isn't sufficient to suppress all jumps, otherwise we'd never jump.

---

More directly replying to the comment, the realigner and/or branch predictor do need some way of informing the frontend that the current fetch result is unusable because we're waiting on the second half of an instruction. I don't think that is otherwise covered by the interface. We could potentially augment the "ovr" flag to include this case. Passive monitoring of the address would also be sufficient, discussed above.

WasabiFan:
> What happens during nonspec mode, where I$ misses need to be confirmed by the BE? Does the realigner pass through the exception? Does the second half of a misaligned nonspec access act as speculative or nonspeculative?

The exception handling is currently entirely broken. The vaddr passed along with exceptions is not correct right now.

Exceptions were the final planned fix and verification step I was working on a few weeks ago. Rather than guessing correct behavior or implementing a half-measure I took the time to write some VM-enabled misaligned instruction tests which stress ITLB and exception handling. These tests are in place and as far as I recall mostly "done" but I have not yet implemented the correct exception handling to make them pass. I still have a to-do item to implement a nonspeculative mode test with misaligned instructions for the same reason. I intend to pick that back up only after I've implemented the rewrite requested above.

Regarding nonspeculative I$ in particular: the theory is that the FE queue output for exceptions isn't suppressed by requests for the first half of an instruction. This means that a misaligned instruction fetch can't ever issue an instruction, but _can_ issue an exception -- access violations, page faults, and, in theory, cache misses if enabled. Similarly, the second half can emit exceptions. In both cases, the exception halts processing of the entire instruction.

As I understand it, the reply to a nonspec cache miss is a cache fill and then a redirect, which would then restart processing of the instruction. The replay register captures the PC rather than the fetched vaddr and thus will replay the entire fetch, first and second halves, even if the first half fetched fine. I opted for this approach because it minimizes the extra logic+verification to ensure that we keep the realigner state coherent while waiting for the resume. If we wanted to avoid this, we could probably cache the realigner buffer in the resume register. I haven't thoroughly evaluated the implications of doing so and what extra logic it would require.

bp_fe/src/v/bp_fe_pc_gen.sv
WasabiFan: My understanding of the suggestion is to select only fetch addresses in IF0/IF1, and pick a concrete PC (essentially, second bit) in IF2. Is that right?

In that case, what would be the logic and inputs for assigning a PC to a given fetch? 

 I think it would be much nicer if we can do it this way. However, I have previously rejected this design for a few reasons. Some of these reasons are no longer applicable if we store branch prediction data indexed by the second half of an instruction rather than the first half as I believe you have suggested elsewhere.

The reasoning is as follows, from my recollection.

We need to both a) ensure that earlier stages can still make effective decisions about following jumps, and b) continue to output full-fidelity final PCs. In the latter case, this is vital to ensure the realigner produces the correct result (it must know whether to realign a misaligned instruction or pass through an aligned one).

For the former goal, we previously needed to know whether we were operating on a misaligned instruction so that we could decide to delay following branch predictions by a cycle when fetching the second half of an instruction. If we store misaligned instructions' metadata with their second half, this would no longer be an issue as the predictions only come when they can be acted upon.

Also relating to the former goal: once C support is added, knowing in IF1 whether we are immediately executing the upper half or waiting a cycle would allow us to ignore a lower-half compressed jump if it isn't going to be executed anyway.

For the latter goal, the realigner or whoever makes the final PC decision would need access to the BTB entry to be able to follow indirect jumps. If this is indeed the realigner it would be a bit strange to be feeding the jump target so far. If the pc_gen module internally is the one that assigns a final PC (perhaps by virtue of the realigner remaining within pc_gen), this would be a non-issue.

Note that we could _also_ go to the extreme of not assigning a PC at all in the frontend and instead letting the backend decide whether it's the first half or the second half being executed. It seems like it would work for our particular situation barring the points above but would no longer be appropriate if the branch predictors gained 16-byte specificity.

WasabiFan:
@dpetrisko Bump on the above questions/replies.

dpetrisko:
>Bump on the above questions/replies.

Hi, thanks for the updates. Sorry, I'm quite behind on PRs. Will get to this before end of quarter...

WasabiFan:
Thanks for the reply, no worries. This quarter I'm taking a class which requires orders of magnitude more time than I've ever spent on a class before, so I'm pretty behind as well. But once I know in particular what changes I need to make I'll find some time to make it happen.

WasabiFan:
@dpetrisko Friendly bump. I would like to work on this and get it done while I'm still around, and your comments would be appreciated. Some of the changes discussed here will be significant refactorings so it'll take me a bit to get back up to speed but I want to make it happen.

dpetrisko:
Hi @WasabiFan, I'll try to pull some time for this. Appreciate the persistence!

bp_fe/src/include/bp_fe_defines.svh
dpetrisko: Hi @WasabiFan 

```
`define bp_align_addr(addr_mp, num_bytes_mp) \
({addr_mp[$bits(addr_mp)-1:$clog2(num_bytes_mp)], {$clog2(num_bytes_mp){1'b0}}}) 
```

See here for use: https://www.edaplayground.com/x/M3q6

PR #988, opened 2021-12-24
Add new pc_gen_tracer to top testbench for logging frontend behavior

WasabiFan:
## Summary

Introduce a tracer module focused on the frontend. It prints the PCs in IF1 and IF2, the current source for the IF1 instruction (pc+4, RAS, etc.), and whether there are any incoming commands or cache misses.

Designed to help validate #986 

PCs are enclosed in parenthesis if the pipeline stage is invalid.

## Issue Fixed

None

## Area

bp_top testbench

There's no existing way to get a clear view into the frontend's behavior. Viewing the raw simulation output in a wave viewer doesn't give a good picture of the overall flow or context. This tracer outputs a log of cycle-by-cycle status in text form, which means it's easy to textually diff the outputs and see how changes affect results.

None

## Analysis

None

## Verification

None

## Additional Context

Attached a sample trace from the `bubble_sort_demo` program.

[pc_gen_trace_0.trace.txt](https://github.com/black-parrot/black-parrot/files/7773190/pc_gen_trace_0.trace.txt)


bp_top/test/common/bp_nonsynth_pc_gen_tracer.sv
```diff
@@ -0,0 +1,143 @@
...
+    end
+
+  string padded_pc_src_if1_name;
+  always_ff @(negedge clk_i)
+    if (!reset_i && !freeze_i)
+    begin
+      // Name of the current source, to output to the log file, but with spaces appended for easy uniform-width substrings
+      padded_pc_src_if1_name = ({pc_src_if1_r.name(), {20{" "}}});
+
+      $fwrite
```
dpetrisko, line 120: Can you keep the same prefix as the other tracers? https://github.com/black-parrot/black-parrot/blob/master/bp_top/test/common/bp_nonsynth_cache_tracer.sv#L189

We use simple grep / awk / sort etc. based on the prefix to combine and divide tracer files

bp_top/test/common/bp_nonsynth_pc_gen_tracer.sv
```diff
@@ -0,0 +1,143 @@
...
+
+     ,.clear_i(1'b0)
+     ,.up_i(1'b1)
+     ,.count_o(cycle_cnt)
+     );
+
+  bp_pc_src_e pc_src_if1_n, pc_src_if1_r;
+  always_ff @(posedge clk_i)
+    if (reset_i | freeze_i)
+      pc_src_if1_r <= e_pc_src_undefined;
```
dpetrisko, line 78: Does this cause the first one to be src_undefined? If this is the only case it's used for, I would just make it whichever one is first

bp_top/test/common/bp_nonsynth_pc_gen_tracer.sv
```diff
@@ -0,0 +1,143 @@
...
+    else
+      pc_src_if1_r <= pc_src_if1_n;
+
+  always_comb
+    begin
+      if (src_redirect_i)
```
dpetrisko, line 84: Does this just mirror the FE condition? Probably want to remove the code duplication here. If we replace the always_comb mux in pc_gen with a bsg_mux, we could use the control signal directly in this tracer

bp_top/test/common/bp_nonsynth_pc_gen_tracer.sv
```diff
@@ -0,0 +1,143 @@
...
+      file      = $fopen(file_name, "w");
+      $fwrite(file, "%s,%s,%s,%s,%s,%s,\n", "cycle", "IF1 PC", "IF2 PC", "IF1 PC src", "state", "events");
+    end
+
+  string padded_pc_src_if1_name;
+  always_ff @(negedge clk_i)
+    if (!reset_i && !freeze_i)
+    begin
+      // Name of the current source, to output to the log file, but with spaces appended for easy uniform-width substrings
+      padded_pc_src_if1_name = ({pc_src_if1_r.name(), {20{" "}}});
```
dpetrisko, line 118: I think you should be able to use the print format itself for this: %4s

bp_top/test/common/bp_nonsynth_pc_gen_tracer.sv
WasabiFan: The cycle counter format? Yeah, I can adopt that.

bp_top/test/common/bp_nonsynth_pc_gen_tracer.sv
WasabiFan: I'll check it out. The substr behavior was really fickle (it seems that going out of bounds gives an empty string) so I got angry and introduced a hack. I'll try and see if that format specifier does what we need.

bp_top/test/common/bp_nonsynth_pc_gen_tracer.sv
WasabiFan: Yes, although note that the first few rows are all X when we come out of reset. The initial redirect happens a few cycles later. So I'm good with defaulting to a +4 and calling it "close enough" if we want to. What the first few rows show probably doesn't matter.

bp_top/test/common/bp_nonsynth_pc_gen_tracer.sv
WasabiFan: How would I use a bsg_mux for this? I'd be happy to reduce duplication in any feasible way if there's a good way of doing it. The control signals/selector is the same between the testbench and the real impl but the signal being passed through is different (an enum indicating the source vs. the actual PC).

I could use a priority encoder or similar then select based on the output enum (i.e., make this enum part of the real datapath) if we want to do that. It would still be an "if" chain since using the bsg priority encoder mode would mean we'd have to cast between the enum and a raw bit pattern.

bp_top/test/common/bp_nonsynth_pc_gen_tracer.sv
dpetrisko: Yeah I would say close enough and get rid of the src_undefined. You can pipeline the reset the length of the FE and that should sync up?

bp_top/test/common/bp_nonsynth_pc_gen_tracer.sv
dpetrisko: Yeah it would probably be a bsg_priority_encode_one_hot + bsg_mux_one_hot. I'm fine merging this without, since we'd need to verify it didn't impact critical path.

Put a TODO comment to remove the duplication by restructuring the datapath, though

bp_top/test/common/bp_nonsynth_pc_gen_tracer.sv
```diff
@@ -0,0 +1,143 @@
...
+  ,e_pc_src_redirect
+  ,e_pc_src_override_ras
+  ,e_pc_src_override_branch
+  ,e_pc_src_btb_taken_branch
+  ,e_pc_src_last_fetch_plus_four
+} bp_pc_src_e;
+
+module bp_nonsynth_pc_gen_tracer
+  import bp_common_pkg::*;
+  import bp_be_pkg::*;
```
dpetrisko, line 17: Do you need anything from be? Should be bp_fe_pkg?

bp_top/test/common/bp_nonsynth_pc_gen_tracer.sv
```diff
@@ -0,0 +1,143 @@
+`include "bp_common_defines.svh"
```
dpetrisko, line 1: Tracer should be in bp_fe/test/common/bp_nonsynth_fe_pc_gen_tracer.sv

bp_top/test/common/bp_nonsynth_pc_gen_tracer.sv
```diff
@@ -0,0 +1,143 @@
...
+typedef enum logic [2:0]
+{
+  e_pc_src_undefined = 3'd0
+  ,e_pc_src_redirect
+  ,e_pc_src_override_ras
+  ,e_pc_src_override_branch
+  ,e_pc_src_btb_taken_branch
+  ,e_pc_src_last_fetch_plus_four
+} bp_pc_src_e;
+
```
dpetrisko, line 14: Put a comment saying how this tracer is intended to be used. Maybe put an example trace so people can see the format without running

bp_top/test/common/bp_nonsynth_pc_gen_tracer.sv
```diff
@@ -0,0 +1,143 @@
...
+typedef enum logic [2:0]
+{
+  e_pc_src_undefined = 3'd0
+  ,e_pc_src_redirect
+  ,e_pc_src_override_ras
+  ,e_pc_src_override_branch
+  ,e_pc_src_btb_taken_branch
+  ,e_pc_src_last_fetch_plus_four
+} bp_pc_src_e;
```
dpetrisko, line 13: bp_fe_pc_gen_src_e

bp_top/test/common/bp_nonsynth_pc_gen_tracer.sv
```diff
@@ -0,0 +1,143 @@
...
+    if (!reset_i && !freeze_i)
+    begin
+      // Name of the current source, to output to the log file, but with spaces appended for easy uniform-width substrings
+      padded_pc_src_if1_name = ({pc_src_if1_r.name(), {20{" "}}});
+
+      $fwrite
+        (file
+        ,"%0d,%s,%s,%s,%s,"
+        ,cycle_cnt
+        ,render_addr_with_validity(if1_pc_i,         if1_top_v_i)
```
dpetrisko, line 124: Generally for tracers, we prefer to have a single commit point where things are recorded. I would suggest making this IF2, when the instruction data goes into the FE queue (or not). Any information in IF1 will be seen the next cycle and possibly squashed, so the redundancy is typically more noisy than valuable

bp_top/test/common/bp_nonsynth_pc_gen_tracer.sv
dpetrisko: $time counter but yes

bp_top/test/common/bp_nonsynth_pc_gen_tracer.sv
dpetrisko: not a huge deal if it doesn't work, but would be more maintainable and portable if it does

dpetrisko:
Very cool! Left feedback, nothing too major

bp_top/test/common/bp_nonsynth_pc_gen_tracer.sv
WasabiFan: In this case, I _did_ design it to show all stages, including squashed entries, since it helps monitor branch prediction choices. This identified a bug. For the version of the tracer included here (i.e., without the misaligned instruction support) removing IF1 is fine since all IF1 entries flow into IF2 unmodified anyway, optionally invalidated. However, _with_ misaligned support (where there's both a fetch and pc address) and especially with "C" support, this will matter.

Pending the changes requested to my misaligned insn PR, the details here may change, but assuming there is additional information shown by the IF1 tracer I would prefer to keep it. As in, I will certainly need it for debugging either way, and feel that what's valuable for implementation probably also ought to be valuable for others later.

WasabiFan:
Superseded by #1098

WasabiFan:
Sorry, wrong one, this needs to be updated but isn't superseded

bp_top/test/common/bp_nonsynth_pc_gen_tracer.sv
WasabiFan: IF1 removed, "source" data now for IF2 to match

WasabiFan:
Updated. All suggestions applied. Block comment explaining tracer with an example output is now included.

dpetrisko:
LGTM!

PR #998, opened 2021-12-30
[WIP - DEFERRED] Downclock L2

dpetrisko:
## Summary
This PR adds the capability to downlock the L2 cache. This is useful in the case that we want a large L2, but don't want to limit the frequency of the core. Of course, the L2 access latency will rise linearly with the downgrade in clock frequency, but overall system performance cares more about L2 hits than L2 latency.

## Area
unicore, tile, L2

We could use async fifos on either end of the cache. However, this incurs a 3 slow cycle penalty on each side, leading to an unnecessary latency hit. Additionally, the extra storage from the async fifos would hurt area gains from a more efficient L2 RAM. Instead, we use a new module, bsg_fifo_1r1w_periodic.sv which modifies the handshakes of ready-valid interfaces to 

- [ ] Fix up aviary test for downclock
- [ ] Run through synthesis flow

## Analysis
This change enables nearly arbitrarily sized L2s. Optimal L2 size will change based on a variety of system parameters, but we should pick reasonable defaults.

## Verification
The RTL has been tested for downclock of 1 (none), 2, 4, 8x core frequency. We need to push this through an ASIC flow so that constraints can be determined. Most likely it will simply be a generated clock constraint and disabling some paths.

### DIFF
--- a/bp_common/src/include/bp_common_aviary_defines.svh
+++ b/bp_common/src/include/bp_common_aviary_defines.svh
@@ -95,6 +95,7 @@
         `BSG_MAX(dcache_sets_p, `BSG_MAX(icache_sets_p, num_cacc_p ? acache_sets_p : '0))          \
     , localparam cce_ucode_p                = proc_param_lp.cce_ucode                              \
                                                                                                    \
+    , localparam l2_downclock_p           = proc_param_lp.l2_downclock                             \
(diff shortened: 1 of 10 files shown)


dpetrisko:
Will revisit this with a PISO/SIPO approach

PR #1011, opened 2022-01-13
Change LCE interface to BedRock Burst

muwyse:


bp_top/src/v/bp_tile.sv
```diff
@@ -105,20 +116,17 @@ module bp_tile
   bp_bedrock_lce_cmd_header_s cce_lce_cmd_header;
   logic [dword_width_gp-1:0] cce_lce_req_data, cce_lce_resp_data, cce_lce_cmd_data;
 
+  // note: data width here doesn't really matter because we only use header and don't deserialize the packet
   `declare_bp_lce_req_wormhole_packet_s(coh_noc_flit_width_p, coh_noc_cord_width_p, coh_noc_len_width_p, coh_noc_cid_width_p, bp_bedrock_lce_req_header_s, cce_block_width_p);
```
dpetrisko, line 120: Should just declare wormhole header then and get rid of the wormhole packet macro

bp_top/src/v/bp_tile.sv
```diff
@@ -137,96 +145,158 @@ module bp_tile
...
-
-         ,.packet_i(lce_resp_packet_lo[i])
-         ,.v_i(lce_resp_v_lo[i])
-         ,.ready_o(lce_resp_ready_li[i])
-
-         ,.link_i(lce_resp_link_li[i])
-         ,.link_o(lce_resp_link_lo[i])
-         );
-    end
+  localparam pr_len_width_lp = 8;
```
dpetrisko, line 148: where's this come from?

muwyse: magic! I forget, I'll trace it down.

bp_top/src/v/bp_tile.sv
```diff
@@ -138,96 +145,158 @@ module bp_tile
...
-         ,.flit_width_p(coh_noc_flit_width_p)
-         )
-       lce_resp_adapter_in
-        (.clk_i(clk_i)
-         ,.reset_i(reset_r)
-
-         ,.packet_i(lce_resp_packet_lo[i])
-         ,.v_i(lce_resp_v_lo[i])
-         ,.ready_o(lce_resp_ready_li[i])
+  localparam pr_len_width_lp = `BSG_SAFE_CLOG2(`BSG_CDIV((1<<e_bedrock_msg_size_128)*8,icache_fill_width_p));
```
dpetrisko, line 148: I would suggest calculating this as a `BSG_MAX of the I$, D$ and A$ fill width in aviary_defines. I agree the extra bits are not a significant overhead, so might as well be flexible for future heterogeneity (and hide it from users a bit).

bedrock_len_width_p?

muwyse: reworked computation and renamed param in bp_tile. I don't think it can be computed in aviary_defines without introducing either a dependence on the bedrock package or using a magic number of 1024 (number of bits in a 128B message)

PR #1012, opened 2022-01-14
merge me_dev to dev

muwyse:
merge me_dev to dev


bp_me/src/v/cce/bp_cce_dir.sv
```diff
@@ -70,12 +68,21 @@ module bp_cce_dir
...
+  wire unused = &{cce_id_i};
+
   // Number of CCEs must be at least as large as the minimal number of sets in any of the
   // LCE types (dcache, icache, acache). This ensures that every tag set is wholly stored
-  // in a *single* CCE (equivalently, tag sets are not split across LCEs).
+  // in a *single* CCE (equivalently, tag sets are not split across CCEs).
   // LCEs and CCEs use the set index bits from the physical address to map address to CCE.
   if (lce_min_sets_lp < num_cce_p)
-    $fatal(0,"Number of CCEs must be at least as large as the minimal number of LCE sets");
+    $fatal(0, "Number of CCEs must be at least as large as the minimal number of LCE sets");
```
dpetrisko, line 79: Not required for this PR, but for future -- $error is better supported in tools than $fatal (and has identical semantics, other than the custom error code)

bp_me/src/v/cce/bp_cce_dir_tag_checker.sv
```diff
@@ -31,10 +31,6 @@ module bp_cce_dir_tag_checker
    , output bp_coh_states_e [tag_sets_per_row_p-1:0]              sharers_coh_states_o
   );
 
-  // parameter checks
```
dpetrisko, line 34: Do we have a config that tests this?

bp_me/src/v/cce/bp_cce_dir.sv
muwyse: sounds good, thanks for the tip

bp_me/src/v/cce/bp_cce_dir_tag_checker.sv
muwyse: this is a "hard-coded" parameter in the CCE directory. I just removed this particular check because it is redundant with the one in bp_cce_dir_segment. We only support a value of 2 for the parameter, which we determined to be reasonable at some point in the past when looking at physical sizing of the directory RAMs.

muwyse: I suppose we could change the parameter to be a localparam in bp_cce_dir_segment to further clarify that it can't be modified via the aviary configs and is a specifically selected constant. I can update the existing comment with the parameter to make it clearer that a value other than 2 will likely break the directory implementation.

dpetrisko (approved):
LGTM!

Issue #1013, opened 2022-01-16
bp_me_axil_client module drops mesg


Yuan-Mao:
## Branch / Commit Hash
top_dev (a6a0d84)

## Area
bp_me_axil_client module:
https://github.com/black-parrot/black-parrot/blob/top_dev/bp_me/src/v/network/bp_me_axil_client.sv#L121

## Current behavior
If state_r == e_wait, and a master is sending an AXIL write to its client, then according to line 125~127, master should be able to get all the required ready signals to complete its write operation. However, if at this moment io_cmd_ready_and_i is not ready yet, then the data that meant to be sent will be lost.

dpetrisko:
Thanks for the report, aware of the issue. Should be fixed this weekend

dpetrisko:
See #1014

Yuan-Mao:
Thank you so much!

dpetrisko:
Fixed

PR #1014, opened 2022-01-18
Fix to axil client

dpetrisko:
This PR fixes the axi lite client for #1013. Writes are able to proceed at 100% throughput, but reads now take an additional cycle. Reads are much less common in our current AXI systems, but future enhancements could pipeline this as well by adding registers for the araddr and additional state

### DIFF
--- a/bp_me/src/v/network/bp_me_axil_client.sv
+++ b/bp_me/src/v/network/bp_me_axil_client.sv
@@ -30,7 +30,7 @@ module bp_me_axil_client
    , input [cce_mem_header_width_lp-1:0]        io_resp_header_i
    , input [axil_data_width_p-1:0]              io_resp_data_i
    , input                                      io_resp_v_i
-   , output logic                               io_resp_ready_and_o
+   , output logic                               io_resp_yumi_o
 
    //====================== AXI-4 LITE =========================
    // WRITE ADDRESS CHANNEL SIGNALS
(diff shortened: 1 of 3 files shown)


Yuan-Mao:
Not sure if I can follow. Are you meant to write 's_axil_awvalid_i & s_axil_wvalid_i'?
https://github.com/black-parrot/black-parrot/blob/61da1f66de2f517bf88e7652afa890c8c36764b1/bp_me/src/v/network/bp_me_axil_client.sv#L135

dpetrisko:
> Not sure if I can follow. Are you meant to write 's_axil_awvalid_i & s_axil_wvalid_i'?
> 
> https://github.com/black-parrot/black-parrot/blob/61da1f66de2f517bf88e7652afa890c8c36764b1/bp_me/src/v/network/bp_me_axil_client.sv#L135

In ready, we send a write if awvalid and wvalid are both high. If only awvalid is high, then we latch it and move to send_write. In send_write, we wait until wvalid is high. This allows us to send writes with 100% throughput in the case of address being sent at the same time as data, as well as max (50%) throughput if address is sent first, then data.

Yuan-Mao:
I see. The reason why I looked at these changes was that, after applying them to zynq-parrot I wasn't able to run simulation on Verilator(AXI timeout), but after looking at the waveform it seems to me it is actually Verilator itself that generated the issue. VCS works well.

dpetrisko:
Yep, Verilator has some issues I'm working on in https://github.com/black-parrot-hdk/zynq-parrot/pull/21. Feel free to pull that in, or just use VCS in the meantime

PR #1016, opened 2022-01-25
Minor fix for bp_axi_top

### LINKED ISSUE #1015: bp_axi_top width conversion
## Branch / Commit Hash
860afac
## Area
bp_axi_top module

## Current behavior
https://github.com/black-parrot/black-parrot/blob/top_dev/bp_top/src/v/bp_axi_top.sv#L394

I am using zynq-parrot, so the axil bus width is 32-bit and the IO width for BP unicore is 64 bit. I found when BP tried to read a value from AXIL (bp_me_axil_master), it will not read out correctly. To be specific, my read value should be 32'h01, but instead I got 0 in my test C program.

I have tried using bsg_bus_pack to expand io_resp_data_li and now it seems to work.

Yuan-Mao:
## Summary
This PR adds I/O response width conversion between BP and axil master.

## Issue Fixed
#1015

## Area
bp_axi_top

## Verification
Able to run in zynq-parrot co-simulation

### DIFF
--- a/bp_me/src/v/network/bp_me_axil_master.sv
+++ b/bp_me/src/v/network/bp_me_axil_master.sv
@@ -8,7 +8,7 @@ module bp_me_axil_master
  #(parameter bp_params_e bp_params_p = e_bp_default_cfg
   `declare_bp_proc_params(bp_params_p)
   `declare_bp_bedrock_mem_if_widths(paddr_width_p, did_width_p, lce_id_width_p, lce_assoc_p, cce)
-
+  , parameter io_data_width_p = multicore_p ? cce_block_width_p : uce_fill_width_p
   // AXI WRITE DATA CHANNEL PARAMS
   , parameter axil_data_width_p = 32
   , parameter axil_addr_width_p  = 32
@@ -24,7 +24,7 @@ module bp_me_axil_master
   , output logic                               io_cmd_ready_and_o
 
   , output logic [cce_mem_header_width_lp-1:0] io_resp_header_o
-  , output logic [axil_data_width_p-1:0]       io_resp_data_o
+  , output logic [io_data_width_p-1:0]         io_resp_data_o
   , output logic                               io_resp_v_o
   , input                                      io_resp_ready_and_i
 
@@ -86,7 +86,6 @@ module bp_me_axil_master
 
       io_cmd_ready_and_o    = 1'b0;
       io_resp_header_cast_o = io_cmd_header_r;
-      io_resp_data_o        = m_axil_rdata_i;
       io_resp_v_o           = '0;
 
       // WRITE ADDRESS CHANNEL SIGNALS
@@ -178,6 +177,16 @@ module bp_me_axil_master
         default: state_n = state_r;
       endcase
     end
+
+  bsg_bus_pack
+   #(.in_width_p(axil_data_width_p)
+     ,.out_width_p(io_data_width_p))
+   bus_pack
+    (.data_i(m_axil_rdata_i)
+     ,.sel_i('b0)
+     ,.size_i(`BSG_SAFE_CLOG2(axil_data_width_p / 8))
+     ,.data_o(io_resp_data_o)
+     );
 
   // synopsys sync_set_reset "reset_i"
   always_ff @(posedge clk_i)


dpetrisko:
Can you put this in the axil_master module instead?

bp_me/src/v/network/bp_me_axil_master.sv
```diff
@@ -179,6 +178,16 @@ module bp_me_axil_master
       endcase
     end
 
+  bsg_bus_pack
+   #(.in_width_p(axil_data_width_p)
+     ,.out_width_p(io_data_width_p))
+   bus_pack
+    (.data_i(m_axil_rdata_i)
+     ,.sel_i('b0)
+     ,.size_i(`BSG_SAFE_CLOG2(axil_data_width_p / 8))
```
dpetrisko, line 187: You need to use the size of the request here: https://github.com/black-parrot/black-parrot/blob/7a1695bce1a0f85ec3e4a5c0d945f7b488eba279/bp_me/src/v/network/bp_me_axil_master.sv#L111. The axil data should be repeated every 32 bits if 32 bits request, 16 bits if 16 bits, etc.

dpetrisko: And the sel_i is based on the address

dpetrisko:
We can expand this test to verify the implementation: https://github.com/black-parrot-sdk/bp-demos/blob/master/src/scratchpad/scratchpad.c

By writing and reading 8b,16b, 32b, 64b accesses to various aligned addresses

bp_me/src/v/network/bp_me_axil_master.sv
Yuan-Mao: Got it! In terms of sel_i, should I expect the client will not do the data shift for the master? For example, if master issues a read on address 0x1, will the client need to right shift the read data by 8 bit or just return the 32-bit read value from address 0x0, and then master will use sel_i to choose the proper subset of data?

Yuan-Mao: I just checked the AXI4-lite spec, so the data access width should always be the same as the bus width, so I guess this confirms that we need sel_i to get rid of the offset in the read value.

dpetrisko: oh can you show me the page in the spec that says that?

Yuan-Mao: https://michaeltaylor.org/edu/papers/AXI4_specification.pdf
B1.1.2 Bus Width

dpetrisko: Oh nice. So we only need to select between 32-bit alignments

Yuan-Mao: Then I guess our axil_master will have limited support for BP commands such as reading a 2-byte data from read address 0x1002. Technically it is doable, we can make all clients support unaligned memory access(read 4-byte from 0x1002) and then BP can filter out our the unused data, but that would be a lot of work/pain.

Otherwise we have to break the spec (or make extension to the spec) by saying that we can have access width smaller than the bus width.

dpetrisko: Yeah not worth it, we can just add an assertion to only support bus width accesses

Yuan-Mao: I think we can still add support for widths smaller than the bus: After receiving such requests, axil_master can instead issue a full bus width read access for axil client and after getting the return value axil_master packs the data using bsg_bus_pack before sending back to BP.

dpetrisko: Possible but I'm not sure it's a useful feature. Since axil clients expect bus-aligned accesses, their register maps are typically configured to be 32-bit mapped. If we need it for a particular client it's fine to add, but I would hold off until then

Yuan-Mao: I think I will need it because when Linux wants to read out the packet from my Ethernet device, it will not always issue 32-bit load instructions. For example if the packet size is 127 then it will at least issue one lb instruction. And you know one of my goal is to use Linux without any custom changes.

PR #1025, opened 2022-02-08
Convert LCE and IO CCE to Burst Interfaces

muwyse:
## Summary
This PR does the following:
- converts the LCE to use BedRock Burst interfaces, enabling critical word first behavior
- converts the IO CCE to use BedRock Burst interfaces
- adds a BedRock Burst gearbox to connect Burst interfaces with different data widths
- refactors burst-wh converters to leverage the burst gearbox
- updates multicore tiles
- reorgs the nonsynth IF verif module to clearly distinguish which area each param check applies to
- updates testbenches
- fixes bug with data PISO in burst_to_wormhole converter

## Area
ME and multicore

The LCE's current interface is deprecated and inefficient. This change implements a more efficient interface for the LCE-CCE channels, which also enables critical word first support in the LCE.

## Analysis
The interface changes throughout should provide minor reduction in buffering and possibly a minor performance bump.

## Verification
Standard regressions.


muwyse:
Note: accelerator configs need further testing

dpetrisko:
Any way we can decompose this? Maybe a first PR to add the fill network?

muwyse:
> Any way we can decompose this? Maybe a first PR to add the fill network?

maybe; I'll investigate

muwyse:
depends on #1027 for some ME branch cleanup

bp_me/src/v/network/bp_me_bedrock_mem_to_link.sv
```diff
@@ -0,0 +1,162 @@
...
+ * Description:
+ *   Converts BedRock Burst Memory Interface to Wormhole Link interface.
+ *   The payload_mask_p parameter indicates which BedRock input messages carry data.
+ *
+ */
+
+`include "bp_common_defines.svh"
+`include "bp_me_defines.svh"
+
+module bp_me_bedrock_mem_to_link
```
dpetrisko, line 15: I think this can work for all bedrock links right, not just mem?

muwyse: With a little refactoring, I think so. The header packet encode module needs to be pulled out, or somehow generalized/parameterized based on the type of bedrock network in use. But otherwise this module is simply wrapping bidirectional burst-wh conversion.

muwyse: I think there are potentially a few changes that could be lumped into a second PR related to generalizing / cleaning up some of the ME network code, especially related to this type of WH-BedRock conversion.

dpetrisko: Yeah maybe let’s do that first. Because that will reduce the code a lot in the tiles, etc

bp_me/src/v/cce/bp_io_cce.sv
```diff
@@ -18,31 +22,49 @@ module bp_io_cce
    `declare_bp_bedrock_lce_if_widths(paddr_width_p, lce_id_width_p, cce_id_width_p, lce_assoc_p)
```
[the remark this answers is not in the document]
muwyse: Converting burst-wh is fairly straightforward, so changing the IO CCE to use stream internally effectively means converting the LCE burst interface to stream, then converting that stream back to WH (~= burst). Converting burst-wh is basically just encoding the header and splitting it across multiple flits if needed. This makes the IO Tile more complex.

Gearboxing the data is required for either stream or burst to wh, but gearboxing is more complicated with stream.

The only other place the IO CCE is used is in the accelerator tiles, where I would argue that the accelerators should actually expose a burst interface for their IO ports to eliminate stream to burst conversions.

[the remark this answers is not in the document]
dpetrisko: I agree totally (maybe too much =)). I sent a more complete message offline, let me know what you think. 
 
But for this, yes let’s keep I/O burst-burst

muwyse:
depends on: https://github.com/bespoke-silicon-group/basejump_stl/pull/552

.gitlab-ci.yml
```diff
@@ -26,6 +26,7 @@ stages:
     - me_dev
     - top_dev
     - sw_dev
+    - me_dev_feature_bedrock
```
muwyse, line 29: note to self: remove before merge

bp_me/src/v/network/bp_me_bedrock_mem_to_link.sv
muwyse: Refactoring and further generalization deferred to future PR.

muwyse:
closes #846

PR #1031, opened 2022-02-11
Fix address/data alignment in LCE Command

muwyse:
## Summary
This change (re-)fixes an alignment issue in the LCE Command module that can occur due to uncached accesses to cacheable memory.

## Area
LCE Command

## Verification
Standard CI


dpetrisko:
LGTM

dpetrisko (approved):

PR #1032, opened 2022-02-11
Merge me_dev to dev

muwyse:
## Summary
This PR merges me_dev to dev in preparation for fixing the BedRock protocol deadlock and implementing BedRock Burst interface in the LCE.

The main changes included here are:
1. cleanup and refactoring in the CCE
2. BedRock interface data width parameterization in cce_to_cache and cache_slice
3. fix address alignment in commands/responses from LCE triggered by uncached access to cacheable memory
4. refactoring of the ME testbench to use the real LCE and a nonsynth cache, run trace-based test executions, and groundwork for automating AXE consistency checking in the ME testbench
5. bump AXE submodule in tools/
6. general cleanup in some ME modules for parameter clarity or constraints
7. update ME wormhole defines to include header struct macros

## Issue Fixed
None

## Area
ME, mostly testing (nonsynth) and code clarity (synth). Small bug fix in LCE.
Very minor changes to some CI scripts, flists, and top testbench.

General prep for upcoming PRs to reduce future PR size.

## Future Work
Some future work that is out of scope of this PR includes:
1. automating AXE consistency tracing and checking
2. adding AXE tracing to the top-level testbench so it can be run with real programs

## Analysis
Bug fix in LCE should result in minimal changes (masking an address). Rest of changes is nonsynth/no impact

## Verification
Standard CI


bp_me/src/v/cce/bp_cce_inst_stall.sv
```diff
@@ -64,27 +64,24 @@ module bp_cce_inst_stall
...
     stall_o |= (decoded_inst_i.poph & (~lce_req_header_v_i & (decoded_inst_i.popq_qsel == e_src_q_sel_lce_req)));
     stall_o |= (decoded_inst_i.poph & (~lce_resp_header_v_i & (decoded_inst_i.popq_qsel == e_src_q_sel_lce_resp)));
-    stall_o |= (decoded_inst_i.lce_cmd_v & ~lce_cmd_header_ready_and_i);
     stall_o |= (decoded_inst_i.poph & (~mem_resp_v_i & (decoded_inst_i.popq_qsel == e_src_q_sel_mem_resp)));
 
-    // Pop Data
-    stall_o |= (decoded_inst_i.popd & (~lce_req_header_v_i & (decoded_inst_i.popq_qsel == e_src_q_sel_lce_req)));
-    stall_o |= (decoded_inst_i.popd & (~lce_resp_header_v_i & (decoded_inst_i.popq_qsel == e_src_q_sel_lce_resp)));
-    stall_o |= (decoded_inst_i.lce_cmd_v & ~lce_cmd_header_ready_and_i);
+    // Pop Data - TODO: not fully implemented
```
dpetrisko, line 75: what's not implemented here?

bp_me/test/common/bp_me_nonsynth_tr_tracer.sv
```diff
@@ -0,0 +1,126 @@
...
+     (.clk_i(clk_i)
+      ,.reset_i(reset_i)
+      ,.clear_i(tr_pkt_yumi_i)
+      ,.up_i(tr_v)
+      ,.count_o(tr_latency)
+      );
+
+  string s;
+  always_comb begin
+    case (tr_cmd.cmd)
```
dpetrisko, line 83: FYI if this is an enum, you can do cmd.name()

tools/Makefile.common
```diff
@@ -8,7 +8,7 @@ BP_TOOLS_INSTALL_DIR := $(BP_TOOLS_DIR)/install
 BP_TOOLS_BIN_DIR     := $(BP_TOOLS_INSTALL_DIR)/bin
 BP_TOOLS_LIB_DIR     := $(BP_TOOLS_INSTALL_DIR)/lib
 BP_TOOLS_INCLUDE_DIR := $(BP_TOOLS_INSTALL_DIR)/include
-BP_TOOLS_TOUCH_DIR   := $(BP_TOOLS_INSTALL_DIR)/touchfiles/tools
+BP_TOOLS_TOUCH_DIR   := $(BP_TOOLS_INSTALL_DIR)/touchfiles
```
dpetrisko, line 11: keep this?

bp_me/test/common/bp_nonsynth_mem.sv
```diff
@@ -23,13 +24,13 @@ module bp_nonsynth_mem
    , input                                          reset_i
 
    , input [mem_header_width_lp-1:0]                mem_cmd_header_i
-   , input [l2_fill_width_p-1:0]                    mem_cmd_data_i
+   , input [cache_data_width_p-1:0]                 mem_cmd_data_i
```
dpetrisko, line 27: keep as L2 fill width (not data width)? Are you overriding this somewhere?

bp_me/src/v/dev/bp_me_cache_slice.sv
```diff
@@ -19,6 +19,7 @@ module bp_me_cache_slice
  import bsg_cache_pkg::*;
  #(parameter bp_params_e bp_params_p = e_bp_default_cfg
    `declare_bp_proc_params(bp_params_p)
+   , parameter cache_data_width_p = l2_data_width_p
```
dpetrisko, line 22: Are you overriding this somewhere? Keep as L2 data width?

dpetrisko:
Great changes, just a few questions. Peripherally related, please "squash and merge" future PRs when merging to me_dev

bp_me/test/common/bp_me_nonsynth_tr_tracer.sv
muwyse: ah, yes, good catch. I was just reading about this actually, but missed implementing it here. Thanks!

tools/Makefile.common
muwyse: Yes. The touchfiles directory is already under TOOLS_INSTALL, so there is no need to create this extra subdirectory. I will double check the tools building to confirm, but I noticed it when building AXE.

bp_me/src/v/dev/bp_me_cache_slice.sv
muwyse: Upcoming PR will override this parameter. This module is used in the unicore, multicore tile, and L2 tiles. Unlike the other ME devices, which have a fixed data width of 64-bits for the bedrock interface, this module should have a parameterized data width because the different instantiations may use different widths. For example, the unicore uses uce_fill_width, while in the multicore tile it will be bedrock_data_width_p.

bp_me/test/common/bp_nonsynth_mem.sv
muwyse: It should be l2_data_width because this interface connects to the BedRock side of bp_me_cache_slice. L2 fill width is used on the DMA side of bp_me_cache_slice. This file is only used in testbenches (FE, BE, ME), so it might be possible to keep as a constant, but it depends if the testbenches all use the same width for the interface.

bp_me/src/v/dev/bp_me_cache_slice.sv
muwyse: This is introduced to allow the data width of bp_me_cce_to_cache and bsg_cache instances to be set from the cache slice instance.

muwyse:
> Peripherally related, please "squash and merge" future PRs when merging to me_dev

Will do.

bp_me/test/common/bp_nonsynth_mem.sv
dpetrisko: oh yeah you're right it should be l2_data_width

bp_me/src/v/dev/bp_me_cache_slice.sv
dpetrisko: It does have a parameterized data width -- l2_data_width. If we're enforcing that the l2_data_width == interconnect width, that's fine but it should be done at an aviary level

bp_me/test/common/bp_nonsynth_mem.sv
dpetrisko: This module is basically just a sub for L2+mem, meaning any testbench that uses it does not have a real L2. So it should be fine to use l2_data_width, and have the test config set it correctly. All the testbenches have l1_fill_width == l2_data_width, so I don't believe there should be any conflicts

bp_me/src/v/cce/bp_cce_inst_stall.sv
muwyse: popping data from incoming messages to registers in the ucode CCE. Currently, data is passed directly from incoming message to outgoing message (LCE Resp/Req to Memory Command; and Memory Response to LCE Command).

tools/Makefile.common
muwyse: Sure that makes sense, but the base install path for tools is `rtl/tools/install`. Then, the tools' touchfiles directory is a subdirectory of that. It is redundant to have `rtl/tools/install/touchfiles/tools`. If we had some progs or etc things that were building under rtl they shouldn't install into `rtl/tools/install`. Everything should install into a generic `rtl/install` directory like the SDK.

muwyse: This change fixes the redundant directory. It does not address the fact that a top-level `rtl/install` directory does not exist/is not used (as the SDK does)

bp_me/src/v/dev/bp_me_cache_slice.sv
muwyse: reverted

dpetrisko (approved):
LGTM!

PR #1033, opened 2022-02-14
[WIP] Gearboxing modules

dpetrisko:
## Summary
This PR adds gearboxing modules to the xbar interconnect. This simplifies the interconnect to module requirements and removes some responsibility from the users

## Issue Fixed
Several related to gearboxing, will find them once finished

## Area
tiles, unicore, xbars

It's non-trivial to up/downconvert bedrock stream, so providing plumbing modules is essential for correctness and usability.

Add gearboxing to burst xbar

## Analysis
There should be minimal PPA impact, since this doesn't add extra buffers, instead using PISO/SIPO passthrough

## Verification
ASIC backend and configuration sweep

## Additional Context
There is some problem with Verilator that should be fixed before this goes to mainline

### DIFF
--- a/bp_common/src/v/bsg_serial_in_parallel_out_passthrough.v
+++ b/bp_common/src/v/bsg_serial_in_parallel_out_passthrough.v
@@ -1,0 +1,88 @@
+/**
+ *  bsg_serial_in_parallel_out_passthrough.v
+ */
+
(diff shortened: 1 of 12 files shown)


dpetrisko:
Deprecated

PR #1034, opened 2022-02-14
Implement LCE Fill network in BedRock

muwyse:
## Summary
This PR fixes a deadlock in the BedRock coherence protocol by adding a fourth coherence network (Fill) to carry cache to cache data transfers.

## Area
ME and multicore designs

There is a deadlock in the BedRock cache coherence protocol. The existing protocol violates two invariants:
1. if receiving message A causes sending message B, then A and B are considered to be in different message classes
2. each message class must be carried on its own network

Cache to Cache transfers currently require a Transfer Command (CCE to LCE) followed by a Data Command (LCE to LCE). This sequence violates the invariants above because both messages are part of the Command network/class, when they should be separate classes.

## Analysis
Area will increase due to the introduction of a fourth physical coherence NoC on the multicore tiles and coherent accelerator tiles.

## Verification
Standard CI


dpetrisko (approved):
LGTM!

muwyse:
Wait for #1032 to close prior to finalizing and merging this PR

PR #1043, opened 2022-02-22
Adding UCE as CCE type

dpetrisko:
## Summary
This PR adds UCE as a CCE type (UCE=LCE+CCE), removing the multicore_p parameter.

## Issue Fixed
none

## Area
Aviary

The multicore parameter is redundant with CCE type, and causes confusion for users who wish to use a single core blackparrot with CCE.

zynq-parrot + bsg_fpga probably need updating if they use this parameter. I will defer that until we bump BP, as it's a trivial change.


muwyse (approved):
LGTM!

This will have minor conflicts with #1025 related to setting io_data_width_p param in files and in bp_nonsynth_if verif (expanded and refactored by other PR), but they should be easy to resolve.

PR #1028, opened 2022-02-09
S-mode external interrupt

Yuan-Mao:
## Summary
This PR brings S-mode external interrupt in BP. 

## Area
Mainly bp_be_csr.

## Verification
1. s_external_interrupt.c from [bp-tests](https://github.com/black-parrot-sdk/bp-tests/pull/29)
2. Have tested it on Zedboard running Ethernet connection in Linux.

## Additional Context
For now I set the S-mode interrupt address to 0x30_a000. Not sure if it really makes sense.
Also, I only considered single-core case, so I didn't do updates for modules like bp_tile, bp_core.

PR #1047, opened 2022-02-25
Bump top_dev to dev

dpetrisko:
- S-mode external interrupt (#1028)
- Adding RTC (#1019)
- Adding aviary test package (#1022)
- Adding UCE as CCE type (#1043)

### DIFF
--- a/bp_be/src/include/bp_be_pkg.sv
+++ b/bp_be/src/include/bp_be_pkg.sv
@@ -1,4 +1,6 @@
 
+  `include "bp_be_defines.svh"
+
 package bp_be_pkg;
 
   import bp_common_pkg::*;
--- a/bp_be/src/v/bp_be_calculator/bp_be_calculator_top.sv
+++ b/bp_be/src/v/bp_be_calculator/bp_be_calculator_top.sv
@@ -58,7 +58,8 @@ module bp_be_calculator_top
 
   , input                                           timer_irq_i
   , input                                           software_irq_i
-  , input                                           external_irq_i
+  , input                                           m_external_irq_i
+  , input                                           s_external_irq_i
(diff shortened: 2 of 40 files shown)


dpetrisko:
CI is failing because of: https://github.com/black-parrot-sdk/black-parrot-sdk/pull/33/files, but I tested locally and it passed. So I'm going to update the SDK version in CI before merging. CI runs on other branches will fail until they pull from dev

muwyse (approved):
Two questions about the new test configs:
1. Is there any documentation about how to use these or if someone wants to create a new testbench with a special testing config how to do that? I see that the testbench makefiles substitute in the bp_common_test_pkg.sv file for bp_common_pkg.sv. Is that all that is required since the test package overloads the bp_common_pkg name?
2. And, to confirm, since the test_pkg file overloads the package naming, the standard import bp_common_pkg::* works in the testbench modules?

Otherwise, LGTM! SDK/CI handling sounds reasonable.

dpetrisko:
>Is there any documentation about how to use these or if someone wants to create a new testbench with a special testing config how to do that? I see that the testbench makefiles substitute in the bp_common_test_pkg.sv file for bp_common_pkg.sv. Is that all that is required since the test package overloads the bp_common_pkg name?

Yeah to switch out the available configs, just need to compile bp_common_test_pkg.sv instead of bp_common_pkg.sv. The other use case is for tapeout builds that have different configurations, same deal, you just compile one versus the other. e_bp_default_cfg and e_bp_custom_cfg are always available regardless of the aviary

> And, to confirm, since the test_pkg file overloads the package naming, the standard import bp_common_pkg::* works in the testbench modules?

Yep!

PR #1049, opened 2022-02-28
merge me_dev to dev

muwyse:
This PR includes:
- Addition of LCE Fill network in BedRock to fix protocol deadlock #1034 
- Implementation of BedRock Burst interfaces in LCE and IO CCE #1025 
- Minor fixes in LCE #1051
- ME documentation update #1052 
- minor cleanup for merges
- #1059 
- #1060


ci/accelerator.sh
```diff
@@ -31,22 +31,23 @@ progs=(
     "streaming_accelerator_loopback"
     "streaming_accelerator_vdp"
     "coherent_accelerator_vdp"
-    "streaming_accelerator_loopback"
-    "coherent_accelerator_vdp"
     )
 
+#    "streaming_accelerator_loopback"
```
dpetrisko, line 36: I believe this doesn't test the 4x configs?

bp_me/src/v/network/bp_me_bedrock_wormhole_header_encode_lce_resp.sv
```diff
@@ -0,0 +1,57 @@
...
+ *
+ *  Description:
+ *    Outputs a bedrock wormhole header given an LCE Command input header.
+ *
+ */
+
+`include "bp_common_defines.svh"
+`include "bp_me_defines.svh"
+
+module bp_me_bedrock_wormhole_header_encode_lce_resp
```
dpetrisko, line 13: Can we raise an issue to combine all of these msg_to_wh modules?

bp_top/src/v/bp_sacc_vdp.sv
```diff
@@ -103,6 +91,7 @@ module bp_sacc_vdp
   state_e state_r, state_n;
 
   assign io_resp_v_o = spm_external_v_lo | resp_v_lo;
+  assign io_resp_last_o = io_resp_v_o;
```
dpetrisko, line 94: Add assert that this only supports <=64b requests?

dpetrisko (changes requested):
About 2/3 of the way through reviewing. Some initial comments, but overall looks great. Feel free to address or wait until full review

bp_me/src/v/network/bp_me_cce_to_mem_link_recv.sv
```diff
@@ -1,105 +0,0 @@
-/**
```
dpetrisko, line 1: how would this affect our FPGA builds that rely on stream -> AXIL converters?

bp_me/src/v/lce/bp_lce_fill.sv
```diff
@@ -0,0 +1,389 @@
...
+      ,.ready_o(lce_fill_header_ready_and_o)
+      ,.data_i({lce_fill_has_data_i, lce_fill_header_cast_i})
+      ,.v_o(lce_fill_header_v_li)
+      ,.yumi_i(lce_fill_header_yumi_lo)
+      ,.data_o({lce_fill_has_data, lce_fill_header_cast_li})
+      );
+
+  // LCE fill data buffer
+  // required to prevent deadlock in multicore networks
+  logic [fill_width_p-1:0] lce_fill_data_li;
```
dpetrisko, line 138: For critical path, this requires being input buffered. I'd add an assertion that fill_data_buffer_els_lp > 1 (2 to maintain full throughput)

dpetrisko: (same applies to lce_cmd

bp_me/src/v/lce/bp_lce.sv
```diff
@@ -258,7 +389,14 @@ module bp_lce
   // - LCE Request module is in ready state and free credits exist
   // - timout signal is low, indicating LCE isn't blocked on using data/tag/stat mem
   // - LCE Command module is ready to process commands (raised after initialization complete)
-  assign cache_req_busy_o = cache_req_credits_full_o | timeout | ~cmd_ready_lo | ~req_ready_lo;
+  // - LCE Fill module is ready to process fill messages
+  assign cache_req_busy_o = cache_req_credits_full_o | timeout
```
dpetrisko, line 393: This seems overly pessimistic. Aren't reqs dependent just on req_ready_lo since we're buffering it?

bp_me/src/v/lce/bp_lce.sv
```diff
@@ -86,45 +88,139 @@ module bp_lce
...
+               ,{cmd_lce_resp_header_v_lo, cmd_lce_resp_has_data_lo, cmd_lce_resp_header_lo}})
+      ,.sel_i(fill_lce_resp_header_v_lo)
+      ,.data_o({lce_resp_header_v_o, lce_resp_has_data_o, lce_resp_header_o})
+      );
+
+  // Headers are single beat transactions
+  // Fill module has priority, and sends if able. Otherwise, Command module can send
+  always_comb begin
+    fill_lce_resp_header_ready_and_li = lce_resp_header_ready_and_i;
+    cmd_lce_resp_header_ready_and_li = ~fill_lce_resp_header_v_lo & lce_resp_header_ready_and_i;
```
dpetrisko, line 222: ready_and_i is implicitly an early interface, and this is a late signal, so not ideal

bp_me/src/v/lce/bp_lce_cmd.sv
```diff
@@ -311,293 +471,350 @@ module bp_lce_cmd
...
+          // Set State and Wakeup
+          e_bedrock_cmd_st_wakeup: begin
+            tag_mem_pkt_cast_o.index = lce_cmd_addr_index;
+            tag_mem_pkt_cast_o.way_id = lce_cmd_way_id;
+            tag_mem_pkt_cast_o.state = lce_cmd_header_cast_li.payload.state;
+            tag_mem_pkt_cast_o.tag = '0;
+            tag_mem_pkt_cast_o.opcode = e_cache_tag_mem_set_state;
+            tag_mem_pkt_v_o = lce_cmd_header_v_li;
+
+            critical_tag_sent = tag_mem_pkt_yumi_i & ~critical_tag_sent_r;
```
dpetrisko, line 564: Don't think we need the ~critical_tag_sent_r here, because once we send, we move onto the next state?

bp_me/src/v/lce/bp_lce_fill.sv
```diff
@@ -0,0 +1,389 @@
...
+ *       a new transaction
+ *
+ *    This module does not perform any reads of the cache's stat, tag, or data memories.
+ *
+ */
+
+`include "bp_common_defines.svh"
+`include "bp_me_defines.svh"
+
+module bp_lce_fill
```
dpetrisko, line 21: Seems like a lot of duplicated logic with bp_lce_cmd. Are there protocol implications to combining the modules and just prioritizing messages from fill in the state machine?

bp_me/src/v/lce/bp_lce_req.sv
```diff
@@ -185,31 +167,74 @@ module bp_lce_req
...
+  // there is an available credit
+  assign ready_o = ~is_reset
+                   & (sync_done_i | (lce_mode_i == e_lce_mode_uncached))
+                   & (~cache_req_v_r | (cache_req_v_r & req_sent))
+                   & ~credits_full_o;
+
+  // note: this could create a crazy loop if the cache uses ready_o to determine when to send a
+  // new cache request to the LCE
+  // consume cache request if valid request on input and ready to accept a new request
+  assign cache_req_yumi_o = cache_req_v_i & ready_o;
```
dpetrisko, line 208: Don't understand this comment. The interface is v->yumi, the ready_o is not exposed on this interface

bp_me/src/v/lce/bp_lce_req.sv
```diff
@@ -223,81 +248,76 @@ module bp_lce_req
...
+          end
+          default: begin
+          end
+        endcase
       end
 
-      // Uncached Request
-      e_send_uncached_req: begin
+      // Uncached Request Data
+      // Requests have a single data beat
```
dpetrisko, line 312: // TODO: For all supported caches, requests have a single data beat

bp_top/src/v/bp_tile.sv
```diff
@@ -105,148 +141,228 @@ module bp_tile
...
+       ,.cid_width_p(coh_noc_cid_width_p)
+       ,.pr_hdr_width_p(lce_req_header_width_lp)
+       ,.pr_payload_width_p(lce_req_payload_width_lp)
+       ,.pr_data_width_p(icache_fill_width_p)
+       )
+     lce_req_burst_to_wh
+     (.clk_i(clk_i)
+      ,.reset_i(reset_r)
+
+      ,.pr_hdr_i(lce_req_wh_header_lo[i][0+:($bits(bp_lce_req_wormhole_header_s)-lce_req_wh_pad_width_lp)])
```
dpetrisko, line 196: why not 0+:$bits(lce_req_header)?

bp_top/src/v/bp_tile.sv
```diff
@@ -105,148 +141,228 @@ module bp_tile
...
-         ,.link_o(lce_resp_link_lo[i])
-         );
-    end
+  // minimum cache fill width
+  localparam cache_fill_width_lp = `BSG_MIN(icache_fill_width_p
+                                            ,`BSG_MIN(dcache_fill_width_p
+                                                      , num_cacc_p ? acache_fill_width_p
+                                                                   : dcache_fill_width_p)
+                                           );
+  localparam bedrock_len_width_lp = `BSG_SAFE_CLOG2(`BSG_CDIV((1<<e_bedrock_msg_size_128)*8,cache_fill_width_lp));
```
dpetrisko, line 169: unused?

bp_top/src/v/bp_tile.sv
```diff
@@ -105,148 +141,228 @@ module bp_tile
...
-         ,.packet_i(lce_resp_packet_lo[i])
-         ,.v_i(lce_resp_v_lo[i])
-         ,.ready_o(lce_resp_ready_li[i])
-
-         ,.link_i(lce_resp_link_li[i])
-         ,.link_o(lce_resp_link_lo[i])
-         );
-    end
+  // minimum cache fill width
+  localparam cache_fill_width_lp = `BSG_MIN(icache_fill_width_p
```
dpetrisko, line 164: unused?

bp_top/src/v/bp_tile.sv
dpetrisko: applies for all tiles

bp_me/src/v/lce/bp_lce_fill.sv
dpetrisko: Another thought: lce_fill messages are only used for data, so they can be considered a subset of lce_cmd. One way to handle this a little more simply would be to have in bp_lce.sv a static arbitration between LCE fill and LCE cmd, and send that message to the bp_lce_cmd.sv module. That way, there's only a single port into the bp_lce_cmd module, and we get rid of all the cache_req interface arbitration

dpetrisko (changes requested):
reviewed all modules

bp_me/src/v/network/bp_me_bedrock_wormhole_header_encode_lce_resp.sv
muwyse: #1053

ci/accelerator.sh
muwyse: I disabled after realizing the confusing accelerator config schemes. I think I fixed the configs now, so I will re-enable and test.

bp_me/src/v/network/bp_me_cce_to_mem_link_recv.sv
muwyse: this module and the matching _send module were replaced by bp_me_bedrock_mem_to_link, which is a bi-directional wh-burst converter. Any existing stream -> AXIL conversion should be changed to be burst -> AXIL.

bp_me/src/v/lce/bp_lce_cmd.sv
muwyse: correct. Same in e_bedrock_cmd_data and e_bedrock_cmd_uc_data, actually.

bp_me/src/v/lce/bp_lce.sv
muwyse: I suppose it depends on whether fill_lce_resp_header_v_lo is considered early or late. I would say it is fairly early since it is connected directly to the early valid signal from the fill header buffer inside bp_lce_fill.

bp_top/src/v/bp_tile.sv
muwyse: yep, good catch! leftover from prior iteration of refactoring the network code

bp_top/src/v/bp_tile.sv
muwyse: yep!

bp_top/src/v/bp_tile.sv
muwyse: the burst-wh converter takes a wormhole encoded header as input, not the plain BedRock burst header. It does this because each BedRock network requires a unique wh header encode, but the wh-burst converter itself only needs to know about some widths of fields in the headers, but not the details of the protocol header.

This is somewhat related to your question above about the wh header encoders and #1053 

For now, I think we leave as is and revisit with the open issue.

muwyse: The current setup is similar to what existed before

ci/accelerator.sh
muwyse: tests seem to pass, re-enabled.

bp_top/src/v/bp_sacc_vdp.sv
muwyse: done

bp_me/src/v/lce/bp_lce.sv
muwyse: Yes, I think it can be simplified a little. I will refactor and update. My understanding is this signal basically passes through the cache and is connected to the cache packet ready_o signal used in the pipeline-cache handshake (for both I$ and D$).

Specifically, fill_ready_lo is not needed as fill is ready once reset goes low. Credits do not need to be considered here, they should be considered with the cache_req_yumi decision. Command ready needs to be considered because we want to block any cache operation until after the command module resets the cache's tag and stat memories.

bp_me/src/v/lce/bp_lce_req.sv
```diff
@@ -185,31 +167,74 @@ module bp_lce_req
...
+  assign req_sent = (lce_req_header_v_o & lce_req_header_ready_and_i & ~lce_req_has_data_o)
+                    | (lce_req_data_v_o & lce_req_data_ready_and_i & lce_req_last_o);
+
+  // LCE is ready for a new request if not in reset, mode is uncached or cached with sync done,
+  // either no current valid request or a valid request that is sending last beat, and
+  // there is an available credit
+  assign ready_o = ~is_reset
+                   & (sync_done_i | (lce_mode_i == e_lce_mode_uncached))
+                   & (~cache_req_v_r | (cache_req_v_r & req_sent))
+                   & ~credits_full_o;
```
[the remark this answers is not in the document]
muwyse: This is related to the cache_req_busy_o discussion above. I think ready_o can be reduced to ~is_reset and the sync done / mode lines. The other two lines should be part of the cache_req_yumi_o signal.

Credits are consumed with the cache request packet yumi, so they need to be considered there to apply backpressure on the cache request interface. Removing the credit check on the yumi is a bug, which I've encountered before, where it becomes possible to issue one more request than there are credits.

bp_me/src/v/lce/bp_lce_req.sv
muwyse: The concern was cache_req_v_i being derived from this module's ready_o (through cache_req_busy_o), which would create a combinational loop. This possibility is eliminated by fixing the ready_o/busy_o/cache_req_yumi_o signals as discussed in other review comments.

bp_me/src/v/lce/bp_lce_fill.sv
muwyse: Yes, fill is basically a subset of the command module. I think they need to be separate FSMs, and therefore having them be separate modules is nice for code clarity.

I would be very hesitant in trying to command these two modules without some detailed system analysis to prove it wouldn't deadlock the coherence protocol. Multiplexing command and fill into one channel would violate the protocol deadlock rule that a lower priority network (command) can not stall a higher priority network (fill). We would have to guarantee that the command message would always be able to make forward progress, in all situations.

I agree there is some potentially undesirable/unoptimal arbitration for the cache fill interface and response network, but with static arbitration that favors the fill module, fill will never get blocked by command.

bp_me/src/v/lce/bp_lce.sv
muwyse: Another option is to add a 2-fifo on the outbound response header to reduce the path lengths, with the obvious cost of storage for two message headers (only header because response data only comes from the command module).

bp_top/src/v/bp_tile.sv
dpetrisko: oh this is a super misleading name then. Should be wh_hdr, not pr_hdr

dpetrisko: Fine with keeping open issues for functionality, but we should rename

dpetrisko (changes requested):

bp_me/src/v/network/bp_me_cce_to_mem_link_recv.sv
dpetrisko: We either need to maintain the stream converters or write a multicore->stream wrapper because we can't break zynq-parrot in dev

bp_me/src/v/lce/bp_lce_fill.sv
muwyse: data buffer required for fill and command; assertions added.

bp_me/src/v/lce/bp_lce.sv
muwyse: Okay, I have adjusted some signals. busy_o is raised on timeout or if request module is not ready. Request module only transitions out of its reset state when cache initialization is finished by Command module (now called cache_init_done_*). A single credit is consumed per LCE request sent, and the credit check is performed when sending the LCE request header. Request module is ready if not in reset and in a valid mode (uncached or cached with sync complete).

The credit counting is now consistent with the UCE in that every request requires a single credit (a request may be multiple flits; currently that is 1 or 2 flits).

bp_me/src/v/lce/bp_lce_req.sv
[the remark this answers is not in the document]
muwyse: See thread above. Credits now consumed on request send, but only one credit per request (not one per flit).

bp_me/src/v/lce/bp_lce.sv
muwyse: If the coh_noc_max_credits_p is meant to be a max flits in the network, then both UCE and LCE need further updating, and this will add complexity since it would be unwise to partially send a request

bp_me/src/v/network/bp_me_cce_to_mem_link_recv.sv
muwyse: Got it. Reverting the deletion of these files and related to keep zynq-parrot functional. I will add an issue for updating bp_axi_top so that we can remove and delete these modules.

bp_top/src/v/bp_tile.sv
muwyse: renamed

dpetrisko (approved):
Awesome work! we should run linux through this on FPGA before merging, but code is good to go

PR #1060, opened 2022-03-10
Fixing RTC reset to not depend on lack of toggling

dpetrisko:
Quick fix for RTC reset. bsg_async_ptr_gray cannot toggle during async_reset, but some systems use oscillators which constantly tick. Rather than gate the clock, we synchronize the reset coming in. This is a more robust reset (and fixes some FPGA builds).

@muwyse, we should incorporate this change into your zynqparrot patch

### DIFF
--- a/bp_common/software/py/nbf.py
+++ b/bp_common/software/py/nbf.py
@@ -69,6 +69,9 @@ cfg_reg_cce_id         = 0x0600
 cfg_reg_cce_mode       = 0x0604
 cfg_mem_base_cce_ucode = 0x8000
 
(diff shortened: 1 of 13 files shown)


muwyse (approved):
LGTM!

muwyse:
@dpetrisko need to add bsg_sync_sync  to bp_top/syn/flist.vcs

muwyse (approved):
LGTM

muwyse:
merge squashed to me_dev. Will reach dev with #1049 after CI finishes.

PR #1062, opened 2022-03-11
hotfix cce_to_mem_link_send|recv

muwyse:
Hotfix to two ME files used by multicore FPGA builds.


dpetrisko:
LGTM, let’s keep this branch open a bit for any remaining FPGA and zynq fixes?

muwyse:
Confirmed this works for multicore_1_cfg on vcu128 FPGA design. Build in progress for multiple core config.

bp_top/src/v/bp_io_link_to_lce.sv
```diff
@@ -119,7 +119,9 @@ module bp_io_link_to_lce
 
   logic [cce_id_width_p-1:0] cce_id_lo;
   bp_me_addr_to_cce_id
-   #(.bp_params_p(bp_params_p))
+   #(.bp_params_p(bp_params_p)
```
dpetrisko, line 122: Do we have a use case for non-default sets_p = lce_sets_p? If not, I would prefer to keep it just bp_params_p

dpetrisko: and if so, sets_p should be the default and only overridden if not. Like bp_be_dcache

muwyse: The io_link_to_lce and lce_req module use different values for sets_p.

muwyse: I can make sets_p have a default of lce_sets_p instead of invalid, and then override it only in lce req, but I thought invalid was clearer as it causes the user to think about what it should be instead of blindly accepting the default.

dpetrisko: Oh, I think I just don’t understand this then. For io link, what is the number of sets representing? For tile, what is it? The striping algorithm is presumably by cache line so I can see why it’s block length. How do the number of sets impact it?

muwyse: Actually, the addition of the sets_p param should be reverted and then the width_p param to bsg_hash_bank inside bp_me_addr_to_cce_id should be (lg_num_cce_lp+1) because the module is only computing the bank and doesn't use the index output of bsg_hash_bank. I think bsg_hash_bank needs at least one bit for the index part.

I think I got this mixed together with another sets/lce_sets thing and the use of the index output in the CCE that I was thinking about.

muwyse:
Realized this was on me_dev directly. me_dev has been reverted to me_dev to dev merge. #1067 opened for these changes.

PR #1067, opened 2022-03-17
Small fixes for me_dev to dev merge

muwyse:
This PR contains a few small fixes for #1049, including:
- fix bp_axi_top for zynq-parrot build
- fix LR/SC bug in D$ where LR was not issuing a cache miss on failure if the block existed in a read-only state
- add single-cycle $ fill configs for unicore and multicore_1, and fix bug related to single cycle fills in LCE
- fix LCE command module cache init sequence to clear proper number of cache sets (was previously clearing sets based on max sets in any cache rather than sets in attached cache)

Depends on:
- https://github.com/black-parrot-sdk/bp-tests/pull/27 (and corresponding merge to SDK)


bp_me/src/v/lce/bp_lce_req.sv
```diff
@@ -291,7 +290,7 @@ module bp_lce_req
             // align address to data width and send address of critical beat
             lce_req_header_cast_o.addr = critical_req_addr;
             lce_req_header_cast_o.msg_type.req = e_bedrock_req_rd_miss;
-            lce_req_header_cast_o.payload.lru_way_id = lg_lce_assoc_lp'(cache_req_metadata_r.hit_or_repl_way);
+            lce_req_header_cast_o.payload.lru_way_id = lce_assoc_width_p'(cache_req_metadata_r.hit_or_repl_way);
```
dpetrisko, line 293: This should be lg, no?

muwyse:
Dual core linux boots on FPGA!

bp_me/src/v/lce/bp_lce_req.sv
muwyse: It is. `lce_assoc_width_p` is defined as log2(lce_assoc_p) in aviary

dpetrisko: Yep made the same mistake reading the code

dpetrisko:
Nice! Let’s merge the new tests and add to regression and then 🛳

muwyse:
Sounds good. Interestingly, the unicore load then LR/SC test passes both before and after the LR/SC fix because the UCE always fills cache blocks in the M state to simplify the I$/D$ coherence and reduce the number of requests to L2/memory. This is a perfectly logical optimization if you expect little to no I$/D$ sharing, but also means that the LR failing to issue an upgrade request never happens because the block will always have read/write permissions.

dpetrisko:
Aha makes sense. Multicore 1 fails before and passes after though, right?

muwyse:
No, it also passes before and after because the CCE will provide the block in the E state with read/write permissions since the request was issued by the D$, which does not set the "non-exclusive" bit in the request. Thus, the CCE optimistically grants the block with read/write permissions.

dpetrisko:
Interesting, well I guess we have the tight loop test to actually trigger the bug

PR #1068, opened 2022-03-25
fix LCE fill when block size == fill width

muwyse:
This change fixes a bug in the LCE when cache block size is equal to cache fill width size. In this case, there is only a single fill packet so the first count input to the stream counter used to generate the proper fill index should be zero.


dpetrisko (approved):

dpetrisko:
We should add configs with single cycle fill for both unicore and multicore so we can add it to weird CI

muwyse:
Merging to me_dev_hotfixes.

PR #1079, opened 2022-03-31
Support multicore configs in bp_axi_top

muwyse:
## Summary
This PR enables multicore configs in bp_axi_top, which is particularly useful for Zynq FPGA platforms. There is a minor reorg in the module, but unicore design should be unchanged. Multicore design logic is changed to use new BedRock Burst to AXIL converters.

## Area
bp_axi_top, with minor change to axil converters and addition of new burst-axil converters

Test changes in zynq-parrot before merge and bumping BP RTL

## Analysis
No changes for unicore, enables multicore designs

## Verification
We do not currently run bp_axi_top in the BP RTL CI, so changes need to be verified in zynq-parrot repo.

## Required Before Merge

- [ ] verify unicore functionality unchanged in zynq-parrot
- [ ] verify multicore configs simulate properly in zynq-parrot cosim flow for > 2 cores
- [ ] verify bp_multicore_1_cfg functional in zynq-parrot on FPGA

### DIFF
--- a/bp_common/src/include/bp_common_aviary_pkgdef.svh
+++ b/bp_common/src/include/bp_common_aviary_pkgdef.svh
@@ -195,7 +195,6 @@
       ,ic_y_dim             : 1
       ,icache_coherent      : 1
       ,l2_amo_support       : '0
-      ,l2_banks             : 1
       ,default : "inv"
       };
   `bp_aviary_derive_cfg(bp_multicore_1_cfg_p
--- a/bp_me/src/v/network/bp_me_axil_client.sv
+++ b/bp_me/src/v/network/bp_me_axil_client.sv
@@ -18,18 +18,20 @@ module bp_me_axil_client
    input                                        clk_i
    , input                                      reset_i
 
-   //==================== BP-LITE SIGNALS ======================
+   //==================== BP-STREAM SIGNALS ====================
    , input [lce_id_width_p-1:0]                 lce_id_i
    , input [did_width_p-1:0]                    did_i
 
    , output logic [mem_header_width_lp-1:0]     io_cmd_header_o
    , output logic [axil_data_width_p-1:0]       io_cmd_data_o
    , output logic                               io_cmd_v_o
+   , output logic                               io_cmd_last_o
    , input                                      io_cmd_ready_and_i
 
    , input [mem_header_width_lp-1:0]            io_resp_header_i
    , input [axil_data_width_p-1:0]              io_resp_data_i
    , input                                      io_resp_v_i
+   , input                                      io_resp_last_i
    , output logic                               io_resp_yumi_o
 
    //====================== AXI-4 LITE =========================
@@ -63,7 +65,8 @@ module bp_me_axil_client
    , input                                      s_axil_rready_i
   );
 
-  wire unused = &{s_axil_awprot_i, s_axil_arprot_i};
+  wire unused = &{s_axil_awprot_i, s_axil_arprot_i, io_resp_last_i};
+  assign io_cmd_last_o = io_cmd_v_o;
 
   // declaring i/o command and response struct type and size
   `declare_bp_bedrock_mem_if(paddr_width_p, did_width_p, lce_id_width_p, lce_assoc_p);
@@ -139,13 +142,12 @@ module bp_me_axil_client
           begin
             s_axil_awready_o = io_cmd_ready_and_i;
             s_axil_wready_o = io_cmd_ready_and_i;
-
             io_cmd_header_cast_o.addr                = s_axil_awaddr_r;
             io_cmd_header_cast_o.msg_type            = e_bedrock_mem_uc_wr;
             io_cmd_header_cast_o.size                = wsize;
             io_cmd_v_o                               = (s_axil_awvalid_i & s_axil_wvalid_i);
 
-            // Send          
+            // Send
             state_n = (s_axil_awvalid_i & ~s_axil_wvalid_i)
                       ? e_send_write_data
                       : (~s_axil_awvalid_i & s_axil_wvalid_i)
@@ -170,7 +172,7 @@ module bp_me_axil_client
         e_send_write_data:
           begin
             s_axil_wready_o = io_cmd_ready_and_i;
-           
+
             io_cmd_header_cast_o.addr                = s_axil_awaddr_r;
             io_cmd_header_cast_o.msg_type            = e_bedrock_mem_uc_wr;
             io_cmd_header_cast_o.size                = wsize;
--- a/bp_me/src/v/network/bp_me_axil_to_burst.sv
+++ b/bp_me/src/v/network/bp_me_axil_to_burst.sv
@@ -1,0 +1,258 @@
+/*
+ * Name:
+ *  bp_me_axil_to_burst.sv
+ *
+ * Description:
+ *  This module converts an AXI4-Lite Subordinate interface to
+ *  BP BedRock Burst Command Out / Response In. This module is a
+ *  minimal AXI4-Lite Subordinate, and assumes IO responses are
+ *  returned in order with respect to IO commands issued.
+ */
+
+`include "bp_common_defines.svh"
+`include "bp_me_defines.svh"
+
+module bp_me_axil_to_burst
+ import bp_common_pkg::*;
+ import bp_me_pkg::*;
+ #(parameter bp_params_e bp_params_p = e_bp_default_cfg
+  `declare_bp_proc_params(bp_params_p)
+  `declare_bp_bedrock_mem_if_widths(paddr_width_p, did_width_p, lce_id_width_p, lce_assoc_p)
+  , parameter io_data_width_p = (cce_type_p == e_cce_uce) ? uce_fill_width_p : bedrock_data_width_p
+
+  , parameter axil_data_width_p = 32
+  , parameter axil_addr_width_p = 32
+  )
+
+  (//==================== GLOBAL SIGNALS =======================
+   input                                        clk_i
+   , input                                      reset_i
+
+   //==================== BP-BURST SIGNALS =====================
+   , input [lce_id_width_p-1:0]                 lce_id_i
+   , input [did_width_p-1:0]                    did_i
+
+   , output logic [mem_header_width_lp-1:0]     io_cmd_header_o
+   , output logic                               io_cmd_header_v_o
+   , output logic                               io_cmd_has_data_o
+   , input                                      io_cmd_header_ready_and_i
+   , output logic [io_data_width_p-1:0]         io_cmd_data_o
+   , output logic                               io_cmd_data_v_o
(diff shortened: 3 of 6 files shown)


bp_me/src/v/network/bp_me_axil_client.sv
```diff
@@ -135,17 +138,20 @@ module bp_me_axil_client
 
       unique casez (state_r)
         // Can send writes immediately if both aw/w are available, else read/write with 1 cycle delay
+        // NOTE: wsize and s_axil_awaddr_r only become valid when r&v axi signals high
```
dpetrisko, line 141: I'm not sure I understand the concern. When axi valid is high, the lines are valid and guaranteed to remain high by the spec. When the r&v occurs, that's when the transaction actually happens

bp_me/src/v/network/bp_me_axil_to_burst.sv
```diff
@@ -0,0 +1,257 @@
...
+    $display("## bp_me_axil_to_burst: instantiating with axil_data_width_p=%d, axil_addr_width_p=%d (%m)\n",axil_data_width_p,axil_addr_width_p);
+  end
+
+  always_ff @(negedge clk_i) begin
+    assert (reset_i !== '0 || s_axil_awprot_i == 3'b000)
+      else $error("AXI4-LITE access permission mode is not supported.");
+    assert (reset_i !== '0 || ~s_axil_wvalid_i || (s_axil_wstrb_i inside {'h1, 'h3, 'hf, 'hff}))
+      else $error("Invalid write strobe encountered");
+    assert (reset_i !== '0 || ~io_resp_header_v_i
+            || io_resp_header_cast_i.size < e_bedrock_msg_size_64
```
dpetrisko, line 244: msg_size_8?

dpetrisko: this should also just check for single beat transaction

bp_me/src/v/network/bp_me_burst_to_axil.sv
```diff
@@ -0,0 +1,284 @@
...
+    $display("## bp_me_burst_to_axil: instantiating with axil_data_width_p=%d, axil_addr_width_p=%d (%m)\n",axil_data_width_p,axil_addr_width_p);
+  end
+
+  always_ff @(negedge clk_i) begin
+    // give a warning if the client device has an error response
+    assert (reset_i !== '0 || ~m_axil_rvalid_i || m_axil_rresp_i == '0)
+      else $error("Client device has an error response to reads");
+    assert (reset_i !== '0 || ~m_axil_bvalid_i || m_axil_bresp_i == '0)
+      else $error("Client device has an error response to writes");
+    assert (reset_i !== '0 || ~io_cmd_header_v_i
```
dpetrisko, line 272: same comments as other

bp_top/src/v/bp_axi_top.sv
```diff
@@ -122,24 +122,29 @@ module bp_axi_top
...
 
   `declare_bsg_cache_dma_pkt_s(daddr_width_p);
-  bsg_cache_dma_pkt_s [l2_banks_p-1:0] dma_pkt_lo;
-  logic [l2_banks_p-1:0] dma_pkt_v_lo, dma_pkt_yumi_li;
-  logic [l2_banks_p-1:0][l2_fill_width_p-1:0] dma_data_lo;
-  logic [l2_banks_p-1:0] dma_data_v_lo, dma_data_yumi_li;
-  logic [l2_banks_p-1:0][l2_fill_width_p-1:0] dma_data_li;
-  logic [l2_banks_p-1:0] dma_data_v_li, dma_data_ready_and_lo;
+  bsg_cache_dma_pkt_s [num_cce_p-1:0][l2_banks_p-1:0] dma_pkt_o;
+  logic [num_cce_p-1:0][l2_banks_p-1:0] dma_pkt_v_o, dma_pkt_ready_and_i;
```
dpetrisko, line 131: keep this as _li / _lo, per convention

bp_top/src/v/bp_axi_top.sv
```diff
@@ -274,29 +354,35 @@ module bp_axi_top
...
+         ,.mem_data_v_o(io_resp_data_v_lo)
+         ,.mem_data_ready_and_i(io_resp_data_ready_and_li)
+         ,.mem_last_o(io_resp_last_lo)
+
+         ,.link_o(send_cmd_link_lo)
+         ,.link_i(send_resp_link_li)
          );
 
-      bp_me_cce_to_mem_link_recv
+      bp_me_bedrock_mem_to_link
```
dpetrisko, line 379: May have been an easier route to just tunnel wormhole to axi-lite. But it's already written so c'est la vie

bp_top/src/v/bp_axi_top.sv
dpetrisko: suggest axi_dma to avoid name conflict

bp_top/src/v/bp_axi_top.sv
```diff
@@ -122,24 +122,29 @@ module bp_axi_top
    , input [1:0]                               m_axi_rresp_i
```
dpetrisko, line 122: This makes me think of our friend's comment that having a wrapper for unicore/multicore would be useful, to avoid rewriting this logic in multiple systems. This would basically be the wrapper currently used in the cache testbench. What do you think about that?

bp_me/src/v/network/bp_me_burst_to_axil.sv
```diff
@@ -0,0 +1,284 @@
...
+ *  to an AXI4-Lite Manager interface. This module is a minimal AXI4-Lite
+ *  Manager, supporting one request at a time. IO write commands are
+ *  restricted to single data beat writes.
+ *
+ */
+
+`include "bp_common_defines.svh"
+`include "bp_me_defines.svh"
+
+module bp_me_burst_to_axil
```
dpetrisko, line 16: nice, these modules seem to follow the handshakes appropriately. Of course, the single beat write restriction makes things simpler

bp_me/src/v/network/bp_me_axil_client.sv
muwyse: Yes, the input lines remain high, but the output io_cmd doesn't actually become valid until the s_axil_awaddr_r signal is valid. Due to how the register enable is wired above, this actually doesn't happen until the io_cmd handshake. So, the io command is being output valid before the contents of the header are actually valid. This causes problems if the consumer examines the header before raising ready_and_i in a following cycle (allowed by BedRock).

bp_me/src/v/network/bp_me_axil_to_burst.sv
muwyse: ah, yes. Mind was probably thinking bits not bytes.

bp_me/src/v/network/bp_me_burst_to_axil.sv
muwyse: Yep. They are not meant to be high performance or fully support all possible transactions (BedRock or AXI). Functionally, they should be the same as bp_me_axil_client|master

bp_top/src/v/bp_axi_top.sv
muwyse: Not sure I follow. The cache testbench wrapper has cache packet on one side and BedRock on the other, both of which are BP specific interfaces. I think this is a very nice top-level wrapper for BP since it is AXIL and AXI only, so should be easy to connect to other systems.

dpetrisko: Sorry, suggestion is to encompass the internals of this block in bp_top.sv. So we have:
testbench->wrapper->bp_top
axi_top->bp_top
bsg_fpga->bp_top

We just have this if (unicore) bp_unicore else bp_multicore code in several places

bp_top/src/v/bp_axi_top.sv
muwyse: Perhaps. It still requires extracting the BedRock message information from the WH network to properly transduce to/from AXI. I think the overhead we should look for here is any gearboxing on the data channels. Ideally, WH flit width == bedrock data width == axi data width.

bp_top/src/v/bp_axi_top.sv
dpetrisko: I won't block this on this comment, but probably next time week change this code it's a good idea

bp_common/src/include/bp_common_aviary_pkgdef.svh
```diff
@@ -393,7 +393,6 @@
       ,cc_y_dim: 3
       ,num_cce : 12
       ,num_lce : 24
-      ,l2_banks: 1
```
dpetrisko, line 396: This breaks certain parts of the infrastructure with the current implementation of bsg_round_robin:   if (dram_type_p == "axi" && (num_cce_p*l2_banks_p) > 16)

dpetrisko (approved):
I would revert the bank change to avoid surprises, but other than that GTG

dpetrisko:
Let's merge into dev directly though, I'm trying to put together a release and we don't need another CI run

bp_me/src/v/network/bp_me_axil_client.sv
dpetrisko: No, I still don't understand. They shouldn't, or it wouldn't be a ready_and_i handshake

bp_common/src/include/bp_common_aviary_pkgdef.svh
dpetrisko: Oh I see the default is 1 for multicore. What is that?

bp_me/src/v/network/bp_me_axil_client.sv
dpetrisko: Oh, I see. That's incorrect then. The flop enables should be

is_ready & s_axil_wvalid_i
is_ready & s_axil_awvalid_i

dpetrisko (changes requested):
Handshake fix requested

bp_common/src/include/bp_common_aviary_pkgdef.svh
muwyse: I don't know why the default for multicore is 1 L2 bank. I don't recall changing that param, but good to know about the round robin supporting up to 16 only.

muwyse: Changed to use default of 2 L2 banks up to 8 cores, then 1 L2 bank for 12 and 16 cores. The multicore CI regression script passed locally for me.

dpetrisko:
With the issue raised for the handshake fix, is this good to go?

muwyse:
> With the issue raised for the handshake fix, is this good to go?

I have not yet confirmed this is working on ZP as expected. I have been splitting out SW/HW changes in ZP to test, and have only confirmed the SDK bump for ZP seems to work as expected.

PR #1084, opened 2022-04-16
Break out tools into its own repo

dpetrisko:
## Summary
This PR removes the tools directory from the RTL repo and moves it to [this repo](https://github.com/black-parrot/black-parrot-tools).

## Area
Tool makefiles and build scripts

Many tool submodules are unused for many users. Some of these submodules are large and contribute to a large cloning time. Finally, these tools are unlikely to need to change along with a specific version of BP, and regressions are unexpected.  For these reasons, it's desirable to have a second parallel repo for the RTL tools.

## Verification
Clean builds of libs, tools and sdk.

## Additional work needed
Current users of BP will need to clone the new tools/ directory and put it in the same directory as the RTL. This also applies to the black-parrot-sim environment, and possibly FPGA environments, if they rely on dramsim3 or dromajo.


muwyse (approved):
LGTM

I will look into ME regression failure tomorrow.

dpetrisko:
Thanks!

dpetrisko:
Merging with raised https://github.com/black-parrot/black-parrot/issues/1086

PR #1091, opened 2022-05-07
Crossbar Control Locking Bug

sripathi-muralitharan:
This PR fixes a bug in the locking crossbar control module.

## Summary
The locking crossbar control module locks an input port for an output port if the input port is granted access to the output port until the input port completes streaming data through to that output port. The valid signal for every output port currently contains information if the output port has valid requestors from any of the input ports but does not contain any information about which input port has locked arbitration. It is possible that this creates a situation where input ports that haven't locked the arbiter can access the output port. The PR attempts to correct this.

The fix is to AND the output request mask with the information about which input has locked the arbiter.

## Issue Fixed
There is no active issue that this PR addresses

## Area
bp_common/src/v/bsg_crossbar_control_locking_o_by_i.sv

This change is required for current functionality.

## Analysis
By using the flopped grant signal (i.e., req_mask_r), we ensure the RR arbiter does not come into the critical path. This needs to be confirmed with a synthesis and/or APR run.

## Verification
Only the hello_world test was run on the BP tile bench in the manycore-BP tapeout directory. Regression testing might be required.

### DIFF
--- a/bp_common/src/v/bsg_crossbar_control_locking_o_by_i.v
+++ b/bp_common/src/v/bsg_crossbar_control_locking_o_by_i.v
@@ -62,6 +62,7 @@ module bsg_crossbar_control_locking_o_by_i
   for (genvar i = 0 ; i < o_els_p; i++) begin: rr
 
     logic [i_els_p-1:0] not_req_mask_r, req_mask_r;
+    logic [i_els_p-1:0] reqs_li;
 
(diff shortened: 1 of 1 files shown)


dpetrisko:
ah yes, this was fixed here instead: https://github.com/bespoke-silicon-group/basejump_stl/pull/536/files. Thanks!

Issue #1080, opened 2022-03-31
Handshakes are broken in VDP demo


dpetrisko:
https://github.com/black-parrot/black-parrot/blob/c07c2d50dbd303760400f00b49b68e3a8724fdcb/bp_top/src/v/bp_cacc_vdp.sv#L76

io_cmd_ready_and_i and io_resp_ready_and_o are not used properly and data will drop under congestion. We should verify all of the accelerators

dpetrisko:
@muwyse Not sure when this cropped up, could have always existed or came in with the fill network changes?

dpetrisko:
Fixed in #1093 #1108

PR #1093, opened 2022-05-12
Fix for broken handshakes in coherent vdp accelerator

vsteltzer:
## Summary
The handshakes in the coherent vdp accelerator were not properly implemented. This proposes a potential fix for these handshakes.

## Issue Fixed
[Issue](https://github.com/black-parrot/black-parrot/issues/1080#issue-1187658361)

## Area
/black-parrot/bp_top/src/v/bp_cacc_vdp.sv

The handshakes are necessary for a correct data transmission to and from the accelerator.

If these changes are correct they could potentially be also implemented in the streaming accelerator version. 

## Verification
The coherent vdp test from the sdk got used to test the changes. Maybe other verification steps necessary.


dpetrisko:
Hi, thanks for the PR. An FSM + datapath diagram of the module before and after the changes would make this much easier to review

vsteltzer:
I will try to supply that in the next days.

dpetrisko:
Hi thanks, the diagrams are excellent. I think there is excess complication because of waiting for I/O resp to be ready. I would instead decouple the cmd / resp sequences by adding an output resp fifo with a depth of the D$ pipeline. Then, simply tie io_cmd_ready_and_o to io_resp_ready_and_i. That will ensure that you never overflow the fifo, while avoiding adding (as many? any?) extra states to the FSM

vsteltzer:
Would the implementation be similar to the one in the canonical example you once referenced? There is a [fifo for the cmd line](https://github.com/black-parrot/black-parrot/blob/2e0013f3e07943c479a39bba12ba25b004311042/bp_me/src/v/dev/bp_me_bedrock_register.sv#L83) implemented. Can i use this as a reference?

vsteltzer:
Does a parameter exist for the D$ Pipeline depth?

dpetrisko:
In fact, since there is only 1 outstanding request to the D$ we can simply have a two element fifo for the output

vsteltzer:
I integrated now a Fifo, not exactly as you suggested, but i hope it solves the problem well enough.
Here the datapath:
image

bp_top/src/v/bp_cacc_vdp.sv
```diff
@@ -115,6 +115,9 @@ module bp_cacc_vdp
   logic [cache_stat_info_width_lp-1:0] stat_mem_o;
   bp_cache_req_metadata_s cache_req_metadata_o;
 
+  // Output data into two element fifo
+  logic [mem_header_width_lp-1:0]      io_resp_header_fifo_i;
```
dpetrisko, line 119: unused?

bp_top/src/v/bp_cacc_vdp.sv
```diff
@@ -250,10 +253,9 @@ module bp_cacc_vdp
   // CCE-IO interface is used for uncached requests-read/write memory mapped CSR
   `declare_bp_bedrock_mem_if(paddr_width_p, did_width_p, lce_id_width_p, lce_assoc_p);
   `bp_cast_i(bp_bedrock_mem_header_s, io_cmd_header);
+  `bp_cast_o(bp_bedrock_mem_header_s, io_cmd_header_fifo);
```
dpetrisko, line 256: unused?

bp_top/src/v/bp_cacc_vdp.sv
```diff
@@ -273,21 +275,63 @@ module bp_cacc_vdp
...
+     ,.v_i(io_cmd_v_i)
+     ,.ready_o(io_cmd_ready_and_o)
+
+     ,.data_o({fifo_data_li, fifo_header_li})
+     ,.v_o(fifo_v_li)
+     ,.yumi_i(fifo_yumi_li)
+     );     
+
+  // D-FF signals
+  logic v_n; // D
```
dpetrisko, line 315: I think this isn't necessary, as long as you used fifo_v_li to start the FSM, and only dequeue when a request is finished

dpetrisko:
Yeah that fifo scheme is fine, left comments on some implementation points

vsteltzer:
Thanks for the review! I integrated your suggestions.

PR #1094, opened 2022-05-19
merge me_dev to dev

muwyse:
Small update with an ME testing fix and a BedRock coherence protocol specification update.


dpetrisko:
Did that bedrock pdf move somewhere else?

muwyse:
> Did that bedrock pdf move somewhere else?

No. The diff does make it look like it is being deleted / moved, but it is definitely present on the me_dev branch and when I do a me_dev to dev merge locally, it is still there on dev post merge...so I think just weird display here.

dpetrisko:
Weird, works for me though. Merge away with CI

muwyse:
Had to revert the bp_top testbench Makefile.cfg change. The NCPUS variable is overloaded in its use as both number of processor cores and number of CCEs (the latter being more important since it is passed to the NBF generator and used to properly configure the CCEs).

PR #1098, opened 2022-07-05
 Support 32-bit instructions aligned to 16 bits (IALIGN=16), take two

WasabiFan:
## Summary

This PR is the new revision of #986. It introduces support for executing 32-bit instructions aligned to an arbitrary 16-bit boundary. It does not widen or otherwise alter the instruction cache output; rather, it performs successive fetches and synthesizes the results via a new "realigner" component.

The key changes from the previous PR are as follows:
- Branch prediction metadata is now indexed by the aligned address of the second (upper) half of an instruction rather than the first (lower) half
  - As a result, I've removed a bunch of state previously used to carry over this information
- Modified and narrowed the I/O interface between the PC gen module and the overall frontend

  Note: Dan and I had previously discussed moving the realigner out of `pc_gen`. I took a deep dive on this approach and found that the interface between the realigner and `pc_gen`/`fe_top` would not be improved as we had hoped; the interaction is somewhat involved and there are a few signals that need to be routed regardless of where the module is housed. Moving it out of pc_gen increased the number of needed signals by one (two, now that I've added one more for ITLB fill handling). I am fairly happy with the middle ground I've arrived at but am happy to discuss if there are concerns.
- Support for exceptions and virtual memory is now fleshed out. This includes supporting instructions which span two pages. See the new tests in https://github.com/black-parrot-sdk/bp-tests/pull/32 for a demonstration.
  - Exception issue packets now include a new bit, indicating whether the exception is specific to the upper half of the instruction
  - Page table walks, ITLB fills and I$ fills are now performed using the appropriate half of the instruction
  - `{m,s}tval` now correctly indicate the portion of the instruction that failed (either the same as the faulting pc or 2 bytes beyond it)
  - `{m,s}epc` now reflects the true pc alignment and doesn't have its second-to-lowest bit masked

## Issue Fixed

IALIGN is now 16

## Area

Mostly frontend, but propagation of half-instruction faults includes backend changes too.

Why is this change required?

Test programs: https://github.com/black-parrot-sdk/bp-tests/pull/32

I also intend to do some updating and porting work for #988 but haven't done so yet.

## Analysis

Test suite does its best to cover relevant cases. Test plan from #986 still applies. But of course, it could have any unknown side-effects.

## Verification

Test suite does its best to cover relevant cases. Test plan from #986 still applies. See tests in https://github.com/black-parrot-sdk/bp-tests/pull/32.

## Additional Context

Previous version of this PR is: #986

### DIFF
--- a/bp_be/src/include/bp_be_ctl_pkgdef.svh
+++ b/bp_be/src/include/bp_be_ctl_pkgdef.svh
@@ -252,7 +252,7 @@
     logic cmd_full;
 
     // Metadata for instruction exceptions (instr_page_fault,instr_access_fault, itlb_miss, icache_miss)
-    logic instr_upper_not_lower_half;
+    logic upper_not_lower_half;
   }  bp_be_exception_s;
 
   typedef struct packed
--- a/bp_be/src/include/bp_be_ctl_pkgdef.svh
+++ b/bp_be/src/include/bp_be_ctl_pkgdef.svh
@@ -251,9 +251,6 @@
     logic dtlb_fill;
     logic _interrupt;
     logic cmd_full;
-
-    // Metadata for instruction exceptions (instr_page_fault,instr_access_fault, itlb_miss, icache_miss)
-    logic instr_partial_v;
   }  bp_be_exception_s;
 
   typedef struct packed
--- a/bp_be/src/include/bp_be_ctl_pkgdef.svh
+++ b/bp_be/src/include/bp_be_ctl_pkgdef.svh
@@ -253,7 +253,7 @@
     logic cmd_full;
 
     // Metadata for instruction exceptions (instr_page_fault,instr_access_fault, itlb_miss, icache_miss)
-    logic upper_not_lower_half;
+    logic instr_partial_v;
   }  bp_be_exception_s;
 
   typedef struct packed
--- a/bp_be/src/include/bp_be_ctl_pkgdef.svh
+++ b/bp_be/src/include/bp_be_ctl_pkgdef.svh
@@ -188,8 +188,6 @@
 
   typedef struct packed
   {
-    logic                             v;
-
     logic                             pipe_ctl_v;
     logic                             pipe_int_v;
     logic                             pipe_mem_early_v;
--- a/bp_be/src/include/bp_be_defines.svh
+++ b/bp_be/src/include/bp_be_defines.svh
@@ -142,7 +142,7 @@
       logic [rv64_priv_width_gp-1:0]  priv_n;                                                      \
       logic                           translation_en_n;                                            \
       logic                           exception;                                                   \
-      logic                           exception_instr_upper_not_lower_half;                        \
+      logic                           exception_instr_partial_v;                                   \
       logic                           _interrupt;                                                  \
       logic                           unfreeze;                                                    \
       logic                           eret;                                                        \
@@ -181,7 +181,7 @@
       logic                          instr_miss_v;                                                 \
       logic                          load_miss_v;                                                  \
       logic                          store_miss_v;                                                 \
-      logic                          instr_upper_not_lower_half;                                   \
+      logic                          instr_partial_v;                                              \
       logic [vaddr_width_mp-1:0]     vaddr;                                                        \
     }  bp_be_ptw_miss_pkt_s;                                                                       \
                                                                                                    \
@@ -193,7 +193,7 @@
       logic instr_page_fault_v;                                                                    \
       logic load_page_fault_v;                                                                     \
       logic store_page_fault_v;                                                                    \
-      logic instr_upper_not_lower_half;                                                            \
+      logic instr_partial_v;                                                                       \
       logic [vaddr_width_mp-1:0] vaddr;                                                            \
       bp_be_pte_leaf_s entry;                                                                      \
     }  bp_be_ptw_fill_pkt_s;                                                                       \
@@ -281,8 +281,8 @@
   `define bp_be_decode_info_width \
     (rv64_priv_width_gp+8)
 
-  `define bp_be_instr_half_address(base_pc_mp, upper_not_lower_half_mp) \
-    (base_pc_mp + (upper_not_lower_half_mp ? 2 : 0))
+  `define bp_be_instr_half_address(base_pc_mp, instr_partial_v_mp) \
+    (base_pc_mp + (instr_partial_v_mp ? 2 : 0))
 
 `endif
 
--- a/bp_be/src/include/bp_be_defines.svh
+++ b/bp_be/src/include/bp_be_defines.svh
@@ -39,6 +39,7 @@
       logic                                    v;                                                  \
       logic                                    queue_v;                                            \
       logic [vaddr_width_mp-1:0]               pc;                                                 \
+      logic                                    instr_v;                                            \
       rv64_instr_s                             instr;                                              \
       bp_be_decode_s                           decode;                                             \
                                                                                                    \
@@ -232,7 +233,7 @@
     (2                                                                                             \
      + vaddr_width_mp                                                                              \
      + rv64_instr_width_gp                                                                         \
-     + 3                                                                                           \
+     + 4                                                                                           \
      + 3 * dpath_width_gp                                                                          \
      + $bits(bp_be_decode_s)                                                                       \
      + $bits(bp_be_exception_s)                                                                    \
--- a/bp_be/src/include/bp_be_defines.svh
+++ b/bp_be/src/include/bp_be_defines.svh
@@ -49,6 +49,7 @@
       logic [dpath_width_gp-1:0]               rs2;                                                \
       logic                                    rs3_fp_v;                                           \
       logic [dpath_width_gp-1:0]               imm;                                                \
+      logic                                    instr_partial_v;                                    \
       bp_be_exception_s                        exception;                                          \
       bp_be_special_s                          special;                                            \
      } bp_be_dispatch_pkt_s;                                                                       \
@@ -113,6 +114,7 @@
       logic [vaddr_width_p-1:0]  vaddr;                                                            \
       logic [dpath_width_gp-1:0] data;                                                             \
       rv64_instr_s               instr;                                                            \
+      logic                      instr_partial_v;                                                  \
       bp_be_exception_s          exception;                                                        \
       bp_be_special_s            special;                                                          \
     }  bp_be_retire_pkt_s;                                                                         \
@@ -236,6 +238,7 @@
      + 4                                                                                           \
      + 3 * dpath_width_gp                                                                          \
      + $bits(bp_be_decode_s)                                                                       \
+     + 1                                                                                           \
      + $bits(bp_be_exception_s)                                                                    \
      + $bits(bp_be_special_s)                                                                      \
      )
@@ -250,7 +253,7 @@
     (3 + vaddr_width_mp)
 
   `define bp_be_retire_pkt_width(vaddr_width_mp) \
-    (3 + dpath_width_gp + 2*vaddr_width_mp + instr_width_gp + $bits(bp_be_exception_s) + $bits(bp_be_special_s))
+    (3 + dpath_width_gp + 2*vaddr_width_mp + instr_width_gp + 1 + $bits(bp_be_exception_s) + $bits(bp_be_special_s))
 
   `define bp_be_pte_leaf_width(paddr_width_mp) \
     (paddr_width_mp - page_offset_width_gp + 7)
--- a/bp_be/src/v/bp_be_calculator/bp_be_pipe_mem.sv
+++ b/bp_be/src/v/bp_be_calculator/bp_be_pipe_mem.sv
@@ -250,7 +250,7 @@ module bp_be_pipe_mem
     '{instr_miss_v  : commit_pkt.itlb_miss
       ,store_miss_v : commit_pkt.dtlb_store_miss
       ,load_miss_v  : commit_pkt.dtlb_load_miss
-      ,instr_upper_not_lower_half: commit_pkt.exception_instr_upper_not_lower_half
+      ,instr_partial_v: commit_pkt.exception_instr_partial_v
       ,vaddr        : commit_pkt.vaddr
       ,mstatus_mxr  : trans_info.mstatus_mxr
       ,mstatus_sum  : trans_info.mstatus_sum
--- a/bp_be/src/v/bp_be_checker/bp_be_director.sv
+++ b/bp_be/src/v/bp_be_checker/bp_be_director.sv
@@ -148,7 +148,7 @@ module bp_be_director
       if (commit_pkt_cast_i.unfreeze)
         begin
           fe_cmd_li.opcode = e_op_state_reset;
-          fe_cmd_li.vaddr  = npc_r;
+          fe_cmd_li.pc     = npc_r;
 
           fe_cmd_pc_redirect_operands.priv           = commit_pkt_cast_i.priv_n;
           fe_cmd_pc_redirect_operands.translation_en = commit_pkt_cast_i.translation_en_n;
@@ -159,7 +159,7 @@ module bp_be_director
       else if (commit_pkt_cast_i.itlb_fill_v)
         begin
           fe_cmd_li.opcode                               = e_op_itlb_fill_response;
-          fe_cmd_li.vaddr                                = commit_pkt_cast_i.npc;
+          fe_cmd_li.pc                                   = commit_pkt_cast_i.npc;
           fe_cmd_li.operands.itlb_fill_response.pte_leaf = commit_pkt_cast_i.pte_leaf;
           fe_cmd_li.operands.itlb_fill_response.partial_instr   = commit_pkt_cast_i.instr;
           fe_cmd_li.operands.itlb_fill_response.partial_instr_v = commit_pkt_cast_i.exception_instr_upper_not_lower_half;
@@ -170,14 +170,14 @@ module bp_be_director
       else if (commit_pkt_cast_i.sfence)
         begin
           fe_cmd_li.opcode = e_op_itlb_fence;
-          fe_cmd_li.vaddr  = commit_pkt_cast_i.npc;
+          fe_cmd_li.pc     = commit_pkt_cast_i.npc;
 
           fe_cmd_v_li = fe_cmd_ready_lo;
         end
       else if (commit_pkt_cast_i.csrw)
         begin
           fe_cmd_li.opcode                            = e_op_pc_redirection;
-          fe_cmd_li.vaddr                             = commit_pkt_cast_i.npc;
+          fe_cmd_li.pc                                = commit_pkt_cast_i.npc;
           fe_cmd_pc_redirect_operands.subopcode       = e_subop_translation_switch;
           fe_cmd_pc_redirect_operands.translation_en  = commit_pkt_cast_i.translation_en_n;
           fe_cmd_li.operands.pc_redirect_operands     = fe_cmd_pc_redirect_operands;
@@ -187,21 +187,21 @@ module bp_be_director
       else if (commit_pkt_cast_i.wfi)
         begin
           fe_cmd_li.opcode = e_op_wait;
-          fe_cmd_li.vaddr  = commit_pkt_cast_i.npc;
+          fe_cmd_li.pc     = commit_pkt_cast_i.npc;
 
           fe_cmd_v_li = fe_cmd_ready_lo;
         end
       else if (commit_pkt_cast_i.fencei)
         begin
           fe_cmd_li.opcode = e_op_icache_fence;
-          fe_cmd_li.vaddr  = commit_pkt_cast_i.npc;
+          fe_cmd_li.pc     = commit_pkt_cast_i.npc;
 
           fe_cmd_v_li = fe_cmd_ready_lo;
         end
       else if (commit_pkt_cast_i.icache_miss)
         begin
           fe_cmd_li.opcode = e_op_icache_fill_response;
-          fe_cmd_li.vaddr  = commit_pkt_cast_i.npc;
+          fe_cmd_li.pc     = commit_pkt_cast_i.npc;
           fe_cmd_li.operands.icache_fill_response.partial_instr   = commit_pkt_cast_i.instr;
           fe_cmd_li.operands.icache_fill_response.partial_instr_v = commit_pkt_cast_i.exception_instr_upper_not_lower_half;
 
@@ -210,7 +210,7 @@ module bp_be_director
       else if (commit_pkt_cast_i.eret)
         begin
           fe_cmd_li.opcode                                 = e_op_pc_redirection;
-          fe_cmd_li.vaddr                                  = commit_pkt_cast_i.npc;
+          fe_cmd_li.pc                                     = commit_pkt_cast_i.npc;
           fe_cmd_pc_redirect_operands.subopcode            = e_subop_eret;
           fe_cmd_pc_redirect_operands.priv                 = commit_pkt_cast_i.priv_n;
           fe_cmd_pc_redirect_operands.translation_en       = commit_pkt_cast_i.translation_en_n;
@@ -221,7 +221,7 @@ module bp_be_director
       else if (commit_pkt_cast_i.exception | commit_pkt_cast_i._interrupt | (is_wait & irq_waiting_i))
         begin
           fe_cmd_li.opcode                                 = e_op_pc_redirection;
-          fe_cmd_li.vaddr                                  = commit_pkt_cast_i.npc;
+          fe_cmd_li.pc                                     = commit_pkt_cast_i.npc;
           fe_cmd_pc_redirect_operands.subopcode            = e_subop_trap;
           fe_cmd_pc_redirect_operands.priv                 = commit_pkt_cast_i.priv_n;
           fe_cmd_pc_redirect_operands.translation_en       = commit_pkt_cast_i.translation_en_n;
@@ -232,7 +232,7 @@ module bp_be_director
       else if (isd_status_cast_i.v & npc_mismatch_v)
         begin
           fe_cmd_li.opcode                                 = e_op_pc_redirection;
-          fe_cmd_li.vaddr                                  = expected_npc_o;
+          fe_cmd_li.pc                                     = expected_npc_o;
           fe_cmd_pc_redirect_operands.subopcode            = e_subop_branch_mispredict;
           fe_cmd_pc_redirect_operands.branch_metadata_fwd  = isd_status_cast_i.branch_metadata_fwd;
           fe_cmd_pc_redirect_operands.misprediction_reason = last_instr_was_branch
@@ -248,7 +248,7 @@ module bp_be_director
       else if (isd_status_cast_i.v & ~npc_mismatch_v & last_instr_was_branch)
         begin
           fe_cmd_li.opcode                               = e_op_attaboy;
-          fe_cmd_li.vaddr                                = expected_npc_o;
+          fe_cmd_li.pc                                   = expected_npc_o;
           fe_cmd_li.operands.attaboy.taken               = last_instr_was_btaken;
           fe_cmd_li.operands.attaboy.branch_metadata_fwd = isd_status_cast_i.branch_metadata_fwd;
 
--- a/bp_be/src/v/bp_be_checker/bp_be_director.sv
+++ b/bp_be/src/v/bp_be_checker/bp_be_director.sv
@@ -161,8 +161,8 @@ module bp_be_director
           fe_cmd_li.opcode                               = e_op_itlb_fill_response;
           fe_cmd_li.pc                                   = commit_pkt_cast_i.npc;
           fe_cmd_li.operands.itlb_fill_response.pte_leaf = commit_pkt_cast_i.pte_leaf;
-          fe_cmd_li.operands.itlb_fill_response.partial_instr   = commit_pkt_cast_i.instr;
-          fe_cmd_li.operands.itlb_fill_response.partial_instr_v = commit_pkt_cast_i.exception_instr_upper_not_lower_half;
+          fe_cmd_li.operands.itlb_fill_response.partial_instr   = compressed_support_p ? commit_pkt_cast_i.instr : '0;
+          fe_cmd_li.operands.itlb_fill_response.partial_instr_v = compressed_support_p & commit_pkt_cast_i.exception_instr_upper_not_lower_half;
           fe_cmd_li.operands.itlb_fill_response.fill_vaddr      = commit_pkt_cast_i.vaddr;
 
           fe_cmd_v_li = fe_cmd_ready_lo;
@@ -202,8 +202,8 @@ module bp_be_director
         begin
           fe_cmd_li.opcode = e_op_icache_fill_response;
           fe_cmd_li.pc     = commit_pkt_cast_i.npc;
-          fe_cmd_li.operands.icache_fill_response.partial_instr   = commit_pkt_cast_i.instr;
-          fe_cmd_li.operands.icache_fill_response.partial_instr_v = commit_pkt_cast_i.exception_instr_upper_not_lower_half;
+          fe_cmd_li.operands.icache_fill_response.partial_instr   = compressed_support_p ? commit_pkt_cast_i.instr : '0;
+          fe_cmd_li.operands.icache_fill_response.partial_instr_v = compressed_support_p & commit_pkt_cast_i.exception_instr_upper_not_lower_half;
 
           fe_cmd_v_li = fe_cmd_ready_lo;
         end
--- a/bp_be/src/v/bp_be_checker/bp_be_director.sv
+++ b/bp_be/src/v/bp_be_checker/bp_be_director.sv
@@ -162,7 +162,7 @@ module bp_be_director
           fe_cmd_li.npc                                  = commit_pkt_cast_i.npc;
           fe_cmd_li.operands.itlb_fill_response.pte_leaf = commit_pkt_cast_i.pte_leaf;
           fe_cmd_li.operands.itlb_fill_response.instr    = compressed_support_p ? commit_pkt_cast_i.instr : '0;
-          fe_cmd_li.operands.vtag                        = commit_pkt_cast_i.vaddr;
+          fe_cmd_li.operands.itlb_fill_response.vtag     = commit_pkt_cast_i.vaddr[vaddr_width_p-1-:vtag_width_p];
 
           fe_cmd_v_li = fe_cmd_ready_lo;
         end
--- a/bp_be/src/v/bp_be_checker/bp_be_director.sv
+++ b/bp_be/src/v/bp_be_checker/bp_be_director.sv
@@ -158,7 +158,7 @@ module bp_be_director
         end
       else if (commit_pkt_cast_i.itlb_fill_v)
         begin
-          fe_cmd_li.opcode                               = (compressed_support_p & commit_pkt_cast_i.exception_instr_upper_not_lower_half) ? e_op_itlb_fill_resume : e_op_itlb_fill_restart;
+          fe_cmd_li.opcode                               = (compressed_support_p & commit_pkt_cast_i.exception_instr_partial_v) ? e_op_itlb_fill_resume : e_op_itlb_fill_restart;
           fe_cmd_li.npc                                  = commit_pkt_cast_i.vaddr;
           fe_cmd_li.operands.itlb_fill_response.pte_leaf = commit_pkt_cast_i.pte_leaf;
           fe_cmd_li.operands.itlb_fill_response.instr    = compressed_support_p ? commit_pkt_cast_i.instr : '0;
@@ -199,7 +199,7 @@ module bp_be_director
         end
       else if (commit_pkt_cast_i.icache_miss)
         begin
-          fe_cmd_li.opcode = (compressed_support_p & commit_pkt_cast_i.exception_instr_upper_not_lower_half) ? e_op_icache_fill_resume : e_op_icache_fill_restart;
+          fe_cmd_li.opcode = (compressed_support_p & commit_pkt_cast_i.exception_instr_partial_v) ? e_op_icache_fill_resume : e_op_icache_fill_restart;
           fe_cmd_li.npc    = commit_pkt_cast_i.vaddr;
           fe_cmd_li.operands.icache_fill_response.instr = compressed_support_p ? commit_pkt_cast_i.instr : '0;
 
--- a/bp_be/src/v/bp_be_checker/bp_be_director.sv
+++ b/bp_be/src/v/bp_be_checker/bp_be_director.sv
@@ -159,10 +159,10 @@ module bp_be_director
       else if (commit_pkt_cast_i.itlb_fill_v)
         begin
           fe_cmd_li.opcode                               = (compressed_support_p & commit_pkt_cast_i.exception_instr_upper_not_lower_half) ? e_op_itlb_fill_resume : e_op_itlb_fill_restart;
-          fe_cmd_li.npc                                  = commit_pkt_cast_i.npc;
+          fe_cmd_li.npc                                  = commit_pkt_cast_i.vaddr;
           fe_cmd_li.operands.itlb_fill_response.pte_leaf = commit_pkt_cast_i.pte_leaf;
           fe_cmd_li.operands.itlb_fill_response.instr    = compressed_support_p ? commit_pkt_cast_i.instr : '0;
-          fe_cmd_li.operands.itlb_fill_response.vtag     = commit_pkt_cast_i.vaddr[vaddr_width_p-1-:vtag_width_p];
+          fe_cmd_li.operands.itlb_fill_response.vtag     = commit_pkt_cast_i.vaddr[vaddr_width_p-1-:vtag_width_p]; // TODO: dedupe?
 
           fe_cmd_v_li = fe_cmd_ready_lo;
         end
@@ -200,7 +200,7 @@ module bp_be_director
       else if (commit_pkt_cast_i.icache_miss)
         begin
           fe_cmd_li.opcode = (compressed_support_p & commit_pkt_cast_i.exception_instr_upper_not_lower_half) ? e_op_icache_fill_resume : e_op_icache_fill_restart;
-          fe_cmd_li.npc = commit_pkt_cast_i.npc;
+          fe_cmd_li.npc    = commit_pkt_cast_i.vaddr;
           fe_cmd_li.operands.icache_fill_response.instr = compressed_support_p ? commit_pkt_cast_i.instr : '0;
 
           fe_cmd_v_li = fe_cmd_ready_lo;
--- a/bp_be/src/v/bp_be_checker/bp_be_director.sv
+++ b/bp_be/src/v/bp_be_checker/bp_be_director.sv
@@ -162,7 +162,6 @@ module bp_be_director
           fe_cmd_li.npc                                  = commit_pkt_cast_i.vaddr;
           fe_cmd_li.operands.itlb_fill_response.pte_leaf = commit_pkt_cast_i.pte_leaf;
           fe_cmd_li.operands.itlb_fill_response.instr    = compressed_support_p ? commit_pkt_cast_i.instr : '0;
-          fe_cmd_li.operands.itlb_fill_response.vtag     = commit_pkt_cast_i.vaddr[vaddr_width_p-1-:vtag_width_p]; // TODO: dedupe?
 
           fe_cmd_v_li = fe_cmd_ready_lo;
         end
--- a/bp_be/src/v/bp_be_checker/bp_be_scheduler.sv
+++ b/bp_be/src/v/bp_be_checker/bp_be_scheduler.sv
@@ -205,6 +205,7 @@ module bp_be_scheduler
       dispatch_pkt = '0;
       dispatch_pkt.v        = (fe_queue_yumi_li & ~poison_isd_i) || be_exc_not_instr_li;
       dispatch_pkt.queue_v  = fe_queue_yumi_li;
+      // TODO: instr_v could be restricted so that we don't need queue_v & instr_v in calculator_top
       dispatch_pkt.instr_v  = instr_v_li;
       dispatch_pkt.pc       = expected_npc_i;
       dispatch_pkt.instr    = fe_queue_lo.instr;
--- a/bp_be/src/v/bp_be_checker/bp_be_scheduler.sv
+++ b/bp_be/src/v/bp_be_checker/bp_be_scheduler.sv
@@ -223,7 +223,7 @@ module bp_be_scheduler
         fe_exc_not_instr_li & fe_queue_lo.msg.exception.exception_code inside {e_itlb_miss};
       dispatch_pkt.exception.icache_miss        |=
         fe_exc_not_instr_li & fe_queue_lo.msg.exception.exception_code inside {e_icache_miss};
-      dispatch_pkt.exception.instr_upper_not_lower_half |=
+      dispatch_pkt.exception.upper_not_lower_half |=
         fe_exc_not_instr_li & fe_queue_lo.msg.exception.upper_not_lower_half;
 
       dispatch_pkt.exception.instr_page_fault |= be_exc_not_instr_li & ptw_fill_pkt_cast_i.instr_page_fault_v;
@@ -231,7 +231,7 @@ module bp_be_scheduler
       dispatch_pkt.exception.store_page_fault |= be_exc_not_instr_li & ptw_fill_pkt_cast_i.store_page_fault_v;
       dispatch_pkt.exception.itlb_fill        |= be_exc_not_instr_li & ptw_fill_pkt_cast_i.itlb_fill_v;
       dispatch_pkt.exception.dtlb_fill        |= be_exc_not_instr_li & ptw_fill_pkt_cast_i.dtlb_fill_v;
-      dispatch_pkt.exception.instr_upper_not_lower_half |= be_exc_not_instr_li & ptw_fill_pkt_cast_i.instr_upper_not_lower_half;
+      dispatch_pkt.exception.upper_not_lower_half |= be_exc_not_instr_li & ptw_fill_pkt_cast_i.instr_upper_not_lower_half;
       dispatch_pkt.exception._interrupt       |= be_exc_not_instr_li & interrupt_v_i & ~unfreeze_i;
       dispatch_pkt.exception.unfreeze         |= be_exc_not_instr_li & unfreeze_i;
 
--- a/bp_be/src/v/bp_be_checker/bp_be_scheduler.sv
+++ b/bp_be/src/v/bp_be_checker/bp_be_scheduler.sv
@@ -218,6 +218,9 @@ module bp_be_scheduler
       dispatch_pkt.imm      = (fe_exc_not_instr_li | be_exc_not_instr_li) ? '0 : issue_pkt.frs3_v ? frf_rs3 : decoded_imm_lo;
       dispatch_pkt.decode   = instr_decoded;
 
+      dispatch_pkt.instr_partial_v = (be_exc_not_instr_li & ptw_fill_pkt_cast_i.instr_partial_v)
+                                   | (fe_exc_not_instr_li & fe_queue_lo.partial_v);
+
       dispatch_pkt.exception.instr_access_fault |=
         fe_exc_not_instr_li & fe_queue_lo.msg_type inside {e_instr_access_fault};
       dispatch_pkt.exception.instr_page_fault   |=
@@ -226,15 +229,12 @@ module bp_be_scheduler
         fe_exc_not_instr_li & fe_queue_lo.msg_type inside {e_itlb_miss};
       dispatch_pkt.exception.icache_miss        |=
         fe_exc_not_instr_li & fe_queue_lo.msg_type inside {e_icache_miss};
-      dispatch_pkt.exception.instr_partial_v |=
-        fe_exc_not_instr_li & fe_queue_lo.partial_v;
 
       dispatch_pkt.exception.instr_page_fault |= be_exc_not_instr_li & ptw_fill_pkt_cast_i.instr_page_fault_v;
       dispatch_pkt.exception.load_page_fault  |= be_exc_not_instr_li & ptw_fill_pkt_cast_i.load_page_fault_v;
       dispatch_pkt.exception.store_page_fault |= be_exc_not_instr_li & ptw_fill_pkt_cast_i.store_page_fault_v;
       dispatch_pkt.exception.itlb_fill        |= be_exc_not_instr_li & ptw_fill_pkt_cast_i.itlb_fill_v;
       dispatch_pkt.exception.dtlb_fill        |= be_exc_not_instr_li & ptw_fill_pkt_cast_i.dtlb_fill_v;
-      dispatch_pkt.exception.instr_partial_v  |= be_exc_not_instr_li & ptw_fill_pkt_cast_i.instr_partial_v;
       dispatch_pkt.exception._interrupt       |= be_exc_not_instr_li & interrupt_v_i & ~unfreeze_i;
       dispatch_pkt.exception.unfreeze         |= be_exc_not_instr_li & unfreeze_i;
 
--- a/bp_be/src/v/bp_be_checker/bp_be_scheduler.sv
+++ b/bp_be/src/v/bp_be_checker/bp_be_scheduler.sv
@@ -164,7 +164,6 @@ module bp_be_scheduler
 
   wire fe_exc_not_instr_li = fe_queue_yumi_li & (fe_queue_lo.msg_type == e_fe_exception);
   wire [vaddr_width_p-1:0] fe_exc_pc_li = fe_queue_lo.msg.exception.pc;
-  // For now, the FE exception is always affiliated with the PC address (aligned)
   wire [vaddr_width_p-1:0] fe_exc_vaddr_li =  `bp_be_instr_half_address(fe_exc_pc_li, fe_queue_lo.msg.exception.upper_not_lower_half);
   wire be_exc_not_instr_li = ptw_fill_pkt_cast_i.v | interrupt_v_i | unfreeze_i;
   wire [vaddr_width_p-1:0] be_exc_vaddr_li = ptw_fill_pkt_cast_i.vaddr;
--- a/bp_be/src/v/bp_be_checker/bp_be_scheduler.sv
+++ b/bp_be/src/v/bp_be_checker/bp_be_scheduler.sv
@@ -225,7 +225,7 @@ module bp_be_scheduler
         fe_exc_not_instr_li & fe_queue_lo.msg_type inside {e_itlb_miss};
       dispatch_pkt.exception.icache_miss        |=
         fe_exc_not_instr_li & fe_queue_lo.msg_type inside {e_icache_miss};
-      dispatch_pkt.exception.upper_not_lower_half |=
+      dispatch_pkt.exception.instr_partial_v |=
         fe_exc_not_instr_li & fe_queue_lo.partial_v;
 
       dispatch_pkt.exception.instr_page_fault |= be_exc_not_instr_li & ptw_fill_pkt_cast_i.instr_page_fault_v;
@@ -233,7 +233,7 @@ module bp_be_scheduler
       dispatch_pkt.exception.store_page_fault |= be_exc_not_instr_li & ptw_fill_pkt_cast_i.store_page_fault_v;
       dispatch_pkt.exception.itlb_fill        |= be_exc_not_instr_li & ptw_fill_pkt_cast_i.itlb_fill_v;
       dispatch_pkt.exception.dtlb_fill        |= be_exc_not_instr_li & ptw_fill_pkt_cast_i.dtlb_fill_v;
-      dispatch_pkt.exception.upper_not_lower_half |= be_exc_not_instr_li & ptw_fill_pkt_cast_i.instr_upper_not_lower_half;
+      dispatch_pkt.exception.instr_partial_v  |= be_exc_not_instr_li & ptw_fill_pkt_cast_i.instr_partial_v;
       dispatch_pkt.exception._interrupt       |= be_exc_not_instr_li & interrupt_v_i & ~unfreeze_i;
       dispatch_pkt.exception.unfreeze         |= be_exc_not_instr_li & unfreeze_i;
 
--- a/bp_be/src/v/bp_be_checker/bp_be_scheduler.sv
+++ b/bp_be/src/v/bp_be_checker/bp_be_scheduler.sv
@@ -197,8 +197,8 @@ module bp_be_scheduler
       isd_status_cast_o.frs3_v   = fe_queue_v_lo & issue_pkt.frs3_v;
       isd_status_cast_o.rs3_addr = fe_queue_lo.msg.fetch.instr.t.fmatype.rs3_addr;
       isd_status_cast_o.rd_addr  = fe_queue_lo.msg.fetch.instr.t.fmatype.rd_addr;
-      isd_status_cast_o.iwb_v    = fe_instr_not_exc_li & instr_decoded.irf_w_v;
-      isd_status_cast_o.fwb_v    = fe_instr_not_exc_li & instr_decoded.frf_w_v;
+      isd_status_cast_o.iwb_v    = instr_decoded.irf_w_v;
+      isd_status_cast_o.fwb_v    = instr_decoded.frf_w_v;
 
       // Form dispatch packet
       dispatch_pkt = '0;
--- a/bp_be/src/v/bp_be_checker/bp_be_scheduler.sv
+++ b/bp_be/src/v/bp_be_checker/bp_be_scheduler.sv
@@ -205,6 +205,7 @@ module bp_be_scheduler
       dispatch_pkt = '0;
       dispatch_pkt.v        = (fe_queue_yumi_li & ~poison_isd_i) || be_exc_not_instr_li;
       dispatch_pkt.queue_v  = fe_queue_yumi_li;
+      dispatch_pkt.instr_v  = instr_v_li;
       dispatch_pkt.pc       = expected_npc_i;
       dispatch_pkt.instr    = fe_queue_lo.instr;
       // If register injection is critical, can be done after bypass
--- a/bp_common/src/include/bp_common_aviary_cfg_pkgdef.svh
+++ b/bp_common/src/include/bp_common_aviary_cfg_pkgdef.svh
@@ -188,6 +188,9 @@
     // Whether to emulate FPU
     //   bit 0: fpu enabled
     integer unsigned fpu_support;
+    // Whether to enable the "c" extension.
+    // Currently, this only enables support for misaligned 32-bit instructions; full "C" support is not yet implemented.
+    integer unsigned compressed_support;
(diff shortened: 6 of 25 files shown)


bp_be/src/v/bp_be_checker/bp_be_scheduler.sv
```diff
@@ -176,7 +176,7 @@ module bp_be_scheduler
       // Calculator status ISD stage
       isd_status_cast_o = '0;
       isd_status_cast_o.v        = fe_queue_yumi_li;
-      isd_status_cast_o.pc       = fe_queue_lo.msg.fetch.pc;
+      isd_status_cast_o.pc       = fe_instr_not_exc_li ? fe_queue_lo.msg.fetch.pc : fe_queue_lo.msg.exception.pc;
```
WasabiFan, line 179: Non-functional change. Fetch and exception address fields happen to alias so the old code was fine, but wasn't obviously correct for both variants of the union. Figure this was a reasonable thing to do.

bp_common/syn/Makefile.vcs
```diff
@@ -72,7 +72,7 @@ $(BUILD_DIR)/simv: | $(BUILD_COLLATERAL)
 	-@tail -n3 $(BUILD_LOG) > $(BUILD_REPORT)
 	-@test -s $(BUILD_ERROR) && echo "FAILED" >> $(BUILD_REPORT) || rm $(BUILD_ERROR)
 
-build_dump.v: VCS_BUILD_OPTS += -debug_pp
+build_dump.v: VCS_BUILD_OPTS += -debug_all
```
WasabiFan, line 75: TODO: revert once ready to merge

(Also, why isn't this the default? Do we _like_ garbage in our debug traces?)

bp_fe/src/v/bp_fe_pc_gen.sv
```diff
@@ -60,31 +62,48 @@ module bp_fe_pc_gen
...
+  logic [vaddr_width_p-1:0] next_pc;
+  logic next_fetch_linear;
   logic ovr_ret, ovr_taken, btb_taken;
   logic [vaddr_width_p-1:0] btb_br_tgt_lo;
   logic [vaddr_width_p-1:0] ras_tgt_lo;
   logic [vaddr_width_p-1:0] br_tgt_lo;
   wire [vaddr_width_p-1:0] pc_plus4  = pc_if1_r + vaddr_width_p'(4);
-  always_comb
+  always_comb begin
+    next_fetch_linear = 1'b0;
```
WasabiFan, line 82: There was some discussion in the other MR about whether this is preferable over retroactively identifying whether the address succeeds the last. My belief so far is that avoiding 39-bit adder-and-comparator seems worthwhile, but I am not strongly of that opinion; if it is better to do it, I can make the change.

bp_fe/src/v/bp_fe_pc_gen.sv
```diff
@@ -29,16 +29,18 @@ module bp_fe_pc_gen
...
 
    , input [instr_width_gp-1:0]                      fetch_i
-   , input                                           fetch_instr_v_i
-   , input                                           fetch_exception_v_i
+   , input                                           fetch_v_i
+   , output [instr_width_gp-1:0]                     fetch_instr_o
+   , output                                          fetch_instr_v_o
    , output logic [branch_metadata_fwd_width_p-1:0]  fetch_br_metadata_fwd_o
    , output logic [vaddr_width_p-1:0]                fetch_pc_o
+   , output                                          fetch_is_second_half_o
```
WasabiFan, line 43: Needed for exception packet servicing faults/I$ fills/ITLB fills.

bp_be/src/v/bp_be_checker/bp_be_scheduler.sv
```diff
@@ -161,7 +161,7 @@ module bp_be_scheduler
      );
 
   wire fe_exc_not_instr_li = fe_queue_yumi_li & (fe_queue_lo.msg_type == e_fe_exception);
-  wire [vaddr_width_p-1:0] fe_exc_vaddr_li = fe_queue_lo.msg.exception.vaddr;
+  wire [vaddr_width_p-1:0] fe_exc_vaddr_li = fe_queue_lo.msg.exception.pc;
```
WasabiFan, line 164: This is unused, can I just remove it? Is it there for debugging?

bp_be/src/include/bp_be_ctl_pkgdef.svh
```diff
@@ -250,6 +250,9 @@
     logic dtlb_fill;
     logic _interrupt;
     logic cmd_full;
+
+    // Metadata for instruction exceptions (instr_page_fault,instr_access_fault, itlb_miss, icache_miss)
+    logic instr_upper_not_lower_half;
```
dpetrisko, line 255: I think just upper_not_lower_half. In a hypothetical extension where we support misaligned D$ (required for Zam extension for misaligned atomics), we can leverage for either

bp_be/src/include/bp_be_defines.svh
```diff
@@ -277,5 +280,8 @@
   `define bp_be_decode_info_width \
     (rv64_priv_width_gp+8)
 
+  `define bp_be_instr_half_address(base_pc_mp, upper_not_lower_half_mp) \
+    (base_pc_mp + (upper_not_lower_half_mp ? 2 : 0))
```
dpetrisko, line 284: use a concatenation rather than addition (in case synthesizer is dumb)

bp_be/src/v/bp_be_checker/bp_be_scheduler.sv
dpetrisko: yeah debugging with verilator is rough with structs sometimes...

bp_be/src/v/bp_be_checker/bp_be_scheduler.sv
dpetrisko, line 179: good catch

bp_common/src/include/bp_common_core_if.svh
```diff
@@ -134,9 +135,21 @@
     typedef struct packed                                                                          \
     {                                                                                              \
       bp_pte_leaf_s              pte_leaf;                                                         \
+      logic instr_upper_not_lower_half;                                                            \
```
dpetrisko, line 138: nit; alignment

bp_common/src/include/bp_common_core_if.svh
```diff
@@ -134,9 +135,21 @@
...
                                  padding;                                                          \
     }  bp_fe_cmd_itlb_map_s;                                                                       \
+    /*                                                                                             \
+     * bp_fe_cmd_icache_fill_s indicates the alignment offset of the original                      \
+     * miss which triggered this fill request. It carries no actual cache data;                    \
+     * the frontend will issue the memory request upon receipt.                                    \
+    */                                                                                             \
+    typedef struct packed                                                                          \
+    {                                                                                              \
+      logic instr_upper_not_lower_half;                                                            \
```
dpetrisko, line 149: nit; alignment

bp_common/src/include/bp_common_rv64_csr_defines.svh
```diff
@@ -213,12 +213,12 @@
                                                                                            \
```
dpetrisko, line 213: noting for later; should we parameterize these for with/without compressed instructions?

bp_common/syn/Makefile.vcs
dpetrisko: :) TL;DR big dump files for big projects. Haven't benchmarked it for BP though...

bp_fe/src/include/bp_fe_decompress.svh
```diff
@@ -1,239 +0,0 @@
-`ifndef BP_FE_DECOMPRESS_SVH
```
dpetrisko, line 1: let's keep this around for now

bp_fe/src/v/bp_fe_top.sv
```diff
@@ -357,19 +377,20 @@ module bp_fe_top
 
   wire icache_miss    = v_if2_r & ~icache_data_v_lo;
   wire queue_miss     = v_if2_r & ~fe_queue_ready_i;
-  wire fe_exception_v = v_if2_r & (instr_misaligned_r | instr_access_fault_r | instr_page_fault_r | itlb_miss_r | icache_miss_v_lo);
-  wire fe_instr_v     = v_if2_r & icache_data_v_lo;
+  wire fe_exception_v = v_if2_r & (instr_access_fault_r | instr_page_fault_r | itlb_miss_r | icache_miss_v_lo);
+  wire fe_instr_v     = v_if2_r & fetch_instr_v_lo;
   assign fe_queue_v_o = fe_queue_ready_i & (fe_instr_v | fe_exception_v) & ~cmd_nonattaboy_v;
+  wire fe_progress    = fe_queue_ready_i & (~icache_miss | fe_exception_v) & ~cmd_nonattaboy_v;
```
dpetrisko, line 383: strictly speaking, if we go into the realigner we don't need the fe_queue_ready_i?

bp_fe/src/v/bp_fe_top.sv
```diff
@@ -215,10 +234,11 @@ module bp_fe_top
   logic [ptag_width_p-1:0] ptag_li;
 
   bp_pte_leaf_s w_tlb_entry_li;
-  wire [vtag_width_p-1:0] w_vtag_li = fe_cmd_cast_i.vaddr[vaddr_width_p-1-:vtag_width_p];
+  wire [vaddr_width_p-1:0] w_tlb_vaddr = fe_cmd_cast_i.vaddr + (fe_cmd_cast_i.operands.itlb_fill_response.instr_upper_not_lower_half ? 2 : 0);
```
dpetrisko, line 237: This is likely critical. If there's no way to avoid this addition here (I don't think there is), I suggest we get rid of the upper_not_lower in fe_cmd and instead calculate the correct address in the BE

bp_fe/src/v/bp_fe_top.sv
```diff
@@ -152,6 +155,22 @@ module bp_fe_top
   wire eret_v        = pc_redirect_v & (fe_cmd_cast_i.operands.pc_redirect_operands.subopcode == e_subop_eret);
   wire translation_v = pc_redirect_v & (fe_cmd_cast_i.operands.pc_redirect_operands.subopcode == e_subop_translation_switch);
 
+  // Conditionally delay I$ fill by one cycle when fill only applies to upper half of instruction
+  // (i.e., the instruction requires two aligned reads and it's the second one that needs a fill)
+  wire icache_fill_response_upper_not_lower_half = fe_cmd_cast_i.operands.icache_fill_response.instr_upper_not_lower_half;
```
dpetrisko, line 160: Not sure I follow this. Can you elaborate on the case?

dpetrisko:
Overall, looks great, excellent work. Much less impact to flow control than before. Made several minor comments. Do you have a link to the google slide that goes along with this? 

Two major comments:
- I think that if we can it's better to allow for compression to be parameterized enable/disable (at the least to benchmark with/without). If this is impossible without substantial work we can skip, but looking at the code I believe that it's possible
- We should look closely at the alignment operations to see which are necessary and which are strictly not. Will discuss offline

bp_fe/src/v/bp_fe_top.sv
WasabiFan: Re: the former, I'm good with this. The drawbacks I see are:
- Now (some) FE commands which are conceptually replies now use a different response PC than the request did
- The "resume" logic would need to store state for the first half of the instruction (I think you mentioned this below) and ensure it's in that state when it receives the reply. Either that, or have _both_ addresses somewhere (the pc and the fault vaddr).

The added complexity on the second point makes me a bit weary, but I don't think there's a reason it can't be done. So I can give it a shot if we want to do that.

bp_fe/src/v/bp_fe_top.sv
WasabiFan: As I understand it: The mechanism for causing a fill is to issue the normal "read" request into the I$ but use the "fill" message type, which triggers a fill if needed. The FE command for a fill just tells it to resume at that PC and flip the "fill" flag when making the I$ request. Then the I$ naturally performs the fill if needed.

In my implementation, when we get an I$ or ITLB fill command (response), we fetch the instruction from scratch, i.e. re-fetch both the first and second half. This means two consecutive I$ requests to fulfil one instruction. If the I$ fill command was for the upper (second) half of the instruction, then we need to issue a) a non-fill I$ request (or, if we'd like, a fill request) _and then_ issue the "real" fill request. In other words, it needs to "delay" forwarding the fill command until after it has performed the fetch of the first half so that the fill applies to the latter half which faulted.

WasabiFan:
> Do you have a link to the google slide that goes along with this?

I am a bit naughty and haven't drawn up a new diagram for the revised implementation. The fetch flow (i.e., which memory fetches are made and their timing) is the same as it used to be. The difference is that I've removed a bunch of state to delay branch prediction. Instead, it just puts the branch prediction metadata in the BTB/BHT according to the latter half of a misaligned instruction. If it'd be worth the time investment I can make an updated data+control block diagram.

The original documents are linked here: https://github.com/black-parrot/black-parrot/pull/986

---

> I think that if we can it's better to allow for compression to be parameterized enable/disable (at the least to benchmark with/without). If this is impossible without substantial work we can skip, but looking at the code I believe that it's possible

Is the goal to eliminate all added state, or purely not use it (and allow the tools to eliminate it if it identifies the undriven regs)? I think either can be done, but I'm not sure how to guarantee the latter with a parameter rather than an `ifdef`. I'm asking, in essence, whether the goal is to allow benchmarking power/area or just cycle counts.

bp_be/src/include/bp_be_defines.svh
WasabiFan: _Can_ I use concatenation here and get the same result? The annoying thing about this operation is that `base_pc_mp` is aligned to 16B (not 32), so e.g. a base PC of `0x0FFE` has its second half at address `0x1000`. In other words, I think this involves a proper adder to propagate the carry, and that's why it ends up being so expensive in the critical path.

For the PTW in the BE, all it needs is the VPN, which I could compute directly and shorten the critical pat a bit. That being said, it has to compute the true value eventually in the case of a miss, since the faulting address is used for exception tracing.

This same operation is currently done in the FE and you commented elsewhere that that one is likely critical, which we'll also need to address and I think we've found a solution in chat. With that solution, we will need to do this addition for I$ fills in the BE.

dpetrisko: Never mind, I misunderstood the usage here. Carry on...

bp_fe/src/v/bp_fe_top.sv
dpetrisko: I see, clever!

dpetrisko:
>Is the goal to eliminate all added state, or purely not use it (and allow the tools to eliminate it if it identifies the undriven regs)? I think either can be done, but I'm not sure how to guarantee the latter with a parameter rather than an ifdef. I'm asking, in essence, whether the goal is to allow benchmarking power/area or just cycle counts.

Power/area as well for sure. if (0) === `ifdef FALSE === my_cool_logic & 0 from a synthesizer perspective, so same principle. Constant propagation is almost disturbingly good in tools nowadays

bp_fe/src/v/bp_fe_top.sv
WasabiFan: Updated. The `upper_not_lower` is sent in the FE->BE, then the scheduler computes the vaddr and replies with that. The I$ fill in the FE uses the computed address.

bp_fe/src/v/bp_fe_top.sv
WasabiFan: This was removed incidentally now that partial instructions are forwarded through the BE.

WasabiFan:
PR updated with the discussed re-architecture.

Key changes:
- The vaddr used for ITLB fills is computed in the BE and sent back with the message, to remove it from critical path in the FE
- I$ and ITLB misses, the two exception types which can occur to only one half of an instruction, now forward the partial realigner state through the backend rather than re-fetching a first half or storing it in the FE.
- The realigner now operates on a "poison" mechanic; it stores validity and is invalidated by control flow and redirects. This is intended to make stalling FE designs easier in the future.

Verification:
- My three new test programs, which verify execution, branch prediction, and exception handling of misaligned instructions, pass and execute in the expected number of cycles
- The output of my "pc gen tracer" on aligned test programs is identical to what it used to be in the unmodified processor. This includes all branch prediction minutiae, stall conditions, etc.
- I have manually inspected the "pc gen tracer" traces for all introduced test programs on misaligned code. I have confirmed that:
  - Misaligning code introduces precisely the expected number of additional cycle delays due to double-fetching, and they occur in the right places
  - Exceptions correctly propagate partial instruction data. Once a "first half" is fetched, it isn't fetched again.

Remaining TODOs (not blocking review):
- The realigner and related signals sometimes use confusing names. "upper" and "lower" halves sometimes refer to those of a fetch, and sometimes refer to those of an instruction. The upper half of a fetch is the lower half of a misaligned instruction, and vice versa. I will do some naming normalization.
- I implemented the parameter for enabling/disabling "C" support but have not used it yet. I need to go through and wrap checks around all new regs.
- As discussed on Slack, I learned that branch prediction metadata is not sent with exceptions but the BE blindly reads the union field anyway. As a temporary workaround, I've zeroed this data in the scheduler to enable my consistency verification. I started but haven't finished introducing branch metadata into exceptions.
- My VM test program doesn't save/restore all registers across ECALLs, and I don't want to leave accidents waiting to happen if the compiler chooses registers which get trampled on. Minor change.
- Misc tracer cleanups. I'd like to add a few more signals to the PC gen tracer.
- There may be some small comments above I haven't addressed yet.

I also haven't changed the FE->BE interface beyond the minimum signalling additions. We discussed a reformulation of those structures and if I make those changes it will be in a separate PR.

bp_be/src/v/bp_be_checker/bp_be_scheduler.sv
```diff
@@ -180,8 +181,9 @@ module bp_be_scheduler
       // Calculator status ISD stage
       isd_status_cast_o = '0;
       isd_status_cast_o.v        = fe_queue_yumi_li;
-      isd_status_cast_o.pc       = fe_queue_lo.msg.fetch.pc;
-      isd_status_cast_o.branch_metadata_fwd = fe_queue_lo.msg.fetch.branch_metadata_fwd;
+      isd_status_cast_o.pc       = fe_instr_not_exc_li ? fe_queue_lo.msg.fetch.pc : fe_queue_lo.msg.exception.pc;
+      // TODO: forward branch metadata for exceptions properly
+      isd_status_cast_o.branch_metadata_fwd = fe_instr_not_exc_li ? fe_queue_lo.msg.fetch.branch_metadata_fwd : '0;
```
WasabiFan, line 186: This is a temporary workaround to avoid forwarding garbage exception metadata back to the FE. It replicates the current behavior because the padding it inadvertently was reading (which is now occupied by a `partial_instr` field) was all-zeroes. I suspended my work to introduce proper exception branch metadata forwarding and haven't picked it up again.

bp_fe/src/v/bp_fe_pc_gen.sv
```diff
@@ -28,17 +28,23 @@ module bp_fe_pc_gen
    , input                                           redirect_br_taken_i
    , input                                           redirect_br_ntaken_i
```
WasabiFan, line 29: This is unused, should I remove it?

bp_fe/src/v/bp_fe_pc_gen.sv
```diff
@@ -152,8 +175,8 @@ module bp_fe_pc_gen
      );
 
   // BHT
-  wire bht_r_v_li = next_pc_yumi_i;
```
WasabiFan, line 155: Note: currently, the `r_v` signals on the BHT and BHT don't mask prediction outputs. The memories are configured to latch the last read when `r_v` goes low. The only impact I've seen of this behavior is that under stall the FE sometimes repeatedly predicts jumps. Since it's stalled, these predictions are ignored.

The below patch could fix it if desired.

<details>

```patch
diff --git a/bp_fe/src/v/bp_fe_bht.sv b/bp_fe/src/v/bp_fe_bht.sv
index 5f823d2f..7889675b 100644
--- a/bp_fe/src/v/bp_fe_bht.sv
+++ b/bp_fe/src/v/bp_fe_bht.sv
@@ -130,8 +130,19 @@ module bp_fe_bht
      assign pred_idx_r = 1'b1;
    end
 
+  logic r_data_v_lo;
+  bsg_dff_reset
+   #(.width_p(1))
+   r_v_reg
+    (.clk_i(clk_i)
+     ,.reset_i(reset_i)
+
+     ,.data_i(r_v_li)
+     ,.data_o(r_data_v_lo)
+     );
+
   assign val_o = r_data_lo;
-  assign pred_o = val_o[pred_idx_r];
+  assign pred_o = r_data_v_lo & val_o[pred_idx_r];
 
 endmodule
 
diff --git a/bp_fe/src/v/bp_fe_btb.sv b/bp_fe/src/v/bp_fe_btb.sv
index 27ebea98..d2b2b6d6 100644
--- a/bp_fe/src/v/bp_fe_btb.sv
+++ b/bp_fe/src/v/bp_fe_btb.sv
@@ -129,8 +129,19 @@ module bp_fe_btb
      ,.data_o(r_tag_r)
      );
 
-  assign br_tgt_v_o   = tag_mem_data_lo.v & (tag_mem_data_lo.tag == r_tag_r);
-  assign br_tgt_jmp_o = tag_mem_data_lo.v & tag_mem_data_lo.jmp;
+  logic tag_mem_data_v_lo;
+  bsg_dff_reset
+   #(.width_p(1))
+   r_v_reg
+    (.clk_i(clk_i)
+     ,.reset_i(reset_i)
+
+     ,.data_i(tag_mem_r_v_li)
+     ,.data_o(tag_mem_data_v_lo)
+     );
+
+  assign br_tgt_v_o   = tag_mem_data_v_lo & tag_mem_data_lo.v & (tag_mem_data_lo.tag == r_tag_r);
+  assign br_tgt_jmp_o = tag_mem_data_v_lo & tag_mem_data_lo.v & tag_mem_data_lo.jmp;
   assign br_tgt_o     = tag_mem_data_lo.tgt;
 
```
</details>

bp_common/src/include/bp_common_aviary_cfg_pkgdef.svh
```diff
@@ -188,6 +188,9 @@
     // Whether to emulate FPU
```
WasabiFan, line 188: Am I missing something, or are neither `muldiv_support` nor `fpu_support` used anywhere?

dpetrisko: Not missing anything. If you try to change them, there's an assertion to stop you

dpetrisko: You can see there's a PR for MULH support where this was intended to be used

bp_common/src/include/bp_common_core_if.svh
```diff
@@ -134,9 +139,24 @@
     typedef struct packed                                                                          \
     {                                                                                              \
       bp_pte_leaf_s              pte_leaf;                                                         \
-      logic [`bp_fe_cmd_itlb_map_padding_width(paddr_width_mp, asid_width_mp, branch_metadata_fwd_width_mp)-1:0] \
+      logic [vaddr_width_mp-1:0] fill_vaddr;                                                       \
```
dpetrisko, line 142: fill_vtag? Don't think we need the full address here

bp_common/src/include/bp_common_rv64_csr_defines.svh
WasabiFan: Done. Bit is suppressed on write so I don't have to change the field widths.

bp_common/syn/Makefile.vcs
WasabiFan, line 75: Still TODO to revert before merge

bp_fe/src/v/bp_fe_pc_gen.sv
```diff
@@ -264,11 +292,38 @@ module bp_fe_pc_gen
   bp_fe_instr_scan
    #(.bp_params_p(bp_params_p))
    instr_scan
-    (.instr_i(fetch_i)
+    (.instr_i(fetch_instr_o)
 
      ,.scan_o(scan_instr)
      );
 
+  generate
```
dpetrisko, line 300: don't use 'generate' explicitly, it's deprecated and tool support is buggy. Make sure to label your blocks so that they come out with predictable names in the ASIC flow.

```
if (compressed_support_p)
  begin : fi1
    bp_fe_realigner ...
  end
else
  begin : fi1
    ...
  end
```

This will always come out as fe.pc_gen.fi1.realigner whereas before it could be (among other things) fe.pc_gen.genblk7.realigner

bp_fe/src/v/bp_fe_pc_gen.sv
```diff
@@ -241,18 +267,20 @@ module bp_fe_pc_gen
   assign br_tgt_lo  = pc_if2_r + scan_instr.imm;
   assign fetch_pc_o = pc_if2_r;
 
+  wire [vaddr_width_p-1:0] branch_prediction_source_addr_if2 = `bp_align_addr_up(pc_if2_r, rv64_instr_width_bytes_gp);
```
dpetrisko, line 270: What's the reason for aligning these up? isn't the truncated PC in IF2 what actually indexed into the predictor?

bp_fe/src/v/bp_fe_pc_gen.sv
dpetrisko: not sure, will have to look at this closely

bp_be/src/v/bp_be_checker/bp_be_director.sv
```diff
@@ -158,11 +158,11 @@ module bp_be_director
         end
       else if (commit_pkt_cast_i.itlb_fill_v)
         begin
-          // TODO: restart vs resume
-          fe_cmd_li.opcode                               = e_op_itlb_fill_restart;
+          fe_cmd_li.opcode                               = (compressed_support_p & commit_pkt_cast_i.exception_instr_upper_not_lower_half) ? e_op_itlb_fill_resume : e_op_itlb_fill_restart;
```
dpetrisko, line 161: Think we can get away without these cases. If we do not enable compressed support in the FE, constant propagation should take care of these control bits

bp_fe/src/v/bp_fe_realigner.sv
```diff
@@ -0,0 +1,69 @@
...
+   , input                           poison_i
+   , input                           restore_lower_half_v_i
+   , input [instr_half_width_gp-1:0] restore_lower_half_i
+
+   , output [instr_width_gp-1:0] fetch_instr_o
+   , output                      fetch_instr_v_o
+   , output                      fetch_is_second_half_o
+   );
+
+  wire [instr_half_width_gp-1:0] icache_data_lower_half_li = fetch_data_i[instr_half_width_gp-1:0];
```
dpetrisko, line 34: nit; BSG style is to always part-select, so:
fetch_data_i[0+:instr_half_width_gp]
fetch_data_i[instr_half_width_gp+:instr_half_width_gp]

bp_fe/src/v/bp_fe_pc_gen.sv
```diff
@@ -264,11 +292,38 @@ module bp_fe_pc_gen
...
+        #(.bp_params_p(bp_params_p))
+        realigner
+        (.clk_i(clk_i)
+        ,.reset_i(reset_i)
+
+        ,.fetch_pc_i    (pc_if2_r)
+        ,.fetch_data_i  (fetch_i)
+        ,.fetch_data_v_i(fetch_v_i)
+
+        ,.poison_i              (realigner_poison_if2_n)
```
dpetrisko, line 312: realigner_poison_if2_n -> realigner_poison_if1_r, no need for the temp variable I think?

bp_be/src/include/bp_be_ctl_pkgdef.svh
```diff
@@ -250,6 +250,9 @@
     logic dtlb_fill;
     logic _interrupt;
     logic cmd_full;
+
+    // Metadata for instruction exceptions (instr_page_fault,instr_access_fault, itlb_miss, icache_miss)
+    logic upper_not_lower_half;
```
dpetrisko, line 255: Would prefer to keep this metadata in bp_be_pipe_sys

bp_be/src/v/bp_be_calculator/bp_be_pipe_mem.sv
```diff
@@ -250,6 +250,7 @@ module bp_be_pipe_mem
     '{instr_miss_v  : commit_pkt.itlb_miss
       ,store_miss_v : commit_pkt.dtlb_store_miss
       ,load_miss_v  : commit_pkt.dtlb_load_miss
+      ,instr_upper_not_lower_half: commit_pkt.exception_instr_upper_not_lower_half
```
dpetrisko, line 253: Would prefer partial_v throughout to keep door open for superscalar

bp_fe/src/v/bp_fe_pc_gen.sv
```diff
@@ -28,17 +28,23 @@ module bp_fe_pc_gen
...
 
    , output logic                                    ovr_o
 
    , input [instr_width_gp-1:0]                      fetch_i
-   , input                                           fetch_instr_v_i
-   , input                                           fetch_exception_v_i
+   , input                                           fetch_v_i
+   , output [instr_width_gp-1:0]                     fetch_instr_o
+   , output                                          fetch_instr_v_o
+   , input                                           fetch_instr_yumi_i
```
dpetrisko, line 44: I haven't looked too closely at this but this is backpressure on a non-stalling pipeline. What happens here when yumi_i is low?

bp_fe/src/v/bp_fe_pc_gen.sv
dpetrisko: Maybe sketch out what the alternative would be?

bp_fe/src/v/bp_fe_pc_gen.sv
dpetrisko: When you say it's ignored under stalls, you mean that the if1/if2 valid bit gates the predictions from being taken?

dpetrisko (commented):
Some comments, really will have to dig into the control flow changes...

bp_fe/src/v/bp_fe_pc_gen.sv
WasabiFan: This signals whether the fetch was actually issued. If it wasn't, we need to avoid updating internal state (e.g. the branch prediction delay reg). The FE used to mask the validity of the I$ fetch output when it wasn't going to issue that fetch, but now there's a distinction for issue and exception purposes whether we're just waiting on a second half or a fetch has entirely failed.

Now, this means I've named a signal `yumi` but it doesn't provide backpressure. It signals whether the output was used, but isn't a proper handshake.

It seemed this was the norm in this FE because e.g. `next_pc_yumi_li` does the same. But if it isn't appropriate or we ought to be moving away from it I can rename it to something else.

WasabiFan:
I don't know why GitHub isn't allowing a direct reply to these last two comments. Replying here.

> Maybe sketch out what the alternative would be?

This would be:

```
assign next_fetch_linear = next_pc == pc_if1_r + 4;
```

Instead of identifying it via which source we pulled `next_pc` from. I currently believe your preference was to have redirects explicitly trigger realigner poisons which I think is more in line with what I've done in the current PR version.

> When you say it's ignored under stalls, you mean that the if1/if2 valid bit gates the predictions from being taken?

Yeah, the `r_v_li` signal for both the BTB and BHT incorporates `if2_v_n` or similar (right now). So it disables reading the SRAMs when it knows that prediction isn't going to be issued. The minor bug is that the frontend still _uses the output_ of the memory to predict jumps even when `r_v_li` is low and the instruction won't be issued. So you have a sequence of e.g. ten cycles where the frontend is stalled or waiting, but it's in a loop blindly predicting jumps to the last thing it predicted a jump to.

The only effects I can see this having are on power and trace clarity. The outputs are probably toggling a fair amount of the other branch prediction logic even though they don't need to, and it _looks_ strange when the FE predicts this bogus in a trace of predictions. It's a very minor thing, so long as there isn't a bug which inadvertently uses the invalid output.

bp_fe/src/v/bp_fe_pc_gen.sv
WasabiFan: ## Short answer

We are saving branch prediction metadata according to the aligned word in which the _final byte_ of the instruction resides, not the _first byte_. This is the primary change vs. the previous instance of this PR (#986) -- rather than storing it with the first half and conditionally delaying the branch prediction state, we store it with the second half.

## Long answer/behavioral explanation

In IF0, it computes the fetch address as:

```
  assign next_fetch_o = incomplete_fetch_if1_n
                        ? `bp_align_addr_down(next_pc, rv64_instr_width_bytes_gp)
                        : `bp_align_addr_up(next_pc, rv64_instr_width_bytes_gp);
```

In other words, the fetch address trails the PC for the first fetch and then precedes it for the rest.

That fetch address is sent into the I$, MMU and BHT in IF0 and otherwise not saved, because it isn't strictly needed -- the fetch address is constructed such that we fetch aligned words in the right order (a linear fetch starting at PC-2). The instruction PC is saved, but the fetch address isn't. This just saves regs if you don't need it.

Logically, the goal is now to use what we had originally computed for `next_fetch_o` in IF0, to compute the relevant indices for branch prediction metadata in IF2.  Since this data is only applicable if we are issuing an instruction, and not if we're doing the "incomplete" first fetch after a redirect, we can ignore the first branch -- it is never applicable to the half of an instruction that allows you to issue it. So rather than recomputing the align down/align up thing above, we just do an align up.

We could equivalently feed `next_fetch_o` through pipeline regs and slice it up in IF2, it's just redundant when you have the PC.

bp_fe/src/v/bp_fe_pc_gen.sv
WasabiFan: This would be:

```
assign next_fetch_linear = next_pc == pc_if1_r + 4;
```

Instead of identifying it via which source we pulled `next_pc` from. I currently believe your preference was to have redirects explicitly trigger realigner poisons which I think is more in line with what I've done in the current PR version.

bp_fe/src/v/bp_fe_pc_gen.sv
WasabiFan: Yeah, the `r_v_li` signal for both the BTB and BHT incorporates `if2_v_n` or similar (right now). So it disables reading the SRAMs when it knows that prediction isn't going to be issued. The minor bug is that the frontend still _uses the output_ of the memory to predict jumps even when `r_v_li` is low and the instruction won't be issued. So you have a sequence of e.g. ten cycles where the frontend is stalled or waiting, but it's in a loop blindly predicting jumps to the last thing it predicted a jump to.

The only effects I can see this having are on power and trace clarity. The outputs are probably toggling a fair amount of the other branch prediction logic even though they don't need to, and it _looks_ strange when the FE predicts this bogus in a trace of predictions. It's a very minor thing, so long as there isn't a bug which inadvertently uses the invalid output.

bp_fe/src/v/bp_fe_pc_gen.sv
dpetrisko: got it, that makes sense

bp_fe/src/v/bp_fe_pc_gen.sv
dpetrisko: >We are saving branch prediction metadata according to the aligned word in which the final byte of the instruction resides, not the first byte.

Thought we were doing the true PC of the final byte of the instruction?
i.e.
IF0: PC=0x6 (ignore BTB), IF1: PC = 0x2, IF2: PC = 0xxx.  BTB: gets 0x06, I$: gets 0x02 but drops bits to fetch 0x00

IF0: PC=BTB[0x06]=0x84, IF1: PC = 0x06, IF2: PC = 0x2

Then the metadata should match because they're both real PCs and not doubly aligned? Do we think that there is an aliasing reason to use the aligned address for prediction? FWIW, it looks like other cores use the true PC

bp_fe/src/v/bp_fe_pc_gen.sv
dpetrisko: no keep it, it's useful as an interface for an alternate BP scheme

bp_fe/src/v/bp_fe_pc_gen.sv
dpetrisko: I think we could mask it at the usage like this? The advantage of not doing it in the predictors is that the current model is desired for a stalling pipeline, where you want the output to last until the stall is over

```
- assign btb_taken = btb_br_tgt_v_lo & (bht_pred_lo | btb_br_tgt_jmp_lo);
+ assign btb_taken = v_if1_r & btb_br_tgt_v_lo & (bht_pred_lo | btb_br_tgt_jmp_lo);
```

WasabiFan: Yes, we can do that. There are three or four places it would need to be done. The difference is just whether you want to have the invalid data floating around for people to make mistakes. But the stalling pipeline makes sense and is a good reason to go with masking at the usages.

bp_fe/src/v/bp_fe_pc_gen.sv
WasabiFan: I believe the suggestion here is to use the address `pc+4` for the instruction at `pc` rather than `round_down(pc+4)`. The latter is what we currently do. Is that right?

Yes, it could be done and I don't think there's an inherent drawback. Right now the lower two bits are ignored for BTB and BHT indexing. If we want to index into them with a misaligned vaddr, we'll need to decide what it does with the lower bits. If we shift the fields over by one bit we'll be wasting a bunch of BTB entries unless alignment is usually a 50/50 mix. I know various strategies exist, we'd just have to pick one. Is that something we want to do?

The benefit to making that change would be we can remove the rounding up. That doesn't _seem_ like a huge benefit, but if that's something we want to do it isn't a big change.

bp_be/src/v/bp_be_checker/bp_be_director.sv
```diff
@@ -158,11 +158,11 @@ module bp_be_director
...
         begin
-          // TODO: restart vs resume
-          fe_cmd_li.opcode                               = e_op_itlb_fill_restart;
-          fe_cmd_li.npc                                  = commit_pkt_cast_i.npc;
+          fe_cmd_li.opcode                               = (compressed_support_p & commit_pkt_cast_i.exception_instr_upper_not_lower_half) ? e_op_itlb_fill_resume : e_op_itlb_fill_restart;
+          fe_cmd_li.npc                                  = commit_pkt_cast_i.vaddr;
           fe_cmd_li.operands.itlb_fill_response.pte_leaf = commit_pkt_cast_i.pte_leaf;
-          fe_cmd_li.operands.itlb_fill_response.instr    = commit_pkt_cast_i.instr;
+          fe_cmd_li.operands.itlb_fill_response.instr    = compressed_support_p ? commit_pkt_cast_i.instr : '0;
+          fe_cmd_li.operands.itlb_fill_response.vtag     = commit_pkt_cast_i.vaddr[vaddr_width_p-1-:vtag_width_p]; // TODO: dedupe?
```
WasabiFan, line 165: TODO. vtag field is now duplicative of npc field.

bp_fe/src/v/bp_fe_pc_gen.sv
```diff
@@ -231,20 +255,39 @@ module bp_fe_pc_gen
      ,.data_i({pred_if2_n, pc_if2_n})
      ,.data_o({pred_if2_r, pc_if2_r})
      );
-  assign return_addr_n = pc_if2_r + vaddr_width_p'(4);
+  assign return_addr_n        = fetch_pc_o + vaddr_width_p'(4);
+  assign if2_second_half_addr = pc_if2_r + vaddr_width_p'(4);
```
WasabiFan, line 259: Might be able to combine these. Still TODO.

bp_fe/src/v/bp_fe_pc_gen.sv
```diff
@@ -231,20 +255,39 @@ module bp_fe_pc_gen
...
   assign ovr_taken  = btb_miss_br & ((is_br & pred_if2_r.pred) | is_jal);
-  assign ovr_o      = ovr_taken | ovr_ret;
-  assign br_tgt_lo  = pc_if2_r + scan_instr.imm;
-  assign fetch_pc_o = pc_if2_r;
+  assign ovr_half   = misaligned_fetch_nonbr & pc_if2_misaligned & fetch_v_i & !fetch_instr_v_o;
+  assign ovr_o      = ovr_taken | ovr_ret | ovr_half;
+  assign br_tgt_lo  = fetch_pc_o + scan_instr.imm;
+  assign fetch_resume_pc_o = pc_if2_r;
+
+`ifndef SYNTHESIS
```
WasabiFan, line 273: Is this how I ought to do the debug-only counter?

bp_fe/src/v/bp_fe_top.sv
```diff
@@ -272,11 +295,11 @@ module bp_fe_top
...
-  assign icache_pkt = '{vaddr: next_pc_lo
+  assign icache_pkt = '{vaddr: next_fetch_lo
                         ,op  : icache_fence_v ? e_icache_fencei : icache_fill_response_v ? e_icache_fill : e_icache_fetch
                         };
   // TODO: Should only ack icache fence when icache_ready
-  wire icache_v_li = next_pc_yumi_li | icache_fence_v;
+  wire icache_v_li = next_fetch_yumi_li | icache_fence_v;
   logic [instr_width_gp-1:0] icache_data_lo;
   logic icache_ready_lo, icache_data_v_lo, icache_miss_v_lo;
   logic icache_poison_tl;
```
WasabiFan, line 305: Still need to reconfigure BTB and BHT indexing to use the extra bit.

bp_fe/src/v/bp_fe_pc_gen.sv
dpetrisko: prefer to add to the pc_gen_profiler and bind in

bp_be/src/v/bp_be_checker/bp_be_scheduler.sv
```diff
@@ -204,6 +205,7 @@ module bp_be_scheduler
       dispatch_pkt = '0;
       dispatch_pkt.v        = (fe_queue_yumi_li & ~poison_isd_i) || be_exc_not_instr_li;
       dispatch_pkt.queue_v  = fe_queue_yumi_li;
+      dispatch_pkt.instr_v  = instr_v_li;
```
dpetrisko, line 208: Use fe_instr_not_exc_li here instead of instr_v_li.

and then in detector, can replace .queue_v & .instr_v with just .instr_v

bp_be/src/v/bp_be_checker/bp_be_director.sv
dpetrisko: makes sense. I think originally this was intended to solve the "resume different than fill" problem, but fill_resume/restart does the same thing

bp_be/src/v/bp_be_checker/bp_be_scheduler.sv
```diff
@@ -223,12 +225,15 @@ module bp_be_scheduler
         fe_exc_not_instr_li & fe_queue_lo.msg_type inside {e_itlb_miss};
       dispatch_pkt.exception.icache_miss        |=
         fe_exc_not_instr_li & fe_queue_lo.msg_type inside {e_icache_miss};
+      dispatch_pkt.exception.upper_not_lower_half |=
```
dpetrisko, line 228: let's stick with partial_v rather than upper_not_lower_half throughout.  48b instructions are on the horizon and then we won't have to rename

bp_be/src/v/bp_be_checker/bp_be_director.sv
WasabiFan: Yeah, it was necessary as a separate field until the most recent revision in the last few days and I haven't yet removed the field.

bp_be/src/v/bp_be_checker/bp_be_scheduler.sv
WasabiFan: 👍 this is one of the remaining TODOs from the previous review round I haven't addressed yet

bp_fe/src/v/bp_fe_pc_gen.sv
```diff
@@ -28,17 +29,24 @@ module bp_fe_pc_gen
    , input                                           redirect_br_taken_i
    , input                                           redirect_br_ntaken_i
    , input                                           redirect_br_nonbr_i
+   , input                                           redirect_restore_insn_lower_half_v_i
```
dpetrisko, line 32: redirect_resume_v_i
redirect_resume_instr_i

to keep the terminology consistent

bp_fe/src/v/bp_fe_pc_gen.sv
```diff
@@ -60,32 +68,47 @@ module bp_fe_pc_gen
...
-  assign pc_if1_n = next_pc_o;
+        next_pc           = btb_br_tgt_lo;
+    else begin
+        next_pc           = pc_plus4;
+        next_pc_nonlinear = 1'b0;
+    end
+  end
+
+  assign pc_if1_n = next_pc;
+  assign next_fetch_o = `bp_align_addr_down(next_pc, rv64_instr_width_bytes_gp);
```
dpetrisko, line 109: let the I$ handle this transparently by passing the true PC

bp_fe/src/v/bp_fe_pc_gen.sv
dpetrisko: Yeah as long as there's a valid signal, garbage data is expected. Garbage data without a valid just ignored implicitly (like we have currently) is bad

WasabiFan:
TODO: If possible, it would be good to add an assert in the FE<->BE interface that verifies all (pc, instr) pairs are correct. This would require out-of-band access to the simulated DRAM.

bp_be/src/v/bp_be_checker/bp_be_scheduler.sv
WasabiFan: I tried this and it degraded performance on the virtual memory test program. I'm going to leave it as-is for now but might revisit later.

bp_fe/src/v/bp_fe_pc_gen.sv
WasabiFan: Is this referring to pc_gen_tracer?

For the purposes of the utility of that tracer, I'd _prefer_ not including incrementing counters in there, since they make it no longer diffable (changes earlier in the trace affect the whole rest of the trace). Is there another place it might make sense? I _can_ put it there if really preferred.

bp_be/src/v/bp_be_checker/bp_be_scheduler.sv
dpetrisko: Hm okay, just put a comment

WasabiFan:
Remaining work, sorted in descending order of impact:
- [ ] BHT+BTB indexing: https://github.com/black-parrot/black-parrot/pull/1098#discussion_r976284717
- [ ] Move exception bit into be_pipe_sys
- [ ] Finalize nonsynth counter of overrides: https://github.com/black-parrot/black-parrot/pull/1098#discussion_r976271510
- [ ] Verify functionality with misaligned insns disabled
- [ ] Cleanup to exception handling in virtual memory test program
- [ ] Revert debug_all in Makefile

WasabiFan:
Some questions @dpetrisko 

---

The BHT and BTB are being fed with potentially misaligned addresses now. Currently they then cut off the lower two bits. There are some aliasing conditions which would be improved by avoiding this truncation. If I just switch it to dropping only one bit, it effectively halves the capacity of these tables for aligned instructions. What should I do here?

---

Question here about nonsynth counter: https://github.com/black-parrot/black-parrot/pull/1098#discussion_r1000179356

WasabiFan:
Also, @dpetrisko do you know how to do hierarchical refs in Verilator when going through a named begin/end block? It seems I've broken the Verilator sim:

```
%Error: testbench.sv:576:67: Can't find definition of 'realigner' in dotted scope/variable: 'pc_gen.fetch_instr_generation.realigner'
  576 |            ,.realigner_v_i         (pc_gen.fetch_instr_generation.realigner.half_buffer_v_r)
      |                                                                   ^~~~~~~~~
        ... Known scopes under 'pc_gen.fetch_instr_generation': <no instances found>
%Error: testbench.sv:577:67: Can't find definition of 'realigner' in dotted scope/variable: 'pc_gen.fetch_instr_generation.realigner'
  577 |            ,.realigner_pc_i        (pc_gen.fetch_instr_generation.realigner.fetch_instr_pc_r)
      |                                                                   ^~~~~~~~~
        ... Known scopes under 'pc_gen.fetch_instr_generation': <no instances found>
%Error: testbench.sv:578:67: Can't find definition of 'realigner' in dotted scope/variable: 'pc_gen.fetch_instr_generation.realigner'
  578 |            ,.realigner_instr_i     (pc_gen.fetch_instr_generation.realigner.half_buffer_r)
      |                                                                   ^~~~~~~~~
        ... Known scopes under 'pc_gen.fetch_instr_generation': <no instances found>
%Error: Exiting due to 3 error(s)
```

bp_fe/src/v/bp_fe_top.sv
dpetrisko: Has this been done?

bp_fe/src/v/bp_fe_pc_gen.sv
```diff
@@ -64,7 +72,9 @@ module bp_fe_pc_gen
   // IF0
   /////////////////////////////////////////////////////////////////////////////////////
   logic [ghist_width_p-1:0] ghistory_n, ghistory_r;
+
   logic [vaddr_width_p-1:0] next_pc;
+  logic next_pc_nonlinear;
```
dpetrisko, line 77: unused?

bp_fe/src/include/bp_fe_defines.svh
```diff
@@ -29,5 +29,18 @@
       logic [ghist_width_mp-1:0]      ghist;                                                      \
     }  bp_fe_branch_metadata_fwd_s;
 
+  `define bp_addr_is_aligned(addr_mp, num_bytes_mp) \
+    (!(|{ addr_mp[$clog2(num_bytes_mp)-1:0] }))
+
+  `define bp_align_addr_down(addr_mp, num_bytes_mp) \
```
dpetrisko, line 35: Are bp_align_addr_down/up used?

bp_fe/src/v/bp_fe_pc_gen.sv
```diff
@@ -290,19 +314,76 @@ module bp_fe_pc_gen
...
+    end
+  else
+    begin : fetch_instr_generation
+      assign fetch_pc_o      = pc_if2_r;
+      assign fetch_instr_o   = fetch_i;
+      assign fetch_instr_v_o = fetch_v_i;
+      assign fetch_is_second_half_o = 0;
+    end
+
+`ifndef SYNTHESIS
```
dpetrisko, line 373: Move to pc_gen tracer

bp_fe/src/v/bp_fe_pc_gen.sv
```diff
@@ -290,19 +314,76 @@ module bp_fe_pc_gen
      );
 
   // Override calculations
+  wire taken_jump_site_if2 = taken_if1_r || ovr_btaken || ovr_jmp || ovr_ret;
+  wire pc_if2_misaligned = !`bp_addr_is_aligned(pc_if2_r, rv64_instr_width_bytes_gp);
   wire btb_miss_ras = pc_if1_r != ras_tgt_lo;
+  // TODO: a jump that skips only one instruction within misaligned code can erroneously pass this check
```
dpetrisko, line 320: what's the code sequence? Will it cause an error? If so, add a test that exposes it

bp_fe/src/v/bp_fe_pc_gen.sv
```diff
@@ -75,25 +85,34 @@ module bp_fe_pc_gen
...
       begin
         next_pred  = redirect_br_taken_i;
         next_taken = redirect_br_taken_i;
         next_pc    = redirect_pc_i;
+        next_pc_nonlinear = !redirect_resume_v_i;
 
         next_metadata = redirect_br_metadata_fwd;
       end
-    else if (ovr_o)
+    else if (ovr_half) begin
```
dpetrisko, line 110: ovr_half->ovr_ntaken. This path may be used in the future for multicycle branch prediction

bp_fe/src/v/bp_fe_pc_gen.sv
```diff
@@ -75,25 +85,34 @@ module bp_fe_pc_gen
   ///////////////////////////
   bp_fe_branch_metadata_fwd_s next_metadata, ovr_metadata;
   logic next_pred, next_taken;
-  logic ovr_ret, ovr_btaken, ovr_jmp, btb_taken;
+  logic ovr_ret, ovr_btaken, ovr_jmp, ovr_half, btb_taken;
   logic [vaddr_width_p-1:0] pc_plus4;
   logic [vaddr_width_p-1:0] ras_tgt_lo;
   logic [vaddr_width_p-1:0] br_tgt_lo;
+  logic [vaddr_width_p-1:0] if2_second_half_addr;
```
dpetrisko, line 92: if2_second_half_addr -> pc_plus4_if2

The convention is base_op_stage

bp_common/src/include/bp_common_core_if.svh
```diff
@@ -211,7 +209,7 @@
     (        1+branch_metadata_fwd_width_mp)
 
   `define bp_fe_cmd_itlb_map_width_no_padding(vaddr_width_mp, paddr_width_mp) \
-    (`bp_pte_leaf_width(paddr_width_mp)+vaddr_width_mp-page_offset_width_gp+instr_width_gp)
+    (`bp_pte_leaf_width(paddr_width_mp)+instr_width_gp)
```
dpetrisko, line 212: Isn't the vtag different for resume/restart commands? Or am I missing something?

bp_fe/src/v/bp_fe_realigner.sv
```diff
@@ -0,0 +1,81 @@
...
+  (input   clk_i
+   , input reset_i
+
+   , input [vaddr_width_p-1:0]   fetch_pc_i
+   , input [instr_width_gp-1:0]  fetch_data_i
+   , input                       fetch_data_v_i
+
+    // poison_v_i takes precedence over store_v_i
+    // restore_lower_half_v_i takes precedence over poison_i
+   , input store_v_i
```
dpetrisko, line 25: Let's try to (logically) reduce the number of asynchronous interfaces here. I think we only need 2 in and 1 out.

store_v_i is always associated with fetch_data_v_i, right? How about fetch_store_i?

poison_v_i and restore are always associated with redirects. How about restore_poison_i?

I would also restore_lower_half->restore for superscalar fetch
Same with fetch_is_second_half -> fetch_partial

bp_fe/src/v/bp_fe_top.sv
WasabiFan: No, see: https://github.com/black-parrot/black-parrot/pull/1098#issuecomment-1285129401

bp_common/src/include/bp_common_core_if.svh
WasabiFan: It is. For FE-resumable exceptions, the "npc" field of the response is the "vaddr" that was used to perform the ptw, which is computed in the BE to conditionally be (pc+2).

bp_fe/src/v/bp_fe_top.sv
```diff
@@ -355,48 +378,57 @@ module bp_fe_top
...
+                                        ? e_instr_page_fault
+                                        : instr_access_fault_r
+                                          ? e_instr_access_fault
+                                            : e_icache_miss;
         fe_queue_cast_o.branch_metadata_fwd = fetch_br_metadata_fwd_lo;
-        // TODO: Partially fetched exceptions
-        fe_queue_cast_o.partial_v = 1'b0;
+        // For now, we leave the upper half of the instr forward field zeroed. Once "C" support is
+        // finished this would contain the full realigner buffer.
+        fe_queue_cast_o.instr     = compressed_support_p ? { (instr_half_width_gp)'(0), realigner_instr_stored_half_lo } : '0;
```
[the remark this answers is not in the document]
WasabiFan: Is there a mux here? I don't see one -- compressed_support_p is a constant so I'd expect the whole reg to be optimized out if it isn't set.

bp_fe/src/v/bp_fe_pc_gen.sv
WasabiFan: No, my belief is that this just causes a poor prediction. A conditional branch by "+4" correctly predicted as taken will result in the same next fetch pc as if it were incorrectly predicted as not-taken. This means they're indistinguishable by this check and thus won't trigger an override/cause a predictor correction. I haven't implemented a proof-of-concept here or validated that it's truly possible, I just suspect it. But I don't see a way it could cause corruption.

bp_common/src/include/bp_common_core_if.svh
dpetrisko: Oh, I do remember this. For posterity (myself). Because we pass the partial instructions, whatever PC we are resuming at is implicitly the same as the vtag we're filling. Thanks!

bp_fe/src/v/bp_fe_pc_gen.sv
dpetrisko: think that's the "correct" behavior in this degenerate case right? Triggering an override will cause an unnecessary bubble. It's a less efficient use of BTB space but it's the same type of case as a regular branch to PC+4. If we get rid of it in the BTB, the BHT+branch identification will still predict it as taken and put it back in the BTB

dpetrisko:
> The BHT and BTB are being fed with potentially misaligned addresses now. Currently they then cut off the lower two bits. There are some aliasing conditions which would be improved by avoiding this truncation. If I just switch it to dropping only one bit, it effectively halves the capacity of these tables for aligned instructions. What should I do here?

Good question! For C extension, it's claimed that misaligned instructions happen at the same frequency as aligned instructions. So I would say parameterize with compressed_support_p? Without compressed support => 2 bits, with compressed => 1?

bp_fe/src/v/bp_fe_top.sv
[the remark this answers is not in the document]
dpetrisko: You're totally right with this statement itself. I'm referring to zero-ing instr for exception and passing instr otherwise

bp_fe/src/v/bp_fe_top.sv
dpetrisko: replied in outer thread

dpetrisko:
```
>Also, @dpetrisko do you know how to do hierarchical refs in Verilator when going through a named begin/end block? It seems I've broken the Verilator sim:

%Error: testbench.sv:576:67: Can't find definition of 'realigner' in dotted scope/variable: 'pc_gen.fetch_instr_generation.realigner'
  576 |            ,.realigner_v_i         (pc_gen.fetch_instr_generation.realigner.half_buffer_v_r)
      |                                                                   ^~~~~~~~~
        ... Known scopes under 'pc_gen.fetch_instr_generation': <no instances found>
%Error: testbench.sv:577:67: Can't find definition of 'realigner' in dotted scope/variable: 'pc_gen.fetch_instr_generation.realigner'
  577 |            ,.realigner_pc_i        (pc_gen.fetch_instr_generation.realigner.fetch_instr_pc_r)
      |                                                                   ^~~~~~~~~
        ... Known scopes under 'pc_gen.fetch_instr_generation': <no instances found>
%Error: testbench.sv:578:67: Can't find definition of 'realigner' in dotted scope/variable: 'pc_gen.fetch_instr_generation.realigner'
  578 |            ,.realigner_instr_i     (pc_gen.fetch_instr_generation.realigner.half_buffer_r)
      |                                                                   ^~~~~~~~~
        ... Known scopes under 'pc_gen.fetch_instr_generation': <no instances found>
%Error: Exiting due to 3 error(s)
```

That's correct; what's happening is it's trying to evaluate the 'else' block where that instance does not exist. Could you use the signals in those blocks rather than inspecting into the conditionally instantiated module? That should fix the issue

bp_fe/src/v/bp_fe_pc_gen.sv
WasabiFan: The instruction sequence I'm imagining is:
- 10: Branch to 18 (taken)
- 14: Whatever
- 18: Whatever

If the jump isn't caught by the BHT, and an override isn't triggered, then we're going to retain the data in the realigner and thus issue 14 before issuing 18. The sequence of "PC" values in the pipeline is the same regardless of whether we override or not, but overriding suppresses issuing of address 14 whereas just continuing as-is does not.

bp_fe/src/v/bp_fe_realigner.sv
WasabiFan: I am having trouble interpreting the suggestions. Are you suggesting these pairs of signals be merged?

> store_v_i is always associated with fetch_data_v_i, right? How about fetch_store_i?

`fetch_data_v_i` is used on its own for the validity of the output (`fetch_instr_v_o`) -- the realigner output is only valid with a valid input fetch. But one does not have to be storing the upper half for the lower half to be valid.

> poison_v_i and restore are always associated with redirects. How about restore_poison_i?

These signals convey one of the following three things at any time:
- Normal execution, fetches incoming, no redirects
- A redirect is occurring and the new stored realigner data should be marked as valid
- A redirect is occurring and the stored state should be cleared

So as-is I don't see a simplification to combine them into one bit. It might be possible to convey the invalidation by artificially asserting `fetch_instr_yumi_i`, but that doesn't seem straightforward to me.

For reference, the connections to these ports are defined as:

```
        ,.poison_v_i (redirect_v_i & ~redirect_resume_v_i)

        ,.restore_lower_half_v_i         (redirect_resume_v_i)
```

They are not negations of each other.

bp_fe/src/v/bp_fe_top.sv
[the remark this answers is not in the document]
WasabiFan: Oh, I see the suggestion. Yes, I could do that instead. The advantage here to me is just clarity and extensibility. The upper half of the current fetch is to become the lower half of the issued instruction, so I muxed it to put the lower half "where it should be" and explicitly zero the upper. I think this fits better with a model in which we had more granularity of partial data and a mask.

WasabiFan:
The BTB/BHT indexing change reduced performance by a handful of cycles on my test programs. I don't think that's too surprising (given _no compressed instructions_ == _strictly half the capacity_) but it's a bit unfortunate given there are a bunch of calls/returns in one of them which I know alias.

bp_fe/src/v/bp_fe_pc_gen.sv
WasabiFan: Resolved via Slack. I'm including the important message below (in screenshot form, sadly, due to Slack's incompetence) for future reference. f395ccc90ed3899bc3b84dcf5f9964aa0289e9d5 is the fix proposed by Dan and I have confirmed it works.

image

WasabiFan:
# Verification methodology

## Test programs

Aside from normal aligned code tests, I've developed three new misaligned test programs to aid in verification for this PR. The first two must only execute successfully and with the correct output to "pass", but also output performance measurements which should be verified manually. The third does not have a meaningful performance score but evaluates correctness as it runs.

For a quick confirmation that nothing drastic has changed, I've also tended to take note of the total cycle count of a simulation run. In general if the cycle count is the same behavior couldn't be much different.

### `misaligned_instructions_basic_jumps`

This program tests _unconditional_ jumps in the following scenarios:

- Aligned -> unaligned unconditional jump, mispredicted as not taken (overridden)
- Unaligned -> unaligned unconditional jump, mispredicted as not taken (overidden)
- Unaligned -> aligned unconditional jump, mispredicted as not taken (overidden)
- Unaligned -> aligned call, return back to unaligned code (call mispredicted as not taken)

#### Expected output

 `Hello Misaligned World!` should be printed. There should then be a hexadecimal integer that follows it. This is the number of cycles taken to execute the relevant part of the program.

Ideally, what you verify here is the _delta_ between this number running misaligned code and this number running aligned code. **This should be done by recompiling the sample program with the `ENABLE_MISALIGNMENT` define disabled, _not_ toggling `compressed_support_p`, as the latter also affects the BTB and BHT.** The increase in cycles would ideally be 19 -- the minimum number of additional cycles required to execute this many jumps to misaligned code, due to double-fetches. Since we have chosen to implement a retroactive override mechanism for correcting erroneous jumps in the middle of an instruction, we lose more cycles to overrides. But the expectation is that the total increase is on that order of magnitude.

As of writing:
- The aligned program gives `0x1fc = 508` cycles
- The misaligned program gives `0x225 = 549` cycles

That's an increase of 41 cycles, which is _fine_. I've verified in the past that behavior is as expected. Confirming this number is fairly straightforward; diff the pc_gen_tracer output between these two runs (see below) to see where the added delays are.

### `misaligned_instructions_advanced_jumps`

This program tests conditional jumps and branch prediction. It runs the same sequence of conditional jumps twice, to train the predictors one way, and then swaps one of the branch results and re-runs the sequence.

#### Expected output

It should print the following:

```
12
12
123
123
```

As with the other program, it will then print a cycle count for the critical portion of execution. And also as with before, the proper way to evaluate this number is relative to the same program and processor configuration with instructions aligned. We expect the measured portion to take 22 cycles longer for misaligned code, barring other mispredicts.

Currently:
- Aligned: `0xb3 = 179`
- Misaligned: `0xd2 = 210`

That's a 31-cycle increase in time taken. As with before, I've confirmed operation is as intended and this can be re-verified by textually diffing the tracer output.

### `misaligned_instructions_virtual_memory`

This program verifies the correctness of exception conditions and the reported `*epc` and `*tval`. Running it and having it terminate successfully is sufficient to confirm that it "passes". However, I have personally inspected the sequence of TLB and cache misses for each case to confirm that they operate as expected, using the pc_gen tracer.

## General technique: diffing pc_gen_tracer output

The new tracer has been vital to verifying that changes have the desired effect when working on the FE. There is a comment at the top of `bp_fe_nonsynth_pc_gen_tracer.sv` which explains its output. It is immensely helpful to diff the output of this tracer. However, since there are a handful of cumulative counters in the raw prints, a one-cycle change effects every line that follows. For this reason I tend to postprocess it to remove the counters:

```
awk -F, -v OFS=, '{$1 = ""; $6 = ""; print }' pc_gen_trace_0.trace 
```

bp_fe/src/v/bp_fe_realigner.sv
dpetrisko: Same number of signals, collecting them into 2 buses to more clearly show the association.  

```
    // Fetch PC and I$ data
    , input                       fetch_v_i
   // Store leftover bits of instruction in the realigner
    , input                       fetch_store_i
    , input [vaddr_width_p-1:0]   fetch_pc_i
    , input [instr_width_gp-1:0]  fetch_data_i
```

```
  // Redirection from backend
  , input                           redirect_v_i
  // Whether to restore the instruction data
  , input                           redirect_resume_i
  , input [instr_half_width_gp-1:0] redirect_partial_i
  , input [vaddr_width_p-1:0]       redirect_vaddr_i
```

- Redirects always take precedence over fetches because they are non-speculative
- On any redirection, we clear the current data in the realigner + update the vaddr
- On resume operations, we store the partial instruction data

I believe this aligns with your current semantics

WasabiFan: I believe I've made this change as requested

dpetrisko (approved):
Excellent work, thanks! Onward to C extension

PR #1100, opened 2022-07-16
Fixing misaligned instruction handling (should be in backend)

### LINKED ISSUE #216: Instruction misaligned exception should raise on branch, not fetch
image

According to the newly ratified spec, misaligned exceptions happen on branches, not their target fetches. 

Two fixes:
1) Raise the exception in the correct place
2) implement compressed instructions, which preclude instruction misaligned exceptions by construction

dpetrisko:
## Summary
This PR fixes the handling of misaligned instructions. Originally, we would raise a misaligned exception upon fetching a misaligned instruction. However, the correct spec behavior is to actually raise the exception on the instruction which causes the misalignment (i.e. the branch).

## Issue Fixed
#216 

## Area
FE-BE interface (danger zone), and exception pipelines.

To comply with the privileged spec:
image

When we implement compressed instructions this check must be disabled entirely

## Analysis
No PPA impact, higher compliance.

## Verification
Run with: https://github.com/black-parrot-sdk/bp-tests/pull/32 to verify that trap is taken as expected


WasabiFan (approved):
LGTM, can confirm it works when branching to a misaligned PC.

PR #1103, opened 2022-08-05
Fix cfg bus addressing to be 64b

dpetrisko:
time. A few details to check:

## Summary
The PR changes the addressing of the config bus to be 64b aligned.

## Issue Fixed
None

## Area
CFG

AXI-lite uses full bus width reads/writes which implies that we need to support the maximum width alignment to be compatible with the maximum width access.

Perch/SDK needs to change to pick up these new addresses

## Analysis
Every change here is risky. We should really pick a solution like SystemRDL to unify the hardware/firmware/software

## Verification
Simple regression to see if config bus picks up new addresses.

## Additional Context
https://github.com/black-parrot-sdk/perch/pull/11
https://github.com/black-parrot-sdk/black-parrot-sdk/pull/49


muwyse (approved):
LGTM

PR #1104, opened 2022-08-09
Only pre-decode issue messages, not exceptions

WasabiFan:
## Summary
The pre-decode logic in the FE queue currently doesn't check whether the issue packet is for an exception or an instruction issue. It just uses the `instr` field of the instruction fetch variant of the fetch/exception union.

This "works" so long as those bits happen to map to zeroes in the exception message; an all-zeroes opcode, reg fields, etc. triggers no hazard conditions nor register reads. The danger is that some in-use bit of the exception payload could put something legitimate in the opcode and pre-decode to an operation which isn't inert.

The effect I've seen is to the hazard detection logic; exceptions can be delayed waiting for arbitrary other instructions due to phantom hazards. It also causes the regfile memories to be read unnecessarily. I haven't traced those signals through the issue packet to know where else they might cause trouble.

I'm finding that, in simulators at least, the current union layout isn't triggering the issue. It's putting garbage into the `rs3_addr` but all the validity signals map to padding in the exception message type which happens to be zero. So, thankfully, it doesn't appear to be having negative impact given the standard union layout. But once you add a handful of bits to an exception message, they start introducing strange side-effects.

## Issue Fixed

No issue.

## Area

`bp_be_issue_queue`.

**Note:** I'm somewhat selfishly opening this PR against `fe_dev` since my other PR into `fe_dev` will want to depend on it. I can re-target it if desired.

Insidious source of bugs in the future.

None.

## Analysis

Introduces some extra logic for "and"ing and a comparison.

Note that I've included zeroing the register address fields. We could probably get by without it so long as the corresponding validity fields are masked appropriately. If the extra 15 bits of mask logic are worth avoiding I can move those outside the "if".

## Verification

Seems to work in VCS.

### DIFF
--- a/bp_be/src/v/bp_be_checker/bp_be_issue_queue.sv
+++ b/bp_be/src/v/bp_be_checker/bp_be_issue_queue.sv
@@ -152,72 +152,74 @@ module bp_be_issue_queue
     begin
       issue_pkt_li = '0;
 
-      // Pre-decode
-      issue_pkt_li.csr_v = instr.opcode inside {`RV64_SYSTEM_OP};
-      issue_pkt_li.mem_v = instr.opcode inside {`RV64_FLOAD_OP, `RV64_FSTORE_OP
-                                                ,`RV64_LOAD_OP, `RV64_STORE_OP
-                                                ,`RV64_AMO_OP, `RV64_SYSTEM_OP
-                                                };
-      issue_pkt_li.fence_v = instr inside {`RV64_FENCE, `RV64_FENCE_I, `RV64_SFENCE_VMA};
-      issue_pkt_li.long_v = instr inside {`RV64_DIV, `RV64_DIVU, `RV64_DIVW, `RV64_DIVUW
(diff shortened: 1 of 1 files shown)


WasabiFan:
For review, remember that you can enable "hide whitespace" in the GitHub UI via the gear icon in the changes view.

dpetrisko:
Love it! Power+performance bug in one.

>Note that I've included zeroing the register address fields. We could probably get by without it so long as the corresponding validity fields are masked appropriately. If the extra 15 bits of mask logic are worth avoiding I can move those outside the "if".

Technically, we could keep the register addresses floating but probably save power with datagating like this (even though the regfile SRAMs are small, they are duplicated for ports). Toggling the input ports of SRAMs (or flops) can cause unnecessary power draw.

>Insidious source of bugs in the future.

I do believe that the predecode bits are only used for regfile reading and hazard detection _for this instruction_. So the worst that could happen is we wait a really long time to issue this exception. Which not great, but not catastrophic. Better to clear though!

PR #1105, opened 2022-08-10
Mask WAW hazards when issuing exceptions

WasabiFan:
Same deal as #1104, but now in the scheduler when detecting register write collisions. The "true" instruction decode had the same issue as the pre-decode.

I don't see other instances of this in the scheduler module. The `iwb_v` and `fwb_v` signals are only used for scoreboard hazard detection as far as I can tell. My test cases suggest this is the last instance of unnecessary issue delays, but I can't guarantee there aren't other cases where the decoded instruction was used by mistake.

### DIFF
--- a/bp_be/src/v/bp_be_checker/bp_be_instr_decoder.sv
+++ b/bp_be/src/v/bp_be_checker/bp_be_instr_decoder.sv
@@ -27,6 +27,7 @@ module bp_be_instr_decoder
    , localparam decode_info_width_lp = `bp_be_decode_info_width
    )
   (input [instr_width_lp-1:0]           instr_i
+   , input logic                        instr_v_i
    , input [decode_info_width_lp-1:0]   decode_info_i
 
    , output logic [decode_width_lp-1:0] decode_o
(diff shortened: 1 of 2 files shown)


dpetrisko:
Nominally, the decoded instr is zeroed here, so anything downstream will see no active control signals. https://github.com/black-parrot/black-parrot/blob/2f03f0230456fd9a833a4e7dae09a21cb802b91a/bp_be/src/v/bp_be_checker/bp_be_scheduler.sv#L210

The place you grabbed from snipes the decoded instr after decode, but before it's zero'd out for invalidity. As long as there's nothing else that uses that intermediate value, it should be okay. The safer way to handle it would probably be to add a valid signal to the decode which zeros out the control signals if it's an exception so there's no intermediate value to accidentally grab from

WasabiFan:
Yeah, I'd be happy to add validity to the instruction decode. That would be cleaner and less prone to error.

WasabiFan:
I've updated it to include a validity signal into the instruction decoder. Note that I opted to not remove the (now unnecessary) "and" ing of `fe_instr_not_exc_li` in the packet construction logic, since I felt the clarity on what state came from the FE vs the BE was worth the redundancy. Happy to change if something else is preferred.

bp_be/src/v/bp_be_checker/bp_be_scheduler.sv
```diff
@@ -141,7 +141,8 @@ module bp_be_scheduler
   bp_be_instr_decoder
    #(.bp_params_p(bp_params_p))
    instr_decoder
-    (.instr_i(fe_queue_lo.msg.fetch.instr)
+    (.instr_i       (fe_queue_lo.msg.fetch.instr)
+     ,.instr_v_i    (fe_queue_lo.msg_type == e_fe_fetch)
```
dpetrisko, line 145: There's still the decode = (fe_exc be_exc, etc) in the always_comb block. I would take that out and make the valid signal:

wire instr_v_li = fe_queue_v_li & ~fe_exc_not_instr_li & ~be_exc_not_instr_lii

This also includes the fe_queue_v_li signal which should prevent further unnecessary toggling

dpetrisko: And we can get rid of this line as well:

dispatch_pkt.decode   = (fe_exc_not_instr_li || be_exc_not_instr_li || illegal_instr_lo) ? '0 : instr_decoded;

bp_be/src/v/bp_be_checker/bp_be_instr_decoder.sv
```diff
@@ -71,559 +72,563 @@ module bp_be_instr_decoder
...
-              `RV64_SUB, `RV64_SUBW : decode_cast_o.fu_op = e_int_op_sub;
-              `RV64_SLL, `RV64_SLLW : decode_cast_o.fu_op = e_int_op_sll;
-              `RV64_SRL, `RV64_SRLW : decode_cast_o.fu_op = e_int_op_srl;
-              `RV64_SRA, `RV64_SRAW : decode_cast_o.fu_op = e_int_op_sra;
-              `RV64_SLT             : decode_cast_o.fu_op = e_int_op_slt;
-              `RV64_SLTU            : decode_cast_o.fu_op = e_int_op_sltu;
-              `RV64_XOR             : decode_cast_o.fu_op = e_int_op_xor;
-              `RV64_OR              : decode_cast_o.fu_op = e_int_op_or;
-              `RV64_AND             : decode_cast_o.fu_op = e_int_op_and;
+      if (instr_v_i) begin
```
dpetrisko, line 77: Let's add this gate explicitly to bottom as 

if (~instr_v_li | illegal_instr_o)
  decode_cast_o = '0;

so that we capture the illegal instr zero-ing as well

WasabiFan:
I believe I've updated it as suggested. Verified that various bp-tests programs still work and their PC trace is the same as before.

bp_be/src/v/bp_be_checker/bp_be_instr_decoder.sv
```diff
@@ -71,560 +72,565 @@ module bp_be_instr_decoder
...
```
dpetrisko, line 77: with the gate below, we can revert this if statement

WasabiFan: `illegal_instr_o` and `imm_o` are also set in this `if`, which I intended to mask. Would you prefer I put them after the block to override the result rather than nesting this in the `if`?

dpetrisko: Got it, if that's the intention then the big if is fine. L633 should then just be if (illegal_instr_o) decode_cast_o = '0

PR #1106, opened 2022-08-15
Refactor fill network / cmd network multiplexing

dpetrisko:
## Summary
This PR refactors the LCE->$ interface implementation, specifically multiplexing the LCE cmd and LCE fill interfaces into a single port. In order to do this, we subsume the LCE fill definitions into a subset of LCE cmd.

## Issue Fixed
No issue, but related to #1034 

## Area
LCE

The cache fill interface is typically critical as it feeds directly into large SRAMs. By multiplexing the signals before they get to LCE cmd / LCE fill, we're able to make so there's only a single write/read port into the cache. No deadlock dependencies have changed since the shared resources of the cache SRAMs remain the same 

None

## Analysis
From other tapeouts, we know this interface is critical so this is a preemptive PPA optimization

## Verification
Standard multicore tests

## Additional Context
None


bp_common/src/include/bp_common_bedrock_if.svh
```diff
@@ -101,13 +101,7 @@
...
     } bp_bedrock_lce_cmd_payload_s;                                                    \
                                                                                        \
-    typedef struct packed                                                              \
-    {                                                                                  \
-      bp_coh_states_e                              state;                              \
-      logic [`BSG_SAFE_CLOG2(lce_assoc_mp)-1:0]    way_id;                             \
-      logic [cce_id_width_mp-1:0]                  src_id;                             \
-      logic [lce_id_width_mp-1:0]                  dst_id;                             \
-    } bp_bedrock_lce_fill_payload_s;                                                   \
+    typedef bp_bedrock_lce_cmd_payload_s bp_bedrock_lce_fill_payload_s;                \
```
muwyse, line 104: Does making the fill and cmd network payloads identical have any implications for other parts of the design and/or potential future uses of the networks?

bp_me/src/v/network/bp_me_xbar_burst.sv
```diff
@@ -50,89 +50,143 @@ module bp_me_xbar_burst
...
+  logic [num_source_p-1:0][data_width_p-1:0] msg_data_li;
+  logic [num_source_p-1:0][lg_num_sink_lp-1:0] msg_dst_li;
+  logic [num_source_p-1:0] msg_data_v_li, msg_data_yumi_lo, msg_last_li;
+
+  logic [num_source_p-1:0][lg_num_sink_lp-1:0] msg_dst_r;
+  logic [num_source_p-1:0] src_is_data_r;
+  logic [num_source_p-1:0] cb_valid_li, cb_yumi_lo;
+  logic [num_source_p-1:0][lg_num_sink_lp-1:0] cb_sel_li;
+
+  for (genvar i = 0; i < num_source_p; i++)
```
muwyse, line 64: Add comment in header of file that the inputs are buffered.

bp_me/src/v/network/bp_me_xbar_burst.sv
```diff
@@ -50,89 +50,143 @@ module bp_me_xbar_burst
...
+         ,.data_i({msg_dst_i[i], msg_has_data_i[i], msg_header_i[i]})
+         ,.v_i(msg_header_v_i[i])
+         ,.ready_o(msg_header_ready_and_o[i])
+ 
+         ,.data_o({msg_dst_li[i], msg_has_data_li[i], msg_header_li[i]})
+         ,.v_o(msg_header_v_li[i])
+         ,.yumi_i(msg_header_yumi_lo[i])
+         );
+
+      bsg_two_fifo
```
muwyse, line 81: A 2-fifo for data does not guarantee that the xbar can buffer even one whole message, depending on the width of the data channel and size of message. This is not needed for correctness of the xbar since we are not saying that the xbar should have that functionality. Perhaps add comment at top saying buffers are for timing isolation / correctness only, not protocol level buffering guarantees.

muwyse (approved):
LGTM. A question and a couple comment requests.

This should be safe so long as:
- fill has static priority over command
- LCE does not process a command unless it has outbound buffer space for a fill message (worst case is command causes a fill and a response - shouldn't need buffer for response since these are always sunk by CCE and CCE implementation guarantees that responses won't cause back pressure into the network from the CCE)

bp_common/src/include/bp_common_bedrock_if.svh
dpetrisko: I could see us move towards having all messages have the same payload to simplify even further, but I think it's unlikely that cmd and fill will need to diverge to specialize

bp_me/src/v/network/bp_me_xbar_burst.sv
dpetrisko: Added comment!

PR #1111, opened 2022-08-22
Introduce compressed_support_p config parameter

WasabiFan:
Currently unused and defaults to disabled.

Looking for this to go in before the main PR.

Precursor to #1098.

### DIFF
--- a/bp_common/src/include/bp_common_aviary_cfg_pkgdef.svh
+++ b/bp_common/src/include/bp_common_aviary_cfg_pkgdef.svh
@@ -307,7 +307,7 @@
       ,fe_cmd_fifo_els   : 4
       ,muldiv_support    : (1 << e_div) | (1 << e_mul)
       ,fpu_support       : 1
-      ,compressed_support: 0
+      ,compressed_support: 1
 
       ,async_coh_clk       : 0
       ,coh_noc_flit_width  : 128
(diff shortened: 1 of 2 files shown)


WasabiFan:
Swapped it to enabled by default because it looks like the ROM tests have default values hard-coded.

dpetrisko:
Can we add 
```
typedef logic [1:0]
{
  e_misaligned_instr_support
  ,e_misaligned_data_support
  ,e_compressed_instr_support
} bp_compressed_support_e
```

Defaulting to (1 << e_misaligned_instr_support) is fine. We can disable the aviary test in CI for now, we'll just put a TODO: to fix before merging FE->DEV. Still not sure what the overhead is here and if having this as default is fine

WasabiFan:
What would `e_misaligned_data_support` do?

WasabiFan:
Oh, also, this would imply you can select multiple variants at once. Is that really desirable? It seems to me that supporting three options -- none, misaligned, compressed -- would model the available configurations better.

dpetrisko:
Misaligned data would be D$ which supports misaligned accesses (required for some hypervisor implementations). 

I think that answers the other question, which is that we may support I$ vs D$ alignment independently in an RV64IAZamisalignedatomics configuration. But you're right that support will likely be monotonic I$ -> C -> D$ in practice

WasabiFan:
I think the thing I'm missing is why the "compressed_support" config parameter affects D$ alignment. Wouldn't the D$ alignment be a toggle independent of the compressed instruction support level? They seem unrelated.

dpetrisko:
Yeah, I guess "alignment support" is probably more what I'm going for, and is orthogonal. You know, let's shelve this until the rest of the support is there anyway

PR #1112, opened 2022-08-22
Clarification for FE<->BE interface

dpetrisko:
## Summary
This PR prepares the core FE<->BE interface for #1098. Mostly renaming changes, but also adds the partial instr and message types to support it

## Issue Fixed
None

## Area
FE<->BE interface, and everything that touches those

Better to separate out these interface changes so that the functional diff is much smaller and easier to review

None

## Analysis

## Verification

## Additional Context

### DIFF
--- a/bp_common/src/include/bp_common_core_if.svh
+++ b/bp_common/src/include/bp_common_core_if.svh
@@ -44,6 +44,7 @@
       logic [vaddr_width_mp-1:0]                pc;                                                \
       rv64_instr_s                              instr;                                             \
       logic [branch_metadata_fwd_width_mp-1:0]  branch_metadata_fwd;                               \
+      logic                                     partial_v;                                         \
       bp_fe_queue_type_e                        msg_type;                                          \
     }  bp_fe_queue_s;                                                                              \
                                                                                                    \
@@ -167,7 +168,7 @@
 
   /* Declare width macros so that clients can use structs in ports before struct declaration */
   `define bp_fe_queue_width(vaddr_width_mp, branch_metadata_fwd_width_mp)                          \
-    ($bits(bp_fe_queue_type_e)+branch_metadata_fwd_width_mp+rv64_instr_width_gp+vaddr_width_mp)
+    ($bits(bp_fe_queue_type_e)+branch_metadata_fwd_width_mp+rv64_instr_width_gp+1+vaddr_width_mp)
 
   `define bp_fe_cmd_width(vaddr_width_mp, paddr_width_mp, asid_width_mp, branch_metadata_fwd_width_mp) \
     (vaddr_width_mp                                                                                \
--- a/bp_common/src/include/bp_common_core_if.svh
+++ b/bp_common/src/include/bp_common_core_if.svh
@@ -30,50 +30,21 @@
   `define declare_bp_core_if(vaddr_width_mp, paddr_width_mp, asid_width_mp, branch_metadata_fwd_width_mp) \
     /*                                                                                             \
      *                                                                                             \
-     * bp_fe_fetch_s contains the pc/instruction pair, along with additional                       \
-     * information including virtual address. The branch metadata is for the branch                \
+     * bp_fe_queue_s contains the pc/instruction pair, along branch metadata for the branch        \
      * predictor to update its internal data based on feedbacks from the BE as to                  \
      * whether this particular PC/instruction pair was correct.  The BE does not look              \
      * at the branch metadata, since this would mean that the BE implementation is                 \
      * tightly coupled to the FE implementation.                                                   \
+     * Exceptions serviced inline with instructions. Otherwise we have no way of knowing if this   \
+     * exception is eclipsed by a preceding branch mispredict. Therefore, we support exceptions as \
+     * alternate message types.                                                                    \
      */                                                                                            \
     typedef struct packed                                                                          \
     {                                                                                              \
       logic [vaddr_width_mp-1:0]                pc;                                                \
       rv64_instr_s                              instr;                                             \
       logic [branch_metadata_fwd_width_mp-1:0]  branch_metadata_fwd;                               \
-                                                                                                   \
-      logic [`bp_fe_fetch_padding_width(vaddr_width_mp, branch_metadata_fwd_width_mp)-1:0]         \
-                                                padding;                                           \
-    }  bp_fe_fetch_s;                                                                              \
-                                                                                                   \
-    /*                                                                                             \
-     *                                                                                             \
-     * bp_fe_exception_s contains FE exception information.  Exceptions should be                  \
-     * serviced inline with instructions. Otherwise we have no way of knowing if this              \
-     * exception is eclipsed by a preceding branch mispredict.  FE does not receive                \
-     * interrupts, but may raise exceptions.                                                       \
-     */                                                                                            \
-    typedef struct packed                                                                          \
-    {                                                                                              \
-      logic [vaddr_width_mp-1:0] vaddr;                                                            \
-      bp_fe_exception_code_e     exception_code;                                                   \
-      logic [`bp_fe_exception_padding_width(vaddr_width_mp, branch_metadata_fwd_width_mp)-1:0]     \
-                                 padding;                                                          \
-    }  bp_fe_exception_s;                                                                          \
-                                                                                                   \
-    /*                                                                                             \
-     * bp_fe_queue_s is the struct that the FE feeds the queue connecting the FE to                \
-     * the BE bp_fe_queue_type_e specifies which type of information is forwarded to               \
-     * the backend, choosing between bp_fe_fetch_s or bp_fe_exception_s.                           \
-     */                                                                                            \
-    typedef struct packed {                                                                        \
-      bp_fe_queue_type_e  msg_type;                                                                \
-      union packed                                                                                 \
-      {                                                                                            \
-        bp_fe_fetch_s       fetch;                                                                 \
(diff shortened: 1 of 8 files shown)


bp_common/src/include/bp_common_core_pkgdef.svh
```diff
@@ -69,11 +49,16 @@
...
-    ,e_icache_miss        = 3
-  } bp_fe_exception_code_e;
+    e_itlb_miss            = 0
+    ,e_instr_page_fault    = 1
+    ,e_instr_access_fault  = 2
+    ,e_icache_miss         = 3
+    // This doesn't particularly scale to superscalar, but we can break this out to a mask
+    //   when that time comes.
+    ,e_instr_partial_lower = 4
+    ,e_instr_partial_upper = 5
```
WasabiFan, line 59: fwiw, right now I only ever do partial_lower.

bp_common/src/include/bp_common_core_pkgdef.svh
```diff
@@ -2,52 +2,32 @@
...
+   * e_op_icache_fill_restart happens when icache non-speculatively misses
+   * e_op_icache_fill_resume happens when icache non-speculatively misses and refetches
    * e_op_icache_fence happens when there is flush in the icache.
-   * e_op_itlb_fill_response happens when itlb populates translation.
+   * e_op_itlb_fill_restart happens when itlb populates translation and restarts fetching
+   * e_op_itlb_fill_resume happens when itlb populates translation and resumes fetching 
    * e_op_itlb_fence issues a fence operation to itlb.
    */
-  typedef enum logic [2:0]
+  typedef enum logic [3:0]
```
WasabiFan, line 19: Is there a way we can more clearly document which of these are at-fault/which trigger predictor updates?

bp_common/src/include/bp_common_core_pkgdef.svh
```diff
@@ -2,52 +2,32 @@
...
    * the cases of an exception. bp_fe_command_queue_opcodes_e explains the reason
    * of why pc is redirected.
    * e_op_state_reset is used after the reset, which flushes all the states.
    * e_op_pc_redirection defines the changes of PC, which happens during the branches.
    * e_op_attaboy informs the frontend that the prediction is correct.
+   * e_op_icache_fill_restart happens when icache non-speculatively misses
+   * e_op_icache_fill_resume happens when icache non-speculatively misses and refetches
    * e_op_icache_fence happens when there is flush in the icache.
-   * e_op_itlb_fill_response happens when itlb populates translation.
+   * e_op_itlb_fill_restart happens when itlb populates translation and restarts fetching
```
WasabiFan, line 15: What's the intended use-case for these? First half is a restart and second half is a resume?

WasabiFan (commented):
Looks good to me. Some minor naming differences vs. what I used but it doesn't matter. The blurring of a "partial" instruction vs. a "real" instruction is a bit spooky to me, but it still seems apt and there isn't a huge room for confusion.

What's your intent re: merging this? Go for it as-is without FE command changes?

When I implemented the metadata forwarding it had a negative performance impact so I put it aside to investigate later; I figure it was a bug/oversight but haven't circled back to it. But I think it's worth profiling this and verifying the forwarding behaves as expected.

bp_common/src/include/bp_common_core_pkgdef.svh
WasabiFan: Ok, that seems fair to me. It should have a clean mapping to what my misaligned instructions PR currently does.

bp_common/src/include/bp_common_core_pkgdef.svh
dpetrisko: I believe only (a subset of) pc-redirect should trigger predictor updates. It comes from EX1 as the PC is verified. Everything else comes in EX3 as the instruction is actually committed

bp_common/src/include/bp_common_core_if.svh
```diff
@@ -30,50 +30,22 @@
...
-     * the backend, choosing between bp_fe_fetch_s or bp_fe_exception_s.                           \
-     */                                                                                            \
-    typedef struct packed {                                                                        \
-      bp_fe_queue_type_e  msg_type;                                                                \
-      union packed                                                                                 \
-      {                                                                                            \
-        bp_fe_fetch_s       fetch;                                                                 \
-        bp_fe_exception_s   exception;                                                             \
-      }  msg;                                                                                      \
+      logic                                     partial_v;                                         \
```
WasabiFan, line 47: I assume the plan here is to just ignore this bit for fetch messages, yeah?

bp_common/src/include/bp_common_core_if.svh
```diff
@@ -130,25 +102,40 @@
...
      * frontend sends the page fill request to the backend, the backend performs the               \
      * page walk, and responds to the frontend with bp_fe_cmd_itlb_map_s.                          \
-    */                                                                                             \
+     */                                                                                            \
     typedef struct packed                                                                          \
     {                                                                                              \
       bp_pte_leaf_s              pte_leaf;                                                         \
-      logic [`bp_fe_cmd_itlb_map_padding_width(paddr_width_mp, asid_width_mp, branch_metadata_fwd_width_mp)-1:0] \
+      logic [vaddr_width_mp-page_offset_width_gp-1:0]                                              \
+                                 vtag;                                                             \
```
WasabiFan, line 110: Where should I compute this? Is it fine to do the manual slicing of the vaddr in the director, or should it be elsewhere?

bp_common/src/include/bp_common_core_if.svh
```diff
@@ -130,25 +102,40 @@
...
+      logic [instr_width_gp-1:0] instr;                                                            \
```
WasabiFan, line 111: I'll need validity of this instr data somewhere in the CMD. Right now because we only ever use the lower half, one bit is sufficient.

bp_common/src/include/bp_common_core_pkgdef.svh
WasabiFan: Speaking of which: isn't it possible that an instruction is issued at the same time as it is committed? The director has the attaboy and redirect cases at the end of the "if" chain, meaning they're the lowest-priority messages. Are there cases where they just get trumped by other messages?

bp_common/src/include/bp_common_core_if.svh
WasabiFan: Oh, right, I forgot about the resume/restart message types. That should be sufficient.

bp_common/src/include/bp_common_core_if.svh
dpetrisko: Will be relevant for 16b fetches only, ignored for now

bp_common/src/include/bp_common_core_pkgdef.svh
dpetrisko: Yes, that's the common case for exceptions. If the exception in EX3, you don't want to redirect to the mispredicted branch because now you're in a whole new basic block

bp_common/src/include/bp_common_core_pkgdef.svh
dpetrisko: fixed my bad understanding offline

bp_common/src/include/bp_common_core_if.svh
dpetrisko: yeah manual slicing is fine (free)

bp_common/src/include/bp_common_core_pkgdef.svh
WasabiFan: Is the pipeline timing intentionally designed so that this _doesn't_ happen for normal instruction issues? Or is it sometimes possible for an attaboy/redirect to just get lost due to being trumped by something else? I guess, something being issued in EX3 is rare, and really only for exceptional control flow, so it doesn't matter?

dpetrisko:
>What's your intent re: merging this? Go for it as-is without FE command changes?

Let's rebase your changes on top of this if it's not too much trouble. This is such a finicky change so I think we need small diffs

bp_common/src/include/bp_common_core_if.svh
WasabiFan: It's more an aesthetics/component responsibilities question. If slicing the tag out of a vaddr is acceptable in the director I'm good with it (and it sounds like it is).

bp_common/src/include/bp_common_core_pkgdef.svh
dpetrisko: Everything is issued in ISD, including interrupts and exceptions. So everything in EX3 is older by construction. It's analogous to EX1 flushing the front end, we don't really care what those instructions are doing once we realize that we've made a mistake

bp_common/src/include/bp_common_core_if.svh
dpetrisko: Ah gotcha, yeah aesthetics-wise I don't think there's much else to do, since we get the full vaddr from the commit_pkt

WasabiFan:
I'm good with that. Should I do so before you merge this PR (i.e., to verify it "works" with the misaligned instructions implementation) or after? After is easier from a merge perspective but I can do either.

bp_common/src/include/bp_common_core_pkgdef.svh
WasabiFan: That makes sense, thanks!

WasabiFan:
Question answered! Will update #1098.

PR #1119, opened 2022-08-31
Rename cmd/resp to fwd/rev

dpetrisko:
This PR renames the mem_cmd/mem_resp networks to mem_fwd/mem_rev the dram network to dma. The motivation for this change is to uniquify the network channels to LCE: [req, cmd, fill, resp]
MEM/IO: [fwd, rev]
DMA: [dma].

This makes it easier to reason about the networking components, since things like lce_cmd/resp cannot be confused with mem_cmd/mem_resp

NOTE: This is a (small) breaking change to downstream implementations as they need to update their wrappers to adapt


dpetrisko:
I suppose it makes sense to patch the CCE ucode to indicate this change, though it's not breaking e.g. legacy code will alias mem_fwd to mem_cmd

PR #1121, opened 2022-09-02
Adding stream pumps to LCE/CCE

dpetrisko:
## Summary
This PR updates LCE and CCE to use stream pumps instead of manually managing the protocol in FSMs

## Issue Fixed
None

## Area
LCE/CCE

This is part of a larger changeset to update the networks to be entirely burst+. This is necessary long term to have a unified bedrock network protocol so that it is easy to integrate accelerators and other SoC components seamlessly.

It appears that the CCE ucode may require changes to support this model of simultaneous header and data. Otherwise, this should be an invisible change

## Analysis
This should have only positive PPA impact as we can optimize the stream pumps for timing and area, simultaneously improving all submodules 

## Verification
Standard multicore regression, should also boot multicore linux to stress test the system

## Additional Context
Depends on #1120  #1119


muwyse (commented):
The ucode CCE in existing designs "pops" the header information into internal registers in order to process the arriving message. This doesn't actually dequeue the header though. Then, if a message has data, the push instruction is issued, which dequeues the data from the inbound message as each outbound message sends. After this, the header of the inbound message can be dequeued safely.

I think this can be adapted as follows: poph instruction still "pops" header information into internal registers for processing. popq is needed if the message had no data and the ucode needs to dequeue the single header-only beat. Otherwise, popq shouldn't be needed after a pushq since the stream pumps effectively present the CCE with a series of header+data beats, each of which is consumed by the pushq command.

muwyse:
ME, top RISC-V, and multicore regressions all passing when running with ucode from here: https://github.com/black-parrot-sdk/bedrock/pull/7

muwyse (approved):
I believe this change is ready to merge to me_dev. Local runs of regressions appear to pass, but it would be good to trigger a full regression in CI if possible.

muwyse:
@dpetrisko merge to me_dev when ready

PR #1127, opened 2022-10-12
Explicitly have a non-zero range in bp_me_addr_to_cce_id.sv

flaviens:
Some widely-used tool has difficulties with line and throws an error saying that the slice width is zero.

## Area
`bp_me/src/v/network/bp_me_addr_to_cce_id.sv`

## Verification
Ran the tool again and the problem went away.

## Additional Context
This seems to have been a problem given the similar syntax of the line below
https://github.com/black-parrot/black-parrot/blob/6ce6523d6b3ab91cdb355d53c419f1b9ee1e1e81/bp_me/src/v/network/bp_me_addr_to_cce_id.sv#L78


dpetrisko:
Thanks, change looks good. Can we add a comment saying which tool/version had the issue? We like to keep track of what code we should try as tools get updated

dpetrisko:
LGTM, merging into dev to be part of the next master release

PR #1128, opened 2022-10-13
Fix fatal syntax

flaviens:
Hi there!

The fatal syntax here was incorrect and made the Morty tool unhappy (cf. [this issue](https://github.com/pulp-platform/morty/issues/46)). This should fix it. Not tested.


dpetrisko:
LGTM. I'm merging this into dev, will be part of the upcoming master release

PR #1129, opened 2022-10-13
Add FuseSoC cores

flaviens:
## Summary
Add FuseSoC cores

## Issue Fixed
None

## Area
The toplevel *.core files

Getting rid of the git submodules and have more clean and easy dependencies

This relies on [this PR](https://github.com/adithyasunil26/basejump_stl_cores/pull/8). We may put the core file in question in this (black-parrot) repo in the end.

## Analysis
I don't think it's great. One reason is [this issue](https://github.com/olofk/fusesoc/issues/589#issue-1402379560), and in general it feels slower than just using git submodules. Also, it may not be easier to maintain.

## Verification
It compiled with a commercial simulator.

### DIFF
--- a/bp_be.core
+++ b/bp_be.core
@@ -1,0 +1,213 @@
+CAPI=2:
+
+name: ::bp_be:0-r1
+description: Black-parrot back-end
+
+filesets:
+  src:
+    files:
+      # Include files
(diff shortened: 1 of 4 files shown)


dpetrisko:
Very cool! Curious on the response to the fusesoc issue

PR #1131, opened 2022-10-14
Replace els_1g_p by els_1g_lp in for loop

flaviens:
It looks that data_1g_high_r may remain undriven in the trivial case where `els_1g_p` < 1. I dont't think it matters so much but Verilator threw a fatal warning regarding it.

## Verification
Not tested.

## Alternatives

Alternatively we could surround it with
```
// verilator lint_off UNDRIVEN
// verilator lint_on UNDRIVEN
```
but it's less elegant.

Another alternative is to do an if/else with the parameter as condition, but it's not elegant either.

## Further remarks

Verilator users could simply suppress all UNDRIVEN warnings with `Wno-UNDRIVEN` but I have the feeling that the UNDRIVEN warnings are usually meaningful.


dpetrisko:
Thanks, good to clean up those kinds of warnings. I'd prefer to be explicit about driving the signals, but not instantiating extra hardware.

```
if (els_1g_p == 0) data_1g_high_r = '0;
else
for ...
```

Does that work?

flaviens:
Updated :+1:

dpetrisko:
Thanks!

PR #1137, opened 2022-11-09
Changes fe_queue_ready_o signal to ready_and interface

BrendenPage:
## Summary
Renames fe_queue_ready_o signal to fe_queue_ready_and_o as the signal is used in a way consistent with the rv->& interface.

## Issue Fixed
None

## Area
BE, FE, Top

Clarifies usage of the ready signal from bp_be_issue_queue and makes the interface more consistent with the style guide.

## Verification
Built with Verilator with no issue, no functional changes so only linting required.

## Additional Context
We see that it is rv->& from its [usage within the bp_be_issue_queue module](https://github.com/black-parrot/black-parrot/blob/master/bp_be/src/v/bp_be_checker/bp_be_issue_queue.sv#L49) as well as the [generation of the ready signal](https://github.com/black-parrot/black-parrot/blob/master/bp_be/src/v/bp_be_checker/bp_be_issue_queue.sv#L119). It can be noted that clr is directed from a signal labelled as valid, but it does not represent a transaction of data, and is more in line with a halting signal which is [generated here](https://github.com/black-parrot/black-parrot/blob/master/bp_be/src/v/bp_be_checker/bp_be_director.sv#L139).

### DIFF
--- a/bp_be/src/v/bp_be_checker/bp_be_issue_queue.sv
+++ b/bp_be/src/v/bp_be_checker/bp_be_issue_queue.sv
@@ -21,7 +21,7 @@ module bp_be_issue_queue
 
    , input [fe_queue_width_lp-1:0]          fe_queue_i
    , input                                  fe_queue_v_i
-   , output logic                           fe_queue_ready_o
+   , output logic                           fe_queue_ready_and_o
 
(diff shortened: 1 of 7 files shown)


dpetrisko (approved):
LGTM!

PR #1152, opened 2023-01-12
Fixing FMA precision for IEEE exact-rounding compliance

### LINKED ISSUE #611: Cosim mismatch: fpu returns a 0 while a -0 is expected
A cosim mismatch occurs at the me_dev branch, in the test of whetstone, beebs. And the corresponding instruction is fmul.d fa2, fa4, fa2. The fpu returns a 0 while a -0 should be returned.
Screen Shot 2020-08-21 at 8 46 53 PM

dpetrisko:
### Summary
This PR adds an additional truncation to SP ops. This eliminates extra precision accumulation in the FPU which can cause IEEE non-compliance

### Issue Fixed
#611, possibly more

### Area
FPU

IEEE compliance is necessary for compatibility with various benchmarks

None

### Analysis
No timing impact, 0.1% area increase, which is not great but we could reduce in the future by some clever reuse of rounding arithmetic

### Verification
Normal FPU regression + https://github.com/black-parrot-sdk/bp-tests/pull/34

### Additional Context
Depends on https://github.com/bsg-external/HardFloat/pull/18


taylor-bsg:
Can you describe qualitative what the key approach is here -- where are you doing the truncation, and for which operators? How do things like sticky bits and guard digits apply?

dpetrisko:
See here for a more detailed explanation: https://github.com/bsg-external/HardFloat/pull/18

We do truncation after FMA and FDIV/FSQRT operations. We do not do truncation after moves or conversions from fp to int/mem. This is consistent with the set of ops described in IEEE:

```
The IEEE standard goes further than just requiring the use of a guard digit. It gives an algorithm for addition, subtraction, multiplication, division and square root, and requires that implementations produce the same result as that algorithm. Thus, when a program is moved from one machine to another, the results of the basic operations will be the same in every bit if both machines support the IEEE standard. This greatly simplifies the porting of programs. Other uses of this precise specification are given in Exactly Rounded Operations.

From https://docs.oracle.com/cd/E19957-01/806-3568/ncg_goldberg.html
```

Specifically in the pipeline, this is done after the final result has been calculated. This does put the re-rounding on the critical path for those modules, but I've confirmed that even in aggressive synthesis in GF12 this does not affect the overall critical path

Sticky bits and guard bits are handled internally by the recoded format (I believe).

mysoreanoop:
https://github.com/black-parrot-sdk/bp-tests/pull/34 -- works as expected in Droamjo cosimulation with zynq-parrot-dromajo now.

taylor-bsg:
@dpetrisko Are you rounding once or twice? We should just round once

To give the analog in decimal

If we have 0.499999999995 and round according to the last digit, we would
get 0.5000000000, and then if we round according to the first digit to
right of decimal point, we get 1.
But if we round directly according to the first digit to the right of the
decimal point, then we get 0.

dpetrisko:
https://hal.archives-ouvertes.fr/hal-01091186/document

My understanding is that double rounding is innocuous at 2x precision if you're rounding in the same direction

taylor-bsg:
What about FMA?

M

dpetrisko:
I believe that FMA follows inductively because the higher precision used in the intermediate calculation with guard bits follows the same property (>=2x precision) that allows for correct double rounding. I’ll look for any other literature supporting / refuting On Jan 16, 2023, at 2:40 PM, taylor-bsg ***@***.***> wrote:﻿
What about FMA?

M

taylor-bsg:
Maybe you can email Pierre and ask?


M

dpetrisko:
Have emailed. For posterity, because it confused me, FMA is only required to be exactly rounded in IEEE 754 2008+. In the original spec, only addition, subtraction, multiplication, division and square root are covered. So we should resolve this open question to be compatible with the latest spec

taylor-bsg:
That's interesting. I think FMA has become much more prevalent in
architecture because people realize it's an easy way to double your
marketing performance numbers, and sometimes you get real performance out
of it too. I think our objective is less about claiming IEEE compliance
(although it sounds like they are doing the right thing) and more about
ensuring binary compatibility.

mysoreanoop:
Regarding the practical aspects: I checked all the benchmarks in beebs and riscv-tests. All the previous precision-mismatches are resolved; no new mismatches -- which seems to indicate nothing else is broken. riscv-tests didn't have precision mismatches to begin with as they segregate their sp and dp programs. They do still have some peculiar, likely unrelated mismatches -- will look into them.

dpetrisko:
Superseded by #1153

Issue #603, opened 2020-08-07
Add Load -> Store Data Forwarding


dpetrisko:
Currently, all forwarding to resolve hazards is done in the ISD stage, before issuing to the functional units. However, cache instructions only require store data in MEM2, before the data is sent to the write buffer.  Thus, we should be able to dispatch a store without fully resolving the dependencies, and then forward the immediately preceding load data within the mem pipe.  This will also require breaking up the dcache_pkt so that data comes in at the same time as the TLB ptag.  There may be some complication with negedges as related to forwarding, but the cycle saving could help accelerate pointer-heavy code.

taylor-bsg:
Once we have some regressions setup, we can put some instrumentation into
place to see how often stalls happens in practice? Generally, the compiler
is well equipped to schedule around this problem, although it may require
unrolling and/or the restrict keyword. Also an interesting question of how
you represent this bypass to the GCC scheduler..

M

PR #1153, opened 2023-02-03
Extract D$ store data from D$

dpetrisko:
### Summary
This PR moves floating point un/recoding out of D$, which simplifies the cache implementation, allows for store data forwarding and offloads recoding from the critical path caused by half-cycle D$ offsetting

### Issue Fixed
None

### Area
D$ and LSU

External contributors have noticed that there can be critical paths related to recoding floating point load data, particularly in FPGA. Additionally, the interface of the cache is fairly confusing, having two exit ports which are somewhat dependent on each other. This PR addresses both issues and simplifies the cache, relegating most floating point logic to the LSU instead.

This PR sets the groundwork for EX2->store forwarding, which is a superset of #603 and also includes e.g. mul->store forwarding.

### Analysis
This change eliminates potential critical paths from recoding the FP load data, as that path becomes full cycle instead of half cycle. 

### Verification
Standard regression

### Additional Context
Depends on #1152


dpetrisko:
Supersedes #1152

PR #1158, opened 2023-04-11
Update BedRock interfaces to Burst and Lite only

muwyse:
### Summary
This PR updates all BedRock networks to use either Burst or Lite interface protocols.

### Issue Fixed
Does this PR address a BlackParrot GitHub issue? Link it here:

### Area
Module/tool/makefile/script in question.

Why is this change required?

Next steps, downstream modules affected, etc.

### Analysis

### Verification

### Additional Context


muwyse:
@dpetrisko this is not ready for full review, but if you have a few minutes to take a look and give high-level feedback that would be appreciated. Current status is unicore lints cleanly but is not yet functional. Multicore top-level has not been revised yet.

dpetrisko:
I think it's a great start. Let's start thinking about some benchmarks which can test performance before / after these changes. Area we can measure with FPGA/ASIC synthesis 

Couple of discussion points:
- what alignment requirements should we have for burst data? (Request size, bus size, arbitrary?)
- what gearboxing restrictions should we assume? (Network > protocol, network == protocol, arbitrary?)
- what restrictions on handshakes should we have? (Client can wait for master, client must have forward progress)

Within these discussions, we want to be able to support critical word first/early restart, minimize shifters in masters/clients, minimize necessary fifos in the network and simplify clients to the maximum degree. Ideally, I'd like to avoid burst2lite style adapters as they introduce fifos and extra code that easily rots...

Probably it makes sense to discuss offline and summarize the results here

PR #1161, opened 2023-04-28
Update FMA to fix neg zero mismatches

mysoreanoop:
Mismatch explained here: https://github.com/black-parrot-sdk/bp-tests/pull/35

We can solve it two ways: 
1. By modifying HardFloat to switch sign based on whether the operation is true FMA or just multiplication.
2. By modifying bp_calculator_pipe_fma to load -0 or +0 to C (in FMA's A X B + C) depending on the sign of A X B (and rounding mode) for pure multiplication only. Sign here needs to be recomputed (it'll be computed again within FMA), so I think it's cleaner to go with (1).

This PR along with https://github.com/bsg-external/HardFloat/pull/24 do (1)

### Analysis
Potential impact of the change: I don't think it affects any other module, it might change the synthesis/placement a little bit.

### Verification
Tested with testcase in https://github.com/black-parrot-sdk/bp-tests/pull/35

### DIFF
--- a/bp_be/src/include/bp_be_pkg.sv
+++ b/bp_be/src/include/bp_be_pkg.sv
@@ -20,6 +20,7 @@ package bp_be_pkg;
 
   localparam [dp_rec_width_gp-1:0] dp_rec_1_0 = 65'h0_80000000_00000000;
   localparam [dp_rec_width_gp-1:0] dp_rec_0_0 = 65'h0_00000000_00000000;
+  localparam [dp_rec_width_gp-1:0] dp_rec_neg_0_0 = 65'h1_00000000_00000000;
 
   localparam [dp_rec_width_gp-1:0] dp_canonical_rec = 65'h0_e0080000_00000000;
 
--- a/bp_be/src/v/bp_be_calculator/bp_be_pipe_fma.sv
+++ b/bp_be/src/v/bp_be_calculator/bp_be_pipe_fma.sv
@@ -176,6 +176,7 @@ module bp_be_pipe_fma
     (.clock(clk_i),
      .control(control_li)
      ,.op(fma_op_li)
+     ,.mul_not_fma(is_fmul_li)
      ,.a(fma_a_li)
      ,.b(fma_b_li)
      ,.c(fma_c_li)
--- a/bp_be/src/v/bp_be_calculator/bp_be_pipe_fma.sv
+++ b/bp_be/src/v/bp_be_calculator/bp_be_pipe_fma.sv
@@ -124,7 +124,11 @@ module bp_be_pipe_fma
 
   wire [dp_rec_width_gp-1:0] fma_a_li = is_imul_li ? rs1 : frs1.rec;
   wire [dp_rec_width_gp-1:0] fma_b_li = is_imul_li ? rs2 : is_faddsub_li ? dp_rec_1_0 : frs2.rec;
-  wire [dp_rec_width_gp-1:0] fma_c_li = is_faddsub_li ? frs2.rec : is_fmul_li ? dp_rec_0_0 : frs3.rec;
+  wire [dp_rec_width_gp-1:0] fma_c_li = is_faddsub_li
+                                          ? frs2.rec
+                                          : is_fmul_li // correct for influence by unintentional sign of C
+                                            ? (frm_li == e_rdn) ? dp_rec_0_0 : dp_rec_neg_0_0
+                                            : frs3.rec;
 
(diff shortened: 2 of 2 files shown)


dpetrisko:
Hi 2 is better

mysoreanoop:
How about now?
Works with the extended fp_signed_zero test

bp_be/src/v/bp_be_calculator/bp_be_pipe_fma.sv
```diff
@@ -124,7 +124,11 @@ module bp_be_pipe_fma
 
   wire [dp_rec_width_gp-1:0] fma_a_li = is_imul_li ? rs1 : frs1.rec;
   wire [dp_rec_width_gp-1:0] fma_b_li = is_imul_li ? rs2 : is_faddsub_li ? dp_rec_1_0 : frs2.rec;
-  wire [dp_rec_width_gp-1:0] fma_c_li = is_faddsub_li ? frs2.rec : is_fmul_li ? dp_rec_0_0 : frs3.rec;
```
dpetrisko, line 127: I think is_fmul_li ? (frs1.rec.sign ^ frs2.rec.sign) : dp_rec_neg_0_0 : dp_rec_0_0 is the more complete case? See comment below:

```
// We emulate A*B with A*B+0
// According to IEEE special arithmetic rules for signed 0 (https://en.wikipedia.org/wiki/Signed_zero):
// x + (+-0) = x for x different from zero
// In order to correct for this we compute the sign of of A*B
//   and negate it to maintain the correct signedness of the result
```



```

File: bp_be/src/include/bp_be_pkg.sv [hunk 2]
```diff
@@ -20,6 +20,7 @@ package bp_be_pkg;
 
   localparam [dp_rec_width_gp-1:0] dp_rec_1_0 = 65'h0_80000000_00000000;
   localparam [dp_rec_width_gp-1:0] dp_rec_0_0 = 65'h0_00000000_00000000;
+  localparam [dp_rec_width_gp-1:0] dp_rec_neg_0_0 = 65'h1_00000000_00000000;
```
dpetrisko, line 23: nit; dp_rec_m0_0

bp_be/src/v/bp_be_calculator/bp_be_pipe_fma.sv
mysoreanoop: I think it's safe to assume rounding mode (RM) is correlated with sign of C, we have:

RM=rounding mode, Exp=expected
```
AB  C   RM    o/p Exp
-   -   ~rne   -   -
-   +   rne    -   -
+   -   ~rne   +   +
+   +   rne    +   +
```

dpetrisko: Why would rounding mode be correlated at all with the sign of C?

dpetrisko: https://github.com/chipsalliance/rocket-chip/blob/97d85c68a4da0a46166d1dd8f2d3fa70ccddd526/src/main/scala/tile/FPU.scala#LL702C1-L702C1

This is how Rocket implements this as well

mysoreanoop: I meant sign of C is correlated with rounding mode -- IOW we can only rely on the rounding mode to determine the sign of C. Do you have an example for which case would cause it to depend on sign of A.B?

mysoreanoop: I see, what we're both suggesting are effectively the same.

I'm saying:
You only need C to be -0 when you're rounding other than towards -inf.
You need C to be +0 when you're rounding towards -inf.
Extended truth table below; RM=rounding mode; (sign of C is correlating with rounding mode)
```
 AB  C   RM     o/p Exp
 -   -   ~rne   -   -
 -   +   rne    -   -
 +   -   ~rne   +   +
 +   +   rne    +   +
```

You're saying:
You only need C to be -0 when A.B is -0
You need C to be +0 when A.B is +0
(Rounding mode doesn't matter)
```
 AB  C   RM     o/p Exp
 -   -   ~rne   -   -
 -   -   rne    -   -
 +   +   ~rne   +   +
 +   +   rne    +   +  
```

** Works for AB~=0

For reference: according to IEEE, for A + B:
```
A  B  A+B
+  +  +
+  -  - if rne, + otherwise
-  +  - if rne, + otherwise
-  -  -
```

Issue #789, opened 2021-03-05
Assertions not working in Verilator


dpetrisko:
To reproduce:

make build.sc sim.sc COSIM_P=1 SUITE=riscv_tests PROG=rv64ua-p-lrsc CFG=e_bp_unicore_l2_atomic_cfg

This will overflow the watchdog timer but not trigger the assertion. Possibly a verilator bug, it seems to hang at test_bp.cpp L50, does not print enabling asssertions

developandplay:
In which directory are you executing this command?
Just asking so that I can do a before and after test with FST on this one.

dpetrisko:
bp_top/syn

dpetrisko:
Do you think FST will affect assertions?

developandplay:
No I would assume the bug remains.
Nevertheless I still wanted to check since with FST the reset is defined in verilog instead of systemc.

developandplay:
I was running this command in bp_top/syn but was missing the riscv_tests.
Then I ran `make -C sdk prog` to manually generate them so it's all fine but I was just wondering if this is by design.

Also for some reason after running that command the tests were in `black-parrot/sdk/prog/riscv-tests` but for
```
make build.sc sim.sc COSIM_P=1 SUITE=riscv_tests PROG=rv64ua-p-lrsc CFG=e_bp_unicore_l2_atomic_cfg
```
they are supposed to be in `black-parrot/sdk/prog/riscv_tests`.

developandplay:
And yes unfortunately nothing changed...
I'm now getting the output before FST:
```
Lowering reset
Running with Dromajo cosimulation
[DEBUG] DUT raised exception 2
csr_read: invalid CSR=0x3b0
[error] HARTID: 0
[error] EMU PC 00000000800001dc, DUT PC 00000000800001dc
[error] EMU INSN 00052583, DUT INSN 00052583
[error] EMU WDATA 0000000000000001, DUT WDATA 0000000000000000
[error] EMU MSTATUS a00000080, DUT MSTATUS a00000080
[error] DUT pending exception -1 pending interrupt -1
NBF loader done!
COSIM_FAIL
- /home/martin/coding/black-parrot/bp_top/test/common/bp_nonsynth_cosim.sv:222: Verilog $finish
Finishing test
Executing final
[CORE0 STATS]
        clk   :                  353
        instr :                   84
        mIPC  :                  237
Exiting
```
and after FST:
```
[...]

PR #1163, opened 2023-05-03
Enable assertions

developandplay:
### Summary
Addresses: https://github.com/black-parrot/black-parrot/issues/789
Just realised in one of my own projects that assertions need to be enabled with the `--assert` flag before they work in Verilator.
https://veripool.org/guide/latest/exe_verilator.html

### Verification

Just place a little `assert(0);` in some `initial` blog.
For example for the `tire_kick` I get this without `--assert` flag:
```
Hello World!
[CORE0 FSH] PASS
All cores finished! Terminating...
```

With  `--assert` flag I get this:
```
Hello World!
[CORE0 FSH] PASS
[25866000] %Error: bp_nonsynth_host.sv:184: Assertion failed in testbench.testbench.host: 'assert' failed.
```
Of course don't forget to `make clean` in between.

### DIFF
--- a/bp_common/syn/Makefile.verilator
+++ b/bp_common/syn/Makefile.verilator
@@ -24,6 +24,7 @@ VV_OPTS += config.vlt
 VV_OPTS += --build --exe
 VV_OPTS += -o simsc
 VV_OPTS += -Wno-timescalemod
+VV_OPTS += --assert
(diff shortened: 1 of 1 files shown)


dpetrisko:
Thanks!

PR #1182, opened 2023-07-22
Catchup ALU

dpetrisko:
### Summary
This PR adds an optional, secondary "catchup" ALU which is used to unblock the pipeline in the case of a load->int/load->branch dependency. With perfect branch prediction, this eliminates all load-use stalls in the pipeline.

### Issue Fixed
None

### Area
BE decoder and hazard detection

Performance enhancement, mainly Coremark

None

### Analysis
PPA has not been analyzed for this feature, but it is unlikely to impact critical path. Area is estimated to be ~2% based on other ALU size.

### Verification
Standard regression and Linux Boot

### Additional Context
None

### DIFF
--- a/bp_be/src/include/bp_be_ctl_pkgdef.svh
+++ b/bp_be/src/include/bp_be_ctl_pkgdef.svh
@@ -232,6 +232,7 @@
     logic dtlb_fill;
     logic _interrupt;
     logic cmd_full;
+    logic mispredict;
   }  bp_be_exception_s;
 
   typedef struct packed
@@ -252,6 +253,7 @@
   {
     logic v;
     logic queue_v;
+    logic ispec_v;
 
     bp_be_exception_s exc;
     bp_be_special_s   spec;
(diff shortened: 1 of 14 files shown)


taylor-bsg:
nice -- performance benefit across Spec? what happens on catchup ALU ->
catchup ALU paths, and catch ALU -> store paths?

dpetrisko:
Will run spec soon. (~7% on coremark)

We add a bypass from EX2 to EX2, so we do load->catchup ALU and get catchup ALU->catchup ALU for free.  

Haven't implemented -> store forwarding because it's not obvious it won't affect critical path. Floating point store data needs to be FPU recoded. 

D$ pipeline is 

````
posedge negedge posedge negedge posedge

 addr ---- TL ------------- TV ----------
    
````
where store data is recoded in the full cycle TL-TV cycle. With forwarding, the store data would arrive in TL+0.5, giving only half a cycle to the recode. Could be fine, but have to do more analysis. 

Current implementation has a full cycle for the forwarding, so only cost is area of the second ALU+1 set of bypass muxes

taylor-bsg:
With Farzam you can add a counter to see how often that event happens? This
seems like an interesting demonstration of that capability?

dpetrisko:
Yep, good idea!

PR #1183, opened 2023-08-04
Miscellaneous hotfixes for unicore, address swizzling, and sipo passthrough dynamic

muwyse:
Miscellaneous hotfixes for unicore and sipo_passthrough_dynamic.

- fix UCE to properly handle back-to-back non-blocking requests from the cache. Due to the negedge to posedge crossing out of the UCE (mem_fwd), it is possible for an uncached store to be mishandled if the receiver stops accepting the mem_fwd messages by lowering ready_and_i. This happens due to the half cycle clock crossing and the dlatch extension of the handshake between the two clock edges. The UCE will see ready_and_i high in the back half of its clock cycle, which it views as a valid handshake, but the posedge clocked logic has already lowered the ready signal and does not actually accept the message.
- remove address swizzling from unicore designs
- add address unswizzling before the dma packet leaves the L2 cache slice for multicore designs
- bug fix sipo_passthrough_dynamic to properly handle non-full length transactions. The sipo now directly connects the ready_i|o signals. This requires the consumer to be ready for the SIPO to accept data. This fix also resolves an issue with stream gearbox where we would otherwise need to buffer a header if the SIPO were allowed to consume the last data word before the consumer can consume the parallel output. In effect, this makes the last word of every transaction, regardless of length, conform to the passthrough behavior.

These fixes have not yet been thoroughly tested.

### DIFF
--- a/bp_me/src/v/dev/bp_me_dram_hash_decode.sv
+++ b/bp_me/src/v/dev/bp_me_dram_hash_decode.sv
@@ -5,6 +5,8 @@
  * Description:
  *   This module reverses the bit swizzling applied by bp_me_dram_hash_encode.
  *
+ *   Unicore designs do not swizzle addresses.
+ *
  *   Encode takes an address of form {A, b...bb, c...cc, D} and outputs
  *   address of {A, c...cc, b...bb, D}. The bit widths of b and c bits may differ and are
  *   specified by the offset_widths_p parameter.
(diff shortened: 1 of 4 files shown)


bp_me/src/v/dev/bp_me_dram_hash_encode.sv
```diff
@@ -28,22 +30,32 @@ module bp_me_dram_hash_encode
...
-     ,daddr_i[(hash_offset_widths_lp[0]+hash_offset_widths_lp[1])+:hash_offset_widths_lp[2]]
-     };
 
-  assign daddr_o =
-    {daddr_i[daddr_width_p-1:offset_width_lp]
-     ,encoded_bits
-     ,daddr_i[0+:hash_offset_widths_lp[0]]
-     };
-  assign bank_o = (l2_banks_p > 1) ? daddr_i[l2_bank_offset_width_lp+:lg_l2_banks_lp] : '0;
+  if (cce_type_p == e_cce_uce) begin : unicore
```
dpetrisko, line 34: Let's swizzle for unicore too

bp_me/src/v/dev/bp_me_dram_hash_decode.sv
```diff
@@ -24,23 +26,27 @@ module bp_me_dram_hash_decode
    , output logic [daddr_width_p-1:0] daddr_o
    );
 
-  localparam l2_block_offset_width_lp = `BSG_SAFE_CLOG2(l2_block_width_p/8);
-  localparam lg_l2_sets_lp            = `BSG_SAFE_CLOG2(l2_sets_p);
-  localparam lg_l2_banks_lp           = `BSG_SAFE_CLOG2(l2_banks_p);
-  localparam lg_num_cce_lp            = `BSG_SAFE_CLOG2(num_cce_p);
-  localparam int hash_offset_widths_lp[2:0] = '{(lg_l2_sets_lp-lg_num_cce_lp), lg_num_cce_lp, l2_block_offset_width_lp};
-  localparam offset_width_lp = hash_offset_widths_lp[0] + hash_offset_widths_lp[1] + hash_offset_widths_lp[2];
+  if (cce_type_p == e_cce_uce) begin : unicore
```
dpetrisko, line 29: Let's swizzle for unicore too

bp_me/src/v/cce/bp_uce.sv
```diff
@@ -549,7 +549,7 @@ module bp_uce
                 cache_req_done = fsm_fwd_ready_and_li & fsm_fwd_v_lo;
               end
 
-            cache_req_yumi_o = cache_req_v_i & ~(cache_req_v_r | cache_req_done);
```
dpetrisko, line 552: Can we rebase on #1184 for a higher performance fix?

dpetrisko (approved):
LGTM with nits addressed

dpetrisko:
I believe these changes have been integrated in latest dev. Feel free to reopen or ping me if you discover problems (one difference is the encode/decode algorithm, but we discussed offline that we may need to overhaul the system entirely)

PR #1185, opened 2023-08-19
FMD disable

Abdul-muheet-ghani:
Summary:
Parameterize the F, M, and D extensions in BlackParrot to enable/disable them, reducing FPGA area requirements and enabling access to the design on smaller FPGAs for academic purposes.

Issue Fixed:
This change addresses an issue with the F, M, and D extensions in BlackParrot related to FPGA area requirements and accessibility for academic purposes.

Area:
BlackParrot module.

Reasoning:
The change is required to parameterize the F, M, and D extensions, allowing them to be enabled or disabled. This is done to reduce the FPGA area requirements of BlackParrot, making it accessible on smaller FPGAs. By disabling these extensions, the design can fit into smaller devices, which is particularly beneficial for academic purposes where resources may be limited.

Additional Changes Required:
No additional changes are required.

Analysis:
The impact of this change is that it allows users to enable or disable the F, M, and D extensions in BlackParrot, reducing the FPGA area requirements. This enables the use of BlackParrot on smaller FPGAs, making it more accessible for academic purposes. The reduction in area requirements may also have benefits in terms of resource utilization and performance.

Verification:
To verify the change, the following steps can be taken:

Set the parameters to enable the F, M, and D extensions.
Verify that the design requires a larger FPGA area.
Set the parameters to disable the F, M, and D extensions.
Verify that the design now fits into a smaller FPGA.
Run any relevant tests or benchmarks to ensure the functionality is preserved despite the disabled extensions.

Additional Context:
The motivation behind disabling the F, M, and D extensions is to reduce the FPGA area requirements of BlackParrot, allowing it to be used on smaller FPGAs. This enhances accessibility, particularly in academic settings where resources may be limited. It is important to consider the trade-off between FPGA area reduction and potential impact on performance or functionality. Logs or additional context about the problem and its impact may be helpful for further investigation.

### DIFF
--- a/bp_be/src/v/bp_be_calculator/bp_be_calculator_top.sv
+++ b/bp_be/src/v/bp_be_calculator/bp_be_calculator_top.sv
@@ -257,22 +257,20 @@ module bp_be_calculator_top
      );
 
   // Aux pipe: 2 cycle latency
-  if(fpu_support_p)begin
-    bp_be_pipe_aux
-     #(.bp_params_p(bp_params_p))
-     pipe_aux
-      (.clk_i(clk_i)
-       ,.reset_i(reset_i)
+  bp_be_pipe_aux
+   #(.bp_params_p(bp_params_p))
+   pipe_aux
+    (.clk_i(clk_i)
+     ,.reset_i(reset_i)
 
-      ,.reservation_i(reservation_r)
-      ,.flush_i(commit_pkt_cast_o.npc_w_v)
-      ,.frm_dyn_i(frm_dyn_lo)
+     ,.reservation_i(reservation_r)
+     ,.flush_i(commit_pkt_cast_o.npc_w_v)
+     ,.frm_dyn_i(frm_dyn_lo)
 
-      ,.data_o(pipe_aux_data_lo)
-      ,.fflags_o(pipe_aux_fflags_lo)
-      ,.v_o(pipe_aux_data_lo_v)
-      );
-  end
+     ,.data_o(pipe_aux_data_lo)
+     ,.fflags_o(pipe_aux_fflags_lo)
+     ,.v_o(pipe_aux_data_lo_v)
+     );
 
   // Memory pipe: 2/3 cycle latency
   bp_be_pipe_mem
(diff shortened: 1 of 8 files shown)


bp_be/src/v/bp_be_calculator/bp_be_csr.sv
```diff
@@ -662,6 +685,7 @@ module bp_be_csr
       // Set FS to dirty if: fflags set, frf written, fcsr written
       mstatus_li.fs |= {2{(csr_w_v_li & csr_fany_li & ~csr_illegal_instr_o)}};
       mstatus_li.fs |= {2{(retire_pkt_cast_i.instret & instr_fany_li)}};
+      mstatus_li.fs |= {2{(fpu_support_p)}};
```
dpetrisko, line 688: This seems odd, if we have fpu support then we should have the normal progression of clean, dirty, disabled. This code will make it always dirty.

Additionally, fpu_support_p is multiple bits, so we have to reduce to 1b as |fpu_support_p

bp_be/src/v/bp_be_checker/bp_be_instr_decoder.sv
```diff
@@ -211,6 +211,7 @@ module bp_be_instr_decoder
             decode_cast_o.ops_v      = instr inside {`RV64_FL_W};
 
             illegal_instr_o = ~decode_info_cast_i.fpu_en;
+            illegal_instr_o = ~fpu_support_p;
```
dpetrisko, line 214: we use the fpu_en signal right above it, so we don't want to override with fpu_support_p. this change will make it so that if we have fpu_support and the fpu is not enabled, it will still successfully execute the instruction.

Instead, making sure decode_info_cast_i.fpu_en depends on fpu_support_p will be cleaner

bp_be/src/v/bp_be_calculator/bp_be_csr.sv
```diff
@@ -166,6 +166,29 @@ module bp_be_csr
...
   assign irq_waiting_o = |interrupt_icode_dec_li;
 
+  // Holding MISA value on different Configuration
+  logic [25:0] MISA_val;
+
+  always_comb
+    begin
+      if((fpu_support_p == 1) && (muldiv_support_p == 0))
+        begin
+          MISA_val = 26'h141105;
```
dpetrisko, line 176: This will be hard to scale to a combination of extensions. Instead, constructing the value from the parameters will be easier.
e.g.
```
MISA_val = ( |fpu_support_p << X)
                    ( |mul_support_p << Y)
...
```

bp_be/src/v/bp_be_calculator/bp_be_calculator_top.sv
```diff
@@ -348,44 +350,50 @@ module bp_be_calculator_top
...
-     ,.flush_i(commit_pkt_cast_o.npc_w_v)
-     ,.frm_dyn_i(frm_dyn_lo)
-
-     ,.imul_data_o(pipe_mul_data_lo)
-     ,.imul_v_o(pipe_mul_data_lo_v)
-     ,.fma_data_o(pipe_fma_data_lo)
-     ,.fma_fflags_o(pipe_fma_fflags_lo)
-     ,.fma_v_o(pipe_fma_data_lo_v)
-     );
+  if(muldiv_support_p)
```
dpetrisko, line 353: Putting the generate statement out here will not allow us to have fine-grained control where we support some instructions but not others. Instead, having the generate statement inside pipe_aux and pipe_fma will allow us to modify the pipe internally without affecting the calculator

bp_be/src/v/bp_be_calculator/bp_be_pipe_aux.sv
```diff
@@ -45,21 +45,26 @@ module bp_be_pipe_aux
...
-     );
-
-  bp_be_nan_unbox
-   #(.bp_params_p(bp_params_p))
-   frs2_unbox
-    (.reg_i(reservation.rs2)
-     ,.unbox_i(decode.ops_v)
-     ,.reg_o(frs2)
-     );
+  generate
```
dpetrisko, line 48: again, don't use the generate keyword

bp_be/src/v/bp_be_calculator/bp_be_pipe_aux.sv
```diff
@@ -393,18 +468,23 @@ module bp_be_pipe_aux
...
-     ,.b(frs2.rec)
-     ,.signaling(signaling_li)
-     ,.lt(flt_lo)
-     ,.eq(feq_lo)
-     ,.gt(fgt_lo)
-     ,.unordered(unordered_lo)
-     ,.exceptionFlags(fcmp_fflags)
-     );
+  generate
+    if(|fpu_support_p)
```
dpetrisko, line 472: I think just having 1 big if will be cleaner

bp_be/src/v/bp_be_calculator/bp_be_pipe_aux.sv
```diff
@@ -495,14 +575,29 @@ module bp_be_pipe_aux
...
+         retiming_chain
+          (.clk_i(clk_i)
+
+           ,.data_i({faux_fflags, faux_result, faux_v_li})
+           ,.data_o({fflags_o, data_o, v_o})
+           );
+      end
+    endgenerate
+
+    always_comb
```
dpetrisko, line 592: make this an else of the above if to promote consistency

bp_common/src/include/bp_common_aviary_cfg_pkgdef.svh
```diff
@@ -303,8 +311,8 @@
 
       ,fe_queue_fifo_els : 8
       ,fe_cmd_fifo_els   : 4
-      ,muldiv_support    : (1 << e_div) | (1 << e_mul) | (1 << e_mulh)
-      ,fpu_support       : 1
+      ,muldiv_support    : 0//(1 << e_idiv) | (1 << e_imul) | (1 << e_imulh) | (1 << e_idiv2b)
```
dpetrisko, line 314: assume this is a temp change; should only change in the tiny parrot config

dpetrisko:
Looks pretty good, mostly style nits! Still need to 
- add the decoder illegal instruction check for multiply
- parameterize out FCSR

since these features are not part of this changes, I assume they are untested as well. Let's add them to the test suite!

dpetrisko:
will also need to rebase on dev, seeing some conflicts

PR #1196, opened 2023-09-29
[WIP] Added L2-bypass

farzamgl:
This PR adds an L2 bypass path for converting CE requests to cache DMA requests in order to be able to completely omit the L2 instead of instantiating a tiny victim cache.

Checklist:
- [ ] Multicore compatibility (currently only tested on unicore) 
- [ ] Handling L2 atomics or adding assertion to restrict access
- [ ] Add to CI regression

### DIFF
--- a/bp_me/src/v/dev/bp_me_cce_to_cache.sv
+++ b/bp_me/src/v/dev/bp_me_cce_to_cache.sv
@@ -69,13 +69,17 @@ module bp_me_cce_to_cache
   bsg_cache_pkt_s cache_pkt;
   assign cache_pkt_o = {l2_banks_p{cache_pkt}};
 
-  enum logic [1:0] {e_reset, e_clear_tag, e_ready} state_n, state_r;
+  enum logic [2:0] {e_reset, e_clear_tag, e_flush_tag, e_flinv_tag, e_ready} state_n, state_r;
   wire is_reset  = (state_r == e_reset);
   wire is_clear  = (state_r == e_clear_tag);
   wire is_ready  = (state_r == e_ready);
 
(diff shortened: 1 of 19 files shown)


bp_me/src/v/dev/bp_me_cce_to_cache_dma.sv
```diff
@@ -0,0 +1,303 @@
+
+`include "bp_common_defines.svh"
```
dpetrisko, line 2: How about combining this with the cce to cache module?

My thought is we want the bypass for performance, but for certain features (AMOs, byte read/writes) we'll need the buffering capability. 

So what we could do is have the FWD message come in and have 3 cases:
- mem_fwd_cached (regular L2 access)
- cache line sized mem_fwd_uncached (use bypass path to DMA)
- byte sized mem_fwd_uncached / AMO (regular L2 access, then tagfl)

Sound reasonable?

bp_me/src/v/dev/bp_me_cce_to_cache.sv
```diff
@@ -294,34 +301,45 @@ module bp_me_cce_to_cache
...
         e_reset:
           begin
+            tag_sent_n = '0;
+            tag_received_n = '0;
             state_n = e_clear_tag;
           end
-        e_clear_tag: begin
-          cache_pkt_v_o = (tagst_sent_r != l2_blocks_lp) << (tagst_sent_r / (l2_sets_p*l2_assoc_p));
-          cache_pkt.opcode = TAGST;
+        e_clear_tag, e_flush_tag, e_flinv_tag: begin
```
dpetrisko, line 316: The address space is broken here because of the hashing (known issue) so we'll have to fix that for multicore

PR #1206, opened 2023-12-04
Update and clarify L2 uncached operations

dpetrisko:
### Summary
This PR differentiates between mem and uc_mem operations in the L2 cache. This allows us to actually do uncached to L1, uncached to L2 operations in additional to normal cached operations. 

### Issue Fixed
None

### Area
LCE/CCE L2 cache controller

Currently mem/uc_mem operations are not differentiated by the L2. There are cases when we truly do not want to cache to L2 (additional levels of cache, etc), and so this is a desired feature to support.

None

### Analysis
No PPA overhead, just control logic.

### Verification
Standard regression and linux boot.

### Additional Context
Requires latest bedrock ucode (cc7521e) which has additional msg types defined.


bp_me/src/v/cce/bp_cce_fsm.sv
```diff
@@ -840,90 +844,47 @@ module bp_cce_fsm
         fsm_cmd_data_lo = fsm_rev_data_li;
 
         // decrement pending bit on mem response dequeue (same as lce cmd send)
-        pending_busy = fsm_rev_yumi_lo & fsm_rev_last_li;
+        pending_busy = fsm_rev_yumi_lo & fsm_rev_last_li & (~fsm_rev_header_li.payload.uncached | rev_pma_cacheable_addr_lo);
```
muwyse, line 847: does this still properly account for cce_normal_mode_r? The CCE has an uncached only mode that effectively overrides the PMA to issue everything as uncached.

bp_me/src/v/network/bp_me_xbar_stream.sv
```diff
@@ -32,27 +30,27 @@ module bp_me_xbar_stream
    , input                                                            reset_i
 
    , input [num_source_p-1:0][xbar_header_width_lp-1:0]               msg_header_i
-   , input [num_source_p-1:0][data_width_p-1:0]                       msg_data_i
+   , input [num_source_p-1:0][bedrock_fill_width_p-1:0]                       msg_data_i
```
muwyse, line 33: formatting/alignment of msg_data_i (same to msg_data_o)

bp_me/test/tb/bp_cce/Makefile.testlist
```diff
@@ -1,13 +1,13 @@
 
 run_testlist.%:
```
muwyse, line 2: is the intention to comment out all the tests or leftover from debugging?

bp_top/test/common/bp_nonsynth_if_verif.sv
```diff
@@ -160,6 +160,13 @@ module bp_nonsynth_if_verif
   if ((cce_type_p != e_cce_uce) && (`BSG_SAFE_CLOG2(icache_block_width_p*icache_sets_p/8) > page_offset_width_gp) && (`BSG_SAFE_CLOG2(dcache_block_width_p*dcache_sets_p/8) > page_offset_width_gp))
     $error("Error: Multicore requires total cache size to be equal to 4kB * associativity");
 
+  if (l2_banks_p > 8)
+    $error("L2 banks must be less than 8");
+  if (l2_assoc_p > 8)
+    $error("L2 associativity must be less than 8");
+  if (l2_block_width_p != 512)
```
muwyse, line 167: 1. must be 512b or must be same as bedrock block width?
2. above, should the num_cce_p check be moved out of the initial block?

bp_top/test/common/bp_nonsynth_nbf_loader.sv
```diff
@@ -137,13 +137,22 @@ module bp_nonsynth_nbf_loader
...
+     );
+
   always_comb
     begin
       mem_fwd_header_cast_o = '0;
       mem_fwd_header_cast_o.payload.lce_id = lce_id_i;
       mem_fwd_header_cast_o.payload.src_did = did_i;
       mem_fwd_header_cast_o.addr = curr_nbf.addr;
-      mem_fwd_header_cast_o.msg_type.fwd = curr_nbf.opcode[5] ? e_bedrock_mem_uc_rd : e_bedrock_mem_uc_wr;
+      mem_fwd_header_cast_o.msg_type.fwd = curr_nbf.opcode[5] ? fwd_mem_cacheable_addr_lo ? e_bedrock_mem_rd : e_bedrock_mem_uc_rd : fwd_mem_cacheable_addr_lo ? e_bedrock_mem_wr : e_bedrock_mem_uc_wr;
```
muwyse, line 155: For the NBF loader, do we want to issue only uncached operations? With the new scheme of using the PMA to determine cached/uncached is it still possible to issue an uncached access to an L2 cacheable region?

docs/platform_guide.md
```diff
@@ -86,10 +86,14 @@ BlackParrot has a configurable physical address width as well as maximum DRAM si
   * Cached, global memory
   * Striped by cache line
   * Cached DRAM region
-* 0x01_0000_0000 - 0x01_FFFF_FFFF
+* 0x01_0000_0000 - 0x01_7FFF_FFFF
   * Uncached, global memory
   * Striped by cache line
-  * Uncached DRAM region
+  * L1-Uncached DRAM region
```
muwyse, line 92: L1-Uncached and L2-Cached DRAM region? (for explicitness)

docs/platform_guide.md
```diff
@@ -86,10 +86,14 @@ BlackParrot has a configurable physical address width as well as maximum DRAM si
...
+* 0x01_8000_0000 - 0x01_FFFF_FFFF
+  * Uncached, global memory
+  * Striped by cache line
+  * L2-Uncached DRAM region
```
muwyse, line 96: L1-Uncached and L2-Uncached DRAM region? (for explicitness)

muwyse (commented):
Added a few comments and questions for now. Need to look a bit closer at cce_fsm and cce_msg still.

bp_me/src/v/cce/bp_cce_fsm.sv
dpetrisko: Uncached mode is just for L1 (bedrock) at the moment, so shouldn't affect L2 PMA? I could be wrong, but that was my understanding

bp_top/test/common/bp_nonsynth_if_verif.sv
dpetrisko: Specifically 512 I believe, for historical reasons (there's a bug we never tracked down)

dpetrisko: 2 Roger

bp_top/test/common/bp_nonsynth_nbf_loader.sv
dpetrisko: That's a good point. I guess it doesn't actually matter, since the type will be overridden by the destination PMA. Perhaps for I/O we only allow WR/RD, as we're saying for microcode?

dpetrisko:
Thanks for feedback, will fix nits in next review cycle!

dpetrisko:
Nits fixed, let me know when you've gone through CCE

bp_me/src/v/cce/bp_cce_fsm.sv
muwyse: Sorry, yes. The CCE starts in an uncached-only mode that is only capable of processing uncached LCE requests, and software controls when the mode switch happens and guarantees the CCE is idle for the mode switch.

bp_top/test/common/bp_nonsynth_if_verif.sv
```diff
@@ -160,6 +155,18 @@ module bp_nonsynth_if_verif
...
+    $error("L2 block width must be 512b");
+
+  /*
+  if (!(num_cce_p inside {1,2,3,4,6,7,8,12,14,15,16,24,28,30,31,32}))
+    $error("Error: unsupported number of CCE's");
+  */
```
muwyse, line 168: This was throwing an error in lint.v seemingly about num_cce_p not being defined, but it is used in the check below, which is fine. My best guess is linting doesn't like the inside keyword.

muwyse (approved):
LGTM. I've not tested the latest changes with the cce_pma, but assuming that works it should be good to go.

bp_top/test/common/bp_nonsynth_if_verif.sv
dpetrisko: ...that's definitely why it was in the initial block :D

dpetrisko:
Thanks for the help!

PR #1208, opened 2024-01-24
Implement correct handling of writebacks and uncached stores in CCE

muwyse:
### Summary
This change modifies the CCE to properly handle writebacks and mem_rev responses to writebacks and uncached store operations. For mem_fwd writes, the payload.uncached bit is set only for uncached stores being issued by the CCE. All writebacks sent to L2/memory have the payload.uncached bit cleared to indicate that no response to the LCE will be required when the mem_rev returns.

### Area
CCE

This supports the removal of the UC operation types from the mem_fwd/rev networks.

### Verification
Standard CI runs.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -26,6 +26,7 @@ stages:
     - me_dev
     - top_dev
     - sw_dev
+    - dev_fix_uncached_rework_rebase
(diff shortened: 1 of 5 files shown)


dpetrisko:
Passed CI locally!

PR #1215, opened 2024-03-09
fence.i decode bugfix

mysoreanoop:
Decoder doesn't recognize the full pattern of `fence.i`
According to the ISA manual: Page 32 (https://drive.google.com/file/d/1s0lZxUZaa7eV_O0_WsZzaurFLLww7ou5/view)
Implementations should ignore some number of bits -- BP instead only recognized zeros in their places (Page 131 above for the actual pattern)

```
The unused fields in the FENCE.I instruction, imm[11:0], rs1, and rd, are reserved for finer-grain
fences in future extensions. For forward compatibility, base implementations shall ignore these
fields, and standard software shall zero these fields.
```


dpetrisko:
Nice, thanks

PR #1216, opened 2024-03-21
re-enable waveform dumping when simulating with Verilator

muwyse:
This change re-enables waveform dumping for Verilator simulations by adding a define of VERILATOR_TRACE to the verilated executable build that is visible to the testbench.


dpetrisko:
I think this is supposed to get set by verilator, but this fix works for me in the meantime

PR #1217, opened 2024-03-22
rework dram hash encode/decode to account for 0-width fields

muwyse:
This PR does two things:
1. fixes the L2/DRAM hashing to account for fields that may have 0-width (CCE, slice, bank, and set indices)
2. prevents aliasing of blocks outside the L2 capacity

The purpose of the hash function is to distribute the L2 blocks evenly across the L2 slices/banks. This algorithm distributes blocks first to CCEs, then across slices and across banks (i.e., slice 0, bank 0 - slice 1, bank 0 - slice 0, bank 1 - slice 1, bank 0, etc.).

By definition (at least for POT number of CCEs), this algorithm guarantees that if there are N total sets across all L2s, then blocks [0, N) will remap to the same range and not alias outside the the L2 capacity.

Number of CCEs can be NPOT, but number of slices, banks, and sets must be POT >= 1.

### DIFF
--- a/bp_me/src/v/dev/bp_me_dram_hash_encode.sv
+++ b/bp_me/src/v/dev/bp_me_dram_hash_encode.sv
@@ -83,6 +83,8 @@ module bp_me_dram_hash_encode
   wire [l2_tag_width_lp-1:0]            tag = daddr[l2_tag_offset_lp+:l2_tag_width_lp];
 
   // assemble the address
+  // note: using concatentation and replication operators would be preferred here,
+  // but Vivado (xsim) v2022.1 throws a fatal error when using them to assemble addr.
   logic [daddr_width_p-1:0] addr;
   assign addr[l2_tag_offset_lp+:l2_tag_width_lp] = tag;
   assign addr[0+:l2_block_offset_width_lp]       = block;
--- a/bp_me/src/v/dev/bp_me_dram_hash_encode.sv
+++ b/bp_me/src/v/dev/bp_me_dram_hash_encode.sv
@@ -83,9 +83,13 @@ module bp_me_dram_hash_encode
   wire [l2_tag_width_lp-1:0]            tag = daddr[l2_tag_offset_lp+:l2_tag_width_lp];
 
   // assemble the address
-  wire [daddr_width_p-1:0] addr = {tag, {(l2_banks_p>1){bank}}, {(l2_slices_p>1){slice}},
-                                   {(num_cce_p>1){cce}}, {(l2_sets_p>1){set}}, block
-                                  };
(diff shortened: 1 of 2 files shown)


bp_me/src/v/dev/bp_me_dram_hash_encode.sv
```diff
@@ -6,8 +6,16 @@
  *   This module swizzles bits in an address, primarily to enable uniform access to L2/memory
  *   in a BlackParrot multicore.
  *
- *   IN:  [ taghi ][         taglo        ][     indexhi     ][ bank ][ slice ][ cce ][ block ]
- *   OUT: [ taghi ][ bank ][ slice ][ cce ][         taglo         ][     indexhi    ][ block ]
+ *   ADDR: [             block number           ][ block ]
+ *   IN:   [ tag ][ set ][ bank ][ slice ][ cce ][ block ]
+ *   OUT:  [ tag ][ bank ][ slice ][ cce ][ set ][ block ]
```
dpetrisko, line 11: I think we want tag to be lower than bank, slice, CCE since those are fixed for a given L2. As an optimization in the future, we could remove those bits from the tag entirely

muwyse: The bits immediately following the block offset (i.e., low order bits of the block number) are used by bsg_hash_bank to select the CCE. Assuming those are fixed for a given CCE (is this true from bsg_hash_bank? For POT, NPOT?), then the my proposal causes blocks to evenly distribute, in a round-robin manner, across an L2's banks, as the raw block number increments.

Moving the slice/bank/set bits into higher bits simply slows down how quickly the slice/bank/set selection changes. What is the benefit of that alone? Using a more complex hash that tries to mitigate the thrashing/aliasing effects of degenerate access patterns (e.g., 4K stride) could be worth investigating in the future.

Removing those bits from the tag could save storage, but they need to be recovered later for decode (likely via inputs or params to the decode module?). I'm not convinced this is worth the effort at this point.

dpetrisko: Aha, yeah NPOT would mean they are not fixed. Shouldn't make a difference then. 

FWIW, I was suggesting moving the output bits to be {BANK, SLICE, CCE, TAG, SET, OFFSET} instead of {TAG, BANK, SLICE, CCE, SET, OFFSET}. Since it's moving L2 tag to L2 tag it doesn't affect the speed of the indexing

dpetrisko (approved):

dpetrisko:
Put a comment with the buggy xsim version

PR #1227, opened 2024-08-22
Parameterizing fetch width

dpetrisko:
### Summary
This PR parameterizes I$ and D$ data widths. For D$, 64b is the only supported mode but I$ now supports flexible widths (tested for 32b and 64b). Additionally, it smooths some fetch edge cases by expanding the fetch width to I$ width + 16b, so that we never need to stall fetch in order to drain the pipe.

### Issue Fixed
None

### Area
Fetch Unit

This is on the pathway to superscalar

None

### Analysis
We see 1% performance improvement on coremark and up to 5% performance improvements on other benchmarks. This is due to reduced aliasing of branches in addition to increased fetch bandwidth in certain conditions (runahead prefetch).

### Verification
Full Linux boot

### Additional Context
None


dpetrisko:
Keeping default fetch as 32b because ZynqParrot doesn't support these kind of mixed width shenanigans

PR #1245, opened 2026-03-11
TB: Add safety check for mem noc cid aliasing (Fixes #391)

### LINKED ISSUE #391: Fix mem noc cid width to x cord width
https://github.com/black-parrot/black-parrot/blob/2715c73292a848e6f848d11091e18052390be285/bp_top/src/v/bp_tile.v#L543

Currently, we have 1 DMC connection per tile column (unlikely to change right now).  The issue with this is if we have x_cord > cid, tiles will alias and the mem responses will go to the wrong tile.  Need to make sure that cid >= x_cord.

RKNAGA18:
Summary
Added a simulation-time parameter check to the tethered wrapper.sv to ensure mem_noc_cid_width_p >= mem_noc_cord_width_p during initialization.

Issue Fixed
Fixes #391

Area
bp_top/test/tb/bp_tethered/wrapper.sv

Reasoning (new feature, inefficient, verbose, etc.)
As discussed in #391, the hardware concentration logic was refactored out of the core tiles and moved to the DUT wrapper. If a user configures a system where the coordinate width requires more bits than the Concentrator ID (CID) width provides, the memory responses will alias and route to the wrong tile. This change adds a "seatbelt" to safely crash the simulation with a clear fatal error instead of silently failing.

Additional Changes Required (if any)
None.

Analysis
Zero impact on synthesized hardware logic. Fails fast in simulation initialization if an invalid parameter combination is provided by the user.

Verification
Verified syntax via Verilator linting.

Built and simulated the default configuration (e_bp_default_cfg) via make -C bp_top/verilator build.verilator sim.verilator to ensure the baseline configuration passes the check cleanly.

### DIFF
--- a/Makefile.env
+++ b/Makefile.env
@@ -7,8 +7,8 @@
 # Project-specific configuration
 ##############################
 
-BP_DIR         ?= $(TOP)
-BP_WORK_DIR    ?= $(BP_DIR)/work
+export BP_DIR         ?= $(TOP)
+export BP_WORK_DIR    ?= $(BP_DIR)/work
 
 # Use meta-installation if available
 ifeq (1,$(IS_SUBMODULE))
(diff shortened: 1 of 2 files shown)


dpetrisko:
I think the restriction should be based on l2_dmas_p since we could have a big coord width but not actually populate the columns enough to alias

Issue #33, opened 2019-03-06
proc_cfg


tommydcjung:
I'm not sure if this proc_cfg struct is really necessary. It just creates more work.
First, I have to find which file defines the struct, declare the struct. BE does not even use icache_id.
Why not just pass in cache id/core id by ports?

taylor-bsg:
Tommy, you are basically saying that instead of having a single port with the struct, to run the individual fields into the processor through separate ports? I think the idea is that there may be a lot of these kinds of parameters (for example, the start PC.)  Maybe adding a comment to the code that mentions what proc_cfg is for, and where it is defined would address the major issues?

dpetrisko:
We probably don’t need this at all once the manycore bridge is implemented

tommydcjung:
I think the intention is good, but I don't think the current use case doesn't justify proc_cfg. It's just too much work. Once, I declare the struct and I have to fill it with values and pass it to the submodules. and inside each submodule that uses proc_cfg I have to include/declare and so on... just passing as ports individually easier in my opinion

dpetrisko:
Each one of those ports will require the same declaration as the original proc_cfg, since they depend on the same parameters. Changing to ports would actually more declarations in total.

dpetrisko:
But, if all of these configuration parameters become set based on the manycore bridge bus, this is a moot point, since it will be just one address and one data port threaded throughout

taylor-bsg:
@tommydcjung I see. If we made the struct fixed size, then we would not have to do all of this struct declaration stuff, but ultimately we are passing around a bunch of fields in this struct that are not used by all of the modules which is also not desirable. Dan's proposal of using the bridge bus (which I assume is an addr + data pair) would address this, although it does mean that these items are time-multiplexed, which means for even simple usage, they may need to do a state machine (or a trace replay!) to configure it, which is not the end of the world, but a slight complexity. And it will use flops instead of wires for true constants (i.e. network IDs, and maybe hart ID) that we will worry about clock gating. (It does make you wish you had a verilog construct which was, take this struct, and tack this new stuff on to, and pass it up. Or, remove this field from the struct, and pass the new struct along.)

dpetrisko:
Keeping this for IDs.  May make sense to have a fixed width for maximum num cores/lces/cces, because then the backend flows wouldn't need to change harden memory sizes, etc. for different #core configurations.  In fact, we could even reuse bp_tile for both a 4 and 8 core hierarchical build

PR #1250, opened 2026-03-14
make: export BP_DIR from Makefile.env

HUNT-001:
Summary
Export BP_DIR from Makefile.env so it is consistently available during Verilator filelist generation.

Issue Fixed
Related to black-parrot-sim issue #33.

Area
Makefile.env

Reasoning
During local bring-up under WSL Ubuntu 24.04, the documented Verilator smoke test failed because the generated flist.vcs retained literal $BP_DIR paths. In issue #33, Dan suggested trying `export BP_DIR` in Makefile.env as the preferred fix. This change makes BP_DIR available earlier and more consistently in the build environment.

Additional Changes Required
None.

Verification
Tested using the documented simulation flow:
- make checkout
- make prep_lite
- make -C black-parrot/bp_top/verilator build.verilator sim.verilator

Confirmed that the first Verilator smoke test completed successfully without manually expanding $BP_DIR in flist.vcs, producing:
- Hello World!
- [CORE FSH] PASS
- [BSG-PASS]

### DIFF
--- a/Makefile.env
+++ b/Makefile.env
@@ -7,9 +7,8 @@
 # Project-specific configuration
 ##############################
 
-BP_DIR         ?= $(TOP)
-export BP_DIR
-BP_WORK_DIR    ?= $(BP_DIR)/work
+export BP_DIR         ?= $(TOP)
+export BP_WORK_DIR    ?= $(BP_DIR)/work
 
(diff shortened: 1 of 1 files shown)


dpetrisko:
cool thanks, edited to make it inline

PR #1254, opened 2026-03-16
bp_be_csr: write faulting PC to mtval/stval on instruction access fault (fixes #1018)

### LINKED ISSUE #1018: [Feature request] Write exception PC to mtval on instruction access fault
## Branch / Commit Hash
dd2d8453a49b0f069e96f175afe97968dac205e1

## Area
bp_be_csr maybe.

## Current behavior
On instruction access exception, BlackParrot doesn't write the latest exception PC to mtval. It seems legal to not write it.

## Desired behavior
Write the latest excepting PC on an instruction access exception. Still legal, and could be helpful for debugging. 

## Analysis
Potential impact of the feature: nothing, from what I understand.

## Verification
Steps to verify the feature.
On instruction access fault, reading mtval (or stval) in the handler should read the excepting PC.

## Additional context
Dromajo does this. Spike also seems to be doing this.

Sairam-Kesanapalli:
**Problem:**
On an instruction access fault (ecode = 1), BlackParrot was writing vaddr (a data address) to mtval/stval instead of the faulting PC (apc_r). Both Dromajo and Spike write the faulting PC here, and while the RISC-V spec allows leaving it zero, writing it is helpful for debugging.

**Fix:**
Updated the mtval_li and stval_li assignments in both the M-mode and S-mode exception handlers in bp_be_csr.sv
The same fix is applied to stval_li in the S-mode delegation block, since it has identical logic.
Fixes #1018

**Questions for reviewer:**
The issue only mentions ecode 1 (instr_access_fault), but ecode 0 (instr_misaligned) also has the same problem — it should also write the faulting PC to mtval/stval. I've included the fix for both in this PR using inside {0, 1}. Is that the right approach, or should ecode 0 be a separate issue?
I also fixed stval_li in the S-mode block since it has the identical bug. Is that within scope?

### DIFF
--- a/ci/sim-bp-tests.sh
+++ b/ci/sim-bp-tests.sh
@@ -20,6 +20,10 @@ progs=(
 	fp_precision
 	hello_world
 	jalr_illegal
+	instr_fault_mtval
+	instr_fault_stval
+	instret_smode
+	instret_smode_deny
 	l2_cache_ops
 	loop
 	mapping


Sairam-Kesanapalli:
For instr_fault_stval, I need to delegate the exception to S-mode using medeleg and drop into S-mode to read stval. Should I follow the infrastructure used in misaligned_instructions_virtual_memory.c (using trapframe_t, pop_tf, vm.h), or is there a simpler way to test S-mode stval without full virtual memory setup?

Issue #790, opened 2021-03-05
Add example quad core mapping to platform address map


dpetrisko:
This will clear confusion for new users looking to connect BlackParrot to an external system. Perhaps a diagram with the regions annotated by addresses would be helpful for clarity.

PR #1255, opened 2026-03-18
Add diagram for platform address maps

nidhibaratam:
### Summary
Added a mermaid diagram to illustrate the platform address maps and their components, including local and global address spaces.This visualization clarifies how the Address Router handles the local tile-shared space versus the global DRAM and I/O regions.

### Issue Fixed
Fixes : #790 

### Area
`docs/platform_guide.md` under **Platform Address Maps**

### Reasoning 
I proposed the solution in Mermaid.js because It is easy for version control and easy to edit rather than static images .

### Analysis
- No functional or RTL impact (documentation-only change)
- Improves clarity of address decoding and system structure for developers
- Helps reduce ambiguity when working with memory-mapped components

### Verification
- Verified Mermaid diagram renders correctly in GitHub preview
- Cross-checked address breakdown (tile, DevID, offset) against existing address map definitions, `bp_common/src/include/bp_common_accelerator_pkgdef.svh` and `bp_common/src/include/bp_common_addr_pkgdef.svh`

### Additional Context
This diagram is intended to complement the existing documentation by providing a structured, visual representation of the address space, which was previously described only textually.

I would be happy to make more changes if required.

Thank you.

### DIFF
--- a/docs/eval_guide.md
+++ b/docs/eval_guide.md
@@ -74,6 +74,20 @@ PARAMS:
 - DEV_TRACE_P: CLINT and CFG tracer
 - DRAM_TRACE_P: DRAM tracer
 
+## bp_top Simulation Examples
+**Hello World**
+```bash
+make -C bp_top/verilator build.verilator sim.verilator
+```
+- Validates: Basic boot flow, UART output, and core-to-memory communication.
+- Expected Output: "Hello World!" appearing in the simulation log.
+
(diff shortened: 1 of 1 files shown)


dpetrisko:
Hi thanks, I'd rather have a generic markdown table for offline viewing, I don't think the flow chart is adding any more information in this case

Issue #343, opened 2020-02-26
Add more information to Interface Verifier


dpetrisko:
bp_nonsynth_if_verif currently prints out the top level parameters and a handful of interface widths. It would be helpful to see some more derived statistics such as:

1) Message header/data field sizes
2) Flits per message type / flits per header
3) Total cache size
4) ... etc.

dpetrisko:
Partially fixed with https://github.com/black-parrot/black-parrot/pull/1241, leaving open for further improvements!

Mayuri-pawar21:
Hi @dpetrisko  , I've opened pull request #1272  to add L2 cache size, total cache size, and number of cores as further enhancements for this issue. Please review when you have time.

dpetrisko:
Closing due to vagueness in definition of "done" -- improvements are always welcome :)

PR #1272, opened 2026-03-24
Add more derived statistics to interface verifier (#1241)

Mayuri-pawar21:
## Summary
Adds L2 cache total size, total cache size (all levels), and number of cores to the derived statistics printed by bp_nonsynth_if_verif.

## Changes
- L2 cache size (when present)
- Combined total cache size (L1 I$, L1 D$, L2)
- Number of cores

## Related Issue
Further improvements for #1241

## Verification
Code compiles; I was unable to run the full simulation due to environment setup, but the added lines follow the existing pattern and use parameters already defined in the file.

PR #1275, opened 2026-03-24
Add L2 and Front End telemetry to bp_nonsynth_if_verif

aarushmohanty2004:
### Summary
Adds L2 cache total size #1272 (accounts for slices as mentioned) to the derived statistics printed by bp_nonsynth_if_verif.sv Adds Front End Throughput Logic to the derived statistics printed by bp_nonsynth_if_verif.sv

### Issue Fixed
fixes the issue #343 and also builds on additions by #1241 and includes #1272 with the minor correction made

### Area
bp_top/test/common/bp_nonsynth_if_verif.sv

The issue #1272 had some changes to be made also did not include expected simulation output. The Front End validation was added to provide an immediate "sanity check" for the processor's instruction fetch capacity. It ensures that the simulated I-Cache size matches the developer's intent and establishes a baseline to see if a test program is large enough to actually stress the hardware's storage limits.

Might require the removal of L2 cache total size part of the code if the implementation is changed and updated in issue #1272 

### Analysis
There is zero impact on the synthesized hardware. The logic adds a negligible amount of overhead to the initial simulation startup to calculate and print these parameters to the console.

### Verification
Verified using the Verilator Simulation workflow. 
Ran make -C bp_top/verilator build.verilator sim.verilator
L2 Validation: Confirmed correct aggregate capacity calculation for multicore and "Not Present" reporting for Unicore (UCE) builds.
Front End Validation: Verified that the I-Cache byte capacity and instruction slot counts correctly reflect the icache_sets_p, icache_assoc_p, and icache_block_width_p parameters.
Ran sanity check using bsg_mem instantiation prints(32 * 128 bits * 2 = 8192 bits = 8KB)
OUTPUT log snippet:
########### DERIVED STATISTICS ##############
L1 I-Cache Total Size:   32 KB
L1 D-Cache Total Size:   32 KB
LCE REQ Header Flits:    1
LCE CMD Header Flits:    1
LCE FILL Header Flits:   1
LCE RESP Header Flits:   1
MEM FWD Header Flits:    1
MEM REV Header Flits:    1
L2 Cache Total Size: 8 KB
L2 Cache: 32 sets, 2-way, 64 B/line, 2 banks, data width 16 B
Fetch Throughput Stats:-
I-Cache Capacity:      32768 Bytes
Instruction Slots:     8192 units
#############################################

### DIFF
--- a/bp_common/src/include/bp_common_aviary_cfg_pkgdef.svh
+++ b/bp_common/src/include/bp_common_aviary_cfg_pkgdef.svh
@@ -358,8 +358,8 @@
       ,l2_slices           : 2
       ,l2_banks            : 1
       ,l2_data_width       : 128
-      ,l2_sets             : 32
-      ,l2_assoc            : 2
+      ,l2_sets             : 256
+      ,l2_assoc            : 4
       ,l2_block_width      : 512
       ,l2_fill_width       : 128
 
--- a/bp_me/src/v/cce/bp_cce_fsm.sv
+++ b/bp_me/src/v/cce/bp_cce_fsm.sv
@@ -483,7 +483,7 @@ module bp_cce_fsm
 
   // Counter for message send/receive
   logic cnt_rst;
-  logic [2:0] cnt_inc, cnt_dec;
+  logic cnt_inc, cnt_dec;
   logic [`BSG_WIDTH(num_lce_p+1)-1:0] cnt;
   bsg_counter_up_down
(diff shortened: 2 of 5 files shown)


dpetrisko (changes requested):
Please be more thorough when researching these parameters. It is unhelpful to add incorrect telemetry

dpetrisko:
Cool thanks

PR #1288, opened 2026-03-30
bp_be: fix interrupt priority ordering (Fixes #1287)

### LINKED ISSUE #1287: CSR interrupt priority path is marked non-compliant and can violate privileged ordering
## Problem
1.The CSR file has a TODO indicating interrupt priority is non-compliant in bp_be_csr.sv
2.Current selection logic can produce non-spec ordering when multiple interrupts are pending with delegation interaction.

## Observation
1.With concurrent pending sources (MEIP, MSIP, MTIP), ordering can be incorrect under specific pending/enable timing.
2.This is most visible in simultaneous-pending and race-style scenarios.

## Expected behavior
1.Interrupt cause selection follows privileged-spec ordering deterministically.
2.Delegated and non-delegated classes are resolved explicitly before selecting M-mode or S-mode cause.

## Proposed solution
Replace generic encoded selection with explicit ordered selection logic for M and S interrupt causes.
Keep delegation-aware gating explicit.
Add directed tests for:
 1.baseline priority ordering
 2.simultaneous pending
 3.pending-before-enable and enable-before-pending race

## Verification
1.Run full sim-bp-tests gate.
2.Confirm directed interrupt-priority tests pass consistently.

Sairam-Kesanapalli:
## Summary:
  - Replace non-compliant interrupt priority selection in CSR path with explicit ordered logic
  - Keep delegation-aware machine/supervisor interrupt selection explicit
  - Update sim-bp-tests list to include interrupt priority tests
  - Include aviary package enum fix required for clean simulation parse

## Validation:
  - Directed tests pass: interrupt_priority, interrupt_priority_simultaneous, interrupt_priority_race
  - Local full sequential sim-bp-tests gate: 29 passed, 0 failed, 0 missing
    (from /tmp/full_gate_seq_driver.log)"

### DIFF
--- a/bp_be/src/v/bp_be_calculator/bp_be_csr.sv
+++ b/bp_be/src/v/bp_be_calculator/bp_be_csr.sv
@@ -151,27 +151,20 @@ module bp_be_csr
   wire ssi_v = mie_r.ssie & mip_r.ssip;
   wire sei_v = mie_r.seie & (mip_r.seip | s_external_irq_i);
 
-  // TODO: interrupt priority is non-compliant with the spec.
-  wire [15:0] interrupt_icode_dec_li =
-    {4'b0
+  // Interrupt priority follows the RISC-V Privileged Spec ordering.
+  wire m_mei_v = mei_v & ~mideleg_lo[11];
+  wire m_msi_v = msi_v & ~mideleg_lo[3];
+  wire m_mti_v = mti_v & ~mideleg_lo[7];
+  wire m_sei_v = sei_v & ~mideleg_lo[9];
+  wire m_ssi_v = ssi_v & ~mideleg_lo[1];
+  wire m_sti_v = sti_v & ~mideleg_lo[5];
 
-     ,mei_v
-     ,1'b0
-     ,sei_v
-     ,1'b0
+  wire s_sei_v = sei_v &  mideleg_lo[9];
+  wire s_ssi_v = ssi_v &  mideleg_lo[1];
+  wire s_sti_v = sti_v &  mideleg_lo[5];
 
-     ,mti_v
-     ,1'b0 // Reserved
-     ,sti_v
-     ,1'b0
-
-     ,msi_v
-     ,1'b0 // Reserved
-     ,ssi_v
-     ,1'b0
-     };
-
-  assign irq_waiting_o = |interrupt_icode_dec_li;
+  assign irq_waiting_o = m_mei_v | m_msi_v | m_mti_v | m_sei_v | m_ssi_v | m_sti_v
+                       | s_sei_v | s_ssi_v | s_sti_v;
 
   rv64_exception_dec_s exception_dec_li;
   assign exception_dec_li =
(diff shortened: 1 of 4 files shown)


bp_be/src/v/bp_be_calculator/bp_be_csr.sv
```diff
@@ -151,27 +151,20 @@ module bp_be_csr
   wire ssi_v = mie_r.ssie & mip_r.ssip;
   wire sei_v = mie_r.seie & (mip_r.seip | s_external_irq_i);
 
-  // TODO: interrupt priority is non-compliant with the spec.
-  wire [15:0] interrupt_icode_dec_li =
-    {4'b0
+  // Interrupt priority follows the RISC-V Privileged Spec ordering.
```
dpetrisko: We should just swizzle the inputs and keep the priority encoder structure the same

dpetrisko (changes requested):

bp_be/src/v/bp_be_calculator/bp_be_csr.sv
Sairam-Kesanapalli: I updated the implementation to keep the priority-encoder structure and swizzle inputs as suggested.
Re-ran directed tests: interrupt_priority, interrupt_priority_simultaneous, interrupt_priority_race (all pass).

Sairam-Kesanapalli:
Hi Dan, quick question on aviary enum formatting in bp_common_aviary_pkgdef.svh:555.

The leading comma here was added to get the interrupt-priority validation flow passing, but I’m now seeing conflicting behavior across flows:

With the leading comma, FP hazard testing fails
Without it, the interrupt-priority path previously failed

Could you confirm the preferred syntax/style for this enum? Is there a tool/parser-specific requirement we should follow here?

dpetrisko:
>Could you confirm the preferred syntax/style for this enum? Is there a tool/parser-specific requirement we should follow here?

Uhhh, that's really odd. As currently written, a leading comma like that should not compile. My guess would be that it was not cleaning / rebuilding properly when you were testing. I'll try it locally...

bp_be/src/v/bp_be_calculator/bp_be_csr.sv
```diff
@@ -151,27 +151,36 @@ module bp_be_csr
...
-     ,1'b0 // Reserved
-     ,sti_v
-     ,1'b0
-
-     ,msi_v
-     ,1'b0 // Reserved
-     ,ssi_v
-     ,1'b0
+  // Swizzled interrupt decode vectors preserve priority-encoder structure while enforcing spec priority ordering.
+  wire m_mei_v = mei_v & ~mideleg_lo[11];
```
dpetrisko: We don't delegate M-mode interrupts to S-mode, so it should just be sei / ssi / sti checking mideleg here

Sairam-Kesanapalli:
Thanks for the help improving this—learned something here.

dpetrisko:
Thank you! This was a longstanding legitimate bug and your tests were extremely helpful in exposing it and verifying the fix

PR #1289, opened 2026-03-30
fix cce-decode-dir-case-bug

durgeshjanapareddy9-collab:
Summary

Fixes an incorrect "unique case" statement in the CCE instruction decoder ("bp_cce_inst_decode.sv") within the "e_op_dir" block.

Problem

A wrong "unique case (minor_op_u.flag_minor_op)" was present before the correct "dir_minor_op" case, leading to invalid decode behavior.

Fix

 Removed the incorrect "flag_minor_op" case
 Kept the correct "unique case (minor_op_u.dir_minor_op)"

Verification

 "bp_top_demo" (Verilator) – PASS
 No errors or assertion failures
 Confirms correct decode behavior in full system context

Impact

 Fixes decode logic for directory operations
 Improves correctness of CCE execution


dpetrisko:
Hi, thanks for the pull 

Is this a casting issue or is the actual field in the struct different?

1) for multicore you'll need to set CFG=e_bp_multicore_1_cce_ucode_cfg, have you done that?
2) can you run the full BP-tests suite including the multicore mc_ tests?

durgeshjanapareddy9-collab:
> Hi, thanks for the pull 
> 
> Is this a casting issue or is the actual field in the struct different?
> 
> 1) for multicore you'll need to set CFG=e_bp_multicore_1_cce_ucode_cfg, have you done that?
> 2) can you run the full BP-tests suite including the multicore mc_ tests?
> 
> 
 Thanks for the review!

1. 
   This is not a casting issue. The problem was due to using "flag_minor_op" instead of "dir_minor_op" in the "e_op_dir" decode logic within "bp_cce_inst_decode.sv", which is inconsistent with the instruction encoding.

2. Multicore Configuration
   I tested the design using the multicore configuration ("CFG=e_bp_multicore_1_cce_ucode_cfg"). The simulation builds and runs successfully, and I observed no errors or assertion failures.

3. Validation:
"bp_top_demo" test: PASS
 Multicore configuration: simulation runs   successfully without failures



Thanks!

dpetrisko:
https://github.com/black-parrot/black-parrot/blob/ca9f76918fccfa48b39725d97937a3bfb80bb47b/bp_me/src/include/bp_me_cce_inst_pkgdef.svh#L179

Okay … it is a casting issue. This is a union meaning that the bits are exactly the same. So there is no functional difference or “invalid decode behavior”

LGTM to clean things up, but for the future:
The demo test passed before your change and after your change. That means it’s either 1) the test is not sufficient to trigger the bug or 2) there is no functional bug. In this case it’s the latter

PR #1290, opened 2026-03-31
TB: Fix associative array nonblocking assignment for Verilator 5.045

binarysorcerer:
### Summary
Fix Verilator 5.045 build failure caused by associative array nonblocking assignment in [bp_nonsynth_dram.sv](cci:7://file:///home/saiyam/PR/SV/black-parrot/bp_top/test/common/bp_nonsynth_dram.sv:0:0-0:0).

### Issue Fixed
No existing issue — discovered during Verilator 5.045 build testing.

### Area
[bp_top/test/common/bp_nonsynth_dram.sv](cci:7://file:///home/saiyam/PR/SV/black-parrot/bp_top/test/common/bp_nonsynth_dram.sv:0:0-0:0), [bp_me/verilator/waiver.vlt](cci:7://file:///home/saiyam/PR/SV/black-parrot/bp_me/verilator/waiver.vlt:0:0-0:0)

### Reasoning
Verilator 5.045+ enforces IEEE 1800-2023 §6.21, which prohibits nonblocking assignments (`<=`) to dynamically-sized variables including associative arrays. This causes `ASSIGNDLY` errors during `make build.verilator`.

None. The fix is self-contained to non-synthesizable testbench code.

### Analysis
- **Zero RTL impact** — `bp_nonsynth_dram` is a testbench-only model
- Blocking vs nonblocking distinction is irrelevant for non-synthesizable associative array writes
- Enables BlackParrot to build cleanly with Verilator 5.045+

### Verification
- `bp_cce` unit test passes with Verilator 5.045 (`make build.verilator TRACE=1`)
- Correct `.fst` waveforms generated and verified

### Additional Context
Verilator error before fix:

### DIFF
--- a/bp_me/verilator/waiver.vlt
+++ b/bp_me/verilator/waiver.vlt
@@ -1,0 +1,5 @@
+`verilator_config
+// Waiver: bp_nonsynth_dram uses associative array (mem [int]) in nonblocking
+// assignments. Verilator 5.045 rejects this per IEEE 1800-2023 6.21.
+// Safe to waive: this is a non-synthesizable testbench model.
+lint_off -rule ASSIGNDLY -file "*/bp_nonsynth_dram.sv"
--- a/bp_me/verilator/waiver.vlt
+++ b/bp_me/verilator/waiver.vlt
@@ -1,5 +1,0 @@
-`verilator_config
-// Waiver: bp_nonsynth_dram uses associative array (mem [int]) in nonblocking
-// assignments. Verilator 5.045 rejects this per IEEE 1800-2023 6.21.
-// Safe to waive: this is a non-synthesizable testbench model.
-lint_off -rule ASSIGNDLY -file "*/bp_nonsynth_dram.sv"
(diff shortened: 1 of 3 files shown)


dpetrisko:
Cool thanks, I'm able to reproduce the compile error and confirm the fix

bp_me/verilator/waiver.vlt
```diff
@@ -0,0 +1,5 @@
+`verilator_config
```
dpetrisko: Remove the waiver, it's unnecessary if we're applying the fix

dpetrisko (changes requested):

dpetrisko:
>Blocking vs nonblocking distinction is irrelevant for non-synthesizable associative array writes

Technically the difference is that the memory is now read-after-write whereas before it was read-before-write. However, the DMA interface should not expose this behavior, so I think this is fine. Supposedly this can reduce total memory usage in Verilator, so should be a net win

dpetrisko:
Thanks!

PR #1292, opened 2026-04-04
Fix/cce pending bits saturation

durgeshjanapareddy9-collab:
Summary

Adds simulation-only assertions to detect overflow and underflow conditions in "bp_cce_pending_bits".

Issue Fixed

N/A

Area

"bp_me/src/v/cce/bp_cce_pending_bits.sv" (Cache Coherence Engine)

Reasoning (new feature, inefficient, verbose, etc.)

The pending bit counters in the CCE (by default 3-bits wide) track outstanding coherence transactions using standard addition/subtraction.

The original design allows wrap-around behavior:

1. Overflow (7 -> 0): Counter wraps to 0 when exceeding max value
2. Underflow (0 -> 7): Counter wraps to max value when decremented at zero

While the design assumes correct protocol behavior, these cases can make debugging difficult if unexpected sequences occur.

This change adds simulation-only ("synopsys translate_off") checks to detect overflow and underflow attempts, helping identify such issues during verification without modifying hardware behavior.

Additional Changes Required (if any)

None.

Analysis

 Behavior remains identical to the original implementation within the valid operating range "[0, 2^width_p - 1]"
 No additional hardware logic introduced
 Assertions are active only in simulation

Verification

1. Custom Module Regression (Icarus Verilog)
A standalone testbench was used to exercise boundary conditions (overflow/underflow scenarios).
Assertions correctly trigger when invalid counter operations are attempted.

2. Full System Simulations (Verilator)
The processor was rebuilt and simulated to ensure no regression in system behavior:

 "e_bp_default_cfg" → "[BSG-PASS]"
 "e_bp_multicore_1_cce_ucode_cfg" → "[BSG-PASS]"
 "e_bp_unicore_cfg" → "[BSG-PASS]"

Results:

 No errors or assertion failures during normal execution
 No impact on functional behavior

Additional Context

The original source code includes a comment noting that counters may overflow/underflow. This PR adds explicit detection for these cases during simulation, improving debug visibility while preserving existing functionality.


dpetrisko:
I would accept a nonsynth block at the bottom that detects overflow and underflow and asserts

Please do not add extra hardware for this as it should be handled by system and programmer constraints

durgeshjanapareddy9-collab:
> I would accept a nonsynth block at the bottom that detects overflow and underflow and asserts
> 
> Please do not add extra hardware for this as it should be handled by system and programmer constraints

@dpetrisko Thanks for the feedback!

I’ve updated the PR accordingly:
 Updated the PR description
 Removed saturation logic to avoid adding hardware changes
 Added nonsynth assertions to detect overflow and underflow conditions.

dpetrisko (approved):
Cool, thanks!

PR #1294, opened 2026-04-04
Preserve uncached AMO msg_type/subop across IO bridge paths + add directed regression(Fixes #1293)

### LINKED ISSUE #1293: I/O bridge drops uncached AMO semantics by forcing store subop and binary read/write mapping
## Problem
 In I/O bridge translation paths, AMO message type and subop were not preserved.
 AMO requests could be downgraded to non-AMO behavior.

## Impact
 Incorrect semantics for uncached atomic operations over IO path.
 Potential software-visible correctness issues for synchronization/MMIO atomics.

## Fix
 Preserve AMO message type and subop across bp_io_link_to_lce.sv and bp_io_cce.sv.

Sairam-Kesanapalli:
## Fixes AMO semantic loss in IO bridge translation:
 Preserve req/mem AMO types and subop fields in both directions.

## Adds directed regression test:
  bp_io_amo_bridge_regression_tb.sv

## Evidence:
 Fails on pre-fix code (AMO downgraded)
 Passes on fixed code (AMO type/subop preserved)


dpetrisko:
We’re not going to maintain a one-off testbench for something like this, please remove. The issue is that the widgets on the other side of the I/O CCE don’t support AMOs. The proper way to implement would be to add AMO support to bp_sacc_scratchpad and test it end to end

Sairam-Kesanapalli:
Implemented the AMO fix end-to-end for the scratchpad accelerator path.
The update preserves AMO metadata through the IO bridges and adds AMO/LR-SC execution at the scratchpad endpoint, with configuration-safe width handling.
I validated this with an end-to-end run on scratchpad config using mc_amo_add, and the simulation now passes.

bp_top/src/v/bp_sacc_scratchpad.sv
```diff
@@ -29,42 +29,112 @@ module bp_sacc_scratchpad
...
-  wire spm_r_v_li = r_v_li && (global_addr_lo.hio == 1);
-  wire [dword_width_gp-1:0] spm_data_li = data_lo;
+  logic mem_rev_v_r;
+  logic [dword_width_gp-1:0] mem_rev_data_r;
+  bp_bedrock_mem_rev_header_s mem_rev_header_r;
+  assign mem_rev_v_o = mem_rev_v_r;
+  assign mem_rev_header_cast_o = mem_rev_header_r;
+
+  // Pack 64b internal response data onto the outgoing Bedrock fill bus width.
+  if (acache_fill_width_p > 0)
```
dpetrisko: I think bsg_bus_pack should handle the 0 case properly, did you run into errors?

bp_top/src/v/bp_sacc_scratchpad.sv
```diff
@@ -29,42 +29,112 @@ module bp_sacc_scratchpad
...
-  assign local_addr_lo = addr_lo;
+  bp_global_addr_s req_global_addr_lo;
+  assign req_global_addr_lo = mem_fwd_header_li.addr;
+
+  wire req_is_wr = (mem_fwd_header_li.msg_type == e_bedrock_mem_wr);
+  wire req_is_rd = (mem_fwd_header_li.msg_type == e_bedrock_mem_rd);
+  wire req_is_amo = (mem_fwd_header_li.msg_type == e_bedrock_mem_amo);
+  wire req_is_spm = (req_global_addr_lo.hio == 1'b1);
+  wire req_is_csr = (mem_fwd_header_li.addr inside {accel_wr_cnt_csr_idx_gp});
+  logic [dword_width_gp-1:0] req_data;
```
dpetrisko: I think for simplicity + hardware, let's just make bp_sacc_scratchpad only handle <= 64b requests

bp_top/src/v/bp_sacc_scratchpad.sv
```diff
@@ -29,42 +29,112 @@ module bp_sacc_scratchpad
   // CCE-IO interface is used for uncached requests-read/write memory mapped CSR
   `declare_bp_bedrock_if(paddr_width_p, lce_id_width_p, cce_id_width_p, did_width_p, lce_assoc_p);
   `declare_bp_memory_map(paddr_width_p, daddr_width_p);
+  `bp_cast_i(bp_bedrock_mem_fwd_header_s, mem_fwd_header);
+  `bp_cast_o(bp_bedrock_mem_rev_header_s, mem_rev_header);
 
-  logic r_v_li, w_v_li;
-  logic [paddr_width_p-1:0] addr_lo;
-  logic [dword_width_gp-1:0] data_li, data_lo;
-  bp_me_bedrock_register
```
dpetrisko, line 36: Can you add this functionality to a new module 'bp_me_bedrock_register_amo'

1) copy-paste the original bp_me_bedrock_register code (this way we don't accidentally break existing code)
2) add a very small FSM to do a read->write on atomic operation
3) nit: do the amo outside of a function, like in https://github.com/bespoke-silicon-group/basejump_stl/blob/a43571d2eaaae2dda7c10490e8350dfdac7da878/bsg_cache/bsg_cache.sv#L710

dpetrisko (commented):
Looks promising!

Sairam-Kesanapalli:
1.Added new module bp_me/src/v/dev/bp_me_bedrock_register_amo.sv by starting from bp_me/src/v/dev/
bp_me_bedrock_register.sv, so existing module behavior remains untouched.
2.Added a small AMO read->write FSM in the new module (capture AMO request, perform read, then conditional writeback).
3.Implemented AMO ALU in always_comb case style (outside function), per the nit.
4.Updated scratchpad to use the new module at bp_top/src/v/bp_sacc_scratchpad.sv and removed inlined AMO control there.
5.Added explicit <=64b guard in scratchpad for simplicity/hardware expectation.
6.Updated filelists so the new module is compiled:
 bp_me/flist.vcs
 bp_top/flist.vcs
 
 Validation run:
 1.AMO/LRSC smoke tests passed in Verilator:
    rv64ua-p-amoadd_d
    rv64ua-p-lrsc
2.mc_amo_add passed when run directly.
3.ci/sim-mc-tests.sh was not used as authoritative in this environment because GNU parallel is not installed.
4.streaming_accelerator_vdp was started but terminated due to long runtime (no final pass/fail recorded).

bp_me/src/v/dev/bp_me_bedrock_register_amo.sv
```diff
@@ -0,0 +1,257 @@
...
+      assign addr_match[i] = (reg_addr inside {base_addr_p[i]});
+    end
+
+  // Minimal AMO tracking FSM: capture the AMO request, then issue a single
+  // write pulse after read data becomes available.
+  logic amo_pending_r, amo_write_done_r;
+  logic [els_p-1:0] amo_match_r;
+  logic [reg_addr_width_p-1:0] amo_addr_r;
+  logic [reg_size_width_p-1:0] amo_size_r;
+  logic [reg_data_width_p-1:0] amo_operand_r;
```
dpetrisko: Don't think we need to register these. Can read directly from the mem_fwd and pop when sending mem_rev

dpetrisko (changes requested):

bp_me/src/v/dev/bp_me_bedrock_register_amo.sv
Sairam-Kesanapalli: Done.

bp_top/src/v/bp_io_link_to_lce.sv
```diff
@@ -206,8 +208,13 @@ module bp_io_link_to_lce
       fsm_req_v_li                      = fsm_req_ready_then_lo & fsm_fwd_v_lo;
       fsm_fwd_yumi_li                   = fsm_req_v_li;
 
-      fsm_rev_header_li.msg_type        = lce_cmd_wr_not_rd ? e_bedrock_mem_wr : e_bedrock_mem_rd;
-      fsm_rev_header_li.subop           = e_bedrock_store; // TODO: support I/O AMOs
+      if (fsm_cmd_header_lo.msg_type == e_bedrock_cmd_uc_st_done)
```
dpetrisko, line 211: I would rearrange this as:

```
if (fsm_cmd_header_lo.msg_type == e_bedrock_mem_wr)
  fsm_rev_header_li.msg_type = e_bedrock_mem_wr;
else if (fsm_cmd_header_lo.msg_type == e_bedrock_mem_amo)
  fsm_rev_header_li.msg_type = e_bedrock_mem_amo;
else
  fsm_rev_header_li.msg_type = e_bedrock_mem_rd;
```

Reason being that it's not part of the contract that AMOs cannot set e_bedrock_store for some hypothetical overload. Safer to check the message type only.

bp_top/src/v/bp_io_link_to_lce.sv
```diff
@@ -206,8 +208,13 @@ module bp_io_link_to_lce
...
+        fsm_rev_header_li.msg_type = e_bedrock_mem_wr;
+      else if (fsm_cmd_header_lo.subop != e_bedrock_store)
```
dpetrisko, line 213: Same point as above (correct me if I'm mistaken and e_bedrock_store is actually needed)

dpetrisko:
Sorry for churn but it's going to take a while to review the accelerator part of this code. Can you break out the scratchpad and amo register changes into a new PR and we can merge just the bp_io_cce and bp_io_link_to_lce changes?

Sairam-Kesanapalli:
Sure

bp_top/src/v/bp_io_link_to_lce.sv
Sairam-Kesanapalli: fsm_cmd_header_lo.msg_type isn't a fwd type here, it's a cmd type — in this bridge it's always either e_bedrock_cmd_uc_data (11) or e_bedrock_cmd_uc_st_done (12). Neither of those ever equals e_bedrock_mem_wr (1) or e_bedrock_mem_amo (2), so both ifs stay false and everything falls into else → gets marked e_bedrock_mem_rd.
Did you mean checking a saved copy of fsm_fwd_header_lo.msg_type from when the request went out, instead of fsm_cmd_header_lo.msg_type?

Sairam-Kesanapalli:
- Deleted bp_me_bedrock_register_amo.sv module (moved to new PR)
- Restored bp_sacc_scratchpad.sv to upstream master (removed AMO changes)
- Removed register_amo entries from bp_me/flist.vcs and bp_top/flist.vcs
- Retained IO CCE and IO link bridge changes for AMO support

PR #1296, opened 2026-04-04
bp_me: align popd queue-op support comment with assembler behavior(#Fixes 1295)

Sairam-Kesanapalli:
## Summary
Synchronize SV ISA package queue-op documentation with current assembler behavior for popd to avoid support ambiguity.

## Changes
Update popd queue-op comment in bp_me_cce_inst_pkgdef.svh:163

## Rationale
1.Keeps package-level support messaging consistent with actual instruction plumbing.
2.Reduces confusion during microcode evolution and review.

### DIFF
--- a/bp_me/src/include/bp_me_cce_inst_pkgdef.svh
+++ b/bp_me/src/include/bp_me_cce_inst_pkgdef.svh
@@ -160,8 +160,7 @@
   //,e_pushqc_op                           = 4'b0001   // Push Queue Custom Message
     ,e_popq_op                             = 4'b0010   // Pop Queue - dequeue the header
     ,e_poph_op                             = 4'b0011   // Pop Header From Queue - does not pop message
-    // TODO: popd not yet fully supported - will be supported after serdes changes
-    ,e_popd_op                             = 4'b0100   // Pop Data From Queue
+    ,e_popd_op                             = 4'b0100   // Pop Data From Queue to GPR (uses popq encoding)
     ,e_specq_op                            = 4'b0101   // Write or read speculative access bits
     ,e_inv_op                              = 4'b1000   // Send all Invalidations based on sharers vector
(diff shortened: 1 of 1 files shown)


dpetrisko:
This and 
https://github.com/black-parrot-sdk/bedrock/pull/11

feel like you just fed the TODOs into Claude and let it run…

The comment in the hardware (https://github.com/black-parrot/black-parrot/blob/master/bp_me/src/v/cce/bp_cce_inst_decode.sv#L721) says that popd is not fully supported until deserialization changes are completed. I’m not totally sure what it means and would have to do a more thorough review, but it’s definitely not just that the software isn’t sufficiently tested

Sairam-Kesanapalli:
Sorry about that — I’ll take a step back and review the hardware comment more carefully before pushing changes.
I may have oversimplified this as a parser/testing issue, so I’ll dig into the deserialization aspect and underlying support first.

Sairam-Kesanapalli:
After going through the files again, this is the problem according to my current understanding (please correct me if I am wrong):
1.popd semantics/documentation describe popping one 64-bit data packet per instruction.
2.In the default configuration, incoming BedRock fill beats are 128 bits.
3.In the current popd data path, I do not see logic to select lower vs upper 64-bit lane within a 128-bit beat.
4.Queue yumi/pop behavior appears to consume a beat, not a half-beat.
5.So this looks like a deeper data-path/serialization contract mismatch, not just assembler-side completeness or testing gaps. 

Questions I have:
1.Is the intended popd unit of consumption 64-bit lane or full fill beat?
2.If lane-based, where should lane index/state live (stream pump, queue wrapper, or CCE message logic)?
3.Is there an expected invariant that bedrock_fill_width equals 64 for popd usage paths, or is width-agnostic behavior required?

dpetrisko:
Fixed by #1311

PR #1308, opened 2026-05-08
bp_uce: route e_cache_binval to e_inval to fix deadlock

flaviens:
Hi there!

e_cache_binval was missing from the e_ready dispatch tree, so bulk invalidate requests incorrectly fell through to e_send_critical and deadlocked the UCE.

This PR routes binval_v_li through the invalidate path alongside single-line invalidates, so bulk invalidation reaches e_inval and clears all cache sets as intended.

Thanks!
Flavien


dpetrisko:
Hi, this is intentional. block invalidation is handled entirely in the D$ for non-coherent systems (e.g. those that contain UCE). Therefore, the binval cache_req will never be sent to the UCE

https://github.com/flaviens/black-parrot/blob/185956278df35ca6fcfac566294fcbfa4a38fd74/bp_be/src/v/bp_be_dcache/bp_be_dcache.sv#L703

Happy to accept an assertion in bp_uce that this is never encountered, but we shouldn't need an FSM change here

flaviens:
Hi Dan, clever, I missed it. Updated with an assertion only just in case it can still be useful.
Thanks!
Flavien

dpetrisko (approved):
Thanks!

PR #1314, opened 2026-08-22
Restrict satp access by privilege

flaviens:
The specific `satp` decode bypassed U-mode checks. Reject U-mode access.


dpetrisko:
Hi, thanks seems reasonable. Please provide an assembly test that fails before and passes after the change. This applies to all PRs

flaviens:
Sure will do 👍

flaviens:
Hi @dpetrisko an assembly test has been added for each PR. It's not super automated (the assembly is self checking at the moment but that's it). Lmk if you want to have it in CI or something like this and maybe I can try to set this up.
Thanks for your quick answer btw!

PR #1316, opened 2026-08-22
Apply MPRV to PTW privilege

flaviens:
The PTW used M-mode privilege for MPRV data accesses. Carry separate instruction and data privilege modes.


dpetrisko:
Hmm, this seems buggy. mprv shouldn't apply to instruction ptw, so masking the trans_info ahead of time would be too early. 

Whether that's an actual issue with this fix or not, I think adding prv to trans_info and handling it in the ptw is cleaner

flaviens:
Let me double check that

flaviens:
Indeed that's right. Attempted a fix.

flaviens:
Updated. The translation state now carries current privilege plus MPRV data privilege. The PTW uses current privilege for instruction walks and data privilege for load or store walks.

dpetrisko:
Awesome, thank you! Seems reasonable though I have to look a bit closer. For the test, can you add conditions for instruction / data / debug mode (if not already there)?

flaviens:
Sure! Will do.

flaviens:
The bp-tests PR now covers normal instruction/data accesses plus debug mode; the debug regression passes here and fails on master.

bp_be/src/v/bp_be_calculator/bp_be_csr.sv
```diff
@@ -759,10 +759,12 @@ module bp_be_csr
   assign commit_pkt_cast_o.iscore_v          = retire_pkt_cast_i.iscore;
   assign commit_pkt_cast_o.fscore_v          = retire_pkt_cast_i.fscore;
 
+  wire mprv_en = (~is_debug_mode | dcsr_lo.mprven) & mstatus_lo.mprv;
   assign trans_info_cast_o.priv_mode      = priv_mode_r;
+  assign trans_info_cast_o.dpriv_mode     = mprv_en ? mstatus_lo.mpp : priv_mode_r;
   assign trans_info_cast_o.base_ppn       = satp_lo.ppn;
   assign trans_info_cast_o.translation_en = translation_en_r
```
dpetrisko: This may be a similar existing bug, but shouldn't translation_en also be split into instruction and data fetches? It looks like we shadow translation_en in the frontend so I this would be applying mprv_en to instruction fetches

dpetrisko (changes requested):

bp_be/src/v/bp_be_calculator/bp_be_csr.sv
flaviens: Hi Dan! Thanks! I’ve made the existing instruction/data translation split clear with only data accesses using MPP when MPRV is active.

PR #1317, opened 2026-08-22
Reject writes to read-only counters

flaviens:
The counter decode shadowed the read-only check. Reject counter writes.


dpetrisko:
https://github.com/black-parrot/black-parrot/blob/master/bp_be/src/v/bp_be_checker/bp_be_instr_decoder.sv#L535

Is that not covered here?

flaviens:
The exact counter arms above match first and only checked counteren. They shadow the generic read-only arm. The cycle-write regression in black-parrot-sdk/bp-tests#51 exposes it.

dpetrisko:
Ahh, okay I see. I think it would be clearer if we could separate out L535-538 into its own case statement, just so that we don't have to add csrw_o to any other ready-only CSRs we end up adding

flaviens:
Hi Dan,

Will make it clearer asap. Thanks for checking and suggesting the
improvements!

flaviens:
Thanks for the reminder!

dpetrisko:
Hi, I was thinking more like:

```
                  // Check for CSR-specific illegal accesses
                  unique casez (instr[31-:12])
                    `CSR_ADDR_FCSR
                    ,`CSR_ADDR_FFLAGS
                    ,`CSR_ADDR_FRM      : illegal_instr_o = !decode_info_cast_i.fpu_en;
                    `CSR_ADDR_CYCLE     : illegal_instr_o = !decode_info_cast_i.cycle_en;
                    `CSR_ADDR_INSTRET   : illegal_instr_o = !decode_info_cast_i.instret_en;
                    `CSR_ADDR_SATP      : illegal_instr_o = decode_info_cast_i.s_mode & decode_info_cast_i.tvm;
                    `CSR_ADDR_PMPCFG0
                    ,`CSR_ADDR_PMPCFG2
                    ,`CSR_ADDR_PMPADDR  : illegal_instr_o = num_pmp_p == 0;
                   endcase
                   
                  // Check for generally illegal CSR accesses
                   unique casez (instr[31-:12])
                    {12'b11??_????_????}: illegal_instr_o |= csrw_o;
                    {12'b??01_????_????}: illegal_instr_o |= decode_info_cast_i.u_mode;
                    {12'b??10_????_????}: illegal_instr_o |= decode_info_cast_i.s_mode | decode_info_cast_i.u_mode;
                    {12'b??11_????_????}: illegal_instr_o |= decode_info_cast_i.s_mode | decode_info_cast_i.u_mode;
                  endcase
```

Would this work for this and similar cases? I think this is a bit more extensible since we could add more specific exceptions to the first clause without worrying about priority in the second clause.

flaviens:
Indeed, keeping the generic checks additive would avoid the priority issue. I'll revise the PR.

flaviens:
Added the checks (used AI, please check)

dpetrisko:
Thank you!

PR #1318, opened 2026-08-22
Check the full superpage PPN

flaviens:
The gigapage check missed the low PPN bits. Check the full alignment range.


dpetrisko:
Good catch, thanks

PR #1319, opened 2026-08-22
Fix user counter gates

flaviens:
PR #1262 left U-mode counter access gated by one CSR. Require both gates.


dpetrisko:
Thanks, looks good. Seems weird that you would need both bits but the spec is clear and I guess I can see the logic

For all PRs, thank you for the assembly tests but we prefer to keep this repo pure RTL so please aggregate the new tests in a new PR to https://github.com/black-parrot-sdk/bp-tests (I can fix them up to fit in the bp-tests build system if needed)

flaviens:
@dpetrisko sure! Makes sense. I'll clean this up and send the stuff to the sdk repo.

flaviens:
Moved all six regressions to black-parrot-sdk/bp-tests#51. Removed them from the RTL PRs.

dpetrisko:
>Moved all six regressions to https://github.com/black-parrot-sdk/bp-tests/pull/51. Removed them from the RTL PRs.

Thank you! I'll be traveling for the next few weeks but these all look great with my commented nits. If you have time, could you verify that the regressions all pass under spike in the SDK? Because these are compliance bugs, I expect it should work out of box, but if there is any BP-specific behavior in implementation we'll want to model that

flaviens:
Hi Dan,

Sure I'll do in the next few days as soon as I have a min. Thanks for
checking all this!

PR #1320, opened 2026-08-23
Fault on reserved PTE bits

flaviens:
Nonzero reserved Sv39 PTE bits are accepted.

Fault when the reserved field is set.

Full-core result: master FAIL, this branch PASS


dpetrisko:
Hi thanks, looks good. Just to confirm my understanding of the spec, is this statement accurate?

'the PTE "reserved bits" can be set by software but fault when implicitly read by PTW, if the corresponding PTE extensions are not implemented'

flaviens:
Yes that's true that software may store those bits but that said, a page-table walk must fault when reserved bits for unsupported extensions are set.

dpetrisko:
Perfect, thank you!

PR #1326, opened 2026-09-11
Fault on reserved non-leaf PTE A/D/U bits

flaviens:
Follow-up to #1320: include non-leaf A/D/U in the reserved-bit fault check, following the same privileged-spec 1.12 rule. Leaf checks are unchanged.

Validation: syntax and diff checks; full-core simulation not rerun.


dpetrisko:
LGTM thanks!

